PR #40, opened 2016-02-23
**Review Only** Incorproate coreclr-debug into omnisharp extension

chuckries:
1. Adds coreclr-debug.ts to implement coreclr-debug acquisition and hooks
   it into omnisharp's activation handler
2. Adds the coreclr-debug directory containing the project.json,
   NuGet.config and dummy.cs required dotnet restore/publish
3. Adds the debugger section to contributes in package.json
4. Adds .vscodeignore to exclude items not needed at runtime from being
   packaged

TODO:
1. Update project.json references
2. Update NuGet.config to not point to http://dbgnuget (switch to
   nuget.org once pacakges are published publicly)
3. Create some completion file that we can check for correct complete
   installation
4. Improve reporting status of acquisition to end user


chuckries:
I don't think this should be merged yet but I was hoping to get some feedback.

chuckries:
@gregg-miskelly for discussion

chuckries:
For the minimum supported VSCode version in the package.json, we would like to bump this up to whatever the next release of VSCode is. Do you have any issue with this?

src/coreclr-debug.ts
```diff
@@ -0,0 +1,245 @@
...
+        return '.exe'
+    }
+
+    return '';
+}
+
+function getPlatformLibExtension() : string {
+    switch (process.platform) {
+        case 'win32':
+            return '.lib';
```
chuckries: this should be .dll

src/coreclr-debug.ts
```diff
@@ -0,0 +1,245 @@
...
+    
+    if (existsSync(_debugAdapterDir)) {
+        console.log('.NET Core Debugger tools already installed');
+        return;
+    }
+
+    _channel.appendLine("Downloading and configuring the .NET Core Debugger...");
+    _channel.show(vscode.ViewColumn.Three);
+    
+    vscode.window.setStatusBarMessage("Downloading and configuring the .NET Core Debugger...");
```
chuckries: Remove.

package.json
```diff
@@ -33,6 +33,7 @@
     },
     "activationEvents": [
         "onLanguage:csharp",
+        "onLanguage:fsharp",
```
DustinCampbell: Why is this needed?

package.json
```diff
@@ -128,6 +129,228 @@
                 "language": "csharp",
                 "path": "./snippets/csharp.json"
             }
+        ],
+        "debuggers": [
+            {
+                "type": "coreclr",
+                "label": ".NET Core",
+                "enableBreakpointsFor": { "languageIds": [ "csharp", "fsharp" ] },
```
DustinCampbell: Note that the C# extension does not actually support editing of F# code. I'm not sure this is the right place for this.

src/coreclr-debug.ts
```diff
@@ -0,0 +1,245 @@
+'use strict';
```
DustinCampbell: Need copyright header. Look at other .ts files in the repo to get the boilerplate.

src/coreclr-debug.ts
```diff
@@ -0,0 +1,245 @@
...
+
+import * as vscode from 'vscode';
+import * as child_process from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+
+var _coreClrDebugDir: string;
+var _debugAdapterDir: string;
+var _channel: vscode.OutputChannel;
```
DustinCampbell: Best to use 'let' rather than 'var'. I'm working on moving things to ES6. Note that there's a tslint gulp task included that you can run in VS Code to find minor issues like this.

chuckries: I'm fairly new to node/js. Do I need to use let throughout or just for these file level variables? My understanding is that let has block scope semantics vs global scope semantics?

DustinCampbell: It's OK, I can change them later when I do a full sweep through the code base. Essentially, in ES6 it's "let" for variables that can be assigned to and "const" for variables that can't be assigned to.

Don't sweat it.

package.json
```diff
@@ -128,6 +128,228 @@
...
+                },
+
+                "configurationAttributes": {
+                "launch": {
+                    "required": [ "program", "cwd" ],
+                    "properties": {
+                    "program": {
+                        "type": "string",
+                        "description": "Path to the program (executable file) to launch. On Windows, a '.exe' suffix is appended if not specified already.",
+                        "default": "${workspaceRoot}/bin/Debug/dnxcore50/<My-Project-Name>"
```
gregg-miskelly: We should merge in the most recent changes that @caslan made to this (removal of 'dnxcore50').

src/coreclr-debug.ts
```diff
@@ -0,0 +1,174 @@
...
+    .then(function() {
+        return spawnChildProcess('dotnet', ['--verbose', 'publish', '-o', _debugAdapterDir], _channel, _coreClrDebugDir);
+    }).then(function() {
+        var promises: Promise<void>[] = [];
+
+        promises.push(renameDummyEntrypoint());
+        promises.push(removeLibCoreClrTraceProvider());
+
+        return Promise.all(promises);
+    }).then(function() {
```
gregg-miskelly: Should we either check the version of 'dotnet' or check that it created the content file items to handle the case that the user has an older version of .NET CLI installed?

DustinCampbell:
Closing in favor of #44.

Issue #42, opened 2016-02-25
package.json license should read "SEE LICENSE IN LICENSE.txt"


chrisdias:
From https://docs.npmjs.com/files/package.json#license

In order for the Marketplace to scrape the license and host it (see [here for an example](https://marketplace.visualstudio.com/items/msjsdiag.debugger-for-chrome/license)), and therefore be at the endpoint of the LICENSE button in the Command Palette (in Feb2016 drop of VS Code) you need to set the license property in package.json to this:

```json
"license" : "SEE LICENSE IN LICENSE.txt"
```

PR #56, opened 2016-03-02
Update licenses for the C# extension

gregg-miskelly:
This commit addresses issue #42 and other licensing concerns --
1. Update the readme to reference licensing
2. Change the package.json to reference the EULA rather than the source license.
3. Checkin the OpenDebugAD7 license so that it has a home on the web. Needed for its nuget package.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -11,3 +11,4 @@ coreclr-debug/bin/**
 coreclr-debug/obj/**
 coreclr-debug/project.lock.json
 coreclr-debug/install.log
++RuntimeLicenses/dependencies/*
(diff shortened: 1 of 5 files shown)


gregg-miskelly:
I think we are having CRLF issues. Let me fix those first.

gregg-miskelly:
Text encoding problems fixed. @DustinCampbell please review.

DustinCampbell:
Are there any other notices that we need to include for the various bits third-party software that are included? c.f. https://github.com/OmniSharp/omnisharp-vscode/blob/master/ThirdPartyNotices.txt

gregg-miskelly:
The EULA isn't a replacement for ThirdPartyNotices.txt - one needs both, there is a ThirdPartyNotices.txt which should list all the third party open source that are contained directly in the package, and the EULA needs to indicate that there is third party code. We can certainly link to the ThirdPartyNotices.txt from readme.md, but I don't think it would actually be a requirement.

DustinCampbell:
Sounds good. Just wanted to check. I'm no lawyer :smile:

PR #66, opened 2016-03-08
Ensure extension is activated based on presence of project.json

DustinCampbell:
cc @chuckries, @gregg-miskelly


gregg-miskelly:
:+1:

PR #67, opened 2016-03-08
Make extension more resilient against the .NET Core-based OmniSharp

DustinCampbell:
The OmniSharp protocol will be changing a bit for .NET Core. Notably, 'MSBuild' is no longer part of the workspace information and 'DotNet' replaces 'Dnx'. This change should handle either variation of the protocol and the extension will no longer error when properties are missing.


gregg-miskelly:
:+1:

PR #68, opened 2016-03-08
Silently fail if .NET CLI isn't installed

gregg-miskelly:
This checkin adds code to the debugger downloader to detect if the .NET CLI is installed. If not, it will silently exit without trying to download the debugger.

### DIFF
--- a/src/coreclr-debug.ts
+++ b/src/coreclr-debug.ts
@@ -24,6 +24,12 @@ export function installCoreClrDebug(context: vscode.ExtensionContext) {
         return;
     }
     
+    if (!isOnPath('dotnet')) {
+        // TODO: In a future release, this should be an error. For this release, we will let it go
+        console.log("The .NET CLI tools are not installed. .NET Core debugging will not be enabled.")
+        return;
+    }
+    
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@DustinCampbell and @chuckries please review.

src/coreclr-debug.ts
```diff
@@ -1,194 +1,228 @@
...
+
+// Determines if the specified command is in one of the directories in the PATH environment variable.
+function isOnPath(command : string) : boolean {
+    let pathValue = process.env['PATH'];
+    if (!pathValue) {
+        return false;
+    }
+    let fileName = command;
+    let seperatorChar = ':';
+    if (process.platform == 'win32') {
```
DustinCampbell: ===

DustinCampbell:
I'm not sure why the diff shows this file as being removed and re-added, but :+1:

chuckries:
:+1:  looks good. did the default encoding or tab/space behavior change somehow? I am also curious why the entire file shows as diff'd.

DustinCampbell:
OK. With @chuckries's sign off, I'll go ahead and merge.

PR #73, opened 2016-03-08
Add --configfile to dotnet commands

chuckries:


chuckries:

gregg-miskelly:
:+1:

chuckries:
Not sure if this is truly necessary but I don't think it can hurt. I think this will prevent us from also pulling in the user default NuGet.config, but we <clear /> the package sources we try to use anyway, so I think it's fine.

gregg-miskelly:
Makes sense to me. I think you should merge.

PR #115, opened 2016-03-21
Prompt user to add build task if missing

DustinCampbell:
When the extension is activated and the workspace contains a project.json in the root path, we now check to see if there exists a `{workspaceRoot}/.vscode/tasks.json` file. If the file doesn't exist, or it exists but doesn't contain a build task, we prompt the user to add the build task.


src/tasks.ts
```diff
@@ -0,0 +1,87 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+'use strict';
+
+import * as fs from 'fs-extra-promise';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as tasks from 'vscode-tasks';
+
+function promptToAddBuildTask() {
```
gregg-miskelly: Define the type of the return?

src/tasks.ts
```diff
@@ -0,0 +1,87 @@
...
+    fs.ensureDirAsync(vscodeFolder).then(() => {
+        fs.existsAsync(tasksJsonPath).then(exists => {
+            if (exists) {
+                fs.readFileAsync(tasksJsonPath).then(text => {
+                    const fileText = text.toString();
+                    let tasksConfig: tasks.TaskConfiguration = JSON.parse(fileText);
+                    let buildTask = tasksConfig.tasks.find(td => td.taskName === 'build');
+                    if (!buildTask) {
+                        promptToAddBuildTask().then(shouldAdd => {
+                            buildTask = createBuildTaskDescription();
```
gregg-miskelly: Does this code need to be in an 'if (shouldAdd) {' ...?

gregg-miskelly:
Signing off with comments.

src/tasks.ts
DustinCampbell: Thanks!

src/tasks.ts
DustinCampbell: Great catch -- thanks!

PR #122, opened 2016-03-24
Add telemetry for OmniSharp server requests

DustinCampbell:
This change adds telemetry tracking for the time elapsed from the point that a request sent to the OmniSharp server to the point that a response is received. To keep telemetry traffic low, we simply track the counts of various ranges of delay times for each type of request.

In addition, I refactored the telemetry a bit in coreclr-debug.ts to ensure that we can share the same TelemetryReporter.

cc @gregg-miskelly, @chuckries, @Pilchie

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,7 @@
         "windows": {
           "program": "./coreclr-debug/debugAdapters/OpenDebugAD7.exe"
         },
+        "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
         "configurationAttributes": {
           "launch": {
             "required": [
--- a/package.json
+++ b/package.json
@@ -157,7 +157,6 @@
         "windows": {
           "program": "./coreclr-debug/debugAdapters/OpenDebugAD7.exe"
         },
-        "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
         "configurationAttributes": {
           "launch": {
             "required": [
--- a/src/omnisharpServer.ts
+++ b/src/omnisharpServer.ts
@@ -13,6 +13,7 @@ import omnisharpLauncher from './omnisharpServerLauncher';
 import {Disposable, CancellationToken, OutputChannel, workspace, window} from 'vscode';
 import {ErrorMessage, UnresolvedDependenciesMessage, MSBuildProjectDiagnostics, ProjectInformationResponse} from './protocol';
 import getLaunchTargets, {LaunchTarget} from './launchTargetFinder';
+import TelemetryReporter from 'vscode-extension-telemetry';
 
 enum ServerState {
 	Starting,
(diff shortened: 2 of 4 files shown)


gregg-miskelly:
:+1:

gregg-miskelly:
Sorry, that comment was supposed to be for the other PR. I haven't had a chance to look at this one yet. Give me a few minutes...

package.json
```diff
@@ -157,7 +157,6 @@
         "windows": {
           "program": "./coreclr-debug/debugAdapters/OpenDebugAD7.exe"
         },
-        "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
```
gregg-miskelly: We need this. VS Code reads this value.

DustinCampbell: It does? The instructions for VS Code telemetry do not say to add this to the debuggers section. Or, is this used for the debug adapter telemetry (which I'm totally unfamiliar with).

src/omnisharpServer.ts
```diff
@@ -57,9 +135,30 @@ export abstract class OmnisharpServer {
 	private _setState(value: ServerState) : void {
 		if (typeof value !== 'undefined' && value !== this._state) {
 			this._state = value;
-			this._fireEvent('stateChanged', this._state);
+			this._fireEvent(Events.StateChanged, this._state);
 		}
 	}
+    
+    private _recordDelay(path: string, elapsedTime: number) {
```
gregg-miskelly: Might be obvious to other people, but I am having trouble figuring out what 'path' means.

src/omnisharpServer.ts
```diff
@@ -57,9 +135,30 @@ export abstract class OmnisharpServer {
...
+            delays = new Delays();
+            this._telemetryDelays[path] = delays;
+        }
+        
+        delays.report(elapsedTime);
+    }
+    
+    public reportAndClearTelemetry() {
+        for (var path in this._telemetryDelays) {
+            const eventName = 'omnisharp' + path;
```
gregg-miskelly: If 'path' is a file path, might that have PII?

DustinCampbell: no it doesn't. It's the "path" to the request. You're right. I'll pick a better name here.

package.json
gregg-miskelly: Yes, it is used with debug adapter telemetry.

gregg-miskelly:
:+1:

package.json
DustinCampbell: OK -- will add that back. Sorry about that! I wasn't sure because the debugger code just dug it out of package.json.

DustinCampbell:
@gregg-miskelly I implemented your feedback. Look better now?

gregg-miskelly:
:+1:

Pilchie:
:+1:

PR #135, opened 2016-03-29
Implement debug proxy to display error message before install completes

chuckries:
This adds a debug proxy that acts as the debug adapter program before the
coreclr-debug are fully downloaded. If the user tries to start debugging
before download is finished, the proxy displays a sane error message. Once
the download is complete, we rewrite the manifest to no longer call the
proxy.

During the VS Code session that downloads the components, the manifest is
rewritten but not reloaded. In this case the proxy spawns the real
debugger as a child process and proxies its stdin/stdout. Once vs code is
restarted the new manifest is loaded and the proxy is no longer called.

Also adds an empty command that can be run to force activation of the C#
extension, which will kick off debugger acquisition.


chuckries:
@caslan @gregg-miskelly @DustinCampbell

src/coreclr-debug/main.ts
```diff
@@ -1,263 +1,262 @@
...
-                reject(new Error(code.toString()));    
-            }
-            else {
-                resolve();
-            }
-        });
-    });
-
-    return promise;
+/*---------------------------------------------------------------------------------------------
```
chuckries: not sure why this file is showing as one big delete and add. possibly because of the rename? I thought github could handle this better. Maybe I'm missing something? Doesn't seem to be a line ending thing.

src/coreclr-debug/proxy.ts
```diff
@@ -0,0 +1,109 @@
...
+// During subsequent code sessions, the rewritten manifest will be loaded and this proxy will no longer be called. 
+function proxy() {
+    util.setExtensionDir(path.resolve(__dirname, '../../'));
+    
+    if (!util.installCompleteExists()) {
+        if (util.existsSync(util.installBeginFilePath())) {
+            process.stdout.write(serializeProtocolEvent(new ProxyErrorResponse('The .NET Core Debugger has not finished installing. See Status Bar for details.')));
+        } else {
+            process.stdout.write(serializeProtocolEvent(new ProxyErrorResponse('Run \'Debugger: Install .NET Core Debugger\' command or open a .NET project directory to download the .NET Core Debugger')));
+        }
```
chuckries: These error strings are not finalized.

src/coreclr-debug/proxy.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    else
+    {
+        new Promise<void>(function(resolve, reject) {
+            let processPath = path.join(util.debugAdapterDir(), "OpenDebugAD7" + util.getPlatformExeExtension());
+            let args = process.argv.slice(2);
+            
+            // do not explicitly set a curretnt working dir
```
caslan: "current"

caslan:
LGTM

src/coreclr-debug/util.ts
```diff
@@ -0,0 +1,117 @@
...
+import * as os from 'os';
+
+let _extensionDir: string = '';
+let _coreClrDebugDir: string = '';
+let _debugAdapterDir: string = '';
+let _installLogPath: string = '';
+let _installBeginFilePath: string = '';
+let _installCompleteFilePath: string = '';
+
+export function setExtensionDir(extensionDir: string) {
```
DustinCampbell: Hmmm... a bunch of state and function that operate on it. Maybe just make a class? :-)

chuckries: probably a good call. still not super used to javascript/typescript.

chuckries:
@DustinCampbell I abstracted util into a class. Is this more what you had in mind?

chuckries:
#136

src/coreclr-debug/main.ts
```diff
@@ -1,263 +1,263 @@
...
+        vscode.window.setStatusBarMessage('Successfully installed .NET Core Debugger.');
+    })
+    .catch(function(error) {
+        const viewLogMessage = "View Log";
+        vscode.window.showErrorMessage('Error while installing .NET Core Debugger.', viewLogMessage).then(function (value) {
+            if (value === viewLogMessage) {
+                _channel.show(vscode.ViewColumn.Three);
+            }
+        });
+        statusBarMessage.dispose();
```
gregg-miskelly: Do we want this in the success case too?

chuckries: Are you asking if we want to dispose the first message before setting the new success message? Probably.

src/coreclr-debug/proxy.ts
```diff
@@ -0,0 +1,109 @@
...
+// This proxy will still be called and launch OpenDebugAD7 as a child process.
+// During subsequent code sessions, the rewritten manifest will be loaded and this proxy will no longer be called. 
+function proxy() {
+    let util = new CoreClrDebugUtil(path.resolve(__dirname, '../../'));
+    
+    if (!CoreClrDebugUtil.existsSync(util.installCompleteFilePath())) {
+        if (CoreClrDebugUtil.existsSync(util.installBeginFilePath())) {
+            process.stdout.write(serializeProtocolEvent(new ProxyErrorResponse('The .NET Core Debugger has not finished installing. See Status Bar for details.')));
+        } else {
+            process.stdout.write(serializeProtocolEvent(new ProxyErrorResponse('Run \'Debugger: Install .NET Core Debugger\' command or open a .NET project directory to download the .NET Core Debugger')));
```
gregg-miskelly: As discussed, I would suggest a different name for this command.

src/coreclr-debug/main.ts
chuckries: The way it's written, the success message will never be disposed. The only thing we could do is set it to be disposed after a certain timeout.

package.json
```diff
@@ -50,6 +51,7 @@
     "onCommand:o.execute-last-command",
     "onCommand:dotnet.restore",
     "onCommand:csharp.addTasksJson",
```
gregg-miskelly: @DustinCampbell should we remove 'csharp.addTasksJson' while we are here?

src/omnisharpMain.ts
```diff
@@ -91,8 +91,12 @@ export function activate(context: vscode.ExtensionContext): any {
 		server.stop();
 	}));
     
+	// register empty handler for csharp.installDebugger
+	// running the command activates the extension, which is all we need for installation to kickoff
+	disposables.push(vscode.commands.registerCommand('csharp.downloadDebugger', () => { }));
```
gregg-miskelly: We registered all the other commands in src\features\commands.ts. Should that be there also?

gregg-miskelly:
Otherwise looks good to me.

package.json
DustinCampbell: Yes, didn't realize I'd missed that.

chuckries: I'll remove it.

PR #188, opened 2016-04-14
Update with latest OmniSharp and pick correct version for current Linux distribution

DustinCampbell:
We've updated OmniSharp to build releases for each Linux distribution that we currently support: CentOS, Debian, Red Hat (RHEL), and Ubuntu. This change adds code to sniff out which of the supported Linux distros we're running on to determine which flavor of OmniSharp we need to download and install.

cc @gregg-miskelly, @chuckries, @caslan, @Pilchie, @david-driscoll

### DIFF
--- a/src/omnisharpServerLauncher.ts
+++ b/src/omnisharpServerLauncher.ts
@@ -7,11 +7,12 @@
 
 import {exists as fileExists} from 'fs';
 import {spawn, ChildProcess} from 'child_process';
-import {workspace} from 'vscode';
+import {workspace, OutputChannel} from 'vscode';
 import {satisfies} from 'semver';
 import {join} from 'path';
 import {getOmnisharpLaunchFilePath} from './omnisharpPath';
 import {downloadOmnisharp} from './omnisharpDownload';
+import {SupportedPlatform, getSupportedPlatform} from './utils';
 
 const isWindows = process.platform === 'win32';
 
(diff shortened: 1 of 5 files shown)


Pilchie:
:+1:

DustinCampbell:
Note, I may rebase this against 'dev' and resubmit if we think this is important enough to reset Escrow for RC 2.

chuckries:
:+1:

gregg-miskelly:
:+1:

caslan:
LGTM

david-driscoll:
Should there be a default when we hit the none case?  I'm thinking of other distros like Arch, this would would then make @markrendle cry if vscode broke for him.

Perhaps at that point prompt the user to select the most compatible distro?  I know many are debian based, including ubuntu.  Perhaps that's a good default.

DustinCampbell:
@david-driscoll -- good suggestion. How do others feel? Would Debian or Ubuntu be a good fallback if the user is on a different distro than one of the "blessed four"? It's likely that OmniSharp may work in those cases, and if not, OmniSharp will simply fail to start.

gregg-miskelly:
I don't know that the chances that it would actually work are terribly high -- coreclr is extremely picky.

Could a user download there own and set the omnisharp path? If so, that would be my recommendation.

DustinCampbell:
@gregg-miskelly, yes, that's how others would use it. Essentially, they could download the Mono version and be certain of having it work.

gregg-miskelly:
Should we spit something to the output window telling them how to do that?

DustinCampbell:
Possibly -- threading that through is a bit tricky because the omnisharpDownload.ts module is used by Gulp as well, which can't have a dependency on VSCode. Let me take a look.

gregg-miskelly:
Optional call back delegate?

DustinCampbell:
No, I can put it elsewhere -- before we even get to this code.

DustinCampbell:
No, I can put this elsewhere.

DustinCampbell:
OK, I added a commit that provides instructions to the user if we fail to find a version of OmniSharp that targets the current Linux distribution. Thoughts?

src/omnisharpServerLauncher.ts
```diff
@@ -20,20 +21,34 @@ export interface LaunchResult {
...
-export function installOmnisharpIfNeeded(): Promise<string> {
+export function installOmnisharpIfNeeded(output: OutputChannel): Promise<string> {
     return getOmnisharpLaunchFilePath().catch(err => {
+        if (getSupportedPlatform() == SupportedPlatform.None && process.platform === 'linux') {
+            output.appendLine("[ERROR] Could not locate an OmniSharp server that supports your Linux distribution.");
+            output.appendLine("");
+            output.appendLine("OmniSharp provides a richer C# editing experience, with features like IntelliSense and Find All References.");
+            output.appendLine("It is recommend that you download the version of OmniSharp that runs on Mono using the following steps:");
+            output.appendLine("    1. If it's not already installed, download and install Mono (http://www.mono-project.com)");
+            output.appendLine("    2. Download and untar https://github.com/OmniSharp/omnisharp-roslyn/releases/download/v1.9-alpha13/omnisharp-linux-mono.tar.gz");
```
DustinCampbell: I'll file an issue to make this a bit more robust for RTM. We don't want to have to update this URL everytime we release.

gregg-miskelly:
LGTM

DustinCampbell:
We have sign off from shiproom as well.

PR #212, opened 2016-04-18
Change attach template to use processId by default

caslan:


gregg-miskelly:
👍

DustinCampbell:
How does setting processId to "0" change the behavior?

caslan:
Debugger will give an error saying you need to configure the process id to attach to (rather than saying "couldn't find process with id ..."

DustinCampbell:
Ah got it. LGTM

PR #216, opened 2016-04-19
Change default value for 'processId'

gregg-miskelly:
If you delete and re-add the "processId" field it would give a default value of:
   "processId": ""

Which might tempt someone to change it to:
   "processId": "1234"

Which is wrong, because processId is an integer rather than a string. But this
distinction is pretty subtle to folks who don't deal with json every day.

The default value to 0 to remove the temptation.


gregg-miskelly:
@caslan please review.

caslan:
LGTM

PR #223, opened 2016-04-20
Move project.json dependency completion & hover to vscode-omnisharp

aeschli:
Until now the code that drove intellisense for dependencies in project.json was sitting in the JSON extension.
We want the omnisharp extension to take ownership.
This pull request implements a completion provider and a hover provider using the newly created 'jsonc-parser' node module that does the JSON file processing.
It gets nuget packages from the nuget API, but is not aware of any nuget configuration files.

In the future the omnisharp extension can also consider to power the completions and hovers directly from the omnisharp server.

We will remove the project.json completion support from the JSON extension for the coming release (end of April).


aeschli:
Close, not sure hat went wrong when merging. Created #224 instead.

PR #233, opened 2016-04-21
Properly extract XML doc comment summary content in hover tips, signature help, and completion list

### LINKED ISSUE #13: Method Intellisense mangled & truncated
_From @bitcrazed on January 21, 2016 23:43_

Playing around with ASP.NET Core and noticed that the pop-up intellisense for one of the API's I am accessing appears to be broken in several ways:
image

The text is:

1. Garbled - the API text doesn't appear to be readable end-user intellisense
1. Truncated - the API text/docs are truncated to the right

Running VS Code 0.10.6 on Windows 10 11099 on Surface Pro 4.

_Copied from original issue: Microsoft/vscode#2165_

### LINKED ISSUE #30: Cannot get member parameters on parameter help properly
Moved from microsoft/vscode#2807
From @tugberkugurlu

Not sure why, here is the screenshot.

image

 * On Windows 8.1
 * ASP.NET 5 application is on `1.0.0-rc1-update1`
 * `GitHubClient` class I am using here is available on `Octokit v0.16.0`
 * VS Code about:

    image

DustinCampbell:
Fixes #13 
Fixes #30

### DIFF
--- a/src/features/documentation.ts
+++ b/src/features/documentation.ts
@@ -2,29 +2,30 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
+
 'use strict';
 
+const summaryStartTag = /<summary>/i;
+const summaryEndTag = /<\/summary>/i;
 
-const _regExp = /<(\S*?).*?>((.|\r|\n)*?)<\/\1>/;
-
-/**
- * remove xml-tags from string
- */
-export function plain(doc: string): string {
-
-	if (!doc) {
-		return doc;
+export function extractSummaryText(xmlDocComment: string): string {
+	if (!xmlDocComment) {
+		return xmlDocComment;
 	}
-
-	let newDoc: string;
-
-	while (true) {
-		newDoc = doc.replace(_regExp,(m, g1, g2, g3) => g2);
-		if (newDoc === doc) {
-			break;
-		}
-		doc = newDoc;
+	
+	let summary = xmlDocComment;
+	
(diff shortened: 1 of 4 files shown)


DustinCampbell:
@Pilchie, do you want to take a look?

src/features/documentation.ts
```diff
@@ -2,29 +2,30 @@
...
 
-const _regExp = /<(\S*?).*?>((.|\r|\n)*?)<\/\1>/;
+'use strict';
 
-/**
- * remove xml-tags from string
- */
-export function plain(doc: string): string {
+const summaryStartTag = /<summary>/i;
+const summaryEndTag = /<\/summary>/i;
```
Pilchie: Why `i`?  xml doc tags are case sensitive?

src/features/documentation.ts
```diff
@@ -2,29 +2,30 @@
...
 
-	if (!doc) {
-		return doc;
+export function extractSummaryText(xmlDocComment: string): string {
```
Pilchie: I realize this may be opening a can of worms, but do we have any facilities for unit testing this stuff?

Pilchie:
:+1:, except the case sensitivity part.

src/features/documentation.ts
DustinCampbell: No. They're case *in*sensitive

image

src/features/documentation.ts
DustinCampbell: It's a can of worms I'll be digging into soon. There are no unit tests on this whole thing, and I'm intending to fully open that can of worms next week.

src/features/documentation.ts
Pilchie: Huh.  I never knew that.  :shipit: then :)

PR #238, opened 2016-04-22
Stop using run-in-terminal for 'dotnet restore'

DustinCampbell:
Stop opening a new terminal window to run 'dotnet restore'. Instead, launch 'dotnet restore' as a child process and capture the output

cc @gregg-miskelly, @chuckries, @caslan, @Pilchie

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -8,11 +8,13 @@
 import {OmnisharpServer} from '../omnisharpServer';
 import * as serverUtils from '../omnisharpUtils';
 import findLaunchTargets from '../launchTargetFinder';
-import {runInTerminal} from 'run-in-terminal';
+import * as cp from 'child_process';
+import * as fs from 'fs-extra-promise';
 import * as path from 'path';
+import * as protocol from '../protocol';
 import * as vscode from 'vscode';
 
-const isWindows = process.platform === 'win32';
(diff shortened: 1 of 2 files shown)


src/features/commands.ts
```diff
@@ -65,23 +87,15 @@ export function dotnetRestoreAllProjects(server: OmnisharpServer) {
...
-                            cwd: path.dirname(project.Path)
-                        });
-                    }
-                });
-            }
-        }
-
-		return vscode.window.showQuickPick(commands).then(command => {
+		return Promise.all(commandPromises).then(commands => {
+			return vscode.window.showQuickPick(commands);
```
gregg-miskelly: I know you didn't change this but curious - why does it ask which project to restore instead of restoring all of them?

gregg-miskelly:
:+1:

src/features/commands.ts
DustinCampbell: I was thinking the same thing. I think we should probably add a separate command that uses the picker. Also, if there's just a single project, it should populate the picker. It should just execute the dotnet restore.

Pilchie:
:+1: Having the extra terminal window pop up has been bugging me :)

src/features/commands.ts
DustinCampbell: File #240 to track a better experience here.

PR #266, opened 2016-05-01
Fix spelling mistake in `dotnet restore` label

scottaddie:
The trailing `e` was missing in the `dotnet restore` command's label. This PR corrects the spelling mistake which appears in Command Palette:

dotnet_restore


gregg-miskelly:
Hi @scottaddie thank you so much for taking the time to send in this PR. We definitely love the help we have been getting from the community. However, this particular issue was actually already fixed in commit https://github.com/OmniSharp/omnisharp-vscode/commit/0440d6ffb0466f111d2b6eb2197d19de589de4cf. We are unfortunately in lock down until .NET Core RC2 ships, and so we haven't merged the 'dev' branch. Shortly after it does, you should expect to see a new C# extension with this problem addressed.

scottaddie:
Thank you for the explanation!

PR #320, opened 2016-05-14
fix typo

Jeffiy:


gregg-miskelly:
Hi @Jeffiy thank you so much for taking the time to send in this PR. We definitely love the help we have been getting from the community. However, this particular issue was actually already fixed in commit  0440d6f . We are unfortunately in lock down until .NET Core RC2 ships, and so we haven't merged the 'dev' branch. Shortly after it does, you should expect to see a new C# extension with this problem addressed.

PR #326, opened 2016-05-16
No longer download OmniSharp from GitHub releases

DustinCampbell:
This removes the code that discovers and downloads OmniSharp from the GitHub and switches over to Azure blog storage.

cc @MattGertz, @gregg-miskelly, @caslan, @chuckries, @jaredpar

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.0.8",
+  "version": "1.0.9",
   "description": "C# for Visual Studio Code (powered by OmniSharp).",
   "displayName": "C#",
   "author": "Microsoft Corporation",
@@ -21,7 +21,6 @@
     "decompress": "^3.0.0",
     "del": "^2.0.2",
     "fs-extra-promise": "^0.3.1",
-    "github-releases": "^0.3.0",
     "run-in-terminal": "*",
     "semver": "*",
     "vscode-debugprotocol": "^1.6.1",
--- a/src/omnisharpDownload.ts
+++ b/src/omnisharpDownload.ts
@@ -7,14 +7,15 @@
 
 import * as fs from 'fs-extra-promise';
 import * as path from 'path';
+import * as https from 'https';
+import * as stream from 'stream';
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -137,9 +136,9 @@
             "csharp"
           ]
         },
-        "runtime": "node",
+        "runtime": "",
         "runtimeArgs": [],
-        "program": "./out/coreclr-debug/proxy.js",
+        "program": "./coreclr-debug/debugAdapters/OpenDebugAD7",
```
gregg-miskelly: You don't want this change

package.json
```diff
@@ -137,9 +136,9 @@
             "csharp"
           ]
         },
-        "runtime": "node",
```
gregg-miskelly: Or this one.

package.json
DustinCampbell: Thanks!

package.json
DustinCampbell: Thanks!

DustinCampbell:
Fixed those bits Gregg.

src/omnisharpDownload.ts
```diff
@@ -45,46 +46,39 @@ function getOmnisharpAssetName(): string {
...
+    });
+}
+
 export function downloadOmnisharp(): Promise<boolean> {
     return new Promise<boolean>((resolve, reject) => {
         console.log(`[OmniSharp]: Installing to ${DefaultInstallLocation}`);
         
-        const repo = new Github({ repo: OmnisharpRepo, token: ApiToken });
         const assetName = getOmnisharpAssetName();
+        const urlString = `${BaseDownloadUrl}/${assetName}`;
```
gregg-miskelly: I am not sure if we really want to fix this since it is already published to blob storage, but should there be a version number as part of this URL?

gregg-miskelly:
Otherwise LGTM

src/omnisharpDownload.ts
DustinCampbell: Almost certainly, but I agree that we could do that next time.

jaredpar:
Can't comment too intelligently on TypeScript, but the logic looks good.

PR #343, opened 2016-05-18
Don't package .nyc_output or coverage directories in VSIX

DustinCampbell:
For some reason, on @gregg-miskelly is able to produce much smaller VSIXs that I am. Digging into the folder diffs, it turns out that, on my machine(s), a large amount of code coverage data from one of the node modules (readable-stream) is being included in the VSIX package. This change updates the .vscodeignore file to just ignore those.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -7,6 +7,9 @@ src/**
 .vscode/**
 .omnisharp/**
 
+**/.nyc_output/**
+**/coverage/**
+
 coreclr-debug/debugAdapters/**
 coreclr-debug/bin/**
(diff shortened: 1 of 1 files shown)


DustinCampbell:
With this change. I'm able to package a VSIX for the extension that is 1.6 MB.

DustinCampbell:
I'm going to go ahead and merge this. I don't see it being an issue at the moment, though it might be a small problem down the road if we want to have a folder for a hypothetical code coverage feature called "coverage".

gregg-miskelly:
👍

PR #350, opened 2016-05-19
Fix respecting a system or user configured proxy.

### LINKED ISSUE #349: Omnisharp not downloading with proxy
The latest release of omnisharp-vscode isn't able to download omnisharp through a proxy on Windows.
Instead, the output remains at `[INFO] Starting OmniSharp at [...]`.

It seems that `download(urlString: string)` in `omnisharpDownload.ts` doesn't use the proxy config from vscode and/or the environment variables http_proxy/https_proxy yet.

jugglingthebits:
Add libraries http-proxy-agent and https-proxy-agent
just like vscode itself does and then use that to
request omnisharp binaries.
Most code is taken from vscode and adapted.

Fixes #349

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
             "csharp"
           ]
         },
-        "runtime": "",
+        "runtime": "node",
         "runtimeArgs": [],
         "program": "./out/coreclr-debug/proxy.js",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
         },
         "runtime": "",
         "runtimeArgs": [],
-        "program": "./coreclr-debug/debugAdapters/OpenDebugAD7.exe",
+        "program": "./out/coreclr-debug/proxy.js",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
         "configurationAttributes": {
           "launch": {
--- a/package.json
+++ b/package.json
@@ -25,11 +25,13 @@
     "decompress": "^3.0.0",
     "del": "^2.0.2",
     "fs-extra-promise": "^0.3.1",
+    "http-proxy-agent": "^1.0.0",
+    "https-proxy-agent": "^1.0.0",
+    "open": "*",
     "semver": "*",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -139,9 +141,9 @@
...
           ]
         },
-        "runtime": "node",
+        "runtime": "",
         "runtimeArgs": [],
-        "program": "./out/coreclr-debug/proxy.js",
+        "program": "./coreclr-debug/debugAdapters/OpenDebugAD7.exe",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
         "configurationAttributes": {
           "launch": {
```
chuckries: You need to undo these changes in the debuggers section. This change occurs because the extension rewrites this portion of its manifest during activation. F5'ing the extension in dev scenarios causes this rewrite to occur, but the changes cannot be checked in.

jugglingthebits:
I just noticed that my change breaks the gulpfile: For the "omnisharp:fetch" task, the plugin itself is loaded and through my path it takes a dependency on the vscode module, which isn't there when gulp runs:
`const omnisharpDownload = require('./out/omnisharpDownload');`

Is that task used somewhere? I would thing a dependency on compiled code shouldn't be in the gulpfile.

DustinCampbell:
Thanks for your contribution! If you can sign the CLA, I'll be happy to merge this in.

WRT the gulp task: Yeah, there was a desire to be able to specifically download the OmniSharp server as part of a gulp task so that we could include it in build on some occasions. For now, I think you can just comment that gulp task.

src/proxy.ts
```diff
@@ -0,0 +1,46 @@
+/*---------------------------------------------------------------------------------------------
```
DustinCampbell: Sigh... I wish we didn't have to copy and paste this from VS Code and could just leverage their that's already running. But, :+1:

package.json
```diff
@@ -139,7 +141,7 @@
             "csharp"
           ]
         },
-        "runtime": "node",
+        "runtime": "",
```
DustinCampbell: was this an intentional change?

DustinCampbell:
Thanks for signing the CLA! Merging...

PR #357, opened 2016-05-20
Update coreclr-debug packages for post rc2 work

chuckries:


chuckries:
@gregg-miskelly @DustinCampbell @caslan 

Let me know if this should be merged against master instead of dev. I smoke tested that restoring the debugger works.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.0.11",
+  "version": "1.0.12",
```
gregg-miskelly: Suggestion: use version 1.1.0 to signify that we have a bigger chunk of changes. Also we still might want to ship a '1.0.12' with some low-risk-but-important fix between now and when we get the next version out.

gregg-miskelly:
Otherwise looks good to me.

PR #376, opened 2016-05-25
Schema for pipe transport options.

rajkumar42:
Schema for pipe transport options.

### DIFF
--- a/package.json
+++ b/package.json
@@ -283,38 +283,44 @@
               },
               "pipeTransport": {
                 "type": "object",
-                "description": "Describes options to launch a program through a pipe transport.",
+                "description": "When present, this tells the debugger to connect to a remote computer using another executable as a pipe that will relay standard input/output between VS Code and the .NET Core debugger backend executable (clrdbg).",
                 "default": {
-                  "pipeProgram": "enter pipe program name, for example plink.exe",
+                  "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
                   "pipeArgs": []
                 },
                 "properties" : {
                   "pipeProgram": {
                     "type": "string",
-                    "description": "The pipe command to execute.",
-                    "default": "enter pipe program name, for example plink.exe"
+                    "description": "The fully qualified pipe command to execute.",
+                    "default": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'"
                   },
                   "pipeArgs": {
                     "type": "array",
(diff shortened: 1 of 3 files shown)


rajkumar42:
@chuckries @gregg-miskelly @caslan @DustinCampbell

gregg-miskelly:
Might be obvious, but just in case - I would recommend waiting till you have an official build of OpenDebugAD7 with your changes that you can change the following line to pull in your changes:

    src\coreclr-debug\main.ts:337:            "Microsoft.VisualStudio.OpenDebugAD7": "1.0.20520-preview-3"

package.json
```diff
@@ -280,6 +280,86 @@
                   "type": "string"
                 },
                 "default": []
+              },
+              "pipeTransport": {
+                "type": "object",
+                "description": "Describes options to launch a program through a pipe transport.",
```
gregg-miskelly: How about:When present, this tells the debugger to connect to a remote computer using another executable as a pipe that will relay standard input/output between VS Code and the .NET Core debugger backend executable (clrdbg).

package.json
```diff
@@ -280,6 +280,86 @@
...
+                "properties" : {
+                  "pipeProgram": {
+                    "type": "string",
+                    "description": "The pipe command to execute.",
+                    "default": "enter pipe program name, for example plink.exe"
+                  },
+                  "pipeArgs": {
+                    "type": "array",
+                    "description": "Command line arguments passed to the pipe program.",
+                    "default": []
```
gregg-miskelly: Are you missing?

                  "items": { 
                    "type": "string" 
                  },

package.json
```diff
@@ -280,6 +280,86 @@
...
+              },
+              "pipeTransport": {
+                "type": "object",
+                "description": "Describes options to launch a program through a pipe transport.",
+                "default": {
+                  "pipeProgram": "enter pipe program name, for example plink.exe",
+                  "pipeArgs": []
+                },
+                "properties" : {
+                  "pipeProgram": {
```
gregg-miskelly: Does the MIEngine search PATH if this is not a full file path? Or does it need to be a file path? Probably worth calling out in the description either way.

gregg-miskelly:
Otherwise looks good to me.

package.json
rajkumar42: done.

package.json
rajkumar42: fixed.

package.json
rajkumar42: fixed.

PR #384, opened 2016-05-29
Update compilationOptions -> buildOptions

jamesqo:
The compilationOptions node is deprecated in RC2, buildOptions should be used instead.

See also: aspnet/Templates#548

cc: @DustinCampbell


gregg-miskelly:
Thanks for the PR! Looks good to me.

DustinCampbell:
Yes -- :+1:

PR #387, opened 2016-05-31
Update OmniSharp to 1.9-beta4

DustinCampbell:


DustinCampbell:
the OmniSharp 1.9-beta4 are now uploaded. Merging this change to let VS Code use it.

PR #397, opened 2016-06-01
Add github issue template

chuckries:
This will add a template for new issues which should make collecting info easier. Adapted from https://github.com/dotnet/cli/blob/rel/1.0.0/ISSUE_TEMPLATE

### DIFF
--- a/ISSUE_TEMPLATE
+++ b/ISSUE_TEMPLATE
@@ -1,0 +1,12 @@
+## Environment data
+`dotnet --info` output:
+VS Code version:
+C# Extension version:
+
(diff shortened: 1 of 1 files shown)


chuckries:
@DustinCampbell

gregg-miskelly:
We might want to link to a instructions page on getting the versions (since it isn't obvious how to do that). Otherwise LGTM.

chuckries:
I think that might be too cumbersome since it will appear as markdown. If people are here to submit an issue I trust them to be able to find version numbers.

Issue #369, opened 2016-05-23
ERROR: Run 'Debug: Download .NET Core Debugger' or open a .NET project...


EdDore:
Installed .NET Core 1.0 RC2. Able to create and run simple console app.
Installed Visual Studio Code (1.1.1), and installed the C# code extension.
Opened console project folder in VSCode, and was prompted to add the missing .vscode files (launch.json/tasks.json).

Set a breakpoint in Main(), and then attempted to launch under debugger (.NET Core Launch (console). Was promptly met with the following error:

Error   Run 'Debug: Download .NET Core Debugger' in the Command Palette or open a .NET projects...  "Open launch.json" "Close".

I suspect it's specific to OS X, as I was able to do the same steps on Linux 14.04 successfully.

chuckries:
What happens if you run "Debug: Download .NET Core Debugger" from the command palette? Open the command palette with command + F1 and type debug and it should appear. 
![image](https://cloud.githubusercontent.com/assets/6598594/15486066/cf3a8c14-20f7-11e6-8789-5368eb419fcc.png)

EdDore:
It doesn't appear to do anything.

chuckries:
@EdDore Is there an install.log file in ~/.vscode/extensions/ms-vscode.csharp-<version>/coreclr-debug?

EdDore:
What would be the full path to that folder on a OS x installation?

chuckries:
Here is the full path on my osx machine:
`/Users/chuckr/.vscode/extensions/ms-vscode.csharp-1.0.11/coreclr-debug/install.log`

langsweirdt:
I'm seeing the exact same problem. No install.log file below the coreclr-debug folder. I tried versions 1.0.11 and 1.0.10 of the extension, same behavior. If I can do something to help reproducing/resolving this, please let me know.

ruuda:
I am having the same issue. When I tried this a few weeks ago I gave up. I tried again just yet, with the same results. Attached is my [install.log](https://github.com/OmniSharp/omnisharp-vscode/files/289958/install.log.txt). When I visit one of the timed-out urls with my browser I do get a response.

If I try to download again in the same session, nothing happens. When I tried in a new session it went on to download some more and eventually it succeeded, but it took a very long time (about 30 minutes total), on a 8 MB/s connection. The coreclr-debug directory is ~60 MB after downloading, so it shouldn’t really have taken that long.

OS: Windows 7 x64
[...]

PR #399, opened 2016-06-02
Improve error handling for 'isOnPath'

chuckries:
isOnPath calls existsSync, which was not handling ENOTDIR. This can arise
if a portion of a path is not a directory, such as
/dir1/dir2/<executable>/foo/bar. existsSync should return false in these
cases. Additionally, this adds a try/catch to isOnPath that should catch
all other errors from existsSync and treat them as basic failures for that
particular path segment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.1.2",
+  "version": "1.1.3",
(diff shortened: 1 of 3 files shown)


chuckries:

gregg-miskelly:
👍

chuckries:
This should address #369

PR #406, opened 2016-06-06
Adding attach process picker.

rajkumar42:
This feature will light up when the required vscode features are available in stable release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   ],
   "main": "./out/omnisharpMain",
   "scripts": {
-    "postinstall": "tsc"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
     "onCommand:o.showOutput",
     "onCommand:dotnet.restore",
     "onCommand:csharp.downloadDebugger",
+    "onCommand:debugger.pickProcess",
```
gregg-miskelly: 'debugger.pickProcess' sounds likely to clash with other extensions. Should this be 'csharp.debuggerPickProcess' so something like that?

gregg-miskelly:
Do we want this in for the next release or not?

Issue #258, opened 2016-04-29
Empty variables windows when stoped in HelloMvc sample on Win7


mikes-gh:
From https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger.md

> •Requires .NET Core rc3-23829 or newer. These are daily builds of .NET Core that can be downloaded from the https://www.myget.org/F/dotnet-core/api/v3/index.json nuget feed. These are really daily builds of .NET Core and associated tools. So please continue if you want to dig in and have fun playing with new bits. But don't expect to go to production with those.

I am using  aspnetcirelease feed.  Does the above mean core debugging wont be available in rc2.
What are the practical steps for the above.

Another article I read misses this step.

Damian git cloned the project from https://github.com/aspnet/cli-samples

But when I try to debug HelloMvc I hit my breakpoint but get no variable info in the left hand panes at all.

![no variables](https://cloud.githubusercontent.com/assets/16208742/14913737/07771fb0-0dfd-11e6-828c-98814de0e23a.png)

gregg-miskelly:
@mikes-gh: sorry things are rather confusing at the moment since RC2 isn't quite out yet. The instructions in the master branch are directed at the version of the C# extension which is in the VS Code gallery. This version works with the version of the CLI on the .NET CLI getting-started page. If you want to use cli-samples, then you should follow these instructions: http://aka.ms/vscclrdogfood.

mikes-gh:
@gregg-miskelly I have downloaded the latest cli already.
I have followed those instructions
If I have other cli installed is that the issue?

gregg-miskelly:
@mikes-gh sorry I read the first part of your bug and thought you were running into mismatch problems (we have had a lot of those :)).

mikes-gh:
```
c:\Users\mikes\Repos\cli-samples\HelloMvc>dotnet --info
.NET Command Line Tools (1.0.0-rc2-002543)

Product Information:
 Version:     1.0.0-rc2-002543
 Commit Sha:  38d0c28a1e

Runtime Environment:
 OS Name:     Windows
 OS Version:  6.1.7601
 OS Platform: Windows
 RID:         win7-x64

c:\Users\mikes\Repos\cli-samples\HelloMvc>
```
I already had 2543 so issue still stands.  
I have insider vscode 1.1.0
I have c# plugin 1.0.4 rc2

My launch.json
```
{
    "version": "0.2.0",
    "configurations": [
        {
[...]

PR #407, opened 2016-06-06
Debugger changes for 1.1.4

gregg-miskelly:
This updates the clrdbg package to the 25406 build. This includes fixes
for Win7 expression evulation and the System.Task breaking change.
This addresses: https://github.com/OmniSharp/omnisharp-vscode/issues/258

This also the MIEngine to pull in https://github.com/Microsoft/MIEngine/pull/353

Lastly this disables the external console by default. We decided to have it off
by default as we have seen some issues with it on OSX (ex: not always closing
when the target exits), and we are hoping it will be possible soon to use the
new VS Code console window instead. So we decided to turn it off by default, at
least for now.


gregg-miskelly:
@chuckries please review

chuckries:
:+1:

PR #416, opened 2016-06-08
Implement task to create offline VSIX's

chuckries:
This implements a gulp task that creates platform specific vsix's that
do not require the secondary download/installation steps.

1. Refactor coreclr-debug/main.ts into activation.ts and install.ts
2. Remove VS Code specific modules from install.ts so that it can be
called from gulp
3. Refactor omnisharpDownload.ts so that it has no VS Code dependencies
and can be called from gulp task.
4. Create packaging gulp task that will perform the download of the
debugger and omnisharp in place, package the extension, and clean it up.
This repeats for as many platforms as we need to ship.


chuckries:
@DustinCampbell @gregg-miskelly @caslan

.vscodeignore
```diff
@@ -5,14 +5,8 @@ src/**
...
-.omnisharp/**
 
 **/.nyc_output/**
 **/coverage/**
 
-coreclr-debug/debugAdapters/**
-coreclr-debug/bin/**
-coreclr-debug/obj/**
-coreclr-debug/project.lock.json
-coreclr-debug/install.log
```
gregg-miskelly: Would it make sense to still exclude install.log?

gulpfile.js
```diff
@@ -5,21 +5,142 @@
...
+
+    var promise = Promise.resolve();
+
+    packages.forEach(function(pair) {
+        promise = promise.then(function() {
+            return doOfflinePackage(pair.rid, pair.omni, packageName);
+        })
+    });
+
+    return promise;
```
DustinCampbell: The common pattern for this sort of thing is to produce an array of promises and then return `Promise.all(...)` at the end. You could do something like this:

```TypeScript
return Promise.all(packages.map(function(pair) doOfflinePackage(pair.rid, pair.omni, packageName)));
```

gulpfile.js
```diff
@@ -5,21 +5,142 @@
...
+function installDebugger(runtimeId) {
+    return getDebugInstaller().install(runtimeId);
+}
+
+function cleanDebugger() {
+    try {
+        getDebugInstaller().clean();
+        console.log('Cleaned Succesfully');
+    } catch (error) {
+        console.error(error);
```
gregg-miskelly: Will this cause the script to fail?

gulpfile.js
```diff
@@ -5,21 +5,142 @@
...
+    return cleanOmnisharp();
+}
+
+gulp.task('clean', ['omnisharp:clean',  'debugger:clean', 'package:clean'], () => {
+
+});
+
+/// Omnisharp Tasks
+function installOmnisharp(omnisharpAssetName) {
+    const logFunction = (message) => { console.log(message); };
```
DustinCampbell: This is arrow function and other callbacks use the `function(...) { ... }` syntax. Should we be consistent?

gulpfile.js
```diff
@@ -5,21 +5,142 @@
...
+    packages.push({rid: 'win7-x64', omni: `omnisharp-${OmniSharpVersion}-win-x64-net451.zip`});
+    packages.push({rid: 'osx.10.11-x64', omni: `omnisharp-${OmniSharpVersion}-osx-x64-netcoreapp1.0.tar.gz`});
+    packages.push({rid: 'centos.7-x64', omni: `omnisharp-${OmniSharpVersion}-centos-x64-netcoreapp1.0.tar.gz`});
+    packages.push({rid: 'debian.8-x64', omni: `omnisharp-${OmniSharpVersion}-debian-x64-netcoreapp1.0.tar.gz`});
+    packages.push({rid: 'rhel.7.2-x64', omni: `omnisharp-${OmniSharpVersion}-rhel-x64-netcoreapp1.0.tar.gz`});
+    packages.push({rid: 'ubuntu.14.04-x64', omni: `omnisharp-${OmniSharpVersion}-ubuntu-x64-netcoreapp1.0.tar.gz`});
+
+    var promise = Promise.resolve();
+
+    packages.forEach(function(pair) {
```
gregg-miskelly: nit: should we avoid the 'function' syntax all together to avoid 'this' capture problems?

gulpfile.js
```diff
@@ -5,21 +5,142 @@
...
+        console.error(error);
+    });
+});
+
+gulp.task('debugger:clean', () => {
+    cleanDebugger();
+});
+
+/// Packaging Tasks
+function doPackage(packageName) {
```
DustinCampbell: doPackageSync?

src/coreclr-debug/install.ts
```diff
@@ -0,0 +1,226 @@
...
+                "System.Threading.Thread": "4.0.0-rc2-24027", 
+                "System.Xml.XDocument": "4.0.11-rc2-24027",
+                "System.Xml.XmlDocument": "4.0.1-rc2-24027",  
+                "System.Xml.XmlSerializer": "4.0.11-rc2-24027",
+                "System.ComponentModel":  "4.0.1-rc2-24027",  
+                "System.ComponentModel.Annotations":  "4.1.0-rc2-24027",  
+                "System.ComponentModel.EventBasedAsync":  "4.0.11-rc2-24027",
+                "System.Runtime.Serialization.Primitives": "4.1.1-rc2-24027",
+                "System.Net.Http":  "4.0.1-rc2-24027",
+                "Microsoft.NetCore.DotNetHostPolicy": "1.0.1-rc-002702"
```
gregg-miskelly: Since the CLI locks down at different times than CoreCLR/FX, I would suggest making HostPolicy specific to the offline installer scenario.

gulpfile.js
chuckries: I think I had tried that and it caused issue because all those promises will execute concurrently, while we __need__ them to execute synchronously.

src/coreclr-debug/util.ts
```diff
@@ -7,75 +7,107 @@
...
+    }
+
+    closeInstallLog(): void {
+        if (this._installLog !== null) {
+            this._installLog.close();
+        }
+    }
+
+    log(message: string): void {
+        console.log(message);
```
gregg-miskelly: In VS Code, if a debugger isn't attached to VS Code, console.log messages just get dropped on the floor. Correct?

src/omnisharpDownload.ts
```diff
@@ -119,7 +118,7 @@ export function downloadOmnisharp(output: OutputChannel): Promise<boolean> {
             })
             .catch(err =>
             {
-                output.appendLine(`[ERROR] ${err}`);
+                log(`[ERROR] ${err}`);
             });
```
gregg-miskelly: Do we want to catch this in the case of offline vsix creation?

gulpfile.js
DustinCampbell: Ah, of course. Got it! I thought there was a `Promise.chain(...)` combinator too, but I guess I was wrong.

src/omnisharpDownload.ts
DustinCampbell: I've actually already changed this in my next PR.

src/coreclr-debug/util.ts
chuckries: Correct.

gregg-miskelly:
Otherwise LGTM

src/coreclr-debug/install.ts
chuckries: Does project.json support arbitrary configurations, or just Debug/Release?

gulpfile.js
chuckries: Based on Gregg's comments and this I will standardize on lambda syntax.

src/coreclr-debug/install.ts
gregg-miskelly: I don't know, but I since we are dynamically generating the file anyway, I am sure the simplest option is to only add the element if we are doing the offline install.

chuckries: duh. good call.

gulpfile.js
DustinCampbell: sweet

gulpfile.js
```diff
@@ -5,21 +5,142 @@
...
+gulp.task('package:online', ['clean'], () => {
+    doPackage();
 });
 
-//TODO: decouple omnisharpDownload (specifically proxy.ts) from vscode 
-// gulp.task('omnisharp:fetch', ['omnisharp:clean'], () => {
-// 	return omnisharpDownload.downloadOmnisharp();
-// });
+gulp.task('package:offline', ['clean'], () => {
+    var json = JSON.parse(fs.readFileSync('package.json'));
```
gregg-miskelly: Should we either backup and restore project.json or make sure that the file is clean before and then ask git to revert it?

chuckries: I don't want to muck around with git inside the gulp tasks. I think we'll just stick with the workflow of 
1. git clean
2. npm install
3. gulp package:offline

PR #423, opened 2016-06-09
Remove myget feed from NuGet.config

gregg-miskelly:
When we released the extension, I forgot to remove the myget feed from our NuGet.config,
I was assuming this wouldn't really matter since nuget.org came first in the list, by
the myget folks reported a traffic spike, so I am assuming we should really remove this.

### DIFF
--- a/coreclr-debug/NuGet.config
+++ b/coreclr-debug/NuGet.config
@@ -4,7 +4,5 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <!-- This dependency can be commented out once we push packages to nuget for a real release -->
-    <add key="coreclrdebug" value="https://www.myget.org/F/coreclr-debug/api/v3/index.json" />    
   </packageSources>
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
We are waiting on a response from the myget.org team to understand the priority of this to understand if we should fix it now, or just before we release the next time.

coreclr-debug/NuGet.config
```diff
@@ -4,7 +4,5 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <!-- This dependency can be commented out once we push packages to nuget for a real release -->
-    <add key="coreclrdebug" value="https://www.myget.org/F/coreclr-debug/api/v3/index.json" />
```
chuckries: do we want to comment this out instead of delete it? Are we still planning on pushing daily builds to myget?

gregg-miskelly: We are still going to push daily builds to myget, but my hope is that I can get git to leave this commit in the master branch by manually resolving the conflict if we merge back to master.

PR #498, opened 2016-06-29
Update package versions

caslan:
@gregg-miskelly @chuckries @rajkumar42 @DustinCampbell

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.1.7",
+  "version": "1.2.0",
   "description": "C# for Visual Studio Code (powered by OmniSharp).",
   "displayName": "C#",
   "author": "Microsoft Corporation",
--- a/src/coreclr-debug/install.ts
+++ b/src/coreclr-debug/install.ts
@@ -162,33 +162,31 @@ export class DebugInstaller
                 emitEntryPoint: true
             },
             dependencies: {
-                "Microsoft.VisualStudio.clrdbg": "14.0.25406-preview-3044032",
-                "Microsoft.VisualStudio.clrdbg.MIEngine": "14.0.30610-preview-1",
-                "Microsoft.VisualStudio.OpenDebugAD7": "1.0.20614-preview-2",
-                "NETStandard.Library": "1.5.0-rc2-24027",
+                "Microsoft.VisualStudio.clrdbg": "14.0.25429-preview-3095034",
+                "Microsoft.VisualStudio.clrdbg.MIEngine": "14.0.30629-preview-1",
+                "Microsoft.VisualStudio.OpenDebugAD7": "1.0.20628-preview-1",
+                "NETStandard.Library": "1.6.0",
                 "Newtonsoft.Json": "7.0.1",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.1.7",
+  "version": "1.1.8",
```
gregg-miskelly: I think we should go to 1.2.0 since:
1. This is a breaking change (we are loosing rc2 debugging)
2. Leaves us space to release more 1.1.x vsix's (ex: to fix OmniSharp)

chuckries:
@caslan These packages have been pushed to nuget.org, correct?

caslan:
The release will come out of master, which references myget so we should be good. Changing the version to 1.2 is a good idea.

DustinCampbell:
Yes, let's flip it to 1.2.

src/coreclr-debug/install.ts
```diff
@@ -162,28 +162,28 @@ export class DebugInstaller
                 emitEntryPoint: true
             },
             dependencies: {
-                "Microsoft.VisualStudio.clrdbg": "14.0.25406-preview-3044032",
-                "Microsoft.VisualStudio.clrdbg.MIEngine": "14.0.30610-preview-1",
-                "Microsoft.VisualStudio.OpenDebugAD7": "1.0.20614-preview-2",
-                "NETStandard.Library": "1.5.0-rc2-24027",
+                "Microsoft.VisualStudio.clrdbg": "14.0.25429-preview-3095034",
+                "Microsoft.VisualStudio.clrdbg.MIEngine": "14.0.30628-preview-1",
```
gregg-miskelly: I noticed that I accidentally missed Wes's MIEngine commit from yesterday in this build. I kicked off a new build with that. We probably should take that too.

src/coreclr-debug/install.ts
```diff
@@ -162,28 +162,28 @@ export class DebugInstaller
...
+                "System.Xml.XmlDocument": "4.0.1",  
+                "System.Xml.XmlSerializer": "4.0.11",
+                "System.ComponentModel":  "4.0.1",  
+                "System.ComponentModel.Annotations":  "4.1.0",  
+                "System.ComponentModel.EventBasedAsync":  "4.0.11",
+                "System.Runtime.Serialization.Primitives": "4.1.1",
+                "System.Net.Http":  "4.1.0"
             },
             frameworks: {
                 "netcoreapp1.0": {
```
gregg-miskelly: I found that we didn't need the 'imports' anymore.

gregg-miskelly:
We should update README.md with a 'What's new' before we post the .vsix. I will send in a PR real quick.

DustinCampbell:
In addition, I hope to have OmniSharp ready today. I'm just finishing up installing the additional distros so I can test.

caslan:
@DustinCampbell, we have Fedora, openSuse, Linux Mint, Oracle Linux distros which we can give you remote access

caslan:
Ok, updated version to 1.2.0, update MIEngine version, removed import definitions from install.ts

PR #505, opened 2016-06-29
It's always something silly. Off by one

DustinCampbell:


DustinCampbell:
Note, string.substring takes a start and an *end*. Not a start and a length. Sigh...

PR #507, opened 2016-06-30
Update debugger.md for .NET Core RTM

gregg-miskelly:


DustinCampbell:
LGTM

debugger.md
```diff
@@ -5,9 +5,12 @@ This page gives you detailed instructions on how to debug code running under .NE
 File bugs and feature requests [here](https://github.com/OmniSharp/omnisharp-vscode/issues) and [join our insiders group](http://landinghub.visualstudio.com/dotnetcoreinsiders) to help us build great tooling for .NET Core.
 
 ####Requirements
-* Requires .NET Core 1.0 RC2 or newer (will not work with earlier versions)
+* Requires .NET Core 1.0 (rc2 are earlier releases are not supported)
 * X64 only
-* Supports OSX, Ubuntu 14.04, Red Hat Enterprise Linux 7.2, Debian 8.2, Centos 7.1, and Windows 7+
+* Supported operating systems: 
+    * macOS: 10.11+ (El Capitan+)
```
wesrupert: Have you tried out the macOS 10.12 developer preview yet? We haven't done any testing on Sierra yet, and IIRC @chuckries is currently investigating #492, so we may want to hold off on that "+" for now.

gregg-miskelly: This is the list of supported platforms from CoreCLR. I think it is find to say we support it because we intend to support it there. It is true that until we merge Chuck's PR it will not actually work, but that should be pretty darn soon.

PR #521, opened 2016-07-01
Hardcode osx.10.11-x64 for osx runtime id

chuckries:
This fixes debugger restoration on osx.10.12.

### DIFF
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -133,7 +133,7 @@ function getPlatformRuntimeId() : string {
         case 'win32':
             return 'win7-x64';
         case 'darwin':
-            return getDotnetRuntimeId();
+            return 'osx.10.11-x64';
         case 'linux':
             return getDotnetRuntimeId();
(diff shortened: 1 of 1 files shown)


chuckries:
fix for #492

wesrupert:
:+1:

DustinCampbell:
:+1:

caslan:
would this continue to work with yosemite?

chuckries:
@caslan Yosemite is 10.10 right? I didn't think we supported Yosemite today, nor did .net. Is that incorrect?

gregg-miskelly:
LGTM

gregg-miskelly:
I think .NET doesn't claim to support 10.10, but it actually works (its what I have on my Mac). I suspect it will continue to work even if we use a 10.11 rid. But I can give it a quick try if we want.

gregg-miskelly:
This works correctly on my OSX 10.10 box.

chuckries:
@gregg-miskelly thanks!

PR #530, opened 2016-07-05
Update MIEngine reference and add step filtering to launch schema

chuckries:


chuckries:
@wesrupert @gregg-miskelly @caslan

src/coreclr-debug/install.ts
```diff
@@ -163,7 +163,7 @@ export class DebugInstaller
             },
             dependencies: {
                 "Microsoft.VisualStudio.clrdbg": "14.0.25430-preview-3098923",
-                "Microsoft.VisualStudio.clrdbg.MIEngine": "14.0.30629-preview-1",
+                "Microsoft.VisualStudio.clrdbg.MIEngine": "14.0.30705-preview-1",
                 "Microsoft.VisualStudio.OpenDebugAD7": "1.0.20628-preview-1",
```
gregg-miskelly: You need to update OpenDebugAD7 for step filtering to work (commit f8714e16c880dac148801ad3bf44f12c879d5ace isn't in release-csharp).

gregg-miskelly:
Otherwise LGTM

src/coreclr-debug/install.ts
chuckries: good catch, thanks.

PR #532, opened 2016-07-06
Update link to .NET Core tools

stevensona:
http://dotnet.github.io/getting-started/ redirects to https://www.microsoft.com/net/core#windows


gregg-miskelly:
Thanks! Looks good to me.

PR #548, opened 2016-07-09
Fixing attach template to show pickProcess by default.

rajkumar42:
In a code path, template processId is 0 instead of command.pickProcess. Fixing it.


rajkumar42:
@DustinCampbell @chuckries @wesrupert

wesrupert:
:+1:

PR #550, opened 2016-07-09
Added Fact xunit snippet

jchannon:
Adds a `[Fact]` xunit snippet

### DIFF
--- a/snippets/csharp.json
+++ b/snippets/csharp.json
@@ -520,5 +520,22 @@
 			"}"
 		],
 		"description": "While loop"
+	},
+	
+	"xUnit Test": {
+
+		"prefix": "fact",
+		"body": [
+			"[Fact]",
+			"${1:public} void ${2:TestName}()", 
+			"{",
+				"//Given",
+				"",
+				"//When",
+				"",
+				"//Then",
+			"}${0}"
+		],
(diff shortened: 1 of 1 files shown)


DustinCampbell:
Should there be a field for "public"? Don't tests have to be public?

jchannon:
It is public :)

DustinCampbell:
No, I mean should there be a "snippet field" defined for the accessibility since it can't be anything other than public?

jchannon:
@DustinCampbell i see, updated 😄

snippets/csharp.json
```diff
@@ -532,5 +532,22 @@
...
 		],
 		"description": "constructor"
+	},
+
+         "xUnit Test": {
+
+		"prefix": "fact",
+		"body": [
+			"[Fact]",
+			"public void ${2:TestName}()",
```
DustinCampbell: Should this be `${1:TestName}`?

jchannon: I changed it but not it wouldnt have effected it

DustinCampbell: I didn't think it would, but it stuck out to me.

DustinCampbell:
Thanks for contributing @jchannon!

PR #566, opened 2016-07-13
Add razor support to configuration

wesrupert:
- Enable breakpoints in razor files
- Add default source map for relative paths (workaround for [aspnet/razor #803](https://github.com/aspnet/Razor/issues/803))


wesrupert:
@chuckries @caslan

caslan:
You need to modify the .ts file that writes out the ASP.NET template

Issue #555, opened 2016-07-10
.NET CLI tools cannot be located (but are on path)


robertpeacock22:
## Environment data
`dotnet --info` output:

> .NET Command Line Tools (1.0.0-preview2-003121)
> 
> Product Information:
>  Version:            1.0.0-preview2-003121
>  Commit SHA-1 hash:  1e9d529bc5
> 
> Runtime Environment:
>  OS Name:     Mac OS X
>  OS Version:  10.11
>  OS Platform: Darwin
>  RID:         osx.10.11-x64

VS Code version:
version 1.3.0

C# Extension version:
version 1.2.2

## Steps to reproduce
1. Install latest Visual Studio Code
2. Install .NET CLI tools
3. Confirm that .NET CLI tools are on path:

> cat $PATH
> cat: /Library/Android/android-sdk-mac_x86/build-tools/21.1.1:/Library/Android/android-sdk-mac_x86/tools:/Library/Android/android-sdk-mac_x86/platform-tools:/Users/robert/.rbenv/plugins/ruby-build/bin:/Users/robert/.rbenv/shims:/Users/robert/.rbenv/bin:/Users/robert/Development/Go/bin:/Users/robert/Library/CustomCommands:/Library/Android/android-sdk-mac_x86/platform-tools:/Library/Android/android-sdk-mac_x86/tools:/Library/Android/android-sdk-mac_x86:/opt/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/share/dotnet:/Users/robert/.rvm/bin: No such file or directory

4. Open C# project in VS Code

## Expected  behavior
C# plugin features should be enabled

## Actual behavior
Error message indicates that .NET CLI tools cannot be located.
![screen shot 2016-07-10 at 5 49 40 pm](https://cloud.githubusercontent.com/assets/608951/16716310/f1dd9078-46c6-11e6-80eb-e1c5e093a6ca.png)

JeremyBYU:
Experiencing Same Issue.  Same Message, but running on Windows 7.

chuckries:
@robertpeacock22 I suspect the issue is that your path variable is malformed. We should handle this better, but in the meantime can you try removing `: No such file or directory` from your `$PATH`?

@JeremyBYU What does your %PATH% variable look like?

JeremyBYU:
Heres what my path says


```
[...]

PR #572, opened 2016-07-14
check if dotnet is on $PATH by executing 'dotnet --info'

chuckries:
Previously we were manually parsing the path environment variable which is
prone to issue when users $PATH variables are malformed. This should be
more robust.


chuckries:
@caslan @wesrupert @caslan @rajkumar42

rajkumar42:
:shipit:

chuckries:
fix for #555

wesrupert:
:+1:

PR #585, opened 2016-07-18
Support using dual OmniSharp servers to improve handling of .csproj and project.json

### LINKED ISSUE #529: Add support for "Dual OmniSharp"
There is a strong need for the extension to be able to support two builds of OmniSharp: one running on the Mono/full framework and one running on .NET Core.

On Windows, the C# extension downloads and uses the version of OmniSharp targeting net451. This works well for csproj/sln support because the Windows desktop version of MSBuild is used. However, it exhibits problems with long filenames and MAX_PATH with many project.json projects (https://github.com/OmniSharp/omnisharp-roslyn/issues/594). These long filename issues are fixed in .NET Core.

On macOS/Linux, the C# extension downloads and uses the version of OmniSharp targeting .NET Core. This works great for project.json, but has a lot of problems with csproj/sln.

By supporting two versions of OmniSharp on Windows and macOS/Linux and intelligently switching between them in the extension, each of the problems above could be resolved.

DustinCampbell:
Fixes #529

This is a fairly large change that largely cleans up the OmniSharp acquisition and launching code, but also adds support for two OmniSharp servers:

1. project.json files will still target a CoreCLR-based .NET Core OmniSharp server.
2. .csproj files will target the desktop OmniSharp server (on Windows) or the Mono OmniSharp server (on OSX and Linux).

Essentially, instead of assuming a specific OmniSharp server on a particular platform, this change includes the notion of an OmniSharp "flavor", which is either CoreCLR, Desktop, or Mono.

cc @gregg-miskelly, @chuckries, @caslan, @rajkumar42, @Pilchie


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.3.0",
+  "version": "1.3.0-beta1",
```
rajkumar42: >-beta1 [](start = 19, length = 6)

remove "-beta1"?

DustinCampbell: Nope! :smile: See my email on releases that I sent last week.

Once I merge this, I'll put up a pre-release out of master for folks to test.

gulpfile.js
```diff
@@ -122,21 +127,21 @@ gulp.task('package:offline', ['clean'], () => {
...
-    packages.push({rid: 'win7-x64', omni: `omnisharp-${OmniSharpVersion}-win-x64-net451.zip`});
-    packages.push({rid: 'osx.10.11-x64', omni: `omnisharp-${OmniSharpVersion}-osx-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'centos.7-x64', omni: `omnisharp-${OmniSharpVersion}-centos-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'debian.8-x64', omni: `omnisharp-${OmniSharpVersion}-debian-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'fedora.23-x64', omni: `omnisharp-${OmniSharpVersion}-fedora-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'opensuse.13.2-x64', omni: `omnisharp-${OmniSharpVersion}-opensuse-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'rhel.7.2-x64', omni: `omnisharp-${OmniSharpVersion}-rhel-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'ubuntu.14.04-x64', omni: `omnisharp-${OmniSharpVersion}-ubuntu14-x64-netcoreapp1.0.tar.gz`});
-    packages.push({rid: 'ubuntu.16.04-x64', omni: `omnisharp-${OmniSharpVersion}-ubuntu16-x64-netcoreapp1.0.tar.gz`});
+    packages.push({rid: 'win7-x64', flavor: Flavor.Desktop, platform: Platform.Windows});
```
gregg-miskelly: Just want to make sure: you are intentionally using the Windows version, and not the CoreCLR version for offline. Correct?

package.json
DustinCampbell: To be clear, I'll change this when we merge to the release branch.

gulpfile.js
DustinCampbell: Good eye! Yes, this _could_ be CoreCLR, but I didn't want to change behavior of the offline installer immediately. The only reason I decided to start using CoreCLR on Windows for project.json projects with this PR is because CoreCLR supports long file paths, while .NET 4.5.1 doesn't.

package.json
rajkumar42: got it! thanks

---
In reply to: [71229831](https://github.com/OmniSharp/omnisharp-vscode/pull/585#discussion_r71229831) [](ancestors = 71229831)

rajkumar42:
:shipit:

src/omnisharp/launcher.ts
```diff
@@ -0,0 +1,306 @@
...
+	//   * It should be possible to choose a .csproj as a launch target
+	//   * It should be possible to choose a .sln file even when no .csproj files are found 
+	//     within the root.
+
+	if (!Array.isArray(resources)) {
+		return [];
+	}
+
+	let targets: LaunchTarget[] = [],
+		hasCsProjFiles = false,
```
gregg-miskelly: tabs

gregg-miskelly: Nevermind.

src/omnisharp/server.ts
```diff
@@ -253,63 +268,72 @@ export abstract class OmnisharpServer {
...
 	// --- start, stop, and connect
 
-	public start(solutionPath: string): Promise<void> {
-		if (!this._start) {
-			this._start = this._doStart(solutionPath);
+	private _start(launchTarget: LaunchTarget): Promise<void> {
+		const options = this._readOptions();
+		
+		let flavor: omnisharp.Flavor;
+		if (options.path !== undefined && options.usesMono === true) {
```
gregg-miskelly: If I am reading this correctly, one can use the 'usesMono' option to force the use of mono OmniSharp, but we don't have a corresponding option to force the use of CoreCLR omnisharp for folks who want to use CoreCLR omnisharp with a .csproj. Do I have that right?

If so, would it make sense to also have a way to force CoreCLR for folks dealing with PCL projects targeting CoreCLR?

src/platform.ts
```diff
@@ -62,26 +62,35 @@ export function getSupportedPlatform() {
...
+    return Platform.Unknown;	
 }
+
+export function isWindows(platform: Platform) {
+    return platform === Platform.Windows;
+}
+
+export function isUnix(platform: Platform) {
+    return platform !== Platform.Windows
+        || (platform !== Platform.Unknown && process.platform === 'linux');
```
gregg-miskelly: Should this check for osx also?

src/platform.ts
```diff
@@ -62,26 +62,35 @@ export function getSupportedPlatform() {
...
 
-    return SupportedPlatform.None;	
+    return Platform.Unknown;	
 }
+
+export function isWindows(platform: Platform) {
+    return platform === Platform.Windows;
+}
+
+export function isUnix(platform: Platform) {
```
gregg-miskelly: Is this function used? I am not seeing where it is.

src/omnisharp/server.ts
DustinCampbell: That's not the intention here. Essentially, there's already an option to specify a path to a specific OmniSharp -- `csharp.omnisharp`. I've added a `csharp.omnisharpUsesMono` option that controls whether the OmniSharp specified via `csharp.omnisharp` should be launched via Mono or not. For CoreCLR, the user just has to point `csharp.omnisharp` to the CoreCLR executable and it'll be launched.

gregg-miskelly:
Otherwise LGTM

src/platform.ts
DustinCampbell: I think I just need to delete this. It's not used anywhere.

src/platform.ts
DustinCampbell: Good eye again. :smile:

gulpfile.js
DustinCampbell: FWIW, you're  right that making this desktop kind of defeats the purpose of the offline installer for Windows. I filed #586 to track fixing this.

Issue #170, opened 2016-04-11
No tasks.json/launch.json created for root folder containing global.json


Pilchie:

shanselman:
FYI I did a workshop with 32 people in Germany this week and at least 10 folks hit this issue.

PR #590, opened 2016-07-19
Improve generation of tasks.json and launch.json assets

### LINKED ISSUE #556: VSCode should self-configure debug settings when added to ASP.NET Core projects
_From @bitcrazed on July 10, 2016 17:44_

- VSCode Version: 1.3.0
- OS Version: Windows 10

Steps to Reproduce:

1. Create ASP.NET Core project using either Visual Studio 2013 or via Yeoman ("yo aspnet") at the command-line
2. At the command-line, navigate to the root of the project (wherever project.json resides)
3. Run `dotnet restore`: Dependencies should be downloaded
4. Run `dotnet run`: Project should build & run
5. Run `code .`: VS Code should load and open the project folder
6. Open debug tab and hit "play button": VS Code displays an error that "launch.json must be configured"
saved image from tweetium

I [asked about this on Twitter](https://twitter.com/bitcrazed/status/751464295121063936) and received help from @jchannon, @shanselman and @damianedwards (thanks guys): Turns out that I had to manually determine the `target-framework` and `project-name` fields. I also had to [set the "debugType": "portable" in project.json](https://github.com/bitcrazed/HelloWorldCore/commit/7c34f86998dc53e0a0d4409e7e130f540890b0b0). (see [request to set this by default](https://github.com/aspnet/Templates/issues/650))

I would have expected that since VS Code injected `launch.json` into the project, and that it had access to `project.json` etc. that it'd be able to determine the target-framework automatically.

_Copied from original issue: Microsoft/vscode#9033_

DustinCampbell:
Fixes #556 and #170

This change allows tasks.json and launch.json to be generated when the workspace root contains a global.json rather than a launch.json. It ensures that appropriate values to the json files to account for the target project being more deeply nested.

cc @gregg-miskelly, @rajkumar42, @caslan, @chuckries

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -329,8 +329,17 @@ function hasWebServerDependency(targetProjectData: TargetProjectData): boolean {
     let projectJson = fs.readFileSync(targetProjectData.projectJsonPath, 'utf8');
     projectJson = projectJson.replace(/^\uFEFF/, '');
 
-    let projectJsonObject = JSON.parse(projectJson);
-    
+    let projectJsonObject: any;
+
+    try {
+        // TODO: This error should be surfaced to the user. If the JSON can't be parsed
+        // (maybe due to a syntax error like an extra comma), the user should be notified
+        // to fix up their project.json.
+        projectJsonObject = JSON.parse(projectJson);
+    } catch (error) {
+        projectJsonObject = null;
+    }
+
     if (projectJsonObject == null) {
         return false;
     }
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -372,17 +372,6 @@ export function addAssetsIfNecessary(server: OmnisharpServer) {
         return;
     }
     
-    // If there is no project.json or global.json, we won't bother to prompt the user for tasks.json.		
-    const projectJsonPath = path.join(vscode.workspace.rootPath, 'project.json');
-    const globalJsonPath = path.join(vscode.workspace.rootPath, 'global.json');
-
-    const projectJsonExists = fs.existsSync(projectJsonPath);
-    const globalJsonExists = fs.existsSync(globalJsonPath);
-
-    if (!fs.existsSync(projectJsonPath) && !fs.existsSync(globalJsonPath)) {
-        return;
-    }
-
     return serverUtils.requestWorkspaceInformation(server).then(info => {
         // If there are no .NET Core projects, we won't bother offering to add assets.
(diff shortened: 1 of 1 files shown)


src/assets.ts
```diff
@@ -202,48 +223,113 @@ function createLaunchJson(targetFramework: string, executableName: string, isWeb
...
+            }
+        });
+    });
+
+    return result;
+}
+
+function hasWebServerDependency(targetProjectData: TargetProjectData) {
+    if (!targetProjectData || !targetProjectData.projectJsonPath) {
+        return;
```
gregg-miskelly: should this be 'return false'? Or does the caller handle undefined?

src/assets.ts
```diff
@@ -202,48 +223,113 @@ function createLaunchJson(targetFramework: string, executableName: string, isWeb
...
+                    result.push(project);
+                }
+            }
+        });
+    });
+
+    return result;
+}
+
+function hasWebServerDependency(targetProjectData: TargetProjectData) {
```
gregg-miskelly: Should we define this to return a bool?

src/assets.ts
```diff
@@ -298,9 +363,14 @@ export function addAssetsIfNecessary(server: OmnisharpServer) {
         return;
     }
     
-    // If there is no project.json, we won't bother to prompt the user for tasks.json.		
-    const projectJsonPath = path.join(vscode.workspace.rootPath, 'project.json');		
-    if (!fs.existsSync(projectJsonPath)) {
+    // If there is no project.json or global.json, we won't bother to prompt the user for tasks.json.
```
gregg-miskelly: >global.json [](start = 38, length = 11)

Now that we know how to build a project lower down, do we need this check? Ex: if I have:

src
   MyExe
   MyLib

And I open the src folder, shouldn't we handle this case?

gregg-miskelly:
Otherwise LGTM

src/assets.ts
DustinCampbell: Good catch. Yes, it should return false.

src/assets.ts
DustinCampbell: Sure

src/assets.ts
DustinCampbell: Yeah, that seems reasonable. This should still work.

DustinCampbell:
Look better now @gregg-miskelly?

gregg-miskelly:
@DustinCampbell LGTM. Thanks!

Issue #587, opened 2016-07-19
Duplicate warnings


Souk21:
## Environment data
`dotnet --info` output: no
VS Code version: 1.3.1
C# Extension version: 1.3.0-beta1
Using a unity .sln

## Steps to reproduce

Open a project

## Expected  behavior

Warnings should show only once

## Actual behavior

![capture d ecran 2016-07-19 a 09 11 37](https://cloud.githubusercontent.com/assets/13599865/16941568/3799078e-4d91-11e6-80ed-c22f008e2f81.png)

This warning is showing multiple time for a single line

Edit: just realized it's a duplicate from https://github.com/OmniSharp/omnisharp-vscode/issues/447

PR #592, opened 2016-07-20
Fix several issues with diagnostics

### LINKED ISSUE #447: "Duplicated" errors & warnings
## Environment data
OS Version: Microsoft Windows [version 10.0.10586]
VS Code version: 1.2.1
C# Extension version: 1.0.11

## Steps to reproduce
- Write a code that cause an error/warning.

## Expected  behavior
- Don't duplicate (quintuplicate ? (and increasing...)) the errors/warning.

## Actual behavior

- (duplicating, triplicating and increasing....) the errors/warnings.

 

By typing `> omniSharp: restart` (Restarting the OmniSharp project)

### LINKED ISSUE #477: Debugger: Improve the error message when vscode attempt to attach to dotnet process with mismatched privilege
## Environment data

`dotnet --info` output:
VS Code version:1.2.1
C# Extension version:1.1.7
## Steps to reproduce
1. Create a sample .dotnet core console application with following code:
   
   ```
   public static void Main(string[] args)
   {
       Console.ReadLine();
       Console.WriteLine("Hello World!");
   }
   ```
2. Launch vscode with **non-admin** and open the project 
3. Launch command line with **admin**, and run the application "dotnet run"
4. Set the target as "attach" and configure the process id of "dotnet" 
5. F5
   ## Expected  behavior

Actually, when I did the testing, I was not aware of launching vscode with non-admin, I suppose this is not supported scenario, however I assume we should at least improve the error message,
1.  At least we should try to remove the internal error message "Unexpected CLRDBG output from command "-target-attach 8508". "  ? 
2. Is it possible to add some useful information similar like "vscode doesn't have the sufficient permission to attach the process" if the permission of vscode and attached process is mismatched ? 
## Actual behavior

Error: No process with the specified id is currently running.

(This is VSDBGE_E_PROCESS_ID_NOT_FOUND)

DustinCampbell:
Fixes #447  and #587

1. The diagnostics provider won't refresh diagnostics while the OmniSharp server is restoring packages. Instead, it waits until it receives an event from OmniSharp afterward. However, that event was only expecting the old DNX protocol, so the diagnostics provider never automatically updated after packages were restored. This change gets rid of the DNX protocol and changes the diagnostic provider to look for .NET Core projects.

2. Hidden diagnostics should have DiagnosticSeverity.Info rather DiagnosticSeverity.Warning.

3. When the processing *all* of the diagnostics (not just those for the current file), the previous diagnostics were not cleared out first. This resulted in new diagnostics being merged with old diagnostics, creating duplicates diagnostics. Now we clear the diagnostics for a file before adding new ones.

### DIFF
--- a/src/features/diagnosticsProvider.ts
+++ b/src/features/diagnosticsProvider.ts
@@ -43,8 +43,8 @@ export class Advisor {
 	}
 
 	private _onProjectChange(info: protocol.ProjectInformationResponse): void {
-		if (info.DnxProject && info.DnxProject.SourceFiles) {
-			this._projectSourceFileCounts[info.DnxProject.Path] = info.DnxProject.SourceFiles.length;
+		if (info.DotNetProject && info.DotNetProject.SourceFiles) {
+			this._projectSourceFileCounts[info.DotNetProject.Path] = info.DotNetProject.SourceFiles.length;
 		}
 		if (info.MsBuildProject && info.MsBuildProject.SourceFiles) {
 			this._projectSourceFileCounts[info.MsBuildProject.Path] = info.MsBuildProject.SourceFiles.length;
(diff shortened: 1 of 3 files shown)


DustinCampbell:
cc @gregg-miskelly, @chuckries, @caslan, @rajkumar42, though this is an area of the code base that you're probably not familiar with.

DustinCampbell:
Merging this now so I can produce a new beta release with this change to allow users to try it out.

Issue #579, opened 2016-07-18
Cannot download OmniSharp behind proxy/firewall


weinand:
_From @TheBlueSky on July 18, 2016 4:37_

Hello everyone,

I'm not sure if this is a VS Code or OmniSharp issue, so feel free to move it to the right repository if necessary.

When I open a .Net project, I get the following in Output window:

```
[INFO] Starting OmniSharp at 'c:\Users\xyz\Documents\My Web Sites\SomeProject'...
[INFO] Installing to C:\Users\xyz\.vscode\extensions\ms-vscode.csharp-1.2.2\.omnisharp
[INFO] Attempting to download omnisharp-1.9-beta11-win-x64-net451.zip...
```

And it gets stuck there. Nothing is downloaded and my project doesn't have any kind of Intellisense (obviously).

Maybe this is relevant, and maybe not, but at work, we are behind a firewall and a proxy server (from Fortinet) that does strange stuff. One of which is using its own certificate for all HTTPS requests, which I faced an issue with when using an application that require a valid  certificate signed by trusted CA, or an application that uses certificate pinning technique.
- VSCode Version: Code 1.3.1 (e6b4afa53e9c0f54edef1673de9001e9f0f547ae, 2016-07-12T13:35:06.227Z)
- OS Version: Windows_NT ia32 6.3.9600

Steps to Reproduce:
1. Go behind our proxy.
2. Open a .Net project :)

_Copied from original issue: Microsoft/vscode#9427_

DustinCampbell:
Hi @TheBlueSky, have specified your proxy settings within VS Code?

TheBlueSky:
No. The kind of firewall/proxy that we have in the company does not require specifying proxy settings. You come in the morning, login to the firewall/proxy, and that's all.

I started to face this issue only after v1.0 of .Net Core/ASP.Net Core (not sure what the corresponding OmniSharp or VS Code versions were). Haven't faced this issue during DNX days.

My memory isn't clear whether I tested it with ASP.Net Core RC2 or not; I just know it never worked after v1.0.

DustinCampbell:
The extension just uses your default proxy settings unless you specified something different. I assume other operations work? For example, `dotnet restore` on a new .NET Core project at the command line?

TheBlueSky:
[...]

PR #605, opened 2016-07-22
Display the full download URL

ltrzesniewski:
This is a very simple change which displays the *full* download URL, to relieve some pain caused by #579. 

I have the same proxy issue, and I'm updating OmniSharp every time by downloading the package manually. Knowing the full download URL obviously helps a lot with that.


gregg-miskelly:
@ltrzesniewski Thanks for the PR! This sounds reasonable to me. @DustinCampbell does this look good to you?

DustinCampbell:
Yup! This looks good to me.

PR #608, opened 2016-07-22
Fix OmniSharp server performance telemetry

DustinCampbell:
Telemetry for the delay times of OmniSharp server responses are now sent on a two-minute timer.


DustinCampbell:
cc @gregg-miskelly, @rajkumar42, @chuckries, @caslan. Would anyone like to take a look?

rajkumar42:
:shipit:

PR #620, opened 2016-07-25
Remove System.Diagnostics.StackTrace from debugger install

gregg-miskelly:
The debugger was installing System.Diagnostics.StackTrace which it doesn't appear to actually need and, in the dev install scenarios at least, was causing a problem with pulling in the wrong version of System.Reflection.Metadata. This removes it so the dev and shipping code pulls in the same set of assemblies.

Testing: Verified that if I create a .vsix the debugger install still succeeds and resulting debugger works for basic things. Also verified, with private OpenDebugAD7 bits, that exception stack traces still work.


gregg-miskelly:
@caslan please review

gregg-miskelly:
@rajkumar42 @caslan ping

rajkumar42:
:+1:

PR #670, opened 2016-08-17
Highlight "string[]", "string []", "string[,]", "string [,]", "string…

### LINKED ISSUE #657: Syntax highlighting incorrect for multi-dimensional arrays in parameter list.
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-preview2-003121)

Product Information:
 Version:            1.0.0-preview2-003121
 Commit SHA-1 hash:  1e9d529bc5

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```
VS Code version: 1.4.0
C# Extension version: 1.3.0

## Steps to reproduce
1. Use a multi-dimensional array (e.g. `int[,]`) as the type for an argument to a function.

## Expected behavior
The type and parameter name should be properly colored like other parameters.

## Actual behavior
vscode-syntaxhighlightingissue

Note that both the type and parameter name for `redFish` are in the default coloring.  I originally suspected the presence of a comma in a type might be the culprit, however the templated argument `ham` is rendered as expected.

ivanz:
Highlight "string[]", "string []", "string[,]", "string [,]", "string[ , ]", etc properly inside method parameter list.

Fixes #657

### Before:
before

### After:

after

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -502,7 +502,7 @@
       ]
     },
     "parameters": {
-      "begin": "\\b(ref|params|out)?\\s*\\b([\\w?\\[\\]]+(?:\\s*<.*?>)?)\\s+(\\w+)\\s*(=)?",
+      "begin": "\\b(ref|params|out)?\\s*\\b([\\w?]+(?:\\s*\\[\\s*,?\\s*\\])?(?:\\s*<.*?>)?)\\s+(\\w+)\\s*(=)?",
       "beginCaptures": {
         "1": {
           "name": "storage.type.modifier.cs"


DustinCampbell:
LGTM. Thanks @ivanz!

seraku24:
Hi, folks.  Not sure if this is the best place to comment, but I reviewed the updated regex and found that it is not complete and, unfortunately, regresses behavior.  Also, the original regex was incorrect with respect to C# syntax in a few ways, so the updated one inherited those flaws.

Here is a program I put together to test out two candidates for the regex:
```csharp
using System;
using System.Linq;
using System.Text.RegularExpressions;

namespace Seraku.SyntaxHighlightingRegexTest
{
    internal static class Program
    {
        private static void Main(string[] args) {
            var regexes = new [] {
                new Regex(@"\b(ref|params|out)?\s*\b(\w+(?:\s*<.*?>)?(?:\s*\?)?(?:\s*\[[\s,]*\])*)\s+(\w+)\s*(=)?"),
                new Regex(@"\b(ref|params|out)?\s*\b(\w+(?:\s*<.*?>)?(?:\s*\?)?(?:\s*\[.*?\])?)\s+(\w+)\s*(=)?"),
            } ;
            var types = "int|double?|IEnumerable<string>|Func<object, bool>|Foo<long>?|Outer<Inner<object>>".Split('|');
            var suffixes = "|[]|[][]|[,]|[,,]|[][,]|[,,][][]".Split('|');
            var query = from type in types from suffix in suffixes select $"{type}{suffix} name";
            foreach (var regex in regexes) {
                Console.Write($"{regex} -> ");
                if (query.All(x => regex.IsMatch(x))) {
                    Console.WriteLine("All match.");
                } else {
                    Console.WriteLine("Some failed to match:");
                    Console.WriteLine(string.Join("\n", query.Where(x => !regex.IsMatch(x))));
                }
            }
        }
    }
}
```

The regexes above both address the following issues:

1. Array specification must be the final element of a type.  (Current behavior matches invalid types like `Func[]<int>` but not valid types like `Func<int>[]`.)
2. Array specification may have any number of dimensions.  (Current behavior matches `int[,]` but not `int[,,]` or beyond.)
3. Array specification may include arrays of arrays.  (Old behavior matches `int[][]`, updated behavior regresses this, but neither support mixed array types like `int[][,]`.)
4. Nullable type specification `?` may appear after any non-nullable type, including templated value types.  (Current behavior matches invalid types like `Foo?<int>` but not valid types like `Foo<int>?`.)

The only difference between the two candidates is how the array syntax is handled.  The first regex is more syntactically correct with respect to array specification; whereas the second regex is much less picky with the arrays, similar to how template types are parsed.  But given that invalid syntax like `Func<:-)> smiley` is currently being highlighted, it is hardly a problem if `string[O_o] what` gets highlighted as a side effect.

ivanz:
@seraku24 Good catch. I suggest creating a pull request and re-opening the original issue linked in this pull request.

seraku24:
_(Off-topic)_ Whew.  First time forking, committing, and submitting a pull request.  Here is hoping I've crossed my I's and dotted by T's.  `(:`

Issue #101, opened 2016-03-17
Open C# colorizing issues


aeschli:
FYI I created the following bugs against https://github.com/atom/language-csharp from issues reported against the c# snytax highlighting in VS Code.
I will close the corresponding bugs in the VSCode repository that we kept open to track any progress.

https://github.com/atom/language-csharp/issues/48
https://github.com/atom/language-csharp/issues/46
https://github.com/atom/language-csharp/issues/43
https://github.com/atom/language-csharp/issues/44
https://github.com/atom/language-csharp/issues/45

ivanz:
@DustinCampbell FYI All but one of those are fixed and I have created a separate issue for it here #687 - so you can close this one. I have an idea how to fix the remaining one and will have a look when I get a chance.

DustinCampbell:
@ivanz, do you think it'd make sense to port your fixes over to the (https://github.com/atom/language-csharp) repo?

DustinCampbell:
cc @david-driscoll as well. I'm not sure if the OmniSharp extension for Atom uses the same grammar or not.

david-driscoll:
Yeah, we use the built in csharp grammar from atom.  The semantic highlighting is additive.

DustinCampbell:
Cool. So, porting these fixes to atom would be good for you too.

david-driscoll:
If you port them (which would be amazing!) feel free to cc me on the pull request.  :+1:

DustinCampbell:
All of these are fixed for the next release.

joshpritchard-fcx:
This release did not fix https://github.com/Microsoft/vscode/issues/848 so it should be re-opened. This is very annoying for me.

DustinCampbell:
@joshpritchard-fcx : Here's what the https://github.com/Microsoft/vscode/issues/848 repro looks like it the latest release of VS Code:

![image](https://user-images.githubusercontent.com/116161/35292276-da236fe8-0024-11e8-88a0-bfd0b6aa66fe.png)

This appears to be correct to me. If you disagree, please feel free to file a new issue at https://github.com/dotnet/csharp-tmLanguage/issues.

joshpritchard-fcx:
Thanks. Sorry for commenting here I guess what I am seeing is a separate issue.

Issue #33, opened 2016-02-09
Method return type syntax highlighting issues


aeschli:
_From @TheColorRed on February 6, 2016 3:18_

Theme: Colorful Default Themes (Dark)

There is still a small issue when syntax highlighting generators, as you can see from this screenshot:

![screenshot from 2016-02-05 21-15-55](https://cloud.githubusercontent.com/assets/1571806/12864356/b47a6542-cc4d-11e5-812b-117e9e6b036b.png)

When using a generator, everything becomes white. Also the class definition in the foreach is also white.

_Copied from original issue: Microsoft/vscode#2756_

aeschli:
@TheColorRed Thanks for reporting this. Can you paste the code snippet as well, to speed up the instigations?

aeschli:
_From @TheColorRed on February 9, 2016 16:48_

Here is the code:

```c#
public GameObject[] GetItems(string name){
    List<GameObject> items = new List<GameObject>();
    foreach(InventoryItem i in inventoryItems){
        if(i.itemName == name){
            items.Add(i.gameObject);
        }
    }
    return items.ToArray();
}

public GameObject[] GetItems<T>(){
    List<GameObject> items = new List<GameObject>();
    foreach(var i in inventoryItems){
        if(i.GetType() == typeof(T)){
            items.Add(i.gameObject);
        }
    }
    return items.ToArray();
}
```

**Note:** the type hinting in the foreach also isn't highlighted.

ivanz:
![image](https://cloud.githubusercontent.com/assets/79742/17788917/be45ea0a-6587-11e6-85b1-923d526b37c7.png)

ivanz:
@DustinCampbell Can you please label this with "Syntax highlighting", so that when I come back to looking at this I can potentially look at fixing it. Or maybe @seraku24 will beat me to it ;)

ivanz:
Also change the title to "Method return type syntax highlighting issues" or something 👍

PR #680, opened 2016-08-18
Improve parameter type matching regex

seraku24:
Fixes (again) #657 as well as issues brought up in #670.

- Corrects the order of type, template arguments, nullable type specification and array specifications.
- Ensures array specification support all forms (simple arrays, multi-dimensional arrays, arrays of arrays).

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -502,7 +502,7 @@
       ]
     },
     "parameters": {
-      "begin": "\\b(ref|params|out)?\\s*\\b(\\w+(?:\\s*<.*?>)?(?:\\s*\\?)?(?:\\s*\\[.*?\\])?)\\s+(\\w+)\\s*(=)?",
+      "begin": "\\b(ref|params|out)?\\s*\\b(\\w+(?:\\s*<.*?>)?(?:\\s*\\*)*(?:\\s*\\?)?(?:\\s*\\[.*?\\])?)\\s+(@?\\w+)\\s*(=)?",
       "beginCaptures": {
         "1": {
           "name": "storage.type.modifier.cs"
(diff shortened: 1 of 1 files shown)


seraku24:
And for comparison, here are the results of the regex on a pathological test case:

_Current behavior_
omnisharp-syntaxhighlighting-currentregex

_Updated behavior_
omnisharp-syntaxhighlighting-updatedregex

ivanz:
@seraku24  You may want to check if it's handling unsafe pointers (string* ,etc) too since you are at it. Relates to those two issues: #101 and https://github.com/Microsoft/vscode/issues/1093

ivanz:
@seraku24  Also one more thing is to handle @ too e.g. `string @class`

ivanz:
@seraku24  Finally might be worth checking wether your regex also fixes the method return value highlighting problem here: #33

seraku24:
Okay, @ivanz.  I have adjusted the regex to support unsafe pointers as well as reserved keyword escaping:

omnisharp-syntaxhighlighting-updatedregexpart2

Regarding return types, it seems that some types are highlighting correctly already but there are others that are not:

omnisharp-syntaxhighlighting-returntypes

That said, we might want to keep the scope of this PR to just addressing the issues with highlighting parameters, which I believe we have done.

ivanz:
Looks good to me, but @DustinCampbell is the one with the review and merge powers and I believe he is based in the US, so it's just about morning there. Congrats on your first GitHub pull request 🎉

seraku24:
Well, compared to other source/version control systems, git and GitHub have probably been the easiest to work with, _especially_ when compared to Product Studio or TFS.  `(:`

Thanks for the help and encouragement.  I am just glad to be able to offer some of my regex experience at the very least.

DustinCampbell:
LGTM

PR #682, opened 2016-08-18
Fix highlighting issues caused by escaped reserved keywords (e.g. @class)

### LINKED ISSUE #614: Keywords escaped with @ breaks highlighting
C# keywords can be used by prefixing the name with "@". But the code highlighting breaks in VS code 1.3.0 breaks when this is done.

Example:
Using `@class` throws off all highlighting for the rest of the file.

image

ivanz:
Fixes #614 and overall two issues:

1. Some escaped keywords were still getting highlighted as keywords (e.g. `@class` or `@namespace`)
2. Highlighting was breaking across the file when @class @namespace @interface @struct , etc was used.

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -28,7 +28,7 @@
       "end": "\\s*(?:$|;)"
     },
     "namespace": {
-      "begin": "^\\s*((namespace)\\s+([\\w.]+))",
+      "begin": "^\\s*[^@]((namespace)\\s+([\\w.]+))",
       "beginCaptures": {
         "1": {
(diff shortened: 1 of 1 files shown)


DustinCampbell:
This works for many of the reserved keywords. Is there a changed needed for predefined types (e.g. `@int`)?

ivanz:
Good point. Let me add those and rebase.

ivanz:
@DustinCampbell Added them and rebased.

DustinCampbell:
Do we not need the `[@]` before using anymore?

DustinCampbell:
Otherwise, LGTM. I'll go ahead and merge.

Issue #381, opened 2016-05-26
Wrong syntax highlight "using" keyword


ishvedov:
Insider 1.1.1
The "using" keyword highlight whole line. In VS and other editors only "using" clause highlighted.

PR #683, opened 2016-08-18
Fix highlighting of nested namespaces and using statements.

### LINKED ISSUE #282: Using is highlighted differently when adding inside and outside namespace.
VSCode version: 1.1.0-alpha
OS: Windows10

I observed that when adding using inside and outside namespace then VSCode is not highlighting using statement uniquely.
I verified the same code in Visual studio and visual studio behavior is consistent.

usingissue

ivanz:
Fixes #282, #849, #381

Also the below which are tracked here: #101

https://github.com/Microsoft/vscode/issues/849
https://github.com/Microsoft/vscode/issues/848

### Before

image

### After

image

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -8,16 +8,26 @@
   "foldingStopMarker": "^\\s*#\\s*endregion|^\\s*\\*/|^\\s*\\}",
   "patterns": [
     {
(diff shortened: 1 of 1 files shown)


DustinCampbell:
LGTM

PR #684, opened 2016-08-18
Update the runtime license with latest text from Microsoft Legal

gregg-miskelly:


gregg-miskelly:
@chuckries please review

gregg-miskelly:
@DustinCampbell could you give this a quick review as I think Chuck missed this?

DustinCampbell:
Seems fine to me. :+1:

chuckries:
:+1:

RuntimeLicenses/license.txt
```diff
@@ -9,7 +9,8 @@ except to the extent those have additional terms.
 IF YOU COMPLY WITH THESE LICENSE TERMS, YOU HAVE THE RIGHTS BELOW.
 
 1. INSTALLATION AND USE RIGHTS. 
-You may install and use any number of copies of the software to develop and test your applications. 
+You may only use the C# Extension for Visual Studio Code with Visual Studio Code, Visual Studio or Xamarin Studio software to
+help you develop and test your applications.
```
andschwa: This restriction doesn't make any sense. It's completely contradictory to the software's [MIT license](https://github.com/OmniSharp/omnisharp-vscode/blob/master/LICENSE.txt), which grants:

> Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so

Key part being rights to use without restriction; yet you're attempting to restrict that right in this second license.

gregg-miskelly: @andschwa let me start with the disclaimer that I am not a lawyer. But to answer your question, this extension has two licenses -- there is a license of the source code itself, which is MIT. There is also a license on the official download of the extension that Microsoft produces and uploads to the extension gallery. That is what this is. It is NOT MIT.

andschwa:
But what's the difference between the extension you download and the software as a whole (meaning the source code and the bits you compile from it)? This seems very... sketchy.

RuntimeLicenses/license.txt
andschwa: But what's the difference between the extension you download and the software as a whole (meaning the source code and the bits you compile from it)? This seems very... sketchy.

I am also not a lawyer, but some quick research leads me to believe they (source code and object code) must indeed be covered by the same license. It's a [Stack Exchange answer](https://opensource.stackexchange.com/questions/2954/do-licenses-actually-apply-to-the-source-code-or-product-itself/3959) but it's enough for me to say this needs to be looked into further.

PR #689, opened 2016-08-19
Fix return value syntax highlighting.

### LINKED ISSUE #33: Method return type syntax highlighting issues

seraku24:
Fixes #33.

omnisharp-syntaxhighlighting-returntypes-fixed

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -371,9 +371,6 @@
               "include": "#storage-modifiers"
             },
             {
-              "include": "#builtinTypes"
-            },
-            {
               "begin": "([\\w]+\\s*(?:<[\\w<>\\s,`?]*>)?)\\s*\\(",
               "beginCaptures": {
                 "1": {


DustinCampbell:
FWIW, I would argue that the delimiters should not be colored as part of the type. It's OK for now, but it's not how C# syntax highlighting works on VS. Also, built-in types should be highlighted as keywords. eventually, I'd like to get to parity with VS.

seraku24:
Based on what I was seeing, it looked like the intention was to colorize the entire type with a uniform color.  It should be possible to get something closer to VS, but I know that some of the existing regex is too simplistic.  And achieving complete parity with VS will probably never be completely possible using a regex-based system, as we really need the full parse tree to do it perfectly.

At this point, I think we really need some C# code and corresponding reference images showing exactly the coloring we are trying to achieve.  That would give us a much better yard stick to determine whether we are making progress, as opposed to waiting for individual issues to be filed.

If I started putting together a sample source file, where would be a good spot in the repository to check it in?  Would it mess up anything to have it sitting along side the `csharp.json`?

DustinCampbell:
I totally agree we'll never get it perfect, and I'm fine with where you're taking it in the interim. I think we'll want to make it better at some point. We inherited csharp.json from Atom, and I think it probably needs a rewrite (with tests! :smile:) in the near future to try and get closer to VS.

As I said, this is totally fine for now. Nice work!

seraku24:
Regarding testing, I see that several of the language extensions included in the main VSCode project have tests for syntax highlighting, so there is almost certainly some work there we can build upon.

@ivanz, as you are also doing work in this space, we should probably sync up at some point to coordinate our efforts, especially if a rewrite is deemed necessary to move forward.

ivanz:
I've opened #696 to track any future discussions on testing syntax highlighting

@seraku24 I have no plans to do more work on the syntax highlighting at this point, so no worries about that.

DustinCampbell:
FWIW, I marked #696 as being on the backlog.

PR #692, opened 2016-08-19
Re-enable the debug console window by default

gregg-miskelly:
When the VSCode team added the 'internalConsoleOptions' property to launch.json,
the default they picked was 'openOnFirstSessionStart'. This seems like the wrong default
for C# users as this hides all the messages we send. This changes our template to enable
'openOnSessionStart'.

For reference, here is the VSCode code to use this option --
https://github.com/Microsoft/vscode/blob/ac630b9b38f49fa7d1019cbb3b7ed36653be5240/src/vs/workbench/parts/debug/node/debugAdapter.ts#L113

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -17,6 +17,7 @@ interface DebugConfiguration {
     name: string,
     type: string,
     request: string,
+    internalConsoleOptions?: string,
     sourceFileMap?: any,
 }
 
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@rajkumar42 please review

rajkumar42:
:shipit:

Issue #603, opened 2016-07-22
Add mechanism to suppress 'Get .NET CLI tools' error message


gregg-miskelly:
## Steps to reproduce

Install the C# extension for the purposes of Unity development
Open a Unity Project
## Expected  behavior

No reference to the .NET CLI since you aren't working on a scenario involving .NET Core
## Possible fixes

The code for this is in [active.ts](https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/coreclr-debug/activate.ts#L30):

```
if (!isDotnetOnPath()) {
    const getDotNetMessage = "Get .NET CLI tools"; 
    vscode.window.showErrorMessage("The .NET CLI tools cannot be located. .NET Core debugging will not be enabled. Make sure .NET CLI tools are installed and are on the path.",
        getDotNetMessage).then(value => {
```

Two possible fixes:
1. Instead of downloading the debugger on C# extension activation, we download it when opening a .NET Core App or when they first try to start debugging.
2. Instead of using an error message, maybe we can use a different piece of VS Code UI that lets the user pick 'never show me this again'.

wesrupert:
PR #700 provides a mechanism to get the message to stop appearing. However, leaving this open to track implementing fix 1.

gregg-miskelly:
With this new download system, I don't think it makes sense to do fix 1 anymore. So closing this.

PR #700, opened 2016-08-23
Add suppressDotnetInstallWarning user setting

wesrupert:
- Add a user option to suppress the "Get .NET CLI tools" popup that appears when the tools are not found.
- Add a navigation point to the settings page to the popup.

Fix for #603.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
       }
     ],
     "configuration": {
-      "title": "Omnisharp configuration",
+      "title": "C# configuration",
       "properties": {
         "csharp.suppressDotnetInstallWarning": {
           "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -70,6 +70,16 @@
         "configuration": "./csharp.configuration.json"
       }
     ],
+    "configuration": {
+      "title": "Omnisharp configuration",
+      "properties": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -70,6 +70,16 @@
         "configuration": "./csharp.configuration.json"
       }
     ],
+    "configuration": {
+      "title": "Omnisharp configuration",
```
gregg-miskelly: Should this be 'C# configuration' instead since that is the name of this extension?

gregg-miskelly:
Otherwise LGTM

package.json
wesrupert: Yeah, I'll change it.

chuckries:
:+1:

caslan:
Can we automatically set the setting instead of having the user search for it in the file? That's how the other prompts behave in VS Code (i.e. Don't show again)

wesrupert:
@caslan, as far as I can tell the [extensions API](https://code.visualstudio.com/docs/extensionAPI/vscode-api#WorkspaceConfiguration) does not expose write access to user settings. Please let me know if I'm missing any information here.

caslan:
@wesrupert Yeah, thinking about this, the ones I know of are all part of VS Code itself.

PR #709, opened 2016-08-24
Remove dotnet-core nuget feed

chuckries:
System.Reflection.Metadata has been pulished to nuget.org so we no longer
need this feed dependency.


chuckries:
@gregg-miskelly

chuckries:
I validated that this works on Windows/Linux/Mac with empty NuGet package caches. There were no missing dependencies besides System.Reflection.Metadata.

Issue #674, opened 2016-08-17
Feature request: Better C# syntax colorizer


aeschli:
_From @nfantone on August 17, 2016 14:57_

On latest Insiders (`1.5.0-55168276e4c0b6ec2404ac20af04eeed5f5b1000`), C# syntax  highlighting look _very_ bland.

<img width="823" alt="screen shot 2016-08-17 at 11 39 26 am" src="https://cloud.githubusercontent.com/assets/476069/17740664/d36e14c0-646f-11e6-93a3-9523e0ca963f.png">

As you can see, almost 70% of the text is just plain white, with only a few keyword being blue. Tried different themes as well, and all of them colorize the same bits leaving what was white-ish in the default theme intact.

Am I doing something wrong? Is there an extension of sorts for enhancing this? I am new to C# in general, so perhaps this is to be expected? Or are there plans to support better syntax highlighting in the future?

_Copied from original issue: Microsoft/vscode#10619_

johanneskopf:
Same here with the latest 1.5.0 insider.
"Vector3", "Input" and "Screen" for example in my picture should definitely be colorized.

![csharp_syntaxhighlighting](https://cloud.githubusercontent.com/assets/13490371/17749831/b0231b64-64bf-11e6-9b9e-0e15948accb5.PNG)

aeschli:
@DustinCampbell You are correct, we only have syntax highlighting through text mate grammars. Some grammars go quite far as to also parse statements and expressions, but depending on the complexity of the language this is not feasible.

cwe1ss:
@nfantone @aeschli seems like this specific screenshot is from using the "Dark" theme instead of "Dark+" (which has nicer highlighting), isn't it? 

Themes can be changed in the command palette by typing "Preferences: Color Theme"

nfantone:
@cwe1ss While it's still has plenty of room for improvement,

 since `1.4.0` release, colorizer greatly improved. That screenshot was taken almost a month ago, now. Not entirely sure, but it should have been taken using `1.3.0`.

However, I do remember that switching themes at that moment did absolutely nothing but change the colors of what was _already colorized_ in the shown theme.

Currently, it looks something like this (using _Dark+_):

<img width="815" alt="screen shot 2016-09-09 at 2 56 44 pm" src="https://cloud.githubusercontent.com/assets/476069/18389387/aacfd14e-769d-11e6-8465-457d00f931fb.png">

[...]

PR #716, opened 2016-08-26
Syntax highlighting for field, variable and new expression types

### LINKED ISSUE #687: Types not highlighted in variable and field declarations
Copy from: https://github.com/atom/language-csharp/issues/48

I'm using the VSCode Version 0.10.6 (0.10.6).
Default dark theme.

Here's a sample code:
```
class MyClass
{

}

class Program
{
    static void Main()
    {        
        //string is highlighted
        string s = "MyClass";

        //MyClass is not highlighted
        MyClass myClass = new MyClass();
    }
}
```

Here's an image:
image

Here's a video demonstration:
https://www.youtube.com/watch?v=VpRPJc42FZQ

ivanz:
Effectively the result is this:

image

and in fact matches the TypeScript highlighting:

image

Thoughts?

Closes #687
Relates to #674

### Before

image

### After

image

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -209,6 +209,10 @@
           "include": "#constants"
         },
         {
+          "include": "#new-expression"
+        },
+
+        {
           "include": "#storage-modifiers"
         },
         {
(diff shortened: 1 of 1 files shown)


ivanz:
I think the blue makes it a bit too colourful. Here is a version with the variable/field name highlighting disabled:

image

ivanz:
Closing this, because it hit me that I am not fully happy with this and that it makes the code less readable without the semantic bit to be able to highlight types and variable usages as well. 😞

PR #720, opened 2016-08-26
Improve OmniSharp server logging

### LINKED ISSUE #560: Add datetime log in Omnisharp Log
## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (1.0.0-preview2-003121)

Product Information:
 Version:            1.0.0-preview2-003121
 Commit SHA-1 hash:  1e9d529bc5

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.11
 OS Platform: Darwin
 RID:         osx.10.11-x64

```
VS Code version: 1.3
C# Extension version: 1.2.2

## Steps to reproduce
Open a solution with Omnisharp

Change git branches

Restart Omnisharp

View Omnisharp Logs

## Expected  behavior

In Omnisharp Logs see a datetime against the log entries to see which solution loading it refers to

## Actual behavior

No datetime against log entries

DustinCampbell:
Fixes #560

Initially, I'd just set out to add a timestamp to the OmniSharp server log for #560, but ended up rationalizing logging during the server start up and download steps.

### DIFF
--- a/src/omnisharp/logger.ts
+++ b/src/omnisharp/logger.ts
@@ -9,8 +9,8 @@ export class Logger {
     private _writer: (message: string) => void;
     private _prefix: string;
 
-    private _indent: string = '    ';
     private _indentLevel: number = 0;
+    private _indentSize: number = 4;
     private _atLineStart: boolean = false;
 
     constructor(writer: (message: string) => void, prefix?: string) {
@@ -21,8 +21,7 @@ export class Logger {
     private _appendCore(message: string): void {
         if (this._atLineStart) {
             if (this._indentLevel > 0) {
-                const repeatCount = this._indentLevel;
-                const indent = this._indent.repeat(repeatCount);
+                const indent = " ".repeat(this._indentLevel * this._indentSize);
                 this._writer(indent);
             }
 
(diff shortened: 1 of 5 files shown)


src/omnisharp/logger.ts
```diff
@@ -0,0 +1,59 @@
...
+    private _atLineStart: boolean = false;
+
+    constructor(writer: (message: string) => void, prefix?: string) {
+        this._writer = writer;
+        this._prefix = prefix;
+    }
+
+    private _appendCore(message: string): void {
+        if (this._atLineStart) {
+            if (this._indentLevel > 0) {
```
gregg-miskelly: Shouldn't we really be taking a dependency on leftpad here :)

gregg-miskelly:
LGTM

src/omnisharp/logger.ts
DustinCampbell: Possibly. Seems like overkill just to repeat some characters. Left pad expects a length to be passed. I suppose I could use it to generate the indent string like so:

```JavaScript
leftPad('', this._indentLevel * 4, ' ');
```

Not sure it's worth it. :smile:

DustinCampbell: I could also do something like this:

```JavaScript
const indent = " ".repeat(this._indentLevel * 4);
```

gregg-miskelly: To be clear, this was supposed to just be a joke based on the fact that the removal of leftpad from NPM broke the internet.

DustinCampbell: I know. I was being ironic. Gosh! Why doesn't anyone understand me?

:smile::smile::smile::smile::smile:

Issue #427, opened 2016-06-12
OmniSharp Stop Working with Odd Multi-Byte characters


nabychan:
## Environment data
`dotnet --info` output:NET Command Line Tools (1.0.0-preview2-002864)/windows 10-10586-x64
VS Code version:1.20
C# Extension version:1.0.11

## Steps to reproduce
1. open a solution
2. type some code like below, with odd Chinese characters
``` C#
ExceptionHelper.ThrowIfTrue(userId==entity.uid, "uid", "不存在");
```
3. reopen the solution, it works(showing references, definitions ...)
4. in the same file, OmniSharp will not work when you type any code
## Expected  behavior
OmniSharp works well with any characters

## Actual behavior
OmniSharp stop working

DustinCampbell:
I suspect this will be addressed with https://github.com/OmniSharp/omnisharp-roslyn/pull/598.

Issue #140, opened 2016-04-05
why the tip of method and parameters is Chinese garbled


alexandrudima:
Moved Microsoft/vscode#4950:

From @mrpengfei 

- VSCode Version:0.10.11
- OS Version:windows 10

Steps to Reproduce:
why the tip of method and parameters is Chinese garbled. 

1. 
![qq 20160405161945](https://cloud.githubusercontent.com/assets/5754718/14276360/12949f8a-fb51-11e5-8fac-4a940491a161.png)

2.

DustinCampbell:
Could you provide some code that I can copy and paste as part of a repro?

mrpengfei:
Thanks for your replay,
I open the project which is created by  visual studio 2013(chinese simple)
[ConsoleApplication1.zip](https://github.com/OmniSharp/omnisharp-vscode/files/207671/ConsoleApplication1.zip)

DustinCampbell:
thanks! I'll take a look

DustinCampbell:
I believe this will be fixed by https://github.com/OmniSharp/omnisharp-roslyn/pull/598.

PR #726, opened 2016-08-27
Specify '--encoding utf-8' when launching OmniSharp server

### LINKED ISSUE #4: Unicode characters show as question marks in method documentation hover window
_From @ktos on November 23, 2015 11:23_

I have a few methods which documentation contains Unicode (UTF-8) characters. When showing documentation by hovering over a method I get garbage question mark characters instead in this window.

Looks like (when hovering over Jednorożec's constructor):

XML documentation for this method:

All files are saved as UTF-8.

_Copied from original issue: Microsoft/vscode#488_

DustinCampbell:
Fixes #4, #140, #427

This addresses issues when communicating with OmniSharp with multibyte characters. Note that this is only half of the fix. Due to a subtle issue in OmniSharp, this change currently only fixes the encoding for requests sent to OmniSharp but responses are still received garbled. I've submitted a PR to OmniSharp which fixes the other side of the problem (https://github.com/OmniSharp/omnisharp-roslyn/pull/628). Once I have a new build of OmniSharp in hand, I'll switch over to that which should fully fix the issues listed above.

image

Note: I decided to make the whitespace consist in server.ts as part of this change. Sorry about that! You can click the specific commit that fixes the issue to see what the fix looks like.

### DIFF
--- a/src/omnisharp/download.ts
+++ b/src/omnisharp/download.ts
@@ -24,7 +24,7 @@ import {Logger} from './logger';
 const decompress = require('decompress');
 
 const BaseDownloadUrl = 'https://omnisharpdownload.blob.core.windows.net/ext';
-const OmniSharpVersion = '1.9-beta13';
+const OmniSharpVersion = '1.9-beta14';
 
 tmp.setGracefulCleanup();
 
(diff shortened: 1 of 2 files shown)


DustinCampbell:
OmniSharp is updated. Pushing this change.

PR #737, opened 2016-08-30
fix to support legacy options

sixpindin:
package.json creates omnisharp.useMono and omnisharp.path with default values. This fixes a bug whereby legacy configuration is only used if there is no value for the new configuration, which is not possible, because the default values will always be present.


DustinCampbell:
Good catch! LGTM

PR #742, opened 2016-09-02
Added test framework and a couple of tests for for syntax highlighting

### LINKED ISSUE #696: Syntax highlighting - needs automated tests
The syntax highlighting is based on a regex-based grammar and because there aren't any automated tests to validate changes - it is easy to cause regressions without knowing.

ivanz:
Closes #696 

The idea here was to wrap `vscode-textmate` in an easy to consume interface and expose a quick way to check if certain token was picked up correctly at a certain line and column number in a given input.

Had to update travis.yml to handle node native dependencies (vscode-textmate pulls in a native regex lib) and also added chai.js and its definitions.

### Example

A minimal example looks like this:

```typescript
const input = `
namespace TestNamespace {
}`;

let tokens: Token[] = TokenizerUtil.tokenize(input);
tokens.should.contain(Tokens.NamespaceKeyword("namespace", /*line*/ 2, /*column*/ 1));
tokens.should.contain(Tokens.NamespaceIdentifier("TestNamespace", 2, 11));
```

To add a new token type `tokenizer.ts` needs to be updated with the token details (this avoids hardcoding token id strings across tests and doing string operations):

```typescript
export namespace Tokens {

   // <..snip...>

    export const NamespaceIdentifier = (text: string, line?: number, column?: number) =>
        createToken(text, "entity.name.type.namespace.cs", line, column);
}
```

_Unrelated note:_ It would be great to re-visit the typings story when TypeScript 2.0 comes out with the `@typings/*` npm support, because having to manually version typings and store them in a directory is less than ideal in the long run. Given TS 2.0 is round the corner there is also probably no point investing time in using something like the `typings` or `tsd` tools. Just my 2c :)


DustinCampbell:
This is very cool @ivanz! That said, I'm not thrilled that "npm install" will require several prerequisites to run to completion with this change. Could you document these? 

FWIW, we should consider the existing Roslyn colorizer tests here: https://github.com/dotnet/roslyn/tree/master/src/EditorFeatures/CSharpTest/Classification. I'm not sure how many we could get working properly with regex-based colorizer, but there's certainly some good tests to start from.

DustinCampbell:
I had to do the following to get `npm install` to work on my Windows machine:

1. Run `npm install --global --production windows-build-tools`
2. Modify my installation of VS 2015 to include the Windows 8.1 SDK
3. Update npm with `npm -g install npm@next`

After that, I could successfully do the following:

`npm install`
`npm run compile`
`npm test`

```
> npm test

> csharp@1.5.0-beta1 test C:\Projects\OmniSharp-VSCode
> mocha --timeout 15000 -u tdd ./out/test/**/*.tests.js



  Sanity Tests
    √ Boolean checks

  Grammar
    Class
      √ has a class keyword, a name and optional storage modifiers (79ms)

  Grammar
    Namespace
      √ has a namespace keyword and a name
      √ can be nested
      √ can contain using statements


  5 passing (108ms)
```

ivanz:
I forgot that those steps would be needed since I already had the VS C++ compiler and Python installed on my machine (I can't recall having ever had to install the windows build tools package that you mention). 😞 

It's not great to have to go through those extra hoops on Windows indeed. Maybe we can add the `windows-build-tools` package as a `peerDependency` given that it's a machine-global thing and as you say add something to the readme or a new `Building.md` or similar doc. I wonder if a Dockerfile for building and running tests on Linux will be of any help for those that don't want to install all the deps required by windows-build-tools...

I am on holiday and will be back in ~2 weeks when I will look at this again.

ivanz:
Hi,

I am back from holiday. I should probably make the windows node build tools package a `peerDependency`, because it's a windows-only global thing anyway. This and adding some docs about it is about the only thing I can do to make this less painful. What are you thoughts? (The only other thing I can think of is pulling out the grammar into its own npm package...)

At this point I am not willing personally to invest time in trying to port the Roslyn semantic-highlighting tests to TypeScript. The way I was thinking about this is that a good short-term win can be adding some tests for the most common regressions (things to do with method signatures seem to be one for example) rather than try to build out a full-blown test suit. Given that the grammar is becoming more and more complex (especially more so with the upcoming C# 7 features) it will be worth probably adding some tests for those in the future.

It's unfortunate that we don't even have a finger in the air estimation of when they will add the semantic highlighting api (https://github.com/Microsoft/vscode/issues/585 ) to know how much effort to put into this.

DustinCampbell:
All of that makes great sense to me @ivanz. I agree that moving the Roslyn semantic-highlighting tests to TypeScript would be wasted effort. I just wanted to point that there's a wealth of expectations about how C# expects to colorize code in those tests, and the syntax-based tests are separate from the semantic-based tests.

ivanz:
Rebased and added info in README

ivanz:
Sorry - forgot to say that I did not end up adding it, because I thought that non-Windows devs will start getting a warning about a Windows-specific thing. Thoughts?

DustinCampbell:
Yeah, I agree.

DustinCampbell:
Let's do it.

PR #751, opened 2016-09-07
Note to update GetClrDbg.sh in MIEngine.

rajkumar42:


rajkumar42:
@gregg-miskelly @caslan

gregg-miskelly:
LGTM

src/platform.ts
```diff
@@ -30,6 +30,7 @@ export function getCurrentPlatform() {
     else if (process.platform === 'linux') {
         // Get the text of /etc/os-release to discover which Linux distribution we're running on.
         // For details: https://www.freedesktop.org/software/systemd/man/os-release.html
+        // When any distro or version is added, please update GetClrDbg.sh in MIEngine
```
DustinCampbell: Where do we do that?

gregg-miskelly: https://github.com/Microsoft/MIEngine/blob/master/scripts/GetClrDbg.sh

Do we want to put some text here that makes it clear that we don't necessarily expect community members to do this? (just Microsoft employees)

PR #760, opened 2016-09-17
Update launch targets

DustinCampbell:


mikehaas763:
Mostly curious, but is this "launch target" something that is exposed in the vscode UI or configuration, or is it a behind the scenes detail of omnisharp?

I'm assuming it's the latter based on this comment "Represents the project or solution that OmniSharp is to be launched with" on the LaunchTarget interface at https://github.com/OmniSharp/omnisharp-vscode/pull/760/files#diff-0f649e373df5c5e30b19fa598a56b55bR24

DustinCampbell:
It's in the VS Code UI. You can click the OmniSharp icon at the bottom-right to bring up a list of launch targets to select.

image

PR #795, opened 2016-09-30
Rework asset generation prompt to all user to specify 'Don't Ask Again'

### LINKED ISSUE #635: Allow permanent dismissal of build/debug assets message
image

I have no desire to add the `.vscode` directory for this project, yet every time I open the folder the extension bugs me. I'd like a way to dismiss it permanently for this PC.

The project in question is a library (no entry point) with a set of tests: https://github.com/gwtw/csharp-data-structures
The message: https://github.com/OmniSharp/omnisharp-vscode/blob/6ecb840f6c9e0d90cf7f10625d0347e6ade6b0b2/src/assets.ts#L126

DustinCampbell:
Fixes #635

Next will be adding a command to generate/re-generate the assets. Otherwise, the user might get into a situation where they click "Don't Ask Again" and have no recourse.

cc @gregg-miskelly

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -76,15 +76,15 @@ interface Operations {
 
 function hasOperations(operations: Operations) {
     return operations.addLaunchJson ||
-           operations.updateTasksJson ||
-           operations.addLaunchJson;
+        operations.updateTasksJson ||
+        operations.addLaunchJson;
 }
 
 function getOperations() {
     const paths = getPaths();
 
     return getBuildOperations(paths.tasksJsonPath).then(operations =>
-           getLaunchOperations(paths.launchJsonPath, operations));
+        getLaunchOperations(paths.launchJsonPath, operations));
 }
 
 function getBuildOperations(tasksJsonPath: string) {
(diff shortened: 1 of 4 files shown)


gregg-miskelly:
:shipit:

src/assets.ts
```diff
@@ -428,4 +428,86 @@ export function addAssetsIfNecessary(server: OmnisharpServer): Promise<AddAssetR
         }).catch(err =>
             reject(err));
     });
+}
+
+function doAssetsExists(paths: Paths) {
```
gregg-miskelly: I might suggest we rename this method to 'doesAnyAssetExist' and/or add a comment that this returns true if we have a launch or tasks.json, since it isn't readily apparent from the name that this is doing a `||` rather than an `&&`.

src/assets.ts
```diff
@@ -428,4 +428,86 @@ export function addAssetsIfNecessary(server: OmnisharpServer): Promise<AddAssetR
...
+                resolve(true);
+            }
+        });
+
+    });
+}
+
+export function generateAssets(server: OmnisharpServer) {
+    serverUtils.requestWorkspaceInformation(server).then(info => {
+        if (info.DotNet && info.DotNet.Projects.length > 0) {
```
gregg-miskelly: If this is false, I wonder if we should output some sort of error about unable to find dotnet project.

DustinCampbell: Good point

src/assets.ts
DustinCampbell: Sounds good. Will do

gregg-miskelly:
LGTM

PR #798, opened 2016-10-03
Adding support for Remote .NET Attach and tsc fix on Windows

WardenGnaw:
**Feature:**
package.json adds listRemoteProcesses as a command, including debuggerPath into pipeTransport. The
command is registered into commands.ts. processPicker.ts has its process listing parser refactored
and now includes calling process listing to remote processes. Only works on Linux and OSX. Windows
remote attach will result in a rejected promise.

**Fix:**
Fixing tsc error TS1252 in platform.ts.
Moved the getValue function within getCurrentPlatform function outside of function scope. Added string array parameter to getValue function because lines was originally a local variable within the getCurrentPlatform function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -685,7 +685,7 @@
                   },
                   "debuggerPath" : {
                     "type" : "string",
-                    "description" : "The full path to the debugger",
+                    "description" : "The full path to the debugger on the target machine.",
                     "default" : "~/clrdbg/clrdbg"
                   },
                   "pipeEnv": {
--- a/package.json
+++ b/package.json
@@ -413,7 +413,7 @@
                   "pipeCwd": "${workspaceRoot}",
                   "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
                   "pipeArgs": [],
-                  "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
+                  "debuggerPath" : "enter the path for the debugger on the target machine, for example ~/clrdbg/clrdbg"
                 },
                 "properties": {
                   "pipeCwd": {
@@ -436,7 +436,7 @@
                   },
                   "debuggerPath" : {
                     "type" : "string",
-                    "description" : "The full path to the debugger",
+                    "description" : "The full path to the debugger on the target machine.",
                     "default" : "~/clrdbg/clrdbg"
                   },
                   "pipeEnv": {
@@ -453,8 +453,7 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
-                      "pipeArgs": [],
-                      "debuggerPath" : "Please enter the debugger path here, for example c:\\clrdbg\\clrdbg"
+                      "pipeArgs": []
                     },
                     "properties": {
                       "pipeCwd": {
@@ -475,11 +474,6 @@
                         },
                         "default": []
                       },
-                      "debuggerPath" : {
-                        "type" : "string",
-                        "description" : "The full path to the debugger",
-                        "default" : "c:/clrdbg/clrdbg"
-                      },
                       "pipeEnv": {
                         "type": "object",
                         "additionalProperties": {
@@ -496,8 +490,7 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
-                      "pipeArgs": [],
-                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
+                      "pipeArgs": []
                     },
                     "properties": {
                       "pipeCwd": {
@@ -517,11 +510,6 @@
                           "type": "string"
                         },
                         "default": []
-                      },
-                      "debuggerPath" : {
-                        "type" : "string",
-                        "description" : "The full path to the debugger",
-                        "default" : "~/clrdbg/clrdbg"
                       },
                       "pipeEnv": {
                         "type": "object",
@@ -539,8 +527,7 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
-                      "pipeArgs": [],
-                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
+                      "pipeArgs": []
                     },
                     "properties": {
                       "pipeCwd": {
@@ -560,11 +547,6 @@
                           "type": "string"
                         },
                         "default": []
-                      },
-                      "debuggerPath" : {
-                        "type" : "string",
-                        "description" : "The full path to the debugger",
-                        "default" : "~/clrdbg/clrdbg"
                       },
                       "pipeEnv": {
                         "type": "object",
@@ -672,6 +654,9 @@
               },
               "pipeTransport": {
                 "type": "object",
+                "required": [
+                    "debuggerPath"
+                ],
                 "description": "When present, this tells the debugger to connect to a remote computer using another executable as a pipe that will relay standard input/output between VS Code and the .NET Core debugger backend executable (clrdbg).",
                 "default": {
                   "pipeCwd": "${workspaceRoot}",
@@ -717,8 +702,7 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
-                      "pipeArgs": [],
-                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
+                      "pipeArgs": []
                     },
                     "properties": {
                       "pipeCwd": {
@@ -739,11 +723,6 @@
                         },
                         "default": []
                       },
-                      "debuggerPath" : {
-                        "type" : "string",
-                        "description" : "The full path to the debugger",
-                        "default" : "~/clrdbg/clrdbg"
-                      },
                       "pipeEnv": {
                         "type": "object",
                         "additionalProperties": {
@@ -760,8 +739,7 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
-                      "pipeArgs": [],
-                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
+                      "pipeArgs": []
                     },
                     "properties": {
                       "pipeCwd": {
@@ -781,11 +759,6 @@
                           "type": "string"
                         },
                         "default": []
-                      },
-                      "debuggerPath" : {
-                        "type" : "string",
-                        "description" : "The full path to the debugger",
-                        "default" : "~/clrdbg/clrdbg"
                       },
                       "pipeEnv": {
                         "type": "object",
@@ -803,8 +776,7 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
-                      "pipeArgs": [],
-                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
+                      "pipeArgs": []
                     },
                     "properties": {
                       "pipeCwd": {
@@ -824,11 +796,6 @@
                           "type": "string"
                         },
                         "default": []
-                      },
-                      "debuggerPath" : {
-                        "type" : "string",
-                        "description" : "The full path to the debugger",
-                        "default" : "~/clrdbg/clrdbg"
                       },
                       "pipeEnv": {
                         "type": "object",
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "onCommand:dotnet.restore",
     "onCommand:csharp.downloadDebugger",
     "onCommand:csharp.listProcess",
+    "onCommand:csharp.listRemoteProcess",
     "workspaceContains:project.json"
   ],
   "contributes": {
@@ -153,6 +154,11 @@
         "command": "csharp.listProcess",
         "title": "List process for attach",
         "category": "CSharp"
+      },
+      {
+        "command" : "csharp.listRemoteProcess",
+        "title" : "List processes on remote connection for attach",
+        "category": "CSharp"
       }
     ],
     "keybindings": [
@@ -193,12 +199,11 @@
             "razor"
           ]
         },
-        "runtime": "node",
         "runtimeArgs": [],
         "variables": {
-          "pickProcess": "csharp.listProcess"
+          "pickProcess": "csharp.listProcess",
+          "pickRemoteProcess": "csharp.listRemoteProcess"
         },
-        "program": "./out/src/coreclr-debug/proxy.js",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
         "configurationAttributes": {
           "launch": {
@@ -405,7 +410,8 @@
                 "default": {
                   "pipeCwd": "${workspaceRoot}",
                   "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
-                  "pipeArgs": []
+                  "pipeArgs": [],
+                  "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
                 },
                 "properties": {
                   "pipeCwd": {
@@ -426,6 +432,11 @@
                     },
                     "default": []
                   },
+                  "debuggerPath" : {
+                    "type" : "string",
+                    "description" : "The full path to the debugger",
+                    "default" : "~/clrdbg/clrdbg"
+                  },
                   "pipeEnv": {
                     "type": "object",
                     "additionalProperties": {
@@ -440,7 +451,8 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
-                      "pipeArgs": []
+                      "pipeArgs": [],
+                      "debuggerPath" : "Please enter the debugger path here, for example c:\\clrdbg\\clrdbg"
                     },
                     "properties": {
                       "pipeCwd": {
@@ -461,6 +473,11 @@
                         },
                         "default": []
                       },
+                      "debuggerPath" : {
+                        "type" : "string",
+                        "description" : "The full path to the debugger",
+                        "default" : "c:/clrdbg/clrdbg"
+                      },
                       "pipeEnv": {
                         "type": "object",
                         "additionalProperties": {
@@ -477,7 +494,8 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
-                      "pipeArgs": []
+                      "pipeArgs": [],
+                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
                     },
                     "properties": {
                       "pipeCwd": {
@@ -497,6 +515,11 @@
                           "type": "string"
                         },
                         "default": []
+                      },
+                      "debuggerPath" : {
+                        "type" : "string",
+                        "description" : "The full path to the debugger",
+                        "default" : "~/clrdbg/clrdbg"
                       },
                       "pipeEnv": {
                         "type": "object",
@@ -514,7 +537,8 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
-                      "pipeArgs": []
+                      "pipeArgs": [],
+                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
                     },
                     "properties": {
                       "pipeCwd": {
@@ -534,6 +558,11 @@
                           "type": "string"
                         },
                         "default": []
+                      },
+                      "debuggerPath" : {
+                        "type" : "string",
+                        "description" : "The full path to the debugger",
+                        "default" : "~/clrdbg/clrdbg"
                       },
                       "pipeEnv": {
                         "type": "object",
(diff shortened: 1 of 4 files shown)


src/features/processPicker.ts
```diff
@@ -35,6 +37,80 @@ export class AttachPicker {
     }
 }
 
+export class RemoteAttachPicker {
+    public static ShowAttachEntries(args : any): Promise<string> {
+        // Grab selected name from UI
+        var name : string = args.name;
```
gregg-miskelly: Do we need to error out if `name` is undefined, not a string, or the empty string?

src/features/processPicker.ts
```diff
@@ -35,6 +37,80 @@ export class AttachPicker {
...
+        // Find correct pipeTransport via selected name
+        var config; 
+        for (var i = 0; i < json.configurations.length; ++i) {
+            if (json.configurations[i].name === name) {
+                config = json.configurations[i];
+                break; 
+            }
+        }
+
+        if (!config.pipeTransport) {
```
gregg-miskelly: Error if we couldn't find the right config?

src/features/processPicker.ts
```diff
@@ -35,6 +37,80 @@ export class AttachPicker {
...
+            // Missing PipeTransport, prompt if user wanted to just do local attach?
+            return new Promise<string>((resolve, reject) => {
+                reject(new Error("Configuration \"" + args.name + "\" in launch.json does not have a " + 
+                "pipeTransport argument for pickRemoteProcess. Use pickProcess for local attach."));
+            });
+        } else {
+            var pipeCmd : string = config.pipeTransport.pipeProgram + " " + config.pipeTransport.pipeArgs[0];
```
gregg-miskelly: You want to join the arguments together, not just take the 0th entry from the array.

src/features/processPicker.ts
```diff
@@ -35,6 +37,80 @@ export class AttachPicker {
...
+            // Clean string of newlines
+            var cleanOutput : string = output.replace(/[\r\n]+/g, '').toLowerCase();
+
+            switch(cleanOutput) {
+                case "darwin":
+                case "linux":
+                    return cleanOutput;
+                // Failure case. TODO: test for windows machine
+                default:
+                    return new Promise<string>((resolve, reject) => {
```
gregg-miskelly: You are going to need an error string here.

src/features/processPicker.ts
```diff
@@ -64,28 +140,47 @@ abstract class DotNetAttachItemsProvider implements AttachItemsProvider {
...
+
 export class PsAttachItemsProvider extends DotNetAttachItemsProvider {
+    protected getInternalProcessEntries(): Promise<Process[]> {
+        const commColumnTitle = Array(PsOutputParser.secondColumnCharacters).join("a");
+        // the BSD version of ps uses '-c' to have 'comm' only output the executable name and not
+        // the full path. The Linux version of ps has 'comm' to only display the name of the executable
+        // Note that comm on Linux systems is truncated to 16 characters:
+        // https://bugzilla.redhat.com/show_bug.cgi?id=429565
+        // Since 'args' contains the full path to the executable, even if truncated, searching will work as desired.
+        const psCommand = `ps -axww -o pid=,comm=${commColumnTitle},args=` + (os.platform() === 'darwin' ? ' -c' : '');
```
gregg-miskelly: as.platform() is going to tell you the host platform, you want the target platform

src/features/processPicker.ts
```diff
@@ -35,6 +37,80 @@ export class AttachPicker {
...
+
+        if (!config.pipeTransport) {
+            // Missing PipeTransport, prompt if user wanted to just do local attach?
+            return new Promise<string>((resolve, reject) => {
+                reject(new Error("Configuration \"" + args.name + "\" in launch.json does not have a " + 
+                "pipeTransport argument for pickRemoteProcess. Use pickProcess for local attach."));
+            });
+        } else {
+            var pipeCmd : string = config.pipeTransport.pipeProgram + " " + config.pipeTransport.pipeArgs[0];
+            return RemoteAttachPicker.getRemoteOS(pipeCmd).then(remoteOS => {
```
gregg-miskelly: At least for some pipes (ex: ssh) you are going to get way better perf if you can execute the pipe once and have it return both the OS arch and the PS output. This might turn out to be too complicated, but it is probably worth a try. One thing you could do is optimize for Linux and just require multiple round trips on OSX. Something like:

uname && [ $(uname) == "Linux" ] && ps ...

src/features/processPicker.ts
```diff
@@ -246,4 +333,16 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
             resolve(stdout);
         });
     });
+}
+
+function execChildProcessWithTimeout(process: string, workingDirectory: string): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
+        child_process.exec(process, { timeout: 3000, cwd: workingDirectory, maxBuffer: 500 * 1024 }, (error: Error, stdout: string, stderr: string) => {
```
gregg-miskelly: 3 seconds sounds much too small

gregg-miskelly: Is 'maxBuffer' the maximum number of characters to hold? If so, do we need to specify a max length?

src/features/processPicker.ts
```diff
@@ -246,4 +333,16 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+            if (stderr && stderr.length > 0) {
+                reject(stderr);
```
gregg-miskelly: We may well need to prepend this with some text for this to be comprehensible. Something like 'Pipe transport program bla failed.\n\n<message>'

src/platform.ts
```diff
@@ -34,33 +56,12 @@ export function getCurrentPlatform() {
         const text = child_process.execSync('cat /etc/os-release').toString();
         const lines = text.split('\n');
 
-        function getValue(name: string) {
```
gregg-miskelly: Is this change because TypeScript wasn't working for you? This should probably go in as a separate PR.

src/features/processPicker.ts
WardenGnaw: processPicker.ts has the code for both local and remote attach listing. getInternalProcessEntries is used for local attach.

src/features/processPicker.ts
WardenGnaw: What would be the recommended timeout for waiting on a remote command?

chuckries:
Please run 'gulp tslint' and fix any linting issues in the files you are touching. Many of your 'var's should be 'let's.

src/features/processPicker.ts
WardenGnaw: I am not sure. I used execChildProcess as an example of calling exec on a child process.

gregg-miskelly: Will this command be async? Is there a way to cancel it? Its always nice to not have timeouts at all, as it is always hard to pick a good one. If you need one, I would probably suggest picking something pretty long since this involves connecting to a remote machine, which could be fairly slow. Maybe 30 seconds?

rajkumar42:
@edmunoz is added to the review. #Closed

package.json
```diff
@@ -638,6 +669,178 @@
...
+                    "default": {}
+                  },
+                  "windows": {
+                    "type": "object",
+                    "description": "Windows-specific pipe launch configuration options",
+                    "default": {
+                      "pipeCwd": "${workspaceRoot}",
+                      "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
+                      "pipeArgs": [],
+                      "debuggerPath" : "enter the path for the debugger, for example ~/clrdbg/clrdbg"
```
gregg-miskelly: Thinking about this some more, there is no reason why 'debuggerPath' would be dependent on the host OS. So I would suggest removing it as a property from within the 'windows', 'osx', and 'linux' elements.

package.json
```diff
@@ -638,6 +669,178 @@
                     "default": false
                   }
                 }
+              },
+              "pipeTransport": {
+                "type": "object",
```
gregg-miskelly: In the attach case, I think we should make 'debuggerPath' a required property.

package.json
```diff
@@ -440,7 +453,8 @@
                     "default": {
                       "pipeCwd": "${workspaceRoot}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
-                      "pipeArgs": []
+                      "pipeArgs": [],
+                      "debuggerPath" : "Please enter the debugger path here, for example c:\\clrdbg\\clrdbg"
```
gregg-miskelly: Same suggestion as in attach: I think we don't want the OS-specific 'debuggerPath' properties.

package.json
```diff
@@ -426,6 +434,11 @@
                     },
                     "default": []
                   },
+                  "debuggerPath" : {
+                    "type" : "string",
+                    "description" : "The full path to the debugger",
```
gregg-miskelly: I would suggest adding to the end of this description 'on the target computer' (or similar if you don't like that description)

src/features/processPicker.ts
```diff
@@ -35,6 +37,111 @@ export class AttachPicker {
...
+            });
+        }
+
+        if (!config.pipeTransport) {
+            // Missing PipeTransport, prompt if user wanted to just do local attach.
+            return new Promise<string>((resolve, reject) => {
+                reject(new Error("Configuration \"" + args.name + "\" in launch.json does not have a " + 
+                "pipeTransport argument for pickRemoteProcess. Use pickProcess for local attach."));
+            });
+        } else {
```
gregg-miskelly: I think I would suggest verifying that the 'debuggerPath' property is present. Since if the user has accidentally included the command line to execute the debugger in the pipeArgs, this is going to do pretty confusing things.

src/features/processPicker.ts
```diff
@@ -35,6 +37,111 @@ export class AttachPicker {
...
+            let pipeCmd : string = config.pipeTransport.pipeProgram + " " + config.pipeTransport.pipeArgs.join(" ");
```
gregg-miskelly: Both pipeArgs and pipeProgram can be overwritten with the host-OS-specific options. So you need to check if those properties exist, and if so, give preference to them. See OpenDebugAD7 for an example.

src/features/processPicker.ts
```diff
@@ -35,6 +37,111 @@ export class AttachPicker {
...
+                return new Promise<AttachItem[]>((resolve, reject) => {
+                    reject(new Error("Pipe transport failed to get OS and processes."));
+                });
+            }
+            else {
+                let remoteOS = lines[0].replace(/[\r\n]+/g, '').toLowerCase();
+                
+                if (remoteOS != "linux" && remoteOS != "darwin") {
+                    return new Promise<AttachItem[]>((resolve, reject) => {
+                        reject(new Error("Pipe Transport attach could not determine os. " + remoteOS));
```
gregg-miskelly: Or the operating system is not supported.

src/features/processPicker.ts
```diff
@@ -35,6 +37,111 @@ export class AttachPicker {
...
+            // Processess will follow if listed
+            let lines = output.split(os.EOL);
+            
+            if (lines.length == 0) {
+                return new Promise<AttachItem[]>((resolve, reject) => {
+                    reject(new Error("Pipe transport failed to get OS and processes."));
+                });
+            }
+            else {
+                let remoteOS = lines[0].replace(/[\r\n]+/g, '').toLowerCase();
```
gregg-miskelly: Might want to NOT convert this to lower case since the bash script doesn't.

src/features/processPicker.ts
```diff
@@ -35,6 +37,111 @@ export class AttachPicker {
...
+                
+                if (remoteOS != "linux" && remoteOS != "darwin") {
+                    return new Promise<AttachItem[]>((resolve, reject) => {
+                        reject(new Error("Pipe Transport attach could not determine os. " + remoteOS));
+                    }); 
+                }
+                
+                // Only got OS from uname
+                if (lines.length == 1) {
+                    return RemoteAttachPicker.getRemoteProcesses(pipeCmd, remoteOS);
```
gregg-miskelly: Shouldn't this really just be an error?

package.json
gregg-miskelly: Attach with pipeTransport. There is a 'required' property that you should be able to add here to do that.

package.json
```diff
@@ -153,6 +154,11 @@
         "command": "csharp.listProcess",
         "title": "List process for attach",
         "category": "CSharp"
+      },
+      {
+        "command" : "csharp.listRemoteProcess",
+        "title" : "List processes on remote connection for attach",
+        "category": "CSharp"
```
DustinCampbell: I notice that this command and "csharp.listProcess" have a category of "CSharp", but chsarp.downloadDebugger" has a category of "Debug". Is that intentional? Also, should the category be "C#"?

src/features/commands.ts
```diff
@@ -36,8 +36,9 @@ export default function registerCommands(server: OmnisharpServer, extensionPath:
     let attachItemsProvider = DotNetAttachItemsProviderFactory.Get();
     let attacher = new AttachPicker(attachItemsProvider);
     let d8 = vscode.commands.registerCommand('csharp.listProcess', () => attacher.ShowAttachEntries());
+    let d9 = vscode.commands.registerCommand('csharp.listRemoteProcess', (args) => RemoteAttachPicker.ShowAttachEntries(args));
 
-    return vscode.Disposable.from(d1, d2, d3, d4, d5, d6, d7, d8);
+    return vscode.Disposable.from(d1, d2, d3, d4, d5, d6, d7, d8, d9);
```
DustinCampbell: This will conflict with #795. First one merged wins! :smile:

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+export class RemoteAttachPicker {
+    public static ShowAttachEntries(args : any): Promise<string> {
+        // Grab selected name from UI
+        let name : string = args.name;
+
+        if (!name) {
+            // Config name not found. 
+            return new Promise<string>((resolve, reject) => {
+                reject(new Error("Name not defined in current configuration."));
+            });
```
DustinCampbell: 1. This could be written simpler as

```JS
return Promise.reject<string>(new Error("Name not defined in current configuration"));
```

DustinCampbell: 2. Should the error message mention the name that was passed in?

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+            if (json.configurations[configIdx].name === name) {
+                config = json.configurations[configIdx];
+                break; 
+            }
+        }
+
+        if (configIdx == json.configurations.length) {
+            // Name not found in list of given configurations. 
+            return new Promise<string>((resolve, reject) => {
+                reject(new Error("Could not find configuration that matches given name"));
```
DustinCampbell: Similar feedback to my comments above.

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+            });
+        }
+
+        if (!config.pipeTransport || !config.pipeTransport.debuggerPath) {
+            // Missing PipeTransport and debuggerPath, prompt if user wanted to just do local attach.
+            return new Promise<string>((resolve, reject) => {
+                reject(new Error("Configuration \"" + args.name + "\" in launch.json does not have a " + 
+                "pipeTransport argument with debuggerPath for pickRemoteProcess. Use pickProcess for local attach."));
+            });
```
DustinCampbell: Same here too.

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+        } else if (osPlatform == "linux" && config.pipeTransport.linux) {
+            return config.pipeTransport.linux;
+        } else if (osPlatform == "win32" && config.pipeTransport.windows) {
+            return config.pipeTransport.windows;
+        }
+        
+        return null;
+    }
+    
+    public static getRemoteOSAndProcesses(pipeCmd :string) : Promise<AttachItem[]> {
```
DustinCampbell: spacing looks wrong here. Might be worth a <kbd>Shift+Alt+F</kbd> in VS Code.

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+        const commColumnTitle = Array(PsOutputParser.secondColumnCharacters).join("a");
+        // Commands to get OS and processes
+        const linuxPsCommand = `ps -axww -o pid=,comm=${commColumnTitle},args=`;
+        const osxPsCommand = `ps -axww -o pid=,comm=${commColumnTitle},args= -c`;
+        const command = `uname && if [ $(uname) == "Linux" ] ; then ` + linuxPsCommand + `; elif [ $(uname) == "Darwin" ] ; ` + 
+         `then ` + osxPsCommand + `; fi`;
```
DustinCampbell: Why use `` string literals, but then not use placeholders?

```TypeScript
const command = `uname && if [ $(uname) == "Linux" ] ; then ${linuxPsCommand}; elif [ $(uname) == "Darwin" ] ; then ${osxPsCommand}; fi`;
```

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+
+        return execChildProcess(pipeCmd + ' "' + command + '"', null).then(output => {
```
DustinCampbell: Good spot for a `` string literal.

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+                    reject(new Error("Pipe transport failed to get OS and processes."));
+                });
+            }
+            else {
+                let remoteOS = lines[0].replace(/[\r\n]+/g, '');
+                
+                if (remoteOS != "Linux" && remoteOS != "Darwin") {
+                    return new Promise<AttachItem[]>((resolve, reject) => {
+                        reject(new Error("Could not determine operating system or operating system not supported. " + remoteOS));
+                    });
```
DustinCampbell:
```TypeScript
return Promise.reject<AttackItem[]>(new Error(`Could not determine operating system or operating system not supported. ${remoteOS}`));
```

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+                    return new Promise<AttachItem[]>((resolve, reject) => {
+                        reject(new Error("Could not determine operating system or operating system not supported. " + remoteOS));
+                    }); 
+                }
+                
+                // Only got OS from uname
+                if (lines.length == 1) {
+                    return new Promise<AttachItem[]>((resolve, reject) => {
+                        reject(new Error("Transport attach could not obtain processes list.")); 
+                    });
```
DustinCampbell: Same as above. You can just return `Promise.reject`.

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+                    let processes = lines.slice(1);
+                    return sortProcessEntries(PsOutputParser.parseProcessFromPsArray(processes), remoteOS);
+                }
+            }
+        });
+    }
+
+    public static getRemoteProcesses(pipeCmd: string, os: string) : Promise<AttachItem[]> {
+        const commColumnTitle = Array(PsOutputParser.secondColumnCharacters).join("a");
+        const psCommand = `ps -axww -o pid=,comm=${commColumnTitle},args=` + (os === 'darwin' ? ' -c' : '');
```
DustinCampbell: Also could be:

```TypeScript
const psCommand = `ps -axww -o pid=,comm=${commColumnTitle},args=${(os === 'darwin' ? ' -c' : '')}`;
```

src/features/processPicker.ts
```diff
@@ -35,6 +37,136 @@ export class AttachPicker {
...
+
+        return execChildProcess(pipeCmd + ' "' + command + '"', null).then(output => {
+            // OS will be on first line
+            // Processess will follow if listed
+            let lines = output.split(os.EOL);
+            
+            if (lines.length == 0) {
+                return new Promise<AttachItem[]>((resolve, reject) => {
+                    reject(new Error("Pipe transport failed to get OS and processes."));
+                });
```
DustinCampbell: Another spot to return `Promise.reject`.

src/features/processPicker.ts
```diff
@@ -64,28 +196,47 @@ abstract class DotNetAttachItemsProvider implements AttachItemsProvider {
...
+
 export class PsAttachItemsProvider extends DotNetAttachItemsProvider {
+    protected getInternalProcessEntries(): Promise<Process[]> {
+        const commColumnTitle = Array(PsOutputParser.secondColumnCharacters).join("a");
+        // the BSD version of ps uses '-c' to have 'comm' only output the executable name and not
+        // the full path. The Linux version of ps has 'comm' to only display the name of the executable
+        // Note that comm on Linux systems is truncated to 16 characters:
+        // https://bugzilla.redhat.com/show_bug.cgi?id=429565
+        // Since 'args' contains the full path to the executable, even if truncated, searching will work as desired.
+        const psCommand = `ps -axww -o pid=,comm=${commColumnTitle},args=` + (os.platform() === 'darwin' ? ' -c' : '');
```
DustinCampbell: Is this code copied and pasted? Could it be shared?

src/features/processPicker.ts
WardenGnaw: This is the case where name is null.

package.json
WardenGnaw: This I am not to sure of, @rajkumar42 or @gregg-miskelly

src/features/processPicker.ts
DustinCampbell: Ah of course. I was reading quickly. :smile:

package.json
gregg-miskelly: I don't know what `category` is supposed to be for. You can see if the schema will tell you. @chuckries created the downladDebugger command, so he might know. Otherwise you can look at the VS Code source to see if that has the answer (http://github.com/Microsoft/VSCode)

src/features/processPicker.ts
```diff
@@ -35,6 +37,125 @@ export class AttachPicker {
...
+
+        return null;
+    }
+
+    public static getRemoteOSAndProcesses(pipeCmd: string): Promise<AttachItem[]> {
+        // Commands to get OS and processes
+        const command = `uname && if [ $(uname) == "Linux" ] ; then ${RemoteAttachPicker.linuxPsCommand} ; elif [ $(uname) == "Darwin" ] ; ` +
+            `then ${RemoteAttachPicker.osxPsCommand}; fi`;
+
+        return execChildProcess(`${pipeCmd} "${command}"`, null).then(output => {
```
gregg-miskelly: I added this comment earlier, but to a spot where it got lost -- currently we will just reject the promise if the command returns stderror text. This might be confusing now that we are using execChildProcess for the pipe program. I would suggest doing something like putting in the wrong ip address in your `plink` or `ssh` command line and see if the experience is compressible.

src/features/processPicker.ts
```diff
@@ -35,6 +37,125 @@ export class AttachPicker {
...
+            let lines = output.split(os.EOL);
+
+            if (lines.length == 0) {
+                return Promise.reject<AttachItem[]>(new Error("Pipe transport failed to get OS and processes."));
+            }
+            else {
+                let remoteOS = lines[0].replace(/[\r\n]+/g, '');
+
+                if (remoteOS != "Linux" && remoteOS != "Darwin") {
+                    return Promise.reject<AttachItem[]>(new Error("Could not determine operating system " +
```
gregg-miskelly: Doesn't this always been the OS is unsupported?

src/platform.ts
WardenGnaw: Done in #801

package.json
DustinCampbell: `category` appears in the Command Palette:

image

src/features/processPicker.ts
```diff
@@ -246,4 +418,27 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+        child_process.exec(process, { cwd: workingDirectory, maxBuffer: 500 * 1024 }, (error: Error, stdout: string, stderr: string) => {
+            if (error) {
+                reject(new Error("See remote-attach output."));
+                channel.append(error.message);
+                channel.show();
+                return;
+            }
+
+            if (stderr && stderr.length > 0) {
+                reject(new Error("See remote-attach output"));
```
WardenGnaw: I'm confused about how to do this. I thought this would be part of the callback function after it finishes the exec.

src/features/commands.ts
WardenGnaw: I lost. 😢

src/features/processPicker.ts
```diff
@@ -246,4 +418,27 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+    return new Promise<string>((resolve, reject) => {
+        child_process.exec(process, { cwd: workingDirectory, maxBuffer: 500 * 1024 }, (error: Error, stdout: string, stderr: string) => {
+            if (error) {
+                reject(new Error("See remote-attach output."));
+                channel.append(error.message);
+                channel.show();
+                return;
+            }
+
+            if (stderr && stderr.length > 0) {
```
WardenGnaw: Something like this? Using the channel as the buffer.
```
            if (stdout && stdout.length > 0) {
                channel.appendLine('stdout:');
                channel.append(stdout);
            }

            if (stderr && stderr.length > 0) {
                channel.appendLine('stderr:');
                channel.append(stderr);
            }

            if (error) {
                channel.appendLine('Error message:');
                channel.append(error.message);
            }


            if (error || (stderr && stderr.length > 0)) {
                reject(new Error("See remote-attach output"));
                channel.show();
                return;
            }

            resolve(stdout);
```

gregg-miskelly: Yes, except that you probably don't want to send it to the channel in the success case. So I would suggest keeping an internal string buffer until there is an error. You probably don't need the stderr/stdout prefix -- if something is an error or normal output is usually apparent to humans by reading the text.

src/features/processPicker.ts
gregg-miskelly: I think you just keep state as to if you saw an error and then you can reject the promise at the end if there was an error.

package.json
gregg-miskelly: Looking at https://github.com/Microsoft/vscode-node-debug/search?utf8=%E2%9C%93&q=pickNodeProcess, I wonder if there is a way to avoid declaring this command entirely since it shouldn't actually show up in the command well.

gregg-miskelly:
This PR has been replaced with https://github.com/OmniSharp/omnisharp-vscode/pull/918

asyschikov:
@WardenGnaw Is there any documentation on remote debugging, including how to use pipeTransport?

gregg-miskelly:
@asyschikov I wrote up https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes. Let me know if you need more information/find problems.

asyschikov:
@gregg-miskelly looks good thanks. Is debugging (locally) over SSH advantageous in any way?

gregg-miskelly:
@asyschikov no - I wouldn't expect it to make a big difference, but it would add a bit of extra latency in every interaction between the UI and the backend.

PR #801, opened 2016-10-06
Fixing tsc error TS1252 in platform.ts

WardenGnaw:
I am getting an error message when I run `tsc` after` npm i`.

**Error Message:**
> src/platform.ts(37,18): error TS1252: Function declarations are not allowed inside blocks in strict mode when targeting 'ES3' or 'ES5'. Modules are automatically in strict mode.

**Fix:**
Moved the getValue function within getCurrentPlatform function outside
of function scope. Added string array parameter to getValue function
because lines was originally a local variable within the getCurrentPlatform function.

**Versions:**
tsc version: 2.0.3 (installed via `npm install -g typescript`, also tried 1.5.3 but there are more errors)
npm version: 3.10.3
node version: 6.7.0
Windows version: Windows 10 Enterprise

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -20,6 +20,28 @@ export enum Platform {
     Ubuntu16
 }
 
+function getValue(name: string, lines: string[]) {
+    for (let line of lines) {
+        line = line.trim();
+        if (line.startsWith(name)) {
+            const equalsIndex = line.indexOf('=');
+            if (equalsIndex >= 0) {
+                let value = line.substring(equalsIndex + 1);
+
+                // Strip double quotes if necessary
+                if (value.length > 1 && value.startsWith('"') && value.endsWith('"')) {
+                    value = value.substring(1, value.length - 1);
+                }
+
+                return value;
+            }
+        }
+    }
+
+    return undefined;
+}
+
+
 export function getCurrentPlatform() {
     if (process.platform === 'win32') {
         return Platform.Windows;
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
BTW: this is the TypeScript change: https://github.com/Microsoft/TypeScript/issues/7122

I _think_ we could maybe also define this as a lambda instead?
@DustinCampbell any thoughts?

DustinCampbell:
I'm fine with this change. I was planning to update us to TypeScript 2.0.3 soon anyway. :smile:

src/platform.ts
```diff
@@ -20,6 +20,28 @@ export enum Platform {
     Ubuntu16
 }
 
+function getValue(name: string, lines: string[]) {
```
chuckries: Now that this function is not in getCurrentPlatform, you might want to give it a better name as getValue is not very descriptive.

src/platform.ts
```diff
@@ -20,25 +20,29 @@ export enum Platform {
...
-                let value = line.substring(equalsIndex + 1);
-
-                // Strip double quotes if necessary
-                if (value.length > 1 && value.startsWith('"') && value.endsWith('"')) {
-                    value = value.substring(1, value.length - 1);
-                }
 
-                return value;
+        let equalsIndex = line.indexOf('=');
+        let key = line.substring(0, equalsIndex);
```
gregg-miskelly: Shouldn't this line be within the if?

gregg-miskelly:
Otherwise LGTM

DustinCampbell:
Same here. Looks fine.

PR #807, opened 2016-10-10
Don't immediately return promise within newly created promise

### LINKED ISSUE #804: Omnisharp is stuck at starting phase with nothing happening
This is probable more of an omnisharp error, but I post the problem here you guys probably know better what to do with this.

## Environment data
`dotnet --info` output:

> .NET Command Line Tools (1.0.0-preview2-003131)
> 
> Product Information:
>  Version:            1.0.0-preview2-003131
>  Commit SHA-1 hash:  635cf40e58
> 
> Runtime Environment:
>  OS Name:     ubuntu
>  OS Version:  16.04
>  OS Platform: Linux
>  RID:         ubuntu.16.04-x64

VS Code version: 1.5.3
C# Extension version: 1.4.1
## Steps to reproduce
Try to install on linux with mono flavour and a mono version <4.0.1 (or maybe no mono installed)

## Expected  behavior
An error should be displayed

> Cannot start Omnisharp because Mono version >=4.0.1 is required.

## Actual behavior
No error is displayed, the omnisharp server just hang on startup stage

> Starting OmniSharp server at 10/8/2016, 7:02:23 PM
>     Target: /blah/blah/blah/blah.sln

## How to fix
in out/omnisharp/launcher.js change 

```javascript
// Original code
function launchNixMono(details) {
    return new Promise(function (resolve, reject) {
        return canLaunchMono().then(function () {
            var args = details.args.slice(0); // create copy of details.args
            args.unshift(details.serverPath);
            var process = child_process_1.spawn('mono', args, {
                detached: false,
                cwd: details.cwd
            });
            return resolve({
                process: process,
                command: details.serverPath,
                usingMono: true
            });
        });
    });
}
```

to

```javascript
// Fixed code
function launchNixMono(details) {
    return new Promise(function (resolve, reject) {
        canLaunchMono().then(function () {
            var args = details.args.slice(0); // create copy of details.args
            args.unshift(details.serverPath);
            var process = child_process_1.spawn('mono', args, {
                detached: false,
                cwd: details.cwd
            });
            resolve({
                process: process,
                command: details.serverPath,
                usingMono: true
            });
        }, reject);
    });
}
```
Why are they returning a promise in a promise? I think this is not how es6 Promise works, the callback of the new Promise thing should not return anything. The whole code return things inside its callback it would probably be better to rewrite them.

DustinCampbell:
Fixes #804

### DIFF
--- a/src/omnisharp/launcher.ts
+++ b/src/omnisharp/launcher.ts
@@ -161,21 +161,15 @@ export interface LaunchResult {
 
 export function launchOmniSharp(details: LaunchDetails): Promise<LaunchResult> {
     return new Promise<LaunchResult>((resolve, reject) => {
-        try {
-            launch(details).then(result => {
-                // async error - when target not not ENEOT
-                result.process.on('error', reject);
+        launch(details).then(result => {
+            // async error - when target not not ENEOT
+            result.process.on('error', reject);
 
-                // success after a short freeing event loop
-                setTimeout(function () {
-                    resolve(result);
-                }, 0);
-            }, err => {
-                reject(err);
-            });
-        } catch (err) {
-            reject(err);
-        }
+            // success after a short freeing event loop
+            setTimeout(function () {
+                resolve(result);
+            }, 0);
+        });
     });
 }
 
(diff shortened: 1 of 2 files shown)


src/omnisharp/launcher.ts
```diff
@@ -161,21 +161,15 @@ export interface LaunchResult {
...
-        }
+        launch(details).then(result => {
+            // async error - when target not not ENEOT
+            result.process.on('error', reject);
+
+            // success after a short freeing event loop
+            setTimeout(function () {
+                resolve(result);
+            }, 0);
+        });
```
narf: Is it normal that rejection of launch promise is not handled here?

src/omnisharp/launcher.ts
```diff
@@ -189,34 +183,31 @@ function launch(details: LaunchDetails): Promise<LaunchResult> {
...
-        return resolve({
-            process,
-            command: details.serverPath,
-            usingMono: false
-        });
+    return Promise.resolve({
+        process,
+        command: details.serverPath,
+        usingMono: false
     });
```
narf: Why not putting all the code in a Promise like previously?
Any error there will end up in a throw which is not really what a function returning a promise should do.

src/omnisharp/launcher.ts
DustinCampbell: It's intended that rejection will be propagated to the caller, which is in server.ts [here](https://github.com/OmniSharp/omnisharp-vscode/pull/807/files#diff-f08c23c62ed01212f1e17a9530885a70R302).

src/omnisharp/launcher.ts
DustinCampbell: Any exceptions should ultimately be propagated to the caller in server.ts. The `catch()` there handles all rejections and exceptions.

src/omnisharp/launcher.ts
narf: If there is no reject callback AND the promise is not returned anywhere, I'm quite sure that the error will be silent.

This should be the correct way to bubble the error to the caller:
```
launch(details).then(result => {
    // async error - when target not not ENEOT
    result.process.on('error', reject);

    // success after a short freeing event loop
    setTimeout(function () {
        resolve(result);
    }, 0);
}).catch(reject);
```

PR #822, opened 2016-10-14
Fix debugger proxy

gregg-miskelly:
The previous change to the directory structure broke the debugger proxy code (but we never noticed since it isn't used much). This fixes it.


gregg-miskelly:
@chuckries please review

PR #826, opened 2016-10-17
Show full URL for downloading file.

slawomirbrys:
When using corporate proxy then extension fails to download files inside. In Visual Studio Code into Output we can see file name only and extension hangs. To help investigation what file to download manually and finish extension init we need to know full url which we should download from. This commit shows full url only.

### DIFF
--- a/src/omnisharp/download.ts
+++ b/src/omnisharp/download.ts
@@ -119,7 +119,7 @@ export function go(flavor: Flavor, platform: Platform, logger: Logger, proxy?: s
 
         const urlString = `${BaseDownloadUrl}/${fileName}`;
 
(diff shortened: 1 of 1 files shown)


slawomirbrys:
Strange, but #605 was fixing that, merged, however don't see that into code.

slawomirbrys:
Ok, this commit overwritten #605.
https://github.com/OmniSharp/omnisharp-vscode/commit/833ea9f5d801fb6fa006a7c9c99ac56dd5241222#diff-56b7f4eb8ed677ea3fa70a22fab72dcaR122

DustinCampbell:
Thanks very much! It looks like I stepped on that commit during some refactoring.

PR #841, opened 2016-10-26
Cleanup platform and Linux distribution detection logic

DustinCampbell:


DustinCampbell:
cc @gregg-miskelly

DustinCampbell:
cc @rajkumar42, @gregg-miskelly, @caslan

src/platform.ts
```diff
@@ -20,96 +95,126 @@ export enum Platform {
...
+                distributionPromise = Promise.resolve(null);
+                break;
 
-            ret.set(key, value);
+            case 'linux':
+                operatingSystem = OperatingSystem.Linux; 
+                architecturePromise = PlatformInformation.GetUnixArchitecture();
+                distributionPromise = LinuxDistribution.GetCurrent();
+                break;
         }
```
gregg-miskelly: Should we do something if it is none of these?

DustinCampbell: Yes, we should probably throw. The only other possible values are `aix`, `freebsd`, `openbsd` or `sunos` , which VS Code doesn't really support either.

Issue #777, opened 2016-09-22
Investigate using Mono-based OmniSharp exclusively


DustinCampbell:

DustinCampbell:
This investigation is now complete. OmniSharp will now run on an embedded Mono runtime on macOS and Linux.

PR #846, opened 2016-10-28
Re-work acquisition experience

DustinCampbell:
Related to #777.

This is a pretty significant change that moves away from using .NET Core-based OmniSharp servers for macOS/Linux. Instead, the C# extension will target the same .NET 4.6 version of OmniSharp, regardless of operating system. On Windows, OmniSharp will run on the desktop .NET Framework. On macOS/Linux, the extension acquires an embedded Mono runtime and framework to run OmniSharp.

A lot of the new code below (notably the PackageManager) is taken from the C/C++ extension for VS Code. This completely replaces the existing download and unzip logic that was already in the C# extension. In addition, I've simplified a lot of the launching and installation code in the OmniSharpServer.

I still have a lot of testing to do for this change yet, but I wanted to get it out for code review.

cc @gregg-miskelly, @caslan, @rajkumar42


package.json
```diff
@@ -198,12 +269,10 @@
             "razor"
           ]
         },
-        "runtime": "node",
```
chuckries: The changes from here down need to be undone as this breaks the debugger proxy work. the debugger proxy allows us to run the debugger in the same session that the extension is installed. We do this by rewriting the manifest on installation, so unfortunately this will change every time you f5 during development of the extension.

DustinCampbell: Oops! I missed that. Thanks!

package.json
```diff
@@ -50,6 +51,76 @@
     "chai": "^3.5.0",
     "vscode-textmate": "^2.1.1"
   },
+  "runtimeDependencies": [
```
DustinCampbell: The package manager does not (yet) allow packages to be specified for particular runtime IDs. This would need to be added in order to enable this scenario. I'll look at adding this.

DustinCampbell: OK. This should work now. You should be able to specify an array of `runtimeIds` for a particular runtime dependency in the package.json file and the PackageManager will download and install it when the current RID matches. A couple of caveats:

1. You package must be a .zip file
2. The .zip file will be extracted in the extension root path. So, any directory structure must be included in the .zip.
3. `gulpfile.js` will need to be updated once the debugger moves its acquisition to the package manager.
4. We need to add telemetry to the package manager.

DustinCampbell: I've added an `installPath` property to make item 2 easier. Essentially, this is combined with the extension root path to compute the install location.

DustinCampbell:
OK. Since it seems there's a strong desire to use the package manager before I'm ready to deliver the embedded Mono OmniSharp, I've reverted the packages back to the old .NET Core-based downloads. I need to upload these packages again for the Linux distros before OmniSharp will work on OSX/Linux. However, the package system *should* be ready for the debugger to use. Please review so that I can get this merged in.

DustinCampbell:
OK. This is no longer about embedded Mono, and is instead all about handling runtime dependencies. I'll close this and submit a new PR specifically for that.

gulpfile.js
```diff
@@ -131,28 +137,31 @@ gulp.task('package:online', ['clean'], () => {
...
-    packages.push({ rid: 'osx.10.11-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.OSX }] });
-    packages.push({ rid: 'centos.7-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.CentOS }] });
-    packages.push({ rid: 'debian.8-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.Debian }] });
-    packages.push({ rid: 'fedora.23-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.Fedora }] });
-    packages.push({ rid: 'opensuse.13.2-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.OpenSUSE }] });
-    packages.push({ rid: 'rhel.7.2-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.RHEL }] });
-    packages.push({ rid: 'ubuntu.14.04-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.Ubuntu14 }] });
-    packages.push({ rid: 'ubuntu.16.04-x64', omnisharps: [{ flavor: Flavor.CoreCLR, coreClrFlavor: CoreClrFlavor.Ubuntu16 }] });
+    packages.push({ rid: 'win7-x64', platform: new PlatformInformation('win32') });
```
DustinCampbell: in that case, I'm not sure there's a lot of value. I don't think anyone is really using this offline installers anyway.

PR #850, opened 2016-10-28
Update debugger packages for 10/28

gregg-miskelly:
Update the debugger packages to the version we expect to go out with 1.5.


gregg-miskelly:
I forgot to mention -- I smoke tested these on Ubuntu 14, OSX, and Windows.
The new packages are uploaded to nuget.org.

PR #862, opened 2016-11-02
Add package management system for runtime dependency acquisition

DustinCampbell:
This change adds a package management system (based on the one used by the C/C++ VS Code extension) for acquiring dependencies at runtime.

Runtime dependencies can be defined in the package.json using various criteria:

1. runtime ID (RID)
2. platform - `'win32'`, `'darwin'`, or `'linux'`
3. architecture - `'32-bit'` and `'64-bit'` on Windows or `'x86'` and `'x86_64'` on OSX/Linux

On activation, the extension will try to acquire runtime dependencies if an 'install.lock' file is missing. Any dependency whose criteria matches the current platform will be downloaded and installed.

TODO:

- [ ] Telemetry. The plan is to send the same telemetry that coreclr-debug does during its current acquisition. We should probably do that when the debugger moves over to this new acquisition model.
- [ ] Offline packaging. This is just work I need to do and it should be done shortly.

cc @caslan, @chuckries, @gregg-miskelly, @rajkumar42


DustinCampbell:
@chuckries: Thanks for taking a look. Should I optimistically check this in so that you can start integrating the debugger bits with it?

chuckries:
I think that would be fine. I took a look at all of the new package manager changes, the the gulpfile changes, and the new common stuff. I didn't look to closely at the new code under the omnisharp directory.

DustinCampbell:
OK. I'll go ahead and merge. I've been testing quite a bit on various distros, so hopefully there won't be too much fall out.

package.json
```diff
@@ -50,6 +51,117 @@
     "chai": "^3.5.0",
     "vscode-textmate": "^2.1.1"
   },
+  "runtimeDependencies": [
```
chuckries: It looks like this pulls in both omnisharp-coreclr and omnisharp-desktop on windows. Is this the desired behavior?

DustinCampbell: It is what it is. :smile: Before this change, acquisition of OmniSharp was deferred until the server launched. So, we could acquire just the necessary OmniSharp server at that time. Now that all acquisition happens up front, both servers get acquired at once.

In the near, I'll be removing the CoreCLR version of OmniSharp completely. So, this is a point in time problem.

chuckries: sounds good.

DustinCampbell:
Offline packaging is fixed with https://github.com/OmniSharp/omnisharp-vscode/pull/864.

PR #867, opened 2016-11-03
Specify binaries relative to installPath

DustinCampbell:


DustinCampbell:
cc @chuckries, @gregg-miskelly 

This should make it slightly easier to specify packages that you want to be installed into a specific folder within the extension root and mark binaries within it as executable.

PR #873, opened 2016-11-04
Reimplement debugger acquisition to use Package Manager

chuckries:
1. Remove all code related to calling dotnet restore/dotnet publish
2. Remove coreclr-debug directory
3. Add debugger runtime packages to package manifest
4. Add an install.Begin file to the package manager so we can detect that
it is in progress
5. Rework the debugger proxy so that it understands the package manager
6. Add knowledge of the dotnet cli on the path to the debugger proxy

TODO:
1. Lots and lots of testing
2. Offline package creation


chuckries:
@gregg-miskelly @wesrupert @DustinCampbell 

I need to test this all over the place and offline package creation is still incoming.

package.json
```diff
@@ -190,6 +190,110 @@
...
+    {
+      "description": ".NET Core Debugger (Windows / x64)",
+      "url": "https://vsdebugger.azureedge.net/coreclr-debug-1-5-0/coreclr-debug-win7-x64.zip",
+      "installPath": ".debugger",
+      "runtimeIds": [
+        "win7-x64"
+      ]
+    },
+    {
+      "description": ".NET Core Debugger (OSX / x64)",
```
DustinCampbell: "macOS"?

src/common.ts
```diff
@@ -21,10 +21,6 @@ export function getExtensionPath() {
     return extensionPath;
 }
 
-export function getBinPath() {
-    return path.resolve(getExtensionPath(), "bin");
-}
-
```
DustinCampbell: I'll be using this again in my next PR.

src/common.ts
```diff
@@ -60,19 +56,37 @@ export function fileExists(filePath: string): Promise<boolean> {
...
-    return fileExists(getInstallLockFilePath());
+export function installFileExists(type: InstallFileType): Promise<boolean> {
+    return fileExists(getInstallFilePath(type));
+}
+
+export function touchInstallFile(type: InstallFileType): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+        fs.writeFile(getInstallFilePath(type), '', err => {
+            if (err) {
+                reject(err);
```
DustinCampbell: I'd return here so that it doesn't carry on to the resolve below.

src/coreclr-debug/install.ts
```diff
@@ -2,12 +2,10 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-'use strict';
 
 import { CoreClrDebugUtil } from './util';
 import * as fs from 'fs';
 import * as path from 'path';
-import * as fs_extra from 'fs-extra-promise';
```
DustinCampbell: is this the last reference? Can we remove the dependency from package.json?

src/coreclr-debug/util.ts
```diff
@@ -123,23 +72,67 @@ export class CoreClrDebugUtil
...
+                dotnetError.ErrorMessage = 'The .NET CLI tools on the path are too old. .NET Core debugging will not be enabled. The minimum supported version is ' + MINIMUM_SUPPORTED_DOTNET_CLI + '.';
+                dotnetError.ErrorString = "dotnet cli is too old";
+                throw dotnetError;
+            }
+
+            return dotnetInfo;
+        });
+    }
+
     public spawnChildProcess(process: string, args: string[], workingDirectory: string, onStdout?: (data: Buffer) => void, onStderr?: (data: Buffer) => void): Promise<void> {
```
DustinCampbell: Does `spawnChildProcess` in common.ts work for this?

src/coreclr-debug/activate.ts
```diff
@@ -7,126 +7,50 @@
...
-                }
-            });
-            statusBarMessage.dispose();
-
-            installStage = error.installStage;
-            installError = error.errorMessage;
-            moreErrors = error.hasMoreErrors ? 'true' : 'false';
-        }).then(() => {
-            // log telemetry and delete install begin file
-            logTelemetry('Acquisition', {
```
DustinCampbell: Should we move this into `main.activate()`?

src/common.ts
chuckries: I'll put it back.

package.json
chuckries: I was basing it off omnisharp: "description": "OmniSharp (.NET Core - OSX / x64)",
I'll change it in both places.

src/main.ts
```diff
@@ -90,7 +99,11 @@ function installRuntimeDependencies(extension: vscode.Extension<any>): Promise<v
         })
         .then(() => {
             installationStage = 'touchLockFile';
-            return util.touchLockFile();
+            return util.touchInstallFile(util.InstallFileType.Lock);
+        })
+        .then(() => {
+            installationStage = 'deleteBeginFile';
+            return util.deleteInstallFile(util.InstallFileType.Begin)
```
gregg-miskelly: Should this be in the catch block (or using the temp file stuff) so that we delete the begin file no matter what?

src/coreclr-debug/activate.ts
```diff
@@ -7,126 +7,50 @@
...
-                dotnetVersion: dotnetInfo.Version,
-                osVersion: dotnetInfo.OsVersion,
-                osRID: dotnetInfo.RuntimeId
+    _logger = logger;
+
+    if (!CoreClrDebugUtil.existsSync(_util.debugAdapterDir())) {
+        // We have been activated but it looks like our package was not installed. This is bad.
+        logger.appendLine("[ERROR]: C# Extension failed to install the debugger package");
+        showInstallErrorMessage();
+    } else if (!CoreClrDebugUtil.existsSync(_util.installCompleteFilePath())) {
```
gregg-miskelly: Since we also have an install complete file for the package manager, I wonder if we should rename some part of this to make the distinction more clear (ex: debugUtil instead of util, or debuggerInstallCompleteFilePath...)

src/coreclr-debug/util.ts
```diff
@@ -123,23 +72,67 @@ export class CoreClrDebugUtil
...
     }
 
+    public defaultDotNetCliErrorMessage(): string {
+        return 'Failed to find up to date dotnet cli on the path.';
+    }
+
+    // This function checks for the presence of dotnet on the path and ensures the Version
+    // is new enough for us. 
+    // Returns: a promise that returns a DotnetInfo class
+    // Throws: An CotNetCliError() from the return promise if either dotnet does not exist or is too old.
```
gregg-miskelly: `CotNetCliError` is wrong

src/coreclr-debug/activate.ts
chuckries: There is still an open question surrounding how we want to handle telemetry now that we have moved to the package manager.

---
In reply to: [86636070](https://github.com/OmniSharp/omnisharp-vscode/pull/873#discussion_r86636070) [](ancestors = 86636070)

src/coreclr-debug/install.ts
chuckries: not quite, looks like this is used in assets.ts and features/commands.ts

---
In reply to: [86635042](https://github.com/OmniSharp/omnisharp-vscode/pull/873#discussion_r86635042) [](ancestors = 86635042)

chuckries:
@DustinCampbell I think this is ready to be merged. Can you take another look?

package.json
```diff
@@ -91,7 +91,7 @@
       ]
     },
     {
-      "description": "OmniSharp (.NET Core - OSX / x64)",
+      "description": "OmniSharp (.NET Core - macOS / x64)",
```
DustinCampbell: Thanks!

PR #880, opened 2016-11-06
Add a setting to configure the server load timeout

### LINKED ISSUE #244: Initializing omnisharp fails for large workspaces
It looks like there is a hard coded timeout of 60 seconds in StdioOmnisharpServer._doConnect.  For large projects I'm developing (800+ projects) initial loading can take ~2 minutes.  Easy fix would be making this configurable per workspace.  Maybe the better fix would start the timeout clock on the last message written (i.e. it wouldn't timeout if projects were still being loaded).

wjk:
Does just what it says on the tin. Tested it with a moderately-large solution of mine that times out under the default one-minute interval. After I set `omnisharp.projectLoadTimeout` in my workspace settings file (to a value of 600 seconds), the solution loads fully, without encountering a timeout error.

Fixes #244.

### DIFF
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -82,6 +82,7 @@ export abstract class OmnisharpServer {
 
     protected _serverProcess: ChildProcess;
     protected _extraArgs: string[];
+    protected _options: Options;
 
     constructor(reporter: TelemetryReporter) {
         this._extraArgs = [];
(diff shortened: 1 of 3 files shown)


src/omnisharp/server.ts
```diff
@@ -240,9 +241,9 @@ export abstract class OmnisharpServer {
             '--encoding', 'utf-8'
         ];
 
-        const options = Options.Read();
+        this._options = Options.Read();
```
DustinCampbell: No, I don't think that's actually the case. This is a good change. Thanks!

Issue #879, opened 2016-11-06
Checking ID_LIKE - KDE Neon support


0xFireball:
## Environment data
.NET Command Line Tools (1.0.0-preview2-003131)

Product Information:
 Version:            1.0.0-preview2-003131
 Commit SHA-1 hash:  635cf40e58

Runtime Environment:
 OS Name:     neon
 OS Version:  16.04
 OS Platform: Linux
 RID:         neon.16.04-x64

## Steps to reproduce
Attempt to start the debugger in VSCode

## Expected  behavior
The debugger will start

## Actual behavior
The VSCode log will display `Error: Unsupported linux platform`

[This code file](https://github.com/OmniSharp/omnisharp-vscode/blob/735f90a299e857fc5de97216f961c4dd16907cac/src/platform.ts#L59) checks the platform by looking at `os-release`. My OS release looks like this:

```
NAME="KDE neon"
VERSION="5.8"
ID=neon                                                                                                             
ID_LIKE="ubuntu debian"                                                                                             
PRETTY_NAME="KDE neon User Edition 5.8"
VERSION_ID="16.04"
HOME_URL="http://neon.kde.org/"
SUPPORT_URL="http://neon.kde.org/"
BUG_REPORT_URL="http://bugs.kde.org/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

Though my distro is clearly Ubuntu based, the code only checks the `id` property. Perhaps the switch could be updated to check the if `ID_LIKE` contains a recognized platform, if it exists, or Ubuntu-based distros like Neon could be mapped to the `Ubuntu` platform.

gregg-miskelly:
@0xFireball Have you tried using the Ubuntu 16.04 version of the .NET CLI? I can build you a custom .vsix that will look at `ID_LIKE`/`VERSION_ID` assuming that the Ubuntu 16 CLI actually works on KDE Neon.

0xFireball:
KDE Neon is LTS Ubuntu with KDE, unlike Kubuntu, which is the latest Ubuntu with KDE.
Everything should be fully compatible.

A custom VSIX would be really helpful, thanks.

0xFireball:
Maybe you could change this in the actual code, assuming everything else works as expected

gregg-miskelly:
@0xFireball yes, that is definitely my goal. I just want to verify that I am actually resolving the issue before I submit a checkin a fix.

[...]

PR #890, opened 2016-11-08
Add support for KDE Neon (and other Linux distros with ID_LIKE)

gregg-miskelly:
This checkin adds support for checking the ID_LIKE field in the os-release file to try and find a compatible base OS. This support is imperfect since ID_LIKE doesn't provide a version number for the downstream OS(s). But this at least works for KDE Neon.

Associated issue: #879

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -18,7 +18,8 @@ const unknown = 'unknown';
 export class LinuxDistribution {
     public constructor(
         public name: string,
-        public version: string) { }
+        public version: string,
+        public idLike?: string[]) { }
 
     public static GetCurrent(): Promise<LinuxDistribution> {
         // Try /etc/os-release and fallback to /usr/lib/os-release per the synopsis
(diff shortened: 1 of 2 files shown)


src/platform.ts
```diff
@@ -69,14 +71,17 @@ export class LinuxDistribution {
                 else if (key === 'VERSION_ID') {
                     version = value;
                 }
+                else if (key === 'ID_LIKE') {
+                    idLike = value.split(" ");
+                }
 
                 if (name !== unknown && version !== unknown) {
```
gregg-miskelly: We ignore ID_LIKE if ID is something we understand. If we don't understand ID, we will iterate through the ID_LIKE values to see if we understand any of them.

DustinCampbell: Is that what it does though? It looks like it will bail if `ID_LIKE` hasn't been encountered and we find `ID` and `VERSION_ID` values -- even if we don't understand them.

gregg-miskelly: Everything looks good to me, but I could be missing something. What line are you thinking of?

DustinCampbell:
Did you force push?

src/platform.ts
```diff
@@ -69,14 +71,17 @@ export class LinuxDistribution {
                 else if (key === 'VERSION_ID') {
                     version = value;
                 }
+                else if (key === 'ID_LIKE') {
+                    idLike = value.split(" ");
+                }
 
-                if (name !== unknown && version !== unknown) {
+                if (name !== unknown && version !== unknown && idLike !== null) {
```
DustinCampbell: Now this looks right. :smile:

PR #891, opened 2016-11-08
Implement telemetry for acquisition

chuckries:
1. Create new PackageError class that captures package info to be thrown
from all package related errors.

2. Send telemetry event with relevant platform/package information.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -12,7 +12,7 @@ import * as coreclrdebug from './coreclr-debug/activate';
 import * as OmniSharp from './omnisharp/extension';
 import * as util from './common';
 import { Logger } from './logger';
-import { PackageManager, Status } from './packages';
+import { PackageManager, Status, PackageError } from './packages';
 import { PlatformInformation } from './platform';
 
 let _channel: vscode.OutputChannel = null;
@@ -31,7 +31,7 @@ export function activate(context: vscode.ExtensionContext): any {
 
     let logger = new Logger(text => _channel.append(text));
 
-    ensureRuntimeDependencies(extension, logger)
+    ensureRuntimeDependencies(extension, logger, reporter)
         .then(() => {
             // activate language services
             OmniSharp.activate(context, reporter);
@@ -41,18 +41,18 @@ export function activate(context: vscode.ExtensionContext): any {
         });
 }
 
-function ensureRuntimeDependencies(extension: vscode.Extension<any>, logger: Logger): Promise<void> {
+function ensureRuntimeDependencies(extension: vscode.Extension<any>, logger: Logger, reporter: TelemetryReporter): Promise<void> {
     return util.installFileExists(util.InstallFileType.Lock)
         .then(exists => {
             if (!exists) {
                 return util.touchInstallFile(util.InstallFileType.Begin).then(() => {
-                    return installRuntimeDependencies(extension, logger);
+                    return installRuntimeDependencies(extension, logger, reporter);
                 });
             }
         });
 }
 
-function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Logger): Promise<void> {
+function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Logger, reporter: TelemetryReporter): Promise<void> {
     logger.append('Updating C# dependencies...');
     _channel.show();
 
@@ -72,6 +72,8 @@ function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Lo
     let packageManager: PackageManager;
     let installationStage = 'touchBeginFile';
     let errorMessage = '';
+
+    let telemetryProps: any = {};
 
     return util.touchInstallFile(util.InstallFileType.Begin)
         .then(() => {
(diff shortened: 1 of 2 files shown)


chuckries:
@gregg-miskelly @DustinCampbell @wesrupert

src/omnisharp/server.ts
```diff
@@ -319,7 +319,8 @@ export abstract class OmnisharpServer {
             cleanupPromise = new Promise<void>((resolve, reject) => {
                 const killer = exec(`taskkill /F /T /PID ${this._serverProcess.pid}`, (err, stdout, stderr) => {
                     if (err) {
-                        return reject(err);
+                        reject(err);
+                        return;
```
DustinCampbell: Any reason you're changing these? It's perfectly legal to return a void-returning function.

src/packages.ts
```diff
@@ -154,22 +162,21 @@ function downloadFile(urlString: string, pkg: Package, logger: Logger, status: S
...
         }
 
         let request = https.request(options, response => {
             if (response.statusCode === 301 || response.statusCode === 302) {
                 // Redirect - download from new location
-                return downloadFile(response.headers.location, pkg, logger, status)
-                    .then(
-                        () => resolve(),
-                        err => reject(err));
+                resolve(downloadFile(response.headers.location, pkg, logger, status));
```
DustinCampbell: Why this change? We're now calling resolve even if the `downloadFile` promise rejects.

src/packages.ts
```diff
@@ -204,15 +211,15 @@ function downloadFile(urlString: string, pkg: Package, logger: Logger, status: S
             });
 
             response.on('error', err => {
-                reject(err)
+                reject(new PackageError('Reponse error:' + err.code || '', pkg, err));
```
DustinCampbell: missed space? Maybe just use a template literal?

src/omnisharp/server.ts
DustinCampbell: FWIW, I personally dislike this being on two lines when it could easily be on one. So, unless there's a real semantic difference I'm unaware of, I'd prefer not making these changes.

src/packages.ts
```diff
@@ -230,12 +237,14 @@ function installPackage(pkg: Package, logger: Logger, status?: Status): Promise<
...
         if (!pkg.tmpFile || pkg.tmpFile.fd == 0) {
-            return reject(new Error('Downloaded file unavailable'));
+            reject(new PackageError('Downloaded file unavailable', pkg));
+            return;
         }
 
         yauzl.fromFd(pkg.tmpFile.fd, { lazyEntries: true }, (err, zipFile) => {
             if (err) {
-                return reject(err);
+                reject(new PackageError('Immediate zip file error:', pkg, err));
```
DustinCampbell: Unnecessary `:`?

src/main.ts
```diff
@@ -101,18 +104,48 @@ function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Lo
...
+            telemetryProps['installStage'] = installationStage;
+            telemetryProps['platform.architecture'] = platformInfo.architecture;
+            telemetryProps['platform.platform'] = platformInfo.platform;
+            if (platformInfo.distribution) { telemetryProps['platform.distribution'] = platformInfo.distribution.toString(); }
+            telemetryProps['platform.runtimeId'] = platformInfo.runtimeId;
+
+            _reporter.sendTelemetryEvent(
+                "Acquisition",
+                telemetryProps
+                );
```
DustinCampbell: Why is this split across four lines?

src/main.ts
```diff
@@ -101,18 +104,48 @@ function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Lo
...
+            }
+
             logger.appendLine(`Failed at stage: ${installationStage}`);
             logger.appendLine(errorMessage);
         })
         .then(() => {
+            telemetryProps['installStage'] = installationStage;
+            telemetryProps['platform.architecture'] = platformInfo.architecture;
+            telemetryProps['platform.platform'] = platformInfo.platform;
+            if (platformInfo.distribution) { telemetryProps['platform.distribution'] = platformInfo.distribution.toString(); }
```
DustinCampbell: Total nit: This should probably be on three lines.

src/main.ts
```diff
@@ -101,18 +104,48 @@ function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Lo
             installationStage = 'touchLockFile';
             return util.touchInstallFile(util.InstallFileType.Lock);
         })
+        .then(() => {
+            installationStage = 'completeSuccess';
+        })
         .catch(error => {
-            errorMessage = error.toString();
+            if (error instanceof PackageError) {
+                // we can log the message in a PackageError to telemetry as we do not put PII in PackageError messages
```
DustinCampbell: Good idea.

src/main.ts
```diff
@@ -12,18 +12,19 @@ import * as coreclrdebug from './coreclr-debug/activate';
 import * as OmniSharp from './omnisharp/extension';
 import * as util from './common';
 import { Logger } from './logger';
-import { PackageManager, Status } from './packages';
+import { PackageManager, Status, PackageError } from './packages';
 import { PlatformInformation } from './platform';
 
 let _channel: vscode.OutputChannel = null;
+let _reporter: TelemetryReporter = null;
```
DustinCampbell: Feels inconsistent to make this a global variable while we thread `logger` through. Should we just thread this through the functions with `logger`?

chuckries: Probably.

src/omnisharp/server.ts
chuckries: I wasn't familiar with this pattern. I can change them back.

src/packages.ts
DustinCampbell: Note: This is the change I'm most interested. Everything else is nits.

chuckries: We are resolving with a promise that may reject, simply continuing the promise chain. There's no need to try to capture errors here as downloadFile captures error in all of it's non recursive code paths. 
https://jakearchibald.com/2014/resolve-not-opposite-of-reject/ 

It looks like both this way and the way this was originally written leaves us open to infinite redirect...

src/main.ts
chuckries: The telemetryProps object was initially created as a block here. I'll compress.

src/packages.ts
DustinCampbell: Ah yes, you're right.

DustinCampbell:
I don't have any concerns with this change. It'd be nice if some of the nits were cleaned up though.

chuckries:
I appreciate the nits, my javascript/typescript conventions are still rusty.

chuckries:
Fixed up the nits, merging.

gregg-miskelly:
@chuckries in the definition of PackageError, I would suggest adding a comment that 'message' should not contain PII

src/packages.ts
```diff
@@ -154,22 +161,19 @@ function downloadFile(urlString: string, pkg: Package, logger: Logger, status: S
...
         }
 
         let request = https.request(options, response => {
             if (response.statusCode === 301 || response.statusCode === 302) {
                 // Redirect - download from new location
-                return downloadFile(response.headers.location, pkg, logger, status)
-                    .then(
-                        () => resolve(),
-                        err => reject(err));
+                resolve(downloadFile(response.headers.location, pkg, logger, status));
```
gregg-miskelly: @chuckries You lost the 'return' here. Is that intentional?

chuckries: good catch, see #893

PR #902, opened 2016-11-09
Allow multiple simultaneous requests to be sent to OmniSharp server

DustinCampbell:
Communication with the OmniSharp server happens in a very serial fashion: make a request and wait for the response. This is done to ensure that critical communication, such as updating a buffer during typing, is processed quickly. However, that means that we'll wait on extremely long running requests that don't need to happen serially, such as waiting for all diagnostics to be returned for a solution.

This change splits requests into three queues based on the type of request. High priority requests, such as buffer changes, are still handled serially. Most other requests are placed on a queue that is allowed to make up to 8 simultaneous requests at a time. Other requests (especially long-running requests) are placed on a "deferred" queue that allows a much smaller number of simultaneous requests (maximum 2).

cc @david-driscoll

### DIFF
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -398,6 +398,8 @@ export abstract class OmnisharpServer {
             return Promise.reject<TResponse>('server has been stopped or not started');
         }
 
+        console.log(`makeRequest: path=${path}`);
+
         let startTime: number;
         let request: Request;
 
@@ -441,7 +443,6 @@ export abstract class OmnisharpServer {
     }
 
     private _processQueue(): void {
-
         if (this._queue.length === 0) {
             // nothing to do
             this._isProcessingQueue = false;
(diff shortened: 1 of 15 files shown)


src/omnisharp/server.ts
```diff
@@ -393,149 +389,66 @@ export abstract class OmnisharpServer {
...
         let request: Request;
 
         let promise = new Promise<TResponse>((resolve, reject) => {
             startTime = Date.now();
 
             request = {
-                path,
+                command,
                 data,
                 onSuccess: value => resolve(value),
```
DustinCampbell: If I recall, this caused a tslint error at one point, but I'l take a look.

DustinCampbell: Yeah, it does confuse tslint if I just pass the functions without creating lambdas -- "A Promise was found that appears to not have resolve or reject invoked on all code paths".

I suppose it's already been this way for several months already. I'll leave it as is for now.

src/omnisharp/server.ts
```diff
@@ -393,149 +389,66 @@ export abstract class OmnisharpServer {
...
+                onError: err => reject(err)
             };
 
-            this._queue.push(request);
+            this._requestQueue.enqueue(request);
 
-            if (this._getState() === ServerState.Started && !this._isProcessingQueue) {
-                this._processQueue();
+            if (this._getState() === ServerState.Started) {
+                this._requestQueue.drain();
```
DustinCampbell: You're right -- thanks!

src/omnisharp/server.ts
```diff
@@ -551,104 +464,108 @@ export class StdioOmnisharpServer extends OmnisharpServer {
...
-        delete this._activeRequest[requestSeq];
-
         if (packet.Success) {
-            entry.onSuccess(packet.Body);
-        } else {
-            entry.onError(packet.Message || packet.Body);
+            request.onSuccess(packet.Body);
+        }
+        else {
+            request.onError(packet.Message || packet.Body);
```
DustinCampbell: Makes sense -- thanks!

Issue #911, opened 2016-11-11
Extension will download multiple versions of the debugger on unknown OSs


gregg-miskelly:
## Steps to reproduce
Install the C# extension on an unknown x64 Linux distribution such as Gallium

## Expected  behavior
No debugger should download, and the user should get a warning

## Actual behavior
All x64 versions get downloaded (until we hit a failure on CentOS due to #910). Unclear what happens after that.

PR #912, opened 2016-11-11
Fix handling of unknown Linux distributions

gregg-miskelly:
This checkin fixes the behavior of the C# extension if we are on a Linux distribution that we don't understand or support. Before we would attempt to download every different x64 package, and then show a very generic error message.


gregg-miskelly:
@chuckries @DustinCampbell please review

gregg-miskelly:
This is the fix for #911

Issue #848, opened 2016-10-28
GalliumOS Omnisharp.path error


Neoakb:
## Environment data

`dotnet --info` output:
.NET Command Line Tools (1.0.0-preview2-003131)

Product Information:
 Version:            1.0.0-preview2-003131
 Commit SHA-1 hash:  635cf40e58

Runtime Environment:
 OS Name:     galliumos
 OS Version:  2.0
 OS Platform: Linux
 RID:         galliumos.2.0-x64

VS Code version: latest
C# Extension version: latest
## Steps to reproduce

Start VC
## Actual behavior

this came first:
[ERROR] Could not locate an OmniSharp server that supports your Linux distribution.

OmniSharp provides a richer C# editing experience, with features like IntelliSense and Find All References.
It is recommend that you download the version of OmniSharp that runs on Mono using the following steps:
    1. If it's not already installed, download and install Mono (https://www.mono-project.com)
    2. Download and untar the latest OmniSharp Mono release from  https://github.com/OmniSharp/omnisharp-roslyn/releases/
    3. In Visual Studio Code, select Preferences->User Settings to open settings.json.
    4. In settings.json, add a new setting: "omnisharp.path": "/path/to/omnisharp/OmniSharp.exe"
    5. In settings.json, add a new setting: "omnisharp.useMono": true
    6. Restart Visual Studio Code.

So I followed the instructions. my setting.json:
// Place your settings in this file to overwrite the default settings
{
    "omnisharp.path": "/home/chrx/Omnisharp/Omnisharp.exe",
    "omnisharp.useMono": true
}

I get this error on startup:

Invalid value specified for "omnisharp.path" ('/home/chrx/Omnisharp/Omnisharp.exe).

Neoakb:
I changed Omnisharp.exe to OmniSharp.exe in the path, that still didn't fix it.

DustinCampbell:
The entire path is case-sensitive. Also, what version of Mono do you have installed?

Neoakb:
http://imgur.com/pDs98FGl.png

There's a screenshot with mono -v, VC settings.json, and the file manager open to the Omnisharp install.

Mono is working fine. I can run the hello test just fine. 

I get this error when I run a simple .cs file though rightclick -> Run Code
(using https://code.visualstudio.com/docs/runtimes/dotnet to create a new dotnet project)

[Running] scriptcs "/home/chrx/Documents/TestingOmni/Program.cs"
[...]

PR #922, opened 2016-11-12
Add platform detection for Gallium

gregg-miskelly:
Update platform.ts to treat Gallium 2.0 as Ubuntu 16.


gregg-miskelly:
@DustinCampbell I don't have a strong opinion on if we take this for 1.5.0, or we wait for vNext. It is certainly a really simple fix.

gregg-miskelly:
This addresses https://github.com/OmniSharp/omnisharp-vscode/issues/848

PR #939, opened 2016-11-15
Configurable project list cap

### LINKED ISSUE #875: project.json project list capped to 100
The project.json project list is capped to 100, for both display and search. I cannot search for my projects that are beyond the capped set.

I most often open a narrow folder so that only a few projects are within scope. When I'm working on a set of cross-repo changes, I sometimes open to my whole "git" directory and then run into this problem. 

I'd like to see one of two user experiences, in order of preference:

1. All projects are visible and all projects are searchable.
2. No projects are listed with some UI explaining why, but all projects are searchable.

Thanks!

Current experience:

screen shot 2016-11-05 at 10 23 57 am

/cc @DustinCampbell @kendrahavens

filipw:
Fixes #875 

Introduces a new setting `omnisharp.maxProjectResults` with a default value of `100` in place of the old hardcoded `100`. User can now set the value to whatever seems fit for their workflow.

Naturally, larger value i.e. `5000` would mean longer start up time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -317,7 +317,7 @@
         },
         "omnisharp.maxProjectResults": {
           "type": "number",
-          "default": 100,
+          "default": 250,
           "description": "The amount of projects to be shown in the 'Select Project' dropdown, discovered in the descendant folders of the working folder. Once the limit is reached, the extension will stop searching for projects."
         }
       }
--- a/package.json
+++ b/package.json
@@ -314,6 +314,11 @@
           "type": "number",
           "default": 60,
           "description": "The time Visual Studio Code will wait for the OmniSharp server to start. Time is expressed in seconds."
+        },
+        "omnisharp.maxProjectResults": {
+          "type": "number",
+          "default": 100,
+          "description": "The amount of projects to be shown in the 'Select Project' dropdown, discovered in the descendant folders of the working folder. Once the limit is reached, the extension will stop searching for projects."
         }
       }
     },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -314,6 +314,11 @@
           "type": "number",
           "default": 60,
           "description": "The time Visual Studio Code will wait for the OmniSharp server to start. Time is expressed in seconds."
+        },
+        "omnisharp.maxProjectResults": {
+          "type": "number",
+          "default": 100,
```
DustinCampbell: Should we increase this number as part of this change?

filipw: we could, perhaps `250`? Thing is you probably want to cap it at a reasonable number cause when you open against i.e. `c:\` it would take ages to scan if the limit is too high

package.json
```diff
@@ -314,6 +314,11 @@
           "type": "number",
           "default": 60,
           "description": "The time Visual Studio Code will wait for the OmniSharp server to start. Time is expressed in seconds."
+        },
+        "omnisharp.maxProjectResults": {
+          "type": "number",
+          "default": 250,
+          "description": "The amount of projects to be shown in the 'Select Project' dropdown, discovered in the descendant folders of the working folder. Once the limit is reached, the extension will stop searching for projects."
```
DustinCampbell: This should read: "The maximum number of projects to be shown in the 'Select Project' dropdown (maximum 250)"

filipw:
Thanks -  done

PR #948, opened 2016-11-15
Print current platform and RID during download

DustinCampbell:


src/main.ts
```diff
@@ -83,6 +83,16 @@ function installRuntimeDependencies(extension: vscode.Extension<any>, logger: Lo
             packageManager = new PackageManager(info, extension.packageJSON);
             logger.appendLine();
 
+            // Displa y platform information and RID followed by a blank line
```
gregg-miskelly: Your comment got messed up

DustinCampbell: Oops

Issue #779, opened 2016-09-23
System.TypeInitializationException during expression evaluation


caslan:
This failure stack is the top most non fatal error we are encountering in the debugger. It should have something to do with OpenSSL dependency on OSX but we need to investigate further to see if this is an issue external to debugger or not:

at Interop.Crypto.EvpSha1()#N#   at Internal.Cryptography.HashProviderDispenser.CreateHashProvider(String hashAlgorithmId)#N#   at System.Security.Cryptography.SHA1.Implementation..ctor()#N#   at System.Security.Cryptography.SHA1.Create()#N#   at Roslyn.Utilities.SHA1CryptoServiceProvider..ctor()#N#   at Microsoft.CodeAnalysis.CryptographicHashProvider.ComputeSha1(Byte[] bytes)#N#   at Microsoft.CodeAnalysis.AssemblyIdentity.CalculatePublicKeyToken(ImmutableArray`1 publicKey)#N#   at Microsoft.CodeAnalysis.AssemblyIdentity.get_PublicKeyToken()#N#   at Microsoft.CodeAnalysis.AssemblyIdentity.KeysEqual(AssemblyIdentity x, AssemblyIdentity y)#N#   at Microsoft.CodeAnalysis.AssemblyIdentity.MemberwiseEqual(AssemblyIdentity x, AssemblyIdentity y)#N#   at Microsoft.CodeAnalysis.AssemblyIdentityComparer.TriviallyEquivalent(AssemblyIdentity x, AssemblyIdentity y)#N#   at Microsoft.CodeAnalysis.AssemblyIdentityComparer.Compare(AssemblyIdentity reference, String referenceDisplayName, AssemblyIdentity definition, Boolean& unificationAppl

gregg-miskelly:
This issue is also tracked with https://github.com/dotnet/roslyn/issues/12681

gregg-miskelly:
This is still our top error hit. Lets see if we can fix this for 1.6. My thought is to disable the installation of the debugger unless we find openssl.

gregg-miskelly:
NOTE: This work around will go away once the debugger is running on https://github.com/dotnet/corefx/issues/9394

PR #986, opened 2016-11-23
Add detection logic for missing OpenSSL symlinks on macOS

gregg-miskelly:
By far our most common EE error is due to folks not having the OpenSSL symlinks created. This checkin adds detection logic so that we will not install the debugger unless they are present.

This resolves https://github.com/OmniSharp/omnisharp-vscode/issues/779

Testing:
1. macOS: Verified we correctly fail if either of the two symlinks are missing
2. macOS: Verified the debug proxy correctly fails if we ignore the problem
3. macOS: Verified the 'channel.open' calls work
4. macOS + Windows: Verified the extension can install when everything is correct


gregg-miskelly:
@chuckries please review

Issue #844, opened 2016-10-27
Debugger doesn't support x86 Windows


gregg-miskelly:
## Steps to reproduce

On x86 (32-bit) Windows, download VS Code, and install the C# extension
## Expected  behavior

32-bit debugger downloads and runs
## Actual behavior

32-bit Windows isn't supported.

gregg-miskelly:
NOTE that when we fix this, we need to remove the code added in https://github.com/OmniSharp/omnisharp-vscode/pull/998

Fetandrey:
Any progress?

gregg-miskelly:
We aren't ready to guess when we will be able release support for this -- we have some hot items which are currently consuming all of our time, but we are much closer to it then we once were. We almost have a new build system in place which among other things, finishes most of the build side of the work for this.

Fetandrey:
Thank you! It's good to know.

jayendranarumugam:
I'm currently using V1.12.2. I couldn't make debug my code with any Break Point,my machine is 32bit. When this be fixed ?

sheikhzubayr:
Is there a fix yet? A x86 user here, not able to debug my c# programs on vs code! But, nonetheless, VS Code is the best editor from Microsoft ever!!

gregg-miskelly:
No news yet I'm afraid.

sheikhzubayr:
@gregg-miskelly This is really unfortunate as I have been eagerly waiting to have C# debugger up and running in VS code for my x86 machine!! Please help us resolve it fast!!

PTescAL:
Any progress on this?
Due to third-party libraries to which we would like to link from .net Core projects, we are limited in our ability to debug our x86 programs, too.

twtyypmb:
any progress?If the problem will not be solved,I have to give up .net core.Is it as design?

gregg-miskelly:
Unfortunately due the low usage of x86 Windows operating systems and many competing priorities, we are unable to fund adding x86 debugging support at this time.  The C# extension will of course continue to function as an editor on x86 Windows. 

We recommend anyone looking for .NET debugging on x86 Windows to use Visual Studio instead. See https://www.visualstudio.com/downloads/ for more information.

Yoplayer-1:
Ya I'm also getting the same error

PR #998, opened 2016-11-28
Add warning when x86 windows is used

gregg-miskelly:
Previously the debugger would add an error to the install log when installing on x86 Windows. With this change we instead add a warning.

This code should be removed once https://github.com/OmniSharp/omnisharp-vscode/issues/844 is fixed.


gregg-miskelly:
@chuckries please review

PR #999, opened 2016-11-29
Fixing sourceFileMaps inner type

WardenGnaw:
SourceFileMaps are restricted to be strings, adding this so that the
hints will show that it needs to be a string type.


WardenGnaw:
@wesrupert

PR #1000, opened 2016-11-30
Add support for new Linux distros in the debugger

gregg-miskelly:
This adds support for Ubuntu 16.10, Fedora 24 and openSUSE 42.1. To do this we are moving the debugger packages to .NET Core 1.1.

Testing: Verified this works on Windows, OSX and Ubuntu 14.04.

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -269,19 +269,25 @@ export class PlatformInformation {
         const centos_7 = 'centos.7-x64';
         const debian_8 = 'debian.8-x64';
         const fedora_23 = 'fedora.23-x64';
+        const fedora_24 = 'fedora.24-x64';
         const opensuse_13_2 = 'opensuse.13.2-x64';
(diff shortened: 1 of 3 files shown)


gregg-miskelly:
@rajkumar42 please review

src/platform.ts
```diff
@@ -309,9 +315,21 @@ export class PlatformInformation {
...
+                } else if (distributionVersion === "24") {
+                    return fedora_24;
+                }
+                break;
+
             case 'opensuse':
-                return opensuse_13_2;
+                if (distributionVersion.startsWith("13.")) {
+                    return opensuse_13_2;
+                } else if (distributionVersion.startsWith("42.")) {
```
rajkumar42: I think we should fully qualify the version 13.2 and 42.1, since coreclr seems to exactly match the version.

Issue #757, opened 2016-09-14
Syntax highlighting bug for fields


natemcmaster:
## Environment data

`dotnet --info` output: n/a
VS Code version: 1.5.2
C# Extension version:  1.4.1
## Steps to reproduce

```
public class Test
{
    public string _field1 = Path.Combine("",method());
    public string _field2 = Path.Combine("", method());
    public string _field3 = "";
}
```

![image](https://cloud.githubusercontent.com/assets/2696087/18524807/e015140c-7a6e-11e6-8a86-4d826aabc7aa.png)

Highlighter gets into an incorrect state trying to highlight `_field2`.

DustinCampbell:
Interesting. :smile:. Usually the regular expressions fail at the starts/ends of tokens. cc @ivanz.

ivanz:
Hmm I am not sure what's going on here, because the "variable" regex (https://github.com/OmniSharp/omnisharp-vscode/blob/master/syntaxes/csharp.json#L85) seems fine outside of the grammar file: http://regexr.com/3e9vo

Issue #960, opened 2016-11-18
Generics with multiple type parameters are colorized incorrectly


pflannery:
VS Code version: 1.7.1
C# Extension version: 1.5.2

![c-sharp-colouring](https://cloud.githubusercontent.com/assets/1727302/20418085/53d3c090-ad42-11e6-9e9d-a2709d3ccf19.gif)

PR #1014, opened 2016-12-04
Syntax highlighting improvements and tests

ivanz:
1. Syntax highlighting support for class/interface generic constrains + tests.
2. Much improved fields syntax highlighting  + tests.

Fixes: #757, #960

## Class Declaration and Inheritance

Screenshots are from before I made the above changes around generics, but you get the gist.

### Before
image

### After
image

### Before

image

### After

image

### Before
image

### After

image

### DIFF
--- a/syntaxes/csharp.json
+++ b/syntaxes/csharp.json
@@ -123,6 +123,43 @@
         }
       ]
     },
+    "genericConstraints": {
+      "begin": "(where)\\s*(\\w+)\\s*:",
+      "end": "(?={)",
+      "beginCaptures": {
+        "1": {
+          "name": "keyword.other.cs"
+        },
+        "2": {
+          "name": "storage.type.cs"
+        }
+      },
+      "patterns": [
+        {
+          "match": "\\b(class|struct)\\b",
+          "name": "keyword.other.cs"
+        },
+        {
+          "match": "(new)\\s*\\(\\s*\\)",
+          "captures": {
+            "1": {
+              "name": "keyword.other.cs"
+            }
+          }
+        },
+        {
+          "match": "([\\w<>]+)\\s*(?=,|where|{)",
+          "captures": {
+            "1": {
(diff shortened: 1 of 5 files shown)


syntaxes/csharp.json
```diff
@@ -123,6 +160,43 @@
         }
       ]
     },
+    "genericConstraints": {
+      "begin": "(where)\\s*(\\w+)\\s*:",
```
DustinCampbell: It probably doesn't matter in practice, but shouldn't the `\\s*` after `(where)` above _technically_ be `\\s+`? Otherwise, wouldn't it match `whereImCool:`?

syntaxes/csharp.json
```diff
@@ -123,6 +160,43 @@
...
+        {
+          "match": "(new)\\s*\\(\\s*\\)",
+          "captures": {
+            "1": {
+              "name": "keyword.other.cs"
+            }
+          }
+        },
+        {
+          "match": "([\\w<>]+)\\s*(?=,|where|{)",
```
DustinCampbell: Does matching `[\\w<>]+` mean that it won't work for generic types with multiple parameters?

syntaxes/csharp.json
ivanz: Good catch. Fixed.

syntaxes/csharp.json
ivanz: You are right. I need to think how to handle this better. Whitespace handling is a pain in the a**. :) Will update the pull request.

ivanz:
I've done some updates and it's now again ready for review.

ivanz:
I saw you approved this pull (thanks!), but heads up that I don't have merge permissions, so you will have to merge it when ready 👍

DustinCampbell:
Yup. I was waiting to merge when I got into the office.

PR #1020, opened 2016-12-06
Add recent fixes to changelog

DustinCampbell:
@gregg-miskelly, you should probably add some of your recent changes here too.


gregg-miskelly:
Will do

Issue #877, opened 2016-11-05
Extract C# grammar into own repo to avoid effort duplication between Atom and VSCode?


ivanz:
Appears to me that there is some ongoing duplicated effort with the C# textmate grammar in VSCode ([gitlog](https://github.com/OmniSharp/omnisharp-vscode/commits/master/syntaxes/csharp.json)) and Atom ([gitlog](https://github.com/atom/language-csharp/commits/master/grammars/c%23.cson)). 

Seems like the only real difference between the two is JSON vs CSON, so why not extract the VSCode textmate C# grammar and [test harness setup](https://github.com/OmniSharp/omnisharp-vscode/tree/master/test/syntaxes) into a separate repo in OmniSharp, port any fixes from Atom and then share the same grammar across the two extensions (outping CSON, etc)?

What do you guys think?

/cc @DustinCampbell  /cc @david-driscoll /cc @damieng

DustinCampbell:
That's a possibility, though we're also pulling the C# grammar from here into Visual Studio itself. Atom could pull from here as well. If we want to create a separate repo, I would probably create it in the dotnet organization.

/cc @Pilchie

ivanz:
Having it in the dotnet org makes sense. I don't really mind. Main thing is to avoid duplicated work between VSCode, Atom and potentially others given that there have been substantial improvements to the C# grammar (imho) and its test coverage in the VSCode repo already

damieng:
I'm up for sharing a common grammar file for C# if we can.  Ideally the repo would be just enough for the grammar and a platform-agnostic test suite.  That way we might be able to include it as a submodule or some kind of other similar mechanism to wrap it with each products necessary additional files.

The other issue to sort out would be the license. The one on this repo doesn't include any of the previous licencing from atom/language-csharp which includes GitHub and TextMates licence info in https://github.com/atom/language-csharp/blob/master/LICENSE.md

ivanz:
@damieng The mentioned license info is in: https://github.com/OmniSharp/omnisharp-vscode/blob/master/ThirdPartyNotices.txt

I've extracted the vscode grammar and tests in a repo here: https://github.com/ivanz/csharp-textmate-grammar 

It currently takes the VSCODE `json` and converts it to Atom `cson` in the `grammars/` folder after `npm i && gulp`. It's an npm package where the `grammars` directory is published only, but I haven't published it yet.

[...]

PR #1054, opened 2016-12-17
Embrace "csharp-textmate-grammar" shared C# effort + fixes from atom, etc

ivanz:
As a follow up from #877 - the C# grammar has been extracted into a repository [here](https://github.com/ivanz/csharp-textmate-grammar), which gets built by Tarvis CI and the output is published as a NPM package [here](https://www.npmjs.com/package/csharp-textmate-grammar). In addition to that @damieng has been porting over fixes and features (e.g. static usings) from the Atom grammar.

The csharp-textmate-grammar NPM package looks like this when installed:

image

Since @DustinCampbell mentioned that Visual Studio team is pulling in somehow the grammar from this repo (I guess they are downloading it?) we can't yet delete the file from this repo and automatically pull the shared one.

So for now:

```
npm run update-grammar
```

will update the C# grammar package to the latest stable version and update the one in `syntaxes/csharp.json`.

What do you think?

This pull request also removes all grammar tests, drops the vscode-textmate dependency (which was pulling in a native package)


DustinCampbell:
@ivanz: Do note that https://github.com/OmniSharp/omnisharp-vscode is a Microsoft repo and I'm a member of the C# team. I'm very keen on moving this someplace more shared, and I don't want to stop any of the passion here. But, would you be OK if we slow down a bit so we can do this properly?

ivanz:
Totally fine with that 👍 No rush at all - I was simply channelling some free time from the last 2-3 days into this before I get busy with other stuff and holidays, etc - hence why the PoC  🎄 🎅  and the rapid updates on my end. Not planning on doing anything further for now.

DustinCampbell:
FWIW, there are implementation reasons to do this too. I'd like to switch our unit testing over to being hosted in VS Code so that we can add tests that use VS Code services and features (i.e. https://code.visualstudio.com/docs/extensions/testing-extensions). However, the oniguruma dependency of the colorizer tests create some challenges.

First of all, there are bitness issues because node-gyp defaults to compiling to x64 on Windows-x64 and the VS Code test bed is built with 32-bit electron. However, even after manually running node-gyp to compile to ia32, it still fails to load the compiled onigurama.node library due to a "DLL initialization failure".

Moving this out to its own repo would certainly make this easier. :smile:

damieng:
@DustinCampbell the VS Code testing stuff will not be in the new C# grammar repo tho right?

DustinCampbell:
correct

DustinCampbell:
The new repo shouldn't have any dependency on VS Code. At least, no *real* dependency. There's the "vscode-textmate" package we use for the tests, but that doesn't have a dependency either.

ivanz:
@damieng In case you haven't seen them - the extracted test bits without any vscode-dependant stuff look like this currently: https://github.com/ivanz/csharp-textmate-grammar/tree/master/test  and the `vscode-textmate` (no actual vscode dependency as Dustin says). I say currently - because this is just taking out what I did already in the vscode repo, but I am open to suggestions for improvements as to how the testing is done in the future. It's quite easy to add regressions from my personal experience haha.

DustinCampbell:
@ivanz: Now that we've got an official place to put it, do you want to move the bits in https://github.com/ivanz/csharp-textmate-grammar over to https://github.com/dotnet/csharp-textmate-grammar? Then, we can update this PR

ivanz:
Sure thing - I don't have committed access to the dotnet repo, so I can open a pull request?

DustinCampbell:
Yup. Open a PR. We'll get commit access set up soon.

ivanz:
Done: https://github.com/dotnet/csharp-textmate-grammar/pull/1

DustinCampbell:
First merge is away!

DustinCampbell:
Closing in favor of #1146

Issue #93, opened 2016-03-15
Code formatting does not obey editor.insertSpaces


alexandrudima:
Moved from Microsoft/vscode#3847
I have verified that we call on our side the format on type provider with the configured indentation options.

---

From @nixbb
It seems that the code formatter does not adhere to the indentation mode specified in the user settings.

Here's a simple example using version 0.10.10
In this case "editor.insertSpaces" is set to "auto" but this issue also manifests if it's set to false.
![codeformattingbug](https://cloud.githubusercontent.com/assets/958979/13603388/929fa584-e545-11e5-9009-8b55a45481d9.gif)

The example shows that all the indentation starts off as tabs.
The newly added line (line 11) is also indented using tabs.
When the statement is finished with a semicolon the current line and the next line is then indented with spaces.

Not shown in the gif is that an undo will revert line 11 and 12 to tabs instead of spaces.

Hope this helps narrow down the issue.

0xdeafcafe:
Any updates on this? It's incredibly annoying.

anfomin:
Is is alive?

pavlos256:
Is this the reason that the "implement interface" uses spaces instead of tabs or should I file a new bug?

0xdeafcafe:
@pavlos256 Same reason. It runs the format command on the relevant lines, which uses spaces.

ivanz:
For a workaround see here (how to use `omnisharp.json`): https://github.com/OmniSharp/omnisharp-roslyn/issues/366

PR #1055, opened 2016-12-18
Pass editor tab/indentation settings to OmniSharp

### LINKED ISSUE #834: editor.formatOnType does not respect editor.insertSpaces setting
_From @j-h-a on October 20, 2016 15:58_
- VSCode Version:
  1.6.1
- OS Version:
  OSX (10.11.6)
### Searchable description

VS Code keeps changing some lines to have spaces instead of tabs in the leading whitespace.
### Steps to Reproduce:

Use these settings:
    "editor.tabSize": 4,
    "editor.insertSpaces": false,
    "editor.detectIndentation": false,
    "editor.formatOnType": true,

Type some code that triggers formatOnType to activate a format change.
### Expected behaviour:

Line is formatted using tabs for leading whitespace, because of the insertSpaces setting.
### Actual behaviour:

Line is formatted with spaces instead of tabs for the leading whitespace.

_Copied from original issue: Microsoft/vscode#14108_

filipw:
This should fix #834 and related issues.

This PR ensures that we pass VS Code `editor.*` settings related to tabs/indentation into OmniSharp (through environment variables, which OmniSharp already supports - no changes needed there). This way, when invoking commands such as `Format Document`, we will respect the user set up in his VS Code, rather than imposing OmniSharp defaults (spaces over tabs, 4 spaces indentation).

This feature can be opt out from by setting `omnisharp.useEditorFormattingSettings` to `false`.

Additionally, OmniSharp has its own tab vs spaces and indentation rules that can be controlled using `omnisharp.json`. If this file is present in the current working directory, it will still take precedence as it was in the past - no change there.

### DIFF
--- a/src/omnisharp/launcher.ts
+++ b/src/omnisharp/launcher.ts
@@ -161,26 +161,25 @@ function launch(cwd: string, args: string[]): Promise<LaunchResult> {
     return PlatformInformation.GetCurrent().then(platformInfo => {
         const options = Options.Read();
 
-        let envVars = {};
         if (options.useEditorFormattingSettings) 
         {
             let editorConfig = vscode.workspace.getConfiguration('editor');
-            envVars['formattingOptions:useTabs'] = !editorConfig.get('insertSpaces', true);
-            envVars['formattingOptions:tabSize'] = editorConfig.get('tabSize', 4);
-            envVars['formattingOptions:indentationSize'] = editorConfig.get('tabSize', 4);
+            args.push(`formattingOptions:useTabs=${!editorConfig.get('insertSpaces', true)}`);
+            args.push(`formattingOptions:tabSize=${editorConfig.get('tabSize', 4)}`);
+            args.push(`formattingOptions:indentationSize=${editorConfig.get('tabSize',4)}`);
         }
 
         if (options.path && options.useMono) {
-            return launchNixMono(options.path, cwd, args, envVars);
+            return launchNixMono(options.path, cwd, args);
         }
 
         const launchPath = options.path || getLaunchPath(platformInfo);
 
         if (platformInfo.isWindows()) {
-            return launchWindows(launchPath, cwd, args, envVars);
+            return launchWindows(launchPath, cwd, args);
         }
         else {
-            return launchNix(launchPath, cwd, args, envVars);
+            return launchNix(launchPath, cwd, args);
         }
     });
 }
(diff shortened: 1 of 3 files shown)


src/omnisharp/launcher.ts
```diff
@@ -193,6 +202,7 @@ function launchWindows(launchPath: string, cwd: string, args: string[]): LaunchR
     }
 
     let argsCopy = args.slice(0); // create copy of args
+    //argsCopy.push('--debug');
```
DustinCampbell: Why is this added and commented out?

src/omnisharp/launcher.ts
```diff
@@ -161,17 +161,26 @@ function launch(cwd: string, args: string[]): Promise<LaunchResult> {
...
         const options = Options.Read();
 
+        let envVars = {};
+        if (options.useEditorFormattingSettings) 
+        {
+            let editorConfig = vscode.workspace.getConfiguration('editor');
+            envVars['formattingOptions:useTabs'] = !editorConfig.get('insertSpaces', true);
+            envVars['formattingOptions:tabSize'] = editorConfig.get('tabSize', 4);
+            envVars['formattingOptions:indentationSize'] = editorConfig.get('tabSize', 4);
+        }
```
DustinCampbell: Can't these just be passed as additional arguments? I though OmniSharp's configuration builder would pick these up from the command line.

DustinCampbell:
This will also address https://github.com/OmniSharp/omnisharp-vscode/issues/93.

src/omnisharp/launcher.ts
filipw: oops my dirty attaching of a debugger 😞

src/omnisharp/launcher.ts
filipw: yes they can - in fact I believe `restore packages` is passed like that already if I recall correctly. I can change that for consistency, there is no real reason to use environment variables indeed

DustinCampbell: That's what I thought -- thanks! Yeah, go ahead and pass them as args for consistency. I have no idea what the logic is for options overriding in the configuration middleware that OmniSharp uses. E.g. do command line args override environment variable or vice versa? I just ran into something similar with MSBuild, so it was on my mind. Hence, the question. :smile:

filipw:
Updated the PR to pass the formatting options using command line instead of env vars. 

OmniSharp now starts like this, with your editor settings passed in: 

`OmniSharp: -s c:\code\csharp-scripting-demos\dotnetcli-and-debug --hostPID 3408 --stdio DotNet:enablePackageRestore=false --encoding utf-8 -v formattingOptions:useTabs=true formattingOptions:tabSize=2 formattingOptions:indentationSize=2`

If you opt out of the formatting settings sharing, then it will start same as before i.e.: 

`OmniSharp: -s c:\code\csharp-scripting-demos\dotnetcli-and-debug --hostPID 4024 --stdio DotNet:enablePackageRestore=false --encoding utf-8 -v`

filipw:
I also created a wiki page https://github.com/OmniSharp/omnisharp-roslyn/wiki/Configuration%20Options - might be useful if folks have questions in the future

filipw:
thanks!

PR #1065, opened 2016-12-21
Fix asset generation for .csproj .NET Core projects

### LINKED ISSUE #767: Update launch.json generation logic for MSBuild-based .NET Core
The current launch.json generation logic needs to be updated to handle MSBuild-based .NET Core. Currently, it only runs on project.json. This requires a few things to happen:

- [x] OmniSharp needs to be updated to understand multiple target frameworks in .csproj ([example](https://github.com/dotnet/sdk/blob/8f09baf9b3c600d7df3495c6eb68814959b4bafb/TestAssets/TestProjects/CrossTargeting/DesktopAndNetStandard/DesktopAndNetStandard.csproj)) files using the latest MSBuild bits.
- [ ] The OmniSharp `/projects` end point needs to be updated to return information for multiple target frameworks (like it does for project.json projects).
- [x] The OmniSharp `/projects` end point needs to be updated to include output file path information.
- [x] The C# extension needs to be updated to a newer version of OmniSharp with the changes above.
- [ ] The launch.json/tasks.json generation logic needs tweaked to work for .csproj projects.

DustinCampbell:
Fixes #767

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -59,7 +59,7 @@ export class AssetGenerator {
 
     private hasProject: boolean;
     private projectPath: string;
-    private projectJsonPath: string;
+    private projectFilePath: string;
     private targetFramework: string;
     private executableName: string;
     private configurationName: string;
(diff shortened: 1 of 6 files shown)


src/assets.ts
```diff
@@ -52,22 +50,284 @@ interface AttachConfiguration extends DebugConfiguration {
...
+            version: '0.1.0',
+            command: 'dotnet',
+            isShellCommand: true,
+            args: [],
+            tasks: [this.createBuildTaskDescription()]
+        };
+    }
+}
+
+function findNetCoreAppTargetFramework(project: protocol.MSBuildProject) {
```
gregg-miskelly: nit: I always like having a return type when it isn't obvious.

src/assets.ts
```diff
@@ -52,22 +50,284 @@ interface AttachConfiguration extends DebugConfiguration {
...
+
+function containsDotNetCoreProjects(workspaceInfo: protocol.WorkspaceInformationResponse) {
+    if (workspaceInfo.DotNet && findExecutableProjectJsonProjects(workspaceInfo.DotNet.Projects, 'Debug').length > 0) {
+        return true;
+    }
+
+    if (workspaceInfo.MsBuild && findExecutableMSBuildProjects(workspaceInfo.MsBuild.Projects).length > 0) {
+        return true;
+    }
+}
```
gregg-miskelly: Do you need a `return false;` here?

DustinCampbell: What? It's JavaScript. Of course not!

And, yes, I clearly missed that. Thanks!

src/assets.ts
DustinCampbell: Thanks!

PR #1069, opened 2016-12-21
Add telemetry for the sorts of projects OmniSharp is started on

### LINKED ISSUE #616: C# extension should have telemetry to record .NET Core vs. mono/desktop usage
Currently the C# extension supports both desktop/mono and .NET Core scenarios. We should have telemetry so that we can understand which type of projects people are using.

DustinCampbell:
Fixes #616

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -29,6 +29,15 @@ export function isBoolean(obj: any): obj is boolean {
     return obj === true || obj === false;
 }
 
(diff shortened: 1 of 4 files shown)


DustinCampbell:
OK. I just tested this with several projects of varying sizes and mixes of Unity, .NET Core and desktop projects. Looks like this should work reasonably well.

PR #1121, opened 2017-01-13
Update README.md to mention Python

gregg-miskelly:
Python 2.7 is required on Windows to install dependencies. I am assuming it is required on other platforms too, but since npm itself requires it on Linux, and I don't have a spare Mac to verify, I can't say for sure.


gregg-miskelly:
Good to know. Is it worth checking this in then? Or will it be a little while still before the TextMate grammer is moved over?

DustinCampbell:
Fine by me. I'll roll it back when we close #877

PR #1143, opened 2017-01-24
Be tolerant of tasks.json files with no 'tasks' node or with an os-specific node

### LINKED ISSUE #1140: "Failed to parse tasks.json file" with OS specific closure
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-preview2-1-003177)

Product Information:
 Version:            1.0.0-preview2-1-003177
 Commit SHA-1 hash:  a2df9c2576

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
```

VS Code version: `1.8.1`
C# Extension version: `1.6.2`

## Steps to reproduce

`$ git clone git@github.com:kriwar/dotnetcore-helloworld.git`

Open project in VS Code.

Notice the Windows specific closure in https://github.com/kriwar/dotnetcore-helloworld/blob/master/.vscode/tasks.json#L3, if this enclosing statement is removed, things work just fine.

The project works despite the error message, I can run tasks from the VS Code Command Palette without problems.

## Expected  behavior

Project opens without errors or warnings.

## Actual behavior

Error prompt `Failed to parse tasks.json file`. Follow-up Warn `Required assets to build and debug are missing...` (I believe this is a follow-on to the error displayed first).

As far as I've been able to tell, this behavior started with v1.6.2 of OmniSharp, error message is emitted from https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/assets.ts#L367

DustinCampbell:
Fixes #1140

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -346,7 +346,7 @@ function getOperations(generator: AssetGenerator) {
         getLaunchOperations(generator.launchJsonPath, operations));
 }
 
-function getBuildTasks(tasksConfiguration: tasks.TaskConfiguration) {
+function getBuildTasks(tasksConfiguration: tasks.TaskConfiguration): tasks.TaskDescription[] {
     let result: tasks.TaskDescription[] = [];
 
     function findBuildTask(tasksDescriptions: tasks.TaskDescription[]) {
@@ -392,6 +392,7 @@ function getBuildOperations(tasksJsonPath: string) {
                     }
                     catch (error) {
                         vscode.window.showErrorMessage(`Failed to parse tasks.json file`);
+                        return resolve({ updateTasksJson: false });
                     }
 
                     let buildTasks = getBuildTasks(tasksConfiguration);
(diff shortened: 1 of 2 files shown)


src/assets.ts
```diff
@@ -356,19 +385,18 @@ function getBuildOperations(tasksJsonPath: string) {
...
+                    let tasksConfiguration: tasks.TaskConfiguration;
 
                     try {
-                        const tasksJson: tasks.TaskConfiguration = tolerantParse(text);
-                        buildTask = tasksJson.tasks.find(td => td.taskName === 'build');
+                        tasksConfiguration = tolerantParse(text);
                     }
                     catch (error) {
                         vscode.window.showErrorMessage(`Failed to parse tasks.json file`);
-                        buildTask = undefined;
```
gregg-miskelly: If we catch an exception here, we will still call getBuildTasks. Is that what we want?

DustinCampbell: probably not. Good catch!

DustinCampbell:
Look better @gregg-miskelly?

gregg-miskelly:
LGTM

PR #1169, opened 2017-02-01
Allow configuring any supported Omnisharp log level

### LINKED ISSUE #993: Support all logging levels via omnisharp.loggingLevel setting
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-preview2-003131)

Product Information:
 Version:            1.0.0-preview2-003131
 Commit SHA-1 hash:  635cf40e58

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
```

VS Code version: 1.7.2
C# Extension version: 1.5.3

## Proposed feature

Allow full control of logging levels via the `omnisharp.loggingLevel` setting. Currently only `default` and `verbose` are supported. It would be good if we could have full control, i.e.

- `fatal`
- `error`
- `warn`
- `info`
- `debug`
- `verbose`

filipw:
Fixes #993 

This PR allows the user to set any log level that Omnisharp supports in the user settings.

The default is `information`.

Note 1: that this is technically a breaking change, since I removed the old `verbose` log level as it doesn't exist in Omnisharp and replaced the allowed values with the ones Omnisharp can understand (if it's critical, we can bring back that `verbose` level and map to `debug` but I think this is a fringe feature and there is no need for such "phase out"). 

Note 2: this only works with latest Omnisharp builds, but that has already been included into the latest C# extension.

### DIFF
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -235,19 +235,16 @@ export class OmniSharpServer {
 
         const solutionPath = launchTarget.target;
         const cwd = path.dirname(solutionPath);
+        this._options = Options.Read();
+        
         let args = [
             '-s', solutionPath,
             '--hostPID', process.pid.toString(),
             '--stdio',
             'DotNet:enablePackageRestore=false',
-            '--encoding', 'utf-8'
+            '--encoding', 'utf-8',
+            '-loglevel', this._options.loggingLevel
         ];
-
-        this._options = Options.Read();
-
-        if (this._options.loggingLevel === 'verbose') {
-            args.push('-v');
-        }
 
         this._logger.appendLine(`Starting OmniSharp server at ${new Date().toLocaleString()}`);
         this._logger.increaseIndent();
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -243,7 +243,7 @@ export class OmniSharpServer {
             '--stdio',
             'DotNet:enablePackageRestore=false',
             '--encoding', 'utf-8',
-            '-loglevel', this._options.loggingLevel
+            '--loglevel', this._options.loggingLevel
         ];
 
         this._logger.appendLine(`Starting OmniSharp server at ${new Date().toLocaleString()}`);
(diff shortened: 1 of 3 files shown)


src/omnisharp/server.ts
```diff
@@ -235,20 +235,17 @@ export class OmniSharpServer {
...
+        this._options = Options.Read();
+        
         let args = [
             '-s', solutionPath,
             '--hostPID', process.pid.toString(),
             '--stdio',
             'DotNet:enablePackageRestore=false',
-            '--encoding', 'utf-8'
+            '--encoding', 'utf-8',
+            '-loglevel', this._options.loggingLevel
```
DustinCampbell: Should this be `--loglevel'? If not, we added the option incorrectly in OmniSharp. Only short arguments should have a single dash. If we did it incorrectly, could we fix that up in omnisharp-roslyn?

src/omnisharp/server.ts
```diff
@@ -235,20 +235,17 @@ export class OmniSharpServer {
...
         ];
 
-        this._options = Options.Read();
-
-        if (this._options.loggingLevel === 'verbose') {
-            args.push('-v');
-        }
```
DustinCampbell: To avoid breaking people, could we just map 'verbose' to 'debug'?

DustinCampbell: Seems like an easy thing to do, and it's consistent with other changes to options in the extension. Just implement a mapping from 'verbose' to 'debug' in Options.read().

src/omnisharp/server.ts
filipw: you are right this should have a double dash cause it's the full word

filipw: https://github.com/OmniSharp/omnisharp-roslyn/pull/743

DustinCampbell:
OK, https://github.com/OmniSharp/omnisharp-roslyn/pull/743 is merged, correcting '-loglevel' to '--loglevel'. I'll merge this once I take a new omnisharp-roslyn.

filipw:
thank you!

DustinCampbell:
OK. I'm going to update to the latest omnisharp-roslyn release this morning and will merge this then.

PR #1170, opened 2017-02-01
Rajkumar42/wsl/pipefix

rajkumar42:
VSCode cannot find the path "c:\windows\system32\bash.exe" as bash.exe is only available on 64bit OS. 
It can be invoked from "c:\windows\sysnative\bash.exe", so adding "c:\windows\sysnative" to path if we identify VSCode is running in windows and doesn't have it in the path.


rajkumar42:
@gregg-miskelly @DustinCampbell @WardenGnaw please review. This fixes remote pipe for WSL.

src/features/processPicker.ts
```diff
@@ -389,33 +390,57 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+            let paths : String[] = currentPath.split(";");
+            for (let path of paths) {
+                if (path.toLowerCase() === sysnative) {
+                    return {'PATH' : currentPath };
+                }
+            }
+        
+            return {'PATH' : currentPath + ";" + sysnative };            
+        }
+    });
```
gregg-miskelly: In the non-Windows case you aren't returning anything. Will that mean that line 419 sees `newEnv` as undefined? Might want to at least be a bit more specific.

src/features/processPicker.ts
```diff
@@ -389,33 +390,57 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+            let windir : String = process.env.WINDIR;
+            let sysnative : String = windir.toLowerCase() + "\\sysnative";
+            let paths : String[] = currentPath.split(";");
+            for (let path of paths) {
+                if (path.toLowerCase() === sysnative) {
+                    return {'PATH' : currentPath };
+                }
+            }
+        
+            return {'PATH' : currentPath + ";" + sysnative };
```
gregg-miskelly: Does node merge in the current environment block? Or will this result in an environment block that only defines 'PATH' and nothing else? Maybe try launching VS Code from a command prompt with an extra environment variable and then launch `c:\windows\syswow64\cmd.exe /c set` to see if it handles this case correctly.

src/features/processPicker.ts
```diff
@@ -389,33 +390,57 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+            let sysnative : String = windir.toLowerCase() + "\\sysnative";
+            let paths : String[] = currentPath.split(";");
+            for (let path of paths) {
+                if (path.toLowerCase() === sysnative) {
+                    return {'PATH' : currentPath };
+                }
+            }
+        
+            return {'PATH' : currentPath + ";" + sysnative };            
+        }
```
WardenGnaw: Do we want to make it clear that if it does not enter the if block, the env is null?

src/features/processPicker.ts
WardenGnaw: Want to return null here. It will be the same as not including an env at all in the ExecOptions.

src/features/processPicker.ts
rajkumar42: Just to be consistent, i'm passing the original environment variables. And for the WSL scenario it should add the sysnative to the end of the path.

src/features/processPicker.ts
```diff
@@ -389,33 +390,53 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
...
+function GetSysNativePathIfNeeded() : Promise<Object> {
+    return PlatformInformation.GetCurrent().then(platformInfo => { 
+        let env = process.env;
+        if (platformInfo.isWindows && platformInfo.architecture === "x86_64") {
+            let sysnative : String = process.env.WINDIR + "\\sysnative";
+            env.Path = process.env.PATH + ";" + sysnative;                   
+            return env;            
+        }
+        else {
+            return env;
```
WardenGnaw: nit: remove else block and return statement in if. Just have return env at the end of the function.

src/features/processPicker.ts
```diff
@@ -389,33 +390,53 @@ function execChildProcess(process: string, workingDirectory: string): Promise<st
     });
 }
 
+// VSCode cannot find the path "c:\windows\system32\bash.exe" as bash.exe is only available on 64bit OS. 
+// It can be invoked from "c:\windows\sysnative\bash.exe", so adding "c:\windows\sysnative" to path if we identify
+// VSCode is running in windows and doesn't have it in the path.
+function GetSysNativePathIfNeeded() : Promise<Object> {
```
WardenGnaw: Seems like process.env is an 'any' type.

src/features/processPicker.ts
rajkumar42: Fixed.

PR #1171, opened 2017-02-02
Removing trace and trace response from logging

WardenGnaw:
New system only supports engine logging

@wesrupert @gregg-miskelly  @chuckries

### DIFF
--- a/package.json
+++ b/package.json
@@ -714,7 +714,7 @@
                   },
                   "pipeArgs": {
                     "type": "array",
-                    "description": "Command line arguments passed to the pipe program.",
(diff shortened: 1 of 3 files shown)


wesrupert:
@gregg-miskelly, this isn't an issue of missing support, but of merged support. Since we will no longer have the mi layer, the `engineLogging` feature no longer had any meaning. `trace` also provided little value by itself without `traceResponse`, so those two features were also merged, leaving us with one logging type that covers all three legacy scenarios. Considering "trace" and "traceResponse" have little semantic meaning compared to "engineLogging", I consolidated them all under engineLogging instead. Hence the "removal" of trace/traceResponse; their behavior has been moved to engineLogging, and not removed.

gregg-miskelly:
@wesrupert I see. That makes sense, but we don't want to merge this yet -- we should wait till we are about ready to switch to the new architecture.

gregg-miskelly:
@WardenGnaw And this one too

PR #1176, opened 2017-02-03
.NET: Restore Packages command should work with .NET Core .csproj projects

DustinCampbell:
fixes #1175

PR #1181, opened 2017-02-03
Fix 'dotnet restore' when the extension tries to run it for a specific project

DustinCampbell:
I fixed running restoring packages in one scenario in #1176 but missed another.


DustinCampbell:
I don't think this is contentious. I'm going to go ahead and merge.

PR #1186, opened 2017-02-06
Update build task to launch "npm run compile"

DustinCampbell:
This uses the new tasks.json support in VS Code 1.9 to allow tasks to run different commands. So, now you should be able to build with <kbd>Cmd+Shift+B</kbd> in VS Code.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,8 +1,14 @@
 {
 	"version": "0.1.0",
-	"command": "gulp",
-	"isShellCommand": true,
 	"tasks": [
+		{
+			"taskName": "build",
+			"command": "npm",
+			"isShellCommand": true,
+			"args": ["run", "compile"],
+			"showOutput": "always",
+			"isBuildCommand": true
+		},
 		{
 			"taskName": "test",
 			"showOutput": "always",
@@ -10,7 +16,9 @@
 		},
 		{
 			"taskName": "tslint",
-			"args": [],
+			"command": "gulp",
+			"isShellCommand": true,
+			"args": ["tslint"],
 			"problemMatcher": {
 				"owner": "tslint",
 				"fileLocation": [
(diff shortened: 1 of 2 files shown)


.vscode/tasks.json
```diff
@@ -1,16 +1,24 @@
...
+			"command": "npm",
+			"isShellCommand": true,
+			"args": ["run", "compile"],
+			"showOutput": "always",
+			"isBuildCommand": true
+		},
+		{
 			"taskName": "test",
 			"showOutput": "always",
 			"isTestCommand": true
```
gregg-miskelly: Is the 'test' command still correct? I would have thought it would need a `command` element now that it is gone from the root.

DustinCampbell: It does. I'll take a look.

.vscode/tasks.json
```diff
@@ -11,7 +11,10 @@
 		},
 		{
 			"taskName": "test",
+			"command": "echo",
 			"showOutput": "always",
+			"isShellCommand": true,
+			"args": ["Run tests in VS Code by launching the debugg with the 'Launch Tests' configuration."],
```
gregg-miskelly: You are missing the 'er' in 'debugger'.

DustinCampbell:
@gregg-miskelly: Because we use VS Code's infrastructure to run tests (allowing us to access VS Code objects inside tests), the right way to run tests is to launch the debugger with the 'Launch Tests' configuration. I'd _like_ to be able to launch the VS Code debug command from the test task with that configuration, but there doesn't appear to be a way to do that. So, I'm echoing a message to explain how to run tests instead. Does that work for you?

gregg-miskelly:
@DustinCampbell sounds reasonable to me

DustinCampbell:
cool -- thanks!

PR #1199, opened 2017-02-07
Fix dotnet restore for project.json

DustinCampbell:
In fixing dotnet restore for .csproj, I unintentionally broke it for project.json. The problem here is that OmniSharp returns different information for `Path` with regard to project.json or .csproj. For project.json, it's the directory that the project.json lives in, and for .csproj, it's the actually file path. This change addresses that by adding a `FilePath` property to our `ProjectDescriptor` which represents the real file path and renamed `Path` to `Directory` to be a bit clearer.

cc @jchannon

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -85,7 +85,7 @@ interface Command {
 
 function projectsToCommands(projects: protocol.ProjectDescriptor[]): Promise<Command>[] {
     return projects.map(project => {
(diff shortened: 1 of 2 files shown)


DustinCampbell:
@gregg-miskelly: Would you mind taking a look?

gregg-miskelly:
Taking a look now...

PR #1205, opened 2017-02-08
Update for 1.7.0

DustinCampbell:
I don't have a line in "What's New in 1.7.0" for the debugger. Do you have anything in mind that I should add?


gregg-miskelly:
The changes we made to the debugger this time around were pretty small. I don't think they need to be called out here.

PR #1207, opened 2017-02-08
Adding reqirement for pipeProgram and debuggerPath

WardenGnaw:
When using PipeTransport, these two properties are required to be passed in.
This adds a green squiggle prompting the user to add these fields if its
missing.

@rajkumar42 @wesrupert @gregg-miskelly


gregg-miskelly:
@WardenGnaw do you want to reopen this now?

Issue #1208, opened 2017-02-08
Remote attach doesn't show the command line arguments in the process listing


gregg-miskelly:
## Environment data
C# Extension version: 1.7.0-beta5

## Steps to reproduce
- Follow steps on https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes to setup SSH remote. I am remoting from MacOS to Ubuntu 14.
- Bring up the attach to process dialog

## Expected  behavior
Command line arguments should show up

## Actual behavior
Command line arguments are dropped

gregg-miskelly:
This looks to be broken by the zsh fix (commit 84b3a3b204940acf077acf0bf82a26f202101922).

DustinCampbell:
Resolve fixed?

PR #1209, opened 2017-02-08
Fix remote process listing #1208

gregg-miskelly:
The zshell fix had the unintended problem that '$(uname)' was getting evaluated locally before being sent over the pipe, which meant that remote process listing didn't work from macOS to Linux. This fixes it.


gregg-miskelly:
@wardengnaw please review and give it a test on your zsh machine.

gregg-miskelly:
@rajkumar42 can you give this a quick test for WSL attach?

PR #224, opened 2016-04-20
Move project.json dependency completion & hover to vscode-omnisharp

aeschli:
Until now the code that drove intellisense for dependencies in project.json was sitting in the JSON extension.
 We want the omnisharp extension to take ownership.
 This pull request implements a completion provider and a hover provider using the newly created 'jsonc-parser' node module that does the JSON file processing.
 It gets nuget packages from the nuget API, but is not aware of any nuget configuration files.

In the future the omnisharp extension can also consider to power the completions and hovers directly from the omnisharp server.

We will remove the project.json completion support from the JSON extension for the coming release (end of April).

PR #1214, opened 2017-02-09
[WIP] - csproj intellisense

### LINKED ISSUE #1156: csproj intellisense plans?
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-rc4-004706)

Product Information:
 Version:            1.0.0-rc4-004706
 Commit SHA-1 hash:  e38bc4950c

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.11
 OS Platform: Darwin
 RID:         osx.10.11-x64
 Base Path:   /usr/local/share/dotnet/sdk/1.0.0-rc4-004706

```
VS Code version: `1.8.1 ee428b0eead68bf0fb99ab5fdc4439be227b6281`
C# Extension version: 1.6.2

## Steps to reproduce
Edit a netcore csproj

## Expected  behavior
Intellisense when trying to get nuget package names and versions.  This is currently possible in project.json.  Editing this file manually without intellisense is a bit painful.

## Actual behavior
No behaviour its just editing a text file

cmurczek:
preliminary task list to fix #1156 

- [x] familiarize with NuGet V4 API
NuGet Gallery still runs on V3. I will get the feeds from the user scoped nuget.config, that way it'll "auto-update" as soon as the new api is referenced there. I'll put a test in for the feed name to make sure smthg breaks when that happens.
- [ ] familiarize with how NuGet intellisense works in project.json
- [x] write tests and ~~enhance PackageManager~~ introduce NuGetClient
- [ ] write tests and provider
- [ ] wire up the provider with the extension


test/packageReferenceProvider.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        should();
+        chai.use(chaiAsPromised);
+        let channel: vscode.OutputChannel = vscode.window.createOutputChannel("test");
+        logger = new Logger(text => channel.append(text));
+        PlatformInformation.GetCurrent().then((pi) => platformInfo = pi);
+    });
+
+    test('Retrieve packages from NuGet Gallery', () => {
+        let partialPackageId = 'newtonsoft';
+        let target = new PackageManager(platformInfo, logger);
```
DustinCampbell: Note that PackageManager in the VS Code extension has nothing to do with NuGet.

cmurczek:
@DustinCampbell: Travis CI fails when I call `PlatformInformation.GetCurrent()` while setting up my tests. Any idea how to fix this? It's working on my (Windows) machine...

src/nuget.ts
```diff
@@ -0,0 +1,121 @@
...
+                // and this http://gunnarpeipman.com/2014/10/running-asp-net-5-on-linux/
+                // TODO: verify the file actually resides there
+                nuGetConfig = path.join('~', '.config', 'NuGet', 'NuGet.config');
+            }
+
+            logger.appendLine('using active feeds from ' + nuGetConfig);
+
+            fs.readFile(nuGetConfig, 'utf-8', (err, data) => {
+                if (err) {
+                    reject('Reading ' + nuGetConfig + ' failed with error ' + err);
```
DustinCampbell: Note that this will just carry below trying to access `data`. I suspect you want to return here.

DustinCampbell:
PlatformInformation.GetCurrent() calls our `execChildProcess` utility, which relies on `setExtensionPath` being initialized [here](https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/common.ts#L12-L14). This function called as part of extension activation.

Since you're just checking whether you're on windows, you might consider just checked `process.platform === 'win32'` instead of using `PlatformInformation`.

DustinCampbell:
You should also be aware of this PR: https://github.com/OmniSharp/omnisharp-vscode/pull/224. We'll be taking this soon to pull project.json completion into C# for VS Code (it lives in VS Code today). We'll want to unify these two to have the same behavior.

cmurczek:
yeah, I'm using that one as a template. The `NuGetClient` class would be my shot at the future improvement that is mention in the PR regarding making the FEED_INDEX_URL configurable.

I'll refactor some more to mock away as many dependencies as possible and the push another update.

DustinCampbell:
> I'll refactor some more to mock away as many dependencies as possible and the push another update.

Please don't over mock if you feel like you're going too far. After all, the tests run inside of VS Code to provide access to various VS Code services. So, if you *must* use `PlatformInformation` (which we don't use everywhere either -- only when we need bitness and Linux distro information), it's just a matter of ensuring that the extension is activated. I'm not sure what that looks like precisely, but it's certainly feasible.

cmurczek:
And we're back on track. Thx for the heads up regarding test environment. I was coming from the fact that, the tests as they were, were dependent on an actual nuget.config file being present on the target system. Which in case of Travis CI is probably not the case. After refactoring processing of the config is separated from loading it from the file, so we're good, even on servers without any nuget installation/config.

There's however the question of testing the `CompletionItemProvider` for `PackageReference` elements. This will derive from `AbstractProvider` which requires me to pass an instance of `OmniSharpServer`. Is there a reasonable way to set up the infrastructure for such tests, or is that the reason why there aren't any tests for all the other providers? If so, I'd keep the `PackageReferenceCompletionProvider` as thin as possible on top of the NuGetClient.

DustinCampbell:
I wouldn't bother deriving from `AbstractProvider`. It really doesn't buy you much and this shouldn't have any OmniSharp dependencies anyway.

DustinCampbell:
I just wanted to point out that I've updated the PR for moving project.json completion and hover support into C# for VS Code: https://github.com/OmniSharp/omnisharp-vscode/pull/1236. I'll likely be merging this soon. When I do, you might want to rebase onto the latest master.

cmurczek:
Thx for letting me know. Will update when you signal. I guess that means I'll be tasked with unifying both of them too?

On a different note: What's your opinion on running live http-requests during tests? Right now Mocha flags them as _slow_, <1s but still. If it bothers you I could look into stubbing https calls using sinon, playing back canned responses. Might be worth the work anyway for the consistency of test data alone. Let me know.

DustinCampbell:
I'm not expecting you to unify them. They can be separate for now and we can a work item to unify them later. You can stay focused. :smile:

DustinCampbell:
FYI that I merged the project.json IntelliSense yesterday.

NinoFloris:
Would this be easy to extend now @DustinCampbell?

NinoFloris:
This being the project.json infra you already merged?

borgdylan:
Will this only read PackageReferences via .csproj or will the extension allow loading of any arbitrary extension as long as it is an MSBuild project?

DustinCampbell:
@borgdylan: I'm not sure I understand your question.

borgdylan:
Will it be hardcoded to project that use ".csproj"? Or will it allow projects that are .".fsproj" etc. Intellisense for NuGet package versions is not C# specific and should be available for any project or else put in a separate extension that is generic.

DustinCampbell:
@borgdylan: For C# for VS Code, it'll be specific to .csproj.

NinoFloris:
So... Is this an abandoned project?

DustinCampbell:
It seems like it might be. I'm awaiting @mlorbetske to deliver me a NuGet package of https://github.com/dotnet/ProjFileTools so we can use that within OmniSharp and provide completion to C# for VS Code.

mlorbetske:
Sorry for the delay, I've been swamped of late. I got the go ahead to make a package for the project file tools earlier this week, just need to update the build and find someone to upload it to NuGet under one of the official accounts. I could also make it available on MyGet earlier (tonight maybe?) if that'd help.

mlorbetske:
Filed https://github.com/dotnet/ProjFileTools/issues/10 to get the packaging going as part of the official build.

For the time being, the package is available as ProjectFileTools.NuGetSearch v1.0.0 on https://dotnet.myget.org/F/temp-projfiletools/api/v3/index.json.

hindsight20-20:
Sorry I went AWOL for a couple of weeks. I'll have a couple of hours to kill while traveling tomorrow. I could try to wrap it up then, but I guess using the same told as VS2017 would be preferable. @DustinCampbell let me know

cmrigney:
Any progress?  This would be very useful.

SoheilAlizade:
any plan? I Wait for this thanks :)

tintoy:
Hi.

Just stumbled across this PR and, in case anybody's still wondering about intellisense for .csproj files (or any MSBuild project file, really), there's a separate extension that does this:

https://marketplace.visualstudio.com/items?itemName=tintoy.msbuild-project-tools

DustinCampbell:
Given that it's been a year, I'm going to go ahead and close this PR. @tintoy's msbuild-project-tools extension is an excellent alternatively.

PR #1236, opened 2017-02-13
Move project.json dependency completion & hover to vscode-omnisharp

DustinCampbell:
This change adds the project.json completion and hover support to C# for Visual Studio Code, allowing it to be removed from VS Code.

This PR supersedes https://github.com/OmniSharp/omnisharp-vscode/pull/224.

cc @aeschli


DustinCampbell:
Hey @aeschli: Both jsonc-parser and request-light appear to be designed for VS Code extension authors. Is there a list somewhere of packages intended for VS Code extensions?

package.json
```diff
@@ -33,9 +33,11 @@
     "fs-extra": "^1.0.0",
     "http-proxy-agent": "^1.0.0",
     "https-proxy-agent": "^1.0.0",
+    "jsonc-parser": "^0.3.0",
```
gregg-miskelly: Do we need to update our ThirdPartyNotices.txt for this?

DustinCampbell: I don't *think* so. That's actually produced by the VS Code team and published by @aeschli [here](https://www.npmjs.com/package/jsonc-parser).

DustinCampbell:
Any comments here @aeschli?

aeschli:
@DustinCampbell Good point, we don't have such a list. I created https://github.com/Microsoft/vscode-docs/issues/846

aeschli:
jsonc-parser and request-light are both owned by Microsoft. No need to list in the ThirdPartyNotices.

DustinCampbell:
@gregg-miskelly: Any other feedback? If not, I'm going to go ahead and take this so that @aeschli can remove the code on the VS Code side for their March release.

@aeschli: Currently, I find that this causes duplicate hover tooltips due to having HoverProviders in both VS Code and C# for VS Code. However, the experience isn't *too* terrible, and it's a point in time problem.

src/features/json/projectJSONContribution.ts
```diff
@@ -0,0 +1,235 @@
...
+        }
+        return null;
+    }
+
+    public collectDefaultSuggestions(resource: string, result: ISuggestionsCollector): Thenable<any> {
+        let defaultValue = {
+            'version': '{{1.0.0-*}}',
+            'dependencies': {},
+            'frameworks': {
+                'dnx451': {},
```
gregg-miskelly: Should we update these to something more modern?

DustinCampbell: Yes. We can do it later though. These will still work for now.

DustinCampbell:
@aeschli: Could you cc me on the corresponding change in VS Code?

aeschli:
@DustinCampbell I removed the project.json support from VSCode. This will be part of the February release (in about 2 weeks)

DustinCampbell:
Got it. Thanks!

DustinCampbell:
Also, @aeschli, sorry that took so long!

PR #1238, opened 2017-02-14
Remove "or use Command Prompt"

joseph4tw:
As noted by @rajkumar42 there is a command that won't work in Command Prompt (`cd ~`). Removing the option to use command prompt to avoid confusing other readers.


gregg-miskelly:
Sorry, just noticed that no one merged this. I will now.

PR #1247, opened 2017-02-16
discover CSX files as launch targets

filipw:
This PR enables CSX to be discovered as valid OmniSharp launch points. 
Allows CSX intellisense / language services to be used without a "fake" `project.json` that's currently needed (for OmniSharp to light up).

screenshot 2017-02-15 23 46 16

### DIFF
--- a/src/features/status.ts
+++ b/src/features/status.ts
@@ -26,7 +26,8 @@ let defaultSelector: vscode.DocumentSelector = [
     'csharp', // c#-files OR
     { pattern: '**/project.json' }, // project.json-files OR
     { pattern: '**/*.sln' }, // any solution file OR
-    { pattern: '**/*.csproj' } // an csproj file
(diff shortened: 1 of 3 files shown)


DustinCampbell:
Is this the right experience? Should OmniSharp be launched on a particular .csx? Or should all csx files just be loaded by default?

filipw:
good point, OmniSharp will just load all the CSX files starting from the entry directory, so we don't need to launch OmniSharp at a specific CSX file.
So maybe the best idea is to always include a root-level single "catch-all" CSX launch target? 

I still think the launch target needs to be there for OmniSharp to light up at all; in many cases (i.e. CSX files under folders containing sln/csproj/project.json) it wouldn't be needed at all, but in case those are not there at all (i.e. folder is CSX only or CSX is outside of the slns/csproj/project.json) it's required

filipw:
I pushed this change, let me know if it makes more sense now

PR #1284, opened 2017-03-03
Clean up and adding required fields.

WardenGnaw:
Requiring pipeProgram and debuggerPath to PipeTransport.
Removing trace and trace response as they have been merged into engine
logging.

This replaces: #1171 #1207 

@gregg-miskelly

### DIFF
--- a/package.json
+++ b/package.json
@@ -703,7 +703,6 @@
                 "description": "When present, this tells the debugger to connect to a remote computer using another executable as a pipe that will relay standard input/output between VS Code and the .NET Core debugger backend executable (clrdbg).",
                 "type": "object",
                 "required": [
-                  "pipeProgram",
                   "debuggerPath"
                 ],
                 "default": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -696,22 +696,16 @@
...
-                    "description": "Optional flag to determine whether diagnostic adapter command and response tracing should be logged to the output window.",
-                    "default": false
                   }
                 }
               },
               "pipeTransport": {
                 "description": "When present, this tells the debugger to connect to a remote computer using another executable as a pipe that will relay standard input/output between VS Code and the .NET Core debugger backend executable (clrdbg).",
                 "type": "object",
+                "required": [
+                  "pipeProgram",
```
gregg-miskelly: >"pipeProgram", [](start = 18, length = 14)

We might want to leave `pipeProgram` optional since it is possible to specify it for each OS rather than using a common one on all OSs

src/tools/OptionsSchema.json
```diff
@@ -44,6 +44,7 @@
         },
         "PipeTransport": {
             "type": "object",
+            "required": ["pipeProgram", "debuggerPath"],
```
gregg-miskelly: >pipeProgram [](start = 26, length = 11)

I guess my comment really belongs here.

Issue #1318, opened 2017-03-13
Need a way to avoid quoting 'debuggerPath' when using pipe transport


weinand:
_From @tanaka-takayoshi on March 13, 2017 9:7_

- VSCode Version:
 1.11.0-insider
- OS Version:
RHEL 7.3

Steps to Reproduce:
1. Generate launch.json to set up remote debug configuration.

```
{
  "version": "0.2.0",
  "configurations": [        
    {
        "name": ".NET Core Docker Remote Attach",
        "type": "coreclr",
        "request": "attach",
        "processId": "21",
        "pipeTransport": {
            "pipeProgram": "docler",
            "pipeArgs": [ "exec", "remotedebug-1-mispn"],
            "debuggerPath": "/opt/app-root/src/clrdbg/clrdbg",
            "pipeCwd": "${workspaceRoot}"
        },
        "sourceFileMap": {
            "/opt/app-root/src": "${workspaceRoot}"
        }
    }
  ]
}
```

2. Start debug ".NET Core Docker Remote Attach"

3. Failed to execute docker command. This is because VS Code escape the debuggerPath with parameters.
```
"docker" exec remotedebug-1-mispn "/opt/app-root/src/clrdbg/clrdbg --interpreter=mi"
```

The docker command recognizes "/opt/app-root/src/clrdbg/clrdbg --interpreter=mi" as one file path, then failed. I'm not sure which version, but the previous version (maybe 1.10.x-insider or 1.9.x-insider) VS Code didn't escape with option but not escaped like below.
```
"docker" exec remotedebug-1-mispn /opt/app-root/src/clrdbg/clrdbg --interpreter=mi
```

I submitted to StackOverflow as a question. But, I found this issue occurred with not only "oc (OpenShit CLI)" command but also "docker" command. So I decided to submit this issue as a bug.
http://stackoverflow.com/questions/42731558/how-not-to-escape-debuggerpath-of-pipetransport-in-launch-json

This issue prevents me from executing remote debug from VS Code to docker.

_Copied from original issue: Microsoft/vscode#22514_

weinand:
@tanaka-takayoshi please file .NET related issues against https://github.com/OmniSharp/omnisharp-vscode/issues

gregg-miskelly:
@tanaka-takayoshi this change was intentional as we had to deal with other transports that needed the other quoting behavior, and we wanted to be able to support scripts and pass though as many command line args as we needed. 

[...]

PR #1322, opened 2017-03-14
Adding QuoteArgs Flag

WardenGnaw:
Addressing #1318 

@gregg-miskelly

### DIFF
--- a/package.json
+++ b/package.json
@@ -727,6 +727,11 @@
                     "description": "Environment variables passed to the pipe program.",
                     "default": {}
                   },
+                  "quoteArgs": {
+                    "type": "boolean",
+                    "description": "Flag to enable or disable quotes around 'debuggerPath'",
+                    "default": true
+                  },
                   "windows": {
                     "description": "Windows-specific pipe launch configuration options",
                     "default": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -727,6 +727,11 @@
                     "description": "Environment variables passed to the pipe program.",
                     "default": {}
                   },
+                  "quoteArgs": {
+                    "type": "boolean",
+                    "description": "Flag to enable or disable quotes around 'debuggerPath'",
```
gregg-miskelly: How about this for the description:
   Should arguments that contain characters that need to be quoted (ex: spaces) be quoted? Defaults to 'true'. If set, the debugger command will no longer be automatically quoted.

src/tools/OptionsSchema.json
```diff
@@ -32,6 +32,11 @@
                     },
                     "default": []
                 },
+                "quoteArgs": {
+                    "type": "boolean",
+                    "description": "Should arguments that contain characters that need to be quoted (example: spaces) be quoted? Defaults to 'true'. If set, the debugger command will no longer be automatically quoted.",
```
gregg-miskelly: >set [](start = 152, length = 3)

I think you want 'to false' here.

PR #1349, opened 2017-03-28
Updating debugger to 1.9.0 which has the macos Sierra fix.

rajkumar42:


rajkumar42:
@gregg-miskelly @chuckries  please review.

Issue #1351, opened 2017-03-29
Stop generating windows style paths in launch.json, task.json


chuckries:
When `launch.json` and `tasks.json` are generated on Windows they are generated with windows style paths (backslashes). These paths won't work on Linux or Mac, but Linux/Mac style paths work on Windows. We should only generated paths with forward slashes. 

examples tasks.json generated on windows:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": ".NET Core Launch (console)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "build",
            "program": "${workspaceRoot}\\bin\\Debug\\netcoreapp1.1\\scratch.dll",
            "args": [],
            "cwd": "${workspaceRoot}",
            "console": "internalConsole",
            "stopAtEntry": false,
            "internalConsoleOptions": "openOnSessionStart"
        },
        {
            "name": ".NET Core Attach",
            "type": "coreclr",
            "request": "attach",
            "processId": "${command:pickProcess}"
        }
    ]
}
```

Changing to this still works on windows:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": ".NET Core Launch (console)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "build",
            "program": "${workspaceRoot}/bin/Debug/netcoreapp1.1/Sudoku.dll",
            "args": [],
            "cwd": "${workspaceRoot}",
            "externalConsole": false,
            "stopAtEntry": false,
            "internalConsoleOptions": "openOnSessionStart"
        },
        {
            "name": ".NET Core Attach",
            "type": "coreclr",
            "request": "attach",
            "processId": "${command:pickProcess}"
        }
    ]
}
```

PR #1354, opened 2017-03-29
Always use posix paths in asset generation

chuckries:
We were using native 'path.join' when creating paths for VS Code assets
(tasks.json, launch.json). On windows this would produce windows style
paths. If the assets are checked into source control and used on posix
systems, they will break. Posix paths will run perfectly fine on
windows, so always use posix paths.


chuckries:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1351

DustinCampbell:
Feel free to update the change log if you like.

Issue #817, opened 2016-10-13
C# generated launch.json template should include help comment


gregg-miskelly:
Currently when the C# extension generates a launch.json we don't include any reference to help. We should probably have something. Here is what the node version has:

![node-launch-json](https://cloud.githubusercontent.com/assets/11078489/19357154/37d53e16-9125-11e6-8ad8-a853b1f3aeb0.png)

Here is how they add the comments:
https://github.com/Microsoft/vscode-node-debug/blob/dead56e043b82387c7f3d6c04d0bdbd66e65db23/src/node/extension.ts#L222

PR #1358, opened 2017-03-31
Adding help text to created launch.json

WardenGnaw:
#817

### DIFF
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -1,0 +1,58 @@
+# Configurating launch.json for C# debugging
+The launch.json file is used to configure the debugger in Visual Studio Code.
+
+Visual Studio Code generates a launch.json with almost all of the required information. To get started debugging you need to fill in the program field with the path to the executable you plan to debug. This must be specified for both the launch and attach (if you plan to attach to a running instance at any point) configurations.
+
+The generated file contains two sections. One that configures debugging for launch and a second that configures debugging for attach.
+
+If your workspace has only one launchable project, the C# extension will offer to automatically generate this file. If you missed this prompt, you can force the generation by executing the command `.NET: Generate Assets for Build and Debug` from the VS Code command pallet.
+
+# Configurating VS Code's debugging behavior
+
+## PreLaunchTask
+The `preLaunchTask` field runs the associated taskName in tasks.json before debugging your program. You can get the default build prelaunch task by executing the command `Tasks: Configure Tasks Runner` from the VS Code command pallet.
+
+This will create a task that runs `dotnet build`. You can read more about tasks at [https://code.visualstudio.com/docs/editor/tasks](https://code.visualstudio.com/docs/editor/tasks).
+
+## Just My Code
+You can optionally disable justMyCode by setting it to "false". You should disable Just My Code when you are trying to debug into a library that you pulled down which doesn't have symbols or is optimized.
+
+    "justMyCode":false*
+
+Just My Code is a set of features that makes it easier to focus on debugging your code by hiding some of the details of optimized libraries that you might be using, like the .NET Framework itself. The most important sub parts of this feature are --
+
+* User-unhandled exceptions: automatically stop the debugger just before exceptions are about to be caught by the framework
+* Just My Code stepping: when stepping, if framework code calls back to user code, automatically stop.
+
+## Source File Map
+You can optionally configure a file by file mapping by providing map following this schema:
+
+    "sourceFileMap": {
+        "C:\foo":"/home/me/foo"
+    }
+
+## Symbol Path
+You can optionally provide paths to symbols following this schema:
+
+    "symbolPath": [ "/Volumes/symbols" ]
+
+## Environment variables
+Environment variables may be passed to your program using this schema:
+
+    "env": {
+        "myVariableName":"theValueGoesHere"
+    }
+
+## Console (terminal) window
+By default, processes are launched with their console output (stdout/stderr) going to the VS Code Debugger Console. This is useful for executables that take their input from the network, files, etc. But this does NOT work for applications that want to read from the console (ex: `Console.ReadLine`). For these applications, use a setting such as the following:
+
+    "console": "integratedTerminal"
+
+When this is set to `integratedTerminal` the target process will run inside [VS Code's integrated terminal](https://code.visualstudio.com/docs/editor/integrated-terminal). Click the 'Terminal' tab in the tab group beneath the editor to interact with your application.
+
+When this is set to `externalTerminal` the target process will run in a separate terminal.
+
+## Stepping into properties and operators
+The debugger steps over properties and operators in managed code by default. In most cases, this provides a better debugging experience. To change this and enable stepping into properties or operators add:
+
+    "enableStepFiltering": false
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -1,11 +1,14 @@
 # Configurating launch.json for C# debugging
 The launch.json file is used to configure the debugger in Visual Studio Code.
 
-Visual Studio Code generates a launch.json with almost all of the required information. To get started debugging you need to fill in the program field with the path to the executable you plan to debug. This must be specified for both the launch and attach (if you plan to attach to a running instance at any point) configurations.
-
+Visual Studio Code generates a launch.json with almost all of the required information. 
+If your workspace has only one launchable project, the C# extension will offer to automatically generate this file. 
+If you missed this prompt, you can force the generation by executing the command `.NET: Generate Assets for Build and Debug` from the VS Code command pallet.
 The generated file contains two sections. One that configures debugging for launch and a second that configures debugging for attach.
 
-If your workspace has only one launchable project, the C# extension will offer to automatically generate this file. If you missed this prompt, you can force the generation by executing the command `.NET: Generate Assets for Build and Debug` from the VS Code command pallet.
+If you have more than one launchable project, then you will need to modify your launch.json file by hand. 
+Visual Studio Code will still generate a basic template, but you will need to fill in the 'program' field to point at the executable dll that you would like to debug.
+
 
 # Configurating VS Code's debugging behavior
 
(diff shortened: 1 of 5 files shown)


src/assets.ts
```diff
@@ -488,7 +481,18 @@ function addLaunchJsonIfNecessary(generator: AssetGenerator, operations: Operati
...
         const launchJson = generator.createLaunchJson(isWebProject);
-        const launchJsonText = JSON.stringify(launchJson, null, '    ');
+
+        const configurationsMassaged = JSON.stringify(launchJson, null, '   ').split('\n').map(line => '   ' + line).join('\n').trim();
+
+        const launchJsonText = [
+			'{',
+			'\t// Use IntelliSense to find out which attributes exist for C# debugging',
+			'\t// Use hover for the description of the existing attributes',
+			'\t// For further information visit https://go.microsoft.com/fwlink/?linkid=830387',
```
gregg-miskelly: I think we should come up with our own URL/Documentation. This is the generic VS Code one which isn't nearly as nice as pointing at something C#-extension specific. I will start a mail thread and include you.

src/assets.ts
```diff
@@ -258,24 +258,17 @@ export class AssetGenerator {
     }
 
     public createLaunchJson(isWebProject: boolean): any {
```
gregg-miskelly: We might want to have this return a JSON'ified string instead of an array of objects. My reasons --
1. minor, but 'any' is a shitty type
2. Hopefully someday we will do a similar fix for adding comments inside the launch target. Bonus points if you want to fix this - I am specifically thinking of 'terminal', since that has gotten more than a few questions, and 'program', since that is probably the most important thing, and it may not be obvious that that needs to be updated after retargeting your app at a new version of the framework.

src/assets.ts
```diff
@@ -488,7 +481,18 @@ function addLaunchJsonIfNecessary(generator: AssetGenerator, operations: Operati
 
         const isWebProject = generator.hasWebServerDependency();
         const launchJson = generator.createLaunchJson(isWebProject);
-        const launchJsonText = JSON.stringify(launchJson, null, '    ');
+
+        const configurationsMassaged = JSON.stringify(launchJson, null, '   ').split('\n').map(line => '   ' + line).join('\n').trim();
+
+        const launchJsonText = [
+			'{',
+			'\t// Use IntelliSense to find out which attributes exist for C# debugging',
```
gregg-miskelly: Are you sure you want '\t' here? My launch.json seems to have spaces.

src/assets.ts
```diff
@@ -488,7 +481,18 @@ function addLaunchJsonIfNecessary(generator: AssetGenerator, operations: Operati
 
         const isWebProject = generator.hasWebServerDependency();
         const launchJson = generator.createLaunchJson(isWebProject);
-        const launchJsonText = JSON.stringify(launchJson, null, '    ');
+
+        const configurationsMassaged = JSON.stringify(launchJson, null, '   ').split('\n').map(line => '   ' + line).join('\n').trim();
+
+        const launchJsonText = [
```
DustinCampbell: Why not just use a template string so you don't have to use an array and join it at the end?

E.g.

```TypeScript
const launchJsonText = `
{
    // Use IntelliSense to find out which attributes exist for C# debugging
    // Use hover for the description of the existing attributes
    // For further information visit https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": ${configurationsMassaged}
}`;
```

debugger-launchjson.md
```diff
@@ -0,0 +1,58 @@
+# Configurating launch.json for C# debugging
+The launch.json file is used to configure the debugger in Visual Studio Code.
+
+Visual Studio Code generates a launch.json with almost all of the required information. To get started debugging you need to fill in the program field with the path to the executable you plan to debug. This must be specified for both the launch and attach (if you plan to attach to a running instance at any point) configurations.
```
gregg-miskelly: I would put your 'If your workspace...' sentence here. This sentence is if you using the crappy template.

Then at the end you can add another sentence that says something like:
If you have more than one launchable project, then you will need to modify your launch.json file by hand. Visual Studio code will still generate a basic template, but you will need to fill in the 'program' field to point at the executable dll that you would like to debug.

gregg-miskelly: This sentence is wrong for C#, so delete it.

debugger-launchjson.md
```diff
@@ -0,0 +1,58 @@
...
+
+If your workspace has only one launchable project, the C# extension will offer to automatically generate this file. If you missed this prompt, you can force the generation by executing the command `.NET: Generate Assets for Build and Debug` from the VS Code command pallet.
+
+# Configurating VS Code's debugging behavior
+
+## PreLaunchTask
+The `preLaunchTask` field runs the associated taskName in tasks.json before debugging your program. You can get the default build prelaunch task by executing the command `Tasks: Configure Tasks Runner` from the VS Code command pallet.
+
+This will create a task that runs `dotnet build`. You can read more about tasks at [https://code.visualstudio.com/docs/editor/tasks](https://code.visualstudio.com/docs/editor/tasks).
+
```
gregg-miskelly: We should add documentation for the other fields too. Looks like we are missing: program, args, cwd, stopAtEntry

debugger.md
```diff
@@ -87,49 +86,7 @@ If your code was built on a different computer from where you would like to run
 * **Portable PDBs**: If the code was built on Windows, it might have been built using Windows PDBs instead of portable PDBs, but the C# extension only supports portable PDBs. See the [portable PDB documentation](https://github.com/OmniSharp/omnisharp-vscode/wiki/Portable-PDBs#how-to-generate-portable-pdbs) for more information.
 * **Debug vs. Release**: It is much easier to debug code which has been compiled in the `Debug` configuration. So unless the issue you are looking at only reproduces with optimizations, it is much better to use Debug bits. If you do need to debug optimized code, you will need to disable [justMyCode](#just-my-code) in launch.json.
```
gregg-miskelly: There are a couple of other links in debugger.md that now need to be updated to point at the new file. This '[justMyCode]' link is one of them.

src/assets.ts
```diff
@@ -198,84 +158,89 @@ export class AssetGenerator {
...
-            }
-        };
+    private createWebLaunchConfiguration(): string {
+        return `
+{
+    "name": ".NET Core Launch (web)",
+    "type": "coreclr",
+    "request": "launch",
+    "preLaunchTask": "build",
+    "program": "${this.convertNativePathToPosix(this.computeProgramPath())}",
```
gregg-miskelly: Add the comment for 'program' here too

test/assets.test.ts
```diff
@@ -7,6 +7,7 @@ import { should } from 'chai';
 import * as path from 'path';
 import * as protocol from '../src/omnisharp/protocol';
 import { AssetGenerator } from '../src/assets';
+import * as json from 'comment-json';
```
gregg-miskelly: >json [](start = 12, length = 4)

We use 'tolerantParse' inside the extension. We might want to use that here also as I believe this is what VS Code uses.

DustinCampbell: Please don't use 'comment-json'. [jsonc-parser](https://www.npmjs.com/package/jsonc-parser) is provided by VS Code to parse JSON files with comments and we already list it as a dependency. Please use that rather than adding another dependency for the same functionality.

debugger-launchjson.md
```diff
@@ -0,0 +1,110 @@
+# Configurating launch.json for C# debugging
+The launch.json file is used to configure the debugger in Visual Studio Code.
+
+Visual Studio Code generates a launch.json with almost all of the required information. 
+If your workspace has only one launchable project, the C# extension will offer to automatically generate this file. 
+If you missed this prompt, you can force the generation by executing the command `.NET: Generate Assets for Build and Debug` from the VS Code command pallet.
```
gregg-miskelly: >pallet [](start = 150, length = 6)

The spelling of you want is: 'palette'.

debugger-launchjson.md
```diff
@@ -0,0 +1,110 @@
...
+The generated file contains two sections. One that configures debugging for launch and a second that configures debugging for attach.
+
+If you have more than one launchable project, then you will need to modify your launch.json file by hand. 
+Visual Studio Code will still generate a basic template, but you will need to fill in the 'program' field to point at the executable dll that you would like to debug.
+
+
+# Configurating VS Code's debugging behavior
+
+## PreLaunchTask
+The `preLaunchTask` field runs the associated taskName in tasks.json before debugging your program. You can get the default build prelaunch task by executing the command `Tasks: Configure Tasks Runner` from the VS Code command pallet.
```
gregg-miskelly: >pallet [](start = 227, length = 6)

Same here

debugger-launchjson.md
```diff
@@ -0,0 +1,110 @@
...
+
+This will create a task that runs `dotnet build`. You can read more about tasks at [https://code.visualstudio.com/docs/editor/tasks](https://code.visualstudio.com/docs/editor/tasks).
+
+## Program
+The program field is set to the path to the application dll or .NET Core host executable to launch. 
+
+Example: "${workspaceRoot}/bin/Debug/\<target-framework\>/\<project-name.dll\>" where:
+
+    <target-framework>: (example: 'netstandard1.5') This is the framework that the app is being built for. It is set in the project.json file.
```
gregg-miskelly: >     [](start = 0, length = 4)

I think you don't want these indents as that puts this in a code block, and I don't think you want that.

debugger-launchjson.md
```diff
@@ -0,0 +1,110 @@
...
+    <project-name>: (example: 'MyApp') The name of the project being debugged.",
+
+## Cwd
+The working directory of the target process.
+
+## Args
+These are the arguments that will be passed to your program.
+
+## Stop at Entry
+If you need to stop at the entry point of the target, you can optionally set stopAtEntry to be "true".
```
gregg-miskelly: >stopAtEntry [](start = 77, length = 11)

nit: I would surround this with back ticks (`stopAtEntry`)

package.json
```diff
@@ -46,13 +46,15 @@
   },
   "devDependencies": {
     "@types/chai": "^3.4.34",
+    "@types/comment-json": "^1.1.0",
```
gregg-miskelly: Do we still need this?

DustinCampbell: I missed that. this should go away.

Issue #1024, opened 2016-12-09
C# extension should provide launch.json templates


gregg-miskelly:
With the November 2016 release of VS Code, there is now support for launch.json templates. See https://github.com/Microsoft/vscode/issues/16937 for examples.

This is less interesting for the C# extension since we normally generate launch.json, but it is useful in a few cases --
* Users with multiple launch projects that would like multiple launch configurations
* Adding templates for SSH

We should add new templates for local launch, local attach, ssh launch and ssh attach.

WardenGnaw:
Do we want to dynamically create the configuration snippets or just list all possible configurations?

If it is dynamically created, we will be able to filter out non-usable configurations. This also reduces the amount of non-usable choices for the user. 

The image below shows how crowded the list becomes with a couple of extensions installed. 
![image](https://cloud.githubusercontent.com/assets/3953714/24767498/fe8d676e-1ab3-11e7-858f-ae9a9a82157b.png)

If we decide to add 4 configurations multiplied by 2 for Windows/Linux differences it will get crowded.

Sadly, creating snippets via a command, like initial configurations, does not work:  https://github.com/Microsoft/vscode/issues/23961

gregg-miskelly:
Here are the node snippets as an example: https://github.com/Microsoft/vscode-node-debug/blob/master/package.json#L117

gregg-miskelly:
@WardenGnaw should we leave this open to react to a possible VS Code fix in the future to make this better? Or should we close this for now?

WardenGnaw:
I think we can close this one and then create a new issue asking for "dynamic launch.json templates with comments". That can be worked on once Microsoft/vscode#23961 is completed.

PR #1366, opened 2017-04-06
Adding in configuration snippets

WardenGnaw:
#1024

Creating snippets for local launch, local attach, local web launch, remote launch and remote attach.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1131,6 +1131,104 @@
             }
           }
         },
+        "configurationSnippets": [
+            {
+                "label": ".NET: Launch .NET Core App",
+                "description": "Launch a .NET Core App with a debugger.",
+                "body": {
+                    "name": ".NET Core Launch (console)",
+                    "type": "coreclr",
+                    "request": "launch",
+                    "preLaunchTask": "build",
+                    "program": "^\"\\${workspaceRoot}/bin/Debug/<target-framework>/<project-name.dll>\"",
+                    "args": [],
+                    "cwd": "^\"\\${workspaceRoot}\"",
+                    "stopAtEntry": false,
+                    "console": "internalConsole"
+                }
+            },
+            {
+                "label": ".NET: Attach to local .NET Core App",
+                "description": "Attach a debugger to a .NET Core App.",
+                "body": {
+                    "name": ".NET Core Attach",
+                    "type": "coreclr",
+                    "request": "attach",
+                    "processId": "^\"\\${command:pickProcess}\""
+                }
+            },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1131,6 +1131,104 @@
...
+        "configurationSnippets": [
+            {
+                "label": ".NET: Launch .NET Core App",
+                "description": "Launch a .NET Core App with a debugger.",
+                "body": {
+                    "name": ".NET Core Launch (console)",
+                    "type": "coreclr",
+                    "request": "launch",
+                    "preLaunchTask": "build",
+                    "program": "^\"\\${workspaceRoot}/bin/Debug/<target-framework>/<project-name.dll>\"",
```
gregg-miskelly: Should we use their syntax to have the user fill in this path?

package.json
```diff
@@ -1131,6 +1131,104 @@
             }
           }
         },
+        "configurationSnippets": [
+            {
+                "label": ".NET: Launch .NET Core App",
```
gregg-miskelly: Should we put 'Console' in the title?

package.json
```diff
@@ -1131,6 +1131,104 @@
...
+                    "request": "launch",
+                    "preLaunchTask": "build",
+                    "program": "^\"\\${workspaceRoot}/bin/Debug/<target-framework>/<project-name.dll>\"",
+                    "args": [],
+                    "cwd": "^\"\\${workspaceRoot}\"",
+                    "stopAtEntry": false,
+                    "console": "internalConsole",
+                    "pipeTransport": {
+                        "pipeCwd": "^\"\\${workspaceRoot}\"",
+                        "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
```
gregg-miskelly: 'pipeProgram' also?

package.json
gregg-miskelly: Ex: 
    "program": "^\"\\${workspaceRoot}/bin/Debug/${1:<target-framework>/<project-name.dll>}\"",

WardenGnaw:
@gregg-miskelly After looking deep into the vscode source.... You can add comments to the snippets. Instead of using the "body" keyword, use "bodyText". Should we change to that and include comments in our snippets?

gregg-miskelly:
@WardenGnaw unless it is a huge amount of work, I think so.

WardenGnaw:
bodyText will be used once snippets can call commands, so we can reuse the configurations in assets.

PR #1367, opened 2017-04-06
Update OmniSharp and add support for running/debugging tests

DustinCampbell:
There's a bit of a dance here with debugging of 'dotnet test':

* On a legacy .NET CLI (e.g. "preview2" or earlier -- project.json stuff), we request the command-line to launch test test runner and simply launch the debugger with that command-line. This is largely unchanged from what we had before.
* On a newer .NET CLI (e.g. "preview3" or newer -- .csproj stuff), we ask OmniSharp to launch the test runner process and it returns the process id to us. Then, we attach the debugger to the process id and signal OmniSharp that we're ready to start debugging. OmniSharp then signals 'dotnet test' to start running tests.

@gregg-miskelly: I've run into a few issues here that I'd like to get some help with if possible:

1. I initially tried launching the debugger for .csproj projects like we do for project.json projects, but I ran into a snag. Once the debugger launches the testhost process, it immediately runs to completion instead of waiting to be signaled. However, if I launch the process myself (via `Process.Start(...)` in OmniSharp or `child_process.span(...)` in VS Code, the process waits until I signal 'dotnet test'.
2. I was able to see debugging working on Windows, but got an error message ("Couldn't attach to process") on my Mac. Are there issues with local attach I should be aware of? How can I debug an issue like that?

### DIFF
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -139,7 +139,7 @@ export function debugDotnetTest(testMethod: string, fileName: string, testFramew
 
     return serverUtils.debugTestCheck(server, request)
         .then(response => {
-            debugType = response.PreferredDebugType;
+            debugType = response.DebugType;
             return getLaunchConfiguration(server, debugType, fileName, testMethod, testFrameworkName);
         })
         .then(config => {
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -3,10 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-'use strict';
-
-import {OmniSharpServer} from '../omnisharp/server';
-import {toRange} from '../omnisharp/typeConvertion';
+import { OmniSharpServer } from '../omnisharp/server';
+import { toRange } from '../omnisharp/typeConvertion';
 import * as vscode from 'vscode';
 import * as serverUtils from "../omnisharp/utils";
 import * as protocol from '../omnisharp/protocol';
@@ -35,23 +33,28 @@ export function registerDotNetTestDebugCommand(server: OmniSharpServer): vscode.
 
 // Run test through dotnet-test command. This function can be moved to a separate structure
 export function runDotnetTest(testMethod: string, fileName: string, testFrameworkName: string, server: OmniSharpServer) {
-    getTestOutputChannel().show();
-    getTestOutputChannel().appendLine(`Running test ${testMethod}...`);
+    const output = getTestOutputChannel();
 
-    let disposable = server.onTestMessage(e =>
-    {
-        getTestOutputChannel().appendLine(e.Message);
+    output.show();
+    output.appendLine(`Running test ${testMethod}...`);
+
+    const disposable = server.onTestMessage(e => {
+        output.appendLine(e.Message);
     });
 
-    serverUtils
-        .runDotNetTest(server, { FileName: fileName, MethodName: testMethod, TestFrameworkName: testFrameworkName })
-        .then(
-        response => {
+    const request: protocol.V2.RunTestRequest = {
+        Filename: fileName,
+        MethodName: testMethod,
+        TestFrameworkName: testFrameworkName
+    };
+
+    serverUtils.runTest(server, request)
+        .then(response => {
             if (response.Pass) {
-                getTestOutputChannel().appendLine('Test passed \n');
+                output.appendLine('Test passed \n');
             }
             else {
-                getTestOutputChannel().appendLine('Test failed \n');
+                output.appendLine('Test failed \n');
             }
 
             disposable.dispose();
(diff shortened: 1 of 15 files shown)


DustinCampbell:
Scratch the comment about my mac. It seems to be working now.

src/features/dotnetTest.ts
```diff
@@ -35,41 +33,124 @@ export function registerDotNetTestDebugCommand(server: OmniSharpServer): vscode.
...
-                "stopAtEntry": false
+    const request: protocol.V2.DebugTestCheckRequest = {
+        Filename: fileName
+    };
+
+    let debugType: string;
+
+    return serverUtils.debugTestCheck(server, request)
+        .then(response => {
+            debugType = response.DebugType;
```
gregg-miskelly: For me to debug this, is there a way I can force it into launch mode or attach mode without rebuilding OmniSharp?

src/features/dotnetTest.ts
```diff
@@ -35,41 +33,124 @@ export function registerDotNetTestDebugCommand(server: OmniSharpServer): vscode.
...
+            return getLaunchConfiguration(server, debugType, fileName, testMethod, testFrameworkName);
+        })
+        .then(config => {
+            return vscode.commands.executeCommand('vscode.startDebug', config);
+        })
+        .then(() => {
+            if (debugType === "attach") {
```
gregg-miskelly: Is this not used in 'launch' mode because of the bug? Or is it really not supposed to send this for launch?

src/features/dotnetTest.ts
```diff
@@ -35,41 +33,124 @@ export function registerDotNetTestDebugCommand(server: OmniSharpServer): vscode.
...
+            });
+
+            return {
+                name: ".NET Test Launch",
+                type: "coreclr",
+                request: "launch",
+                program: response.Executable,
+                args: args,
+                cwd: response.WorkingDirectory,
+                stopAtEntry: false
```
gregg-miskelly: Should we be including the environment variables also? Or is that collection always empty?

src/features/dotnetTest.ts
DustinCampbell: You should just need to change

```
return getLaunchConfigurationForAttach(server, fileName, testMethod, testFrameworkName);
```

to

```
return getLaunchConfigurationForLaunch(server, fileName, testMethod, testFrameworkName);
```

gregg-miskelly: Might want to add:
// TODO: Need to find out when the attach is really complete from the debugger. This is currently a race.

src/features/dotnetTest.ts
DustinCampbell: You should just need to change

```
return getLaunchConfigurationForAttach(server, fileName, testMethod, testFrameworkName);
```

to

```
return getLaunchConfigurationForLaunch(server, fileName, testMethod, testFrameworkName);
```

src/features/dotnetTest.ts
DustinCampbell: This is just some bad naming on my part. Right now, OmniSharp will report back "attach" for a .NET CLI newer than preview2, and "launch" for a legacy .NET CLI. The legacy .NET CLI code does not use the same signaling.

src/features/dotnetTest.ts
DustinCampbell: I can return them as well, but it my tests they've always been empty.

src/features/dotnetTest.ts
DustinCampbell: yes it is. Not sure how to sort that out either. The debugger seems a bit of a black box from the point of view of the VS Code extension.

DustinCampbell:
I should also mention that, as I've played with it, attaching to the testhost process has a somewhat inferior experience to just launching under the debugger. I think that'd probably be ideal.

src/features/dotnetTest.ts
DustinCampbell: I just verified that making that change results in the launch behavior I described.

gregg-miskelly:
@DustinCampbell @singhsarab here is what seems to be going on (or at least what goes on when I try to debug this) -- 

First, there is a process spawned that with a command line something like this:

    "dotnet.exe" exec "C:\Program Files\dotnet\sdk\1.0.2\vstest.console.dll" --Port:26036 --ParentProcessId:93188

In the case I was debugging, this had a process id of 91372.

Next the debugger is asked to spawn a process that is something like this:

    d:\proj\clitest\bin\Debug\netcoreapp1.1>"C:\Program Files\dotnet\dotnet.exe" exec --runtimeconfig d:\proj\clitest\bin\Debug\netcoreapp1.1\clitest.runtimeconfig.json --depsfile d:\proj\clitest\bin\Debug\netcoreapp1.1\clitest.deps.json C:\Users\greggm\.nuget\packages\microsoft.testplatform.testhost/15.0.0\lib/netstandard1.5/testhost.dll --port 25855 --parentprocessid 91372

The problem is that by the time the second process starts, the first process is no longer running, so it hits an exception:

	System.Diagnostics.Process.dll!System.Diagnostics.Process.GetProcessById(int processId, string machineName)	Unknown
 	System.Diagnostics.Process.dll!System.Diagnostics.Process.GetProcessById(int processId)	Unknown
 	testhost.dll!Microsoft.VisualStudio.TestPlatform.TestHost.DefaultEngineInvoker.WaitForParentProcessExitAsync(int parentProcessId)	Unknown
 	testhost.dll!Microsoft.VisualStudio.TestPlatform.TestHost.DefaultEngineInvoker.Invoke(System.Collections.Generic.IDictionary<string, string> argsDictionary)	Unknown
 	testhost.dll!Microsoft.VisualStudio.TestPlatform.TestHost.Program.Main(string[] args)	Unknown

Presumably the first process should be sticking around, but it doesn't seem to be.

DustinCampbell:
I've got launching working. It seems that OmniSharp was not keeping the "dotnet test" process alive after passing back the process start info to the debugger. Keeping the process alive allows the debugger to successfully launch and hit breakpoints. I now need to figure out a way to signal OmniSharp when debugging is complete so it can tear down the "dotnet test" process.

DustinCampbell:
OK. I've got things working. I allow OmniSharp to continue until the debugged process signals 'dotnet test' that the test run is complete. Since OmniSharp has a socket connection with the 'dotnet test' process, this works OK. _However_, I still need to figure out a way to determine how to tear down the 'dotnet test' process when the user just clicks "stop" on the debugger. That causes the debugged process to be killed and OmniSharp *does* get a message from 'dotnet test' that the test run was aborted. However, that only comes by way of a textual message (i.e. "The test run was aborted"). It never receives another message after that, and OmniSharp doesn't know what the PID of the debugged process was to watch it.

@gregg-miskelly: Any ideas?

DustinCampbell:
Also, there's still a race between launching the debugger and signaling OmniSharp to start the test run.

gregg-miskelly:
@DustinCampbell do you know if VS Code has any events that extensions can listen to for when debug sessions start or stop? If not, I guess I would suggest we have OmniSharp and/or the language service open a named pipe or Unix domain socket and the debugger can send back the process id of the target, and let OmniSharp know when debugging has stopped.

DustinCampbell:
I don't think they have any eventing. I've looked around and found nothing.

I suspect providing someway for the debugger to pass the process ID of the target is going to be the way (yuck! :smile:)

gregg-miskelly:
@DustinCampbell Okay, I can work on the debugger side of that. Would you prefer to listen in the .ts code or in OmniSharp?

DustinCampbell:
It's probably cleaner to listen on the .ts side and I can update the `/v2/debugtest/run` OmniSharp endpoint (where we call `serverUtils.debugTestRun`) to take a process ID.

gregg-miskelly:
According to [this stackoverflow](http://stackoverflow.com/questions/11750041/how-to-create-a-named-pipe-in-node-js) article, looks like the 'net' library in node can support both named pipes on Windows and Unix domain sockets on non-Windows, so I will try that.

DustinCampbell:
I was expecting node to have the right cross-platform facilities to handle that. :smile:

Are you OK if I merge this in? At the moment, it has two issues that I'm aware of:

* The race condition around notifying OmniSharp after the debugger had launched.
* Forcibly stopping the debugger will leave OmniSharp in a bad state, requiring a restart.

gregg-miskelly:
@DustinCampbell that is fine with me.

Issue #1361, opened 2017-03-31
Create setting to manually configure system linux distro


shealey:
I'm building and running .Net core apps on GalliumOS - a lightweight Linux distro based on Ubuntu 16.10 (http://www.galliumos.org).

When I try to launch the debugger, I get the error message "The current Linux distribution 'galliumos' version '2.1' is not currently supported by the .NET Core debugger. Debugging will not be available."

Could we add a VsCode setting for Omnisharp so that users can specify what debugger version to use, in my instance ubuntu 16.10 so that it will work on more Linux distros?

Thanks

gregg-miskelly:
@shealey that sounds like a reasonable idea. Though if you wouldn't mind, can you try editing your platform.js file to see if things work if you change the following line from '2.0' to '2.1'?

https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/platform.ts#L331

shealey:
Hi Gregg,

Changing the version from 2.0 to 2.1 did the trick and I'm now able to debug as a temporary workaround.

Thanks

WardenGnaw:
@gregg-miskelly Should we have a setting to specify the version of ID_LIKE? E.g. csharp.idLikeVersion: { "ubuntu": "14.04" }? Since several OS's, like Zorin, utilize VERSION to be associated with ID, but KDE Neon uses a version of Ubuntu.

Using this page as reference: https://github.com/zyga/os-release-zoo

Issue #1323, opened 2017-03-14
Crash: 'Debug adapter process has terminated unexpectedly' on arch and manjaro


NikkyAI:
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.1)

Product Information:
 Version:            1.0.1
 Commit SHA-1 hash:  005db40cd1

Runtime Environment:
 OS Name:     manjaro
 OS Version:
 OS Platform: Linux
 RID:         ubuntu.16.10-x64
 Base Path:   /opt/dotnet/sdk/1.0.1
```
VS Code version:
```
Version 1.10.2
Commit 8076a19fdcab7e1fc1707952d652f0bb6c6db331
Date 2017-03-08T14:00:46.854Z
Shell 1.4.6
Renderer 53.0.2785.143
Node 6.5.0
```
C# Extension version:
1.8.0
## Steps to reproduce
- install dotnet, dotnet-cli and dotnet-sdk on archlinux (see additional info for building from aur)
- install C# extension
- try to debug anything 

## Expected  behavior
running the debugger

## Actual behavior
instant crash when trying to run the debugger

![Debug adapter process has terminated unexpectedly](https://aww.moe/ux8pns.png)
`Debug adapter process has terminated unexpectedly`

## Additional info
`Linux railgun 4.9.13-1-rt11-MANJARO #1 SMP PREEMPT RT Wed Mar 1 17:46:29 UTC 2017 x86_64 GNU/Linux`
<details> 
  <summary>System Info</summary>
  <pre><code> OS: Manjaro 17.1-dev Gellivara
 Kernel: x86_64 Linux 4.9.13-1-rt11-MANJARO
 Uptime: 5d 12h 23m
 Packages: 1655
 Shell: zsh 5.3.1
 Resolution: 1920x1080
 DE: XFCE4
 WM: Xfwm4
 WM Theme: Vertex-Maia
 GTK Theme: Vertex-Maia-Square [GTK2]
 Icon Theme: Vertex-Maia
 Font: Cantarell 10
 CPU: Intel Core i5-6200U @ 4x 2.8GHz
 GPU: Mesa DRI Intel(R) HD Graphics 520 (Skylake GT2)</code></pre>
</details>
<br>

to make dotnet build on archlinux/manjaro
i remember i had to patch a single line in corefx build because of 
```
ignoring return value of function declared with 'warn_unused_result'
attribute [-Werror,-Wunused-result]
[...]

PR #1371, opened 2017-04-07
Improve dependancy handling / Linux distro handling

gregg-miskelly:
https://github.com/OmniSharp/omnisharp-vscode/issues/1361
https://github.com/OmniSharp/omnisharp-vscode/issues/1323

Changes:
1. Add a setting to control what version of the debugger to use on Linux
2. We no longer automaticially select a debugger on Arch. Instead, we point folks to a web page telling them how to install it.
3. Added logic to the package manager so that it can detect if a package is already installed so it will not be redownloaded. This was needed since I wanted to trigger redownloads in the case that the user added the Linux distro setting. But it seemed like a useful feature anyway for folks on slow internet connections.
4. Moved the install code to its own .ts file


package.json
```diff
@@ -1179,8 +1212,17 @@
...
-        ]
+        ],
+        "windows": {
+          "program": "./.debugger/vsdbg-ui.exe"
+        },
+        "osx": {
+          "program": "./.debugger/vsdbg-ui"
+        },
+        "linux": {
+          "program": "./.debugger/vsdbg-ui"
```
DustinCampbell: Do you want these?

src/vscodePlatform.ts
```diff
@@ -0,0 +1,41 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+'use strict';
```
DustinCampbell: Perhaps we should just add `alwaysStrict: true` to the tsconfig.json now that TypeScript 2.1 supports that flag?

package.json
```diff
@@ -79,7 +79,8 @@
       "binaries": [
         "./mono.linux-x86",
         "./run"
-      ]
+      ],
+      "installTestPath": "./bin/mono.linux-x86"
```
DustinCampbell: This is a little unfortunate. Kind of wish we didn't have to specially list this for every dependency.

package.json
gregg-miskelly: Nope. Thanks for catching that.

CHANGELOG.md
```diff
@@ -8,13 +8,16 @@
 
 #### Debugger
 
+* **Arch Linux change**: before, the debugger would automatically use the Ubuntu 16 debugger on Arch. Now we require the debugger to be explicitly set. See https://aka.ms/vscode-csext-arch for more information.
 * Several bug fixes that addressed problems with launch ([#1318](https://github.com/OmniSharp/omnisharp-vscode/issues/1318), [#1335](https://github.com/OmniSharp/omnisharp-vscode/issues/1335), [#1336](https://github.com/OmniSharp/omnisharp-vscode/issues/1336))
 * Fix issue where VS Code would incorrectly display threads as paused ([#1317](https://github.com/OmniSharp/omnisharp-vscode/issues/1317))
+* Added new 'csharp.fallbackDebuggerLinuxRuntimeId' configuration setting to control the version of the debugger used on Liunx ([#1361](https://github.com/OmniSharp/omnisharp-vscode/issues/1361)).
```
gregg-miskelly: >Liunx [](start = 121, length = 5)

Spelling

src/vscodePlatform.ts
gregg-miskelly: Sounds good to me.

src/CSharpExtDownloader.ts
```diff
@@ -0,0 +1,148 @@
...
+                packageManager = new PackageManager(info, this.packageJSON);
+                this.logger.appendLine();
+
+                // Display platform information and RID followed by a blank line
+                this.logger.append(`Platform: ${info.toString()}`);
+                if (info.runtimeId) {
+                    this.logger.appendLine(` (${info.runtimeId})`);
+                }
+                else {
+                    this.logger.appendLine();
```
rajkumar42: appendLine('unknown')?

gregg-miskelly: This code isn't new to be clear. But sure, how about ' (no runtime id)'.

DustinCampbell: That sounds good to me. One thing that's also been a small point of confusion is that this 'runtimeId' isn't _really_ the current runtimeId. It's the effective runtimeId to locate the correct package. For example, on macOS Sierra, it reports osx.10.11-x64 instead of osx.10.12-x64. Should we tweak the text to make it clearer?

gregg-miskelly: Do we just want to remove this since the debugger packages will already tell us what version we thought the target was?

DustinCampbell: That's probably OK.

PR #1373, opened 2017-04-10
Fix RemoteProcessPicker

WardenGnaw:
Allowing quoteArgs to handle the pipeProgram and pipeArgs quotes. 

Moving the uname and ps command into a temp file so it can be redirected into the pipeProgram so the current shell does not mangle the command string.

Tested:
Windows -> Linux
Windows -> Docker Linux
Windows -> Mac
Mac -> Linux

+ Fixing configuration snippet formatting

### DIFF
--- a/scripts/remoteProcessPickerScript
+++ b/scripts/remoteProcessPickerScript
@@ -1,0 +1,1 @@
+uname && if [ "$(uname)" = "Linux" ] ; then ps -axww -o pid=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= ; elif [ "$(uname)" = "Darwin" ] ; then ps -axww -o pid=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= -c; fi
--- a/src/features/processPicker.ts
+++ b/src/features/processPicker.ts
@@ -82,17 +82,15 @@ export class RemoteAttachPicker {
             let scriptShellCmd: string = "sh -s";
             pipeCmdList.push(pipeProgram);
 
-            if (pipeArgs.filter(arg => arg === "${debuggerCommand}").length > 0) {
-                for (let arg of pipeArgs)
-                {
-                    if (arg === "${debuggerCommand}")
-                    {
-                        pipeCmdList.push(scriptShellCmd);
+            const debuggerCommandString: string = "${debuggerCommand}";
+
+            if (pipeArgs.indexOf(debuggerCommandString) > 0) {
+                for (let arg of pipeArgs) {
+                    while (arg.indexOf("${debuggerCommand}") >= 0) {
+                        arg = arg.replace("${debuggerCommand}", scriptShellCmd);
                     }
-                    else
-                    {
-                        pipeCmdList.push(arg);
-                    }
+                    
+                    pipeCmdList.push(arg);
                 }
             }
             else {
--- a/src/features/processPicker.ts
+++ b/src/features/processPicker.ts
@@ -5,8 +5,8 @@
 
 import * as os from 'os';
 import * as vscode from 'vscode';
-import * as tmp from 'tmp';
-import * as fs from 'fs-extra';
+import { getExtensionPath } from '../common';
+import * as path from 'path';
 import * as child_process from 'child_process';
 import { PlatformInformation } from '../platform';
 
@@ -80,11 +80,13 @@ export class RemoteAttachPicker {
 
             let pipeCmdList: string[] = [];
             pipeCmdList.push(pipeProgram);
-            pipeCmdList = pipeCmdList.concat(pipeArgs);
 
-            const scriptShellCmdList: string[] = ["sh", "-s"];
+            // Remove debuggerCommand for remoteProcessPicker
+            pipeCmdList = pipeCmdList.concat(pipeArgs.filter(arg => arg !== "${debuggerCommand}"));
 
-            pipeCmdList = pipeCmdList.concat(scriptShellCmdList);
+            let scriptShellCmd: string = "sh -s";
+
+            pipeCmdList.push(scriptShellCmd);
 
             let pipeCmd: string = quoteArgs ? this.createArgumentList(pipeCmdList) : pipeCmdList.join(' ');
 
@@ -129,19 +131,9 @@ export class RemoteAttachPicker {
     }
 
     public static getRemoteOSAndProcesses(pipeCmd: string): Promise<AttachItem[]> {
+        const scriptPath = path.join(getExtensionPath(), 'scripts', 'remoteProcessPickerScript');
 
-        // Commands to get OS and processes
-        const command = `uname && if [ "$(uname)" = "Linux" ] ; then ${RemoteAttachPicker.linuxPsCommand} ; elif [ "$(uname)" = "Darwin" ] ; ` +
-            `then ${RemoteAttachPicker.osxPsCommand}; fi`;
-
-        // Create a temp file to redirect commands to the pipeProgram to solve quoting issues.
-        const tempFile = tmp.fileSync();
-        fs.write(tempFile.fd, command);
-
-        return execChildProcessAndOutputErrorToChannel(`${pipeCmd} < ${tempFile.name}`, null, RemoteAttachPicker._channel).then(output => {
-            // Remove temp file since the uname and ps commands have been executed.
-            tempFile.removeCallback();
-
+        return execChildProcessAndOutputErrorToChannel(`${pipeCmd} < ${scriptPath}`, null, RemoteAttachPicker._channel).then(output => {
             // OS will be on first line
             // Processess will follow if listed
             let lines = output.split(/\r?\n/);
--- a/src/features/processPicker.ts
+++ b/src/features/processPicker.ts
@@ -129,7 +129,7 @@ export class RemoteAttachPicker {
     }
 
     public static getRemoteOSAndProcesses(pipeCmd: string): Promise<AttachItem[]> {
-        
+
         // Commands to get OS and processes
         const command = `uname && if [ "$(uname)" == "Linux" ] ; then ${RemoteAttachPicker.linuxPsCommand} ; elif [ "$(uname)" == "Darwin" ] ; ` +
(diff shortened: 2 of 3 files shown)


src/features/processPicker.ts
```diff
@@ -65,17 +67,27 @@ export class RemoteAttachPicker {
...
+                quoteArgs = platformSpecificPipeTransportOptions.pipeTransport.quoteArgs != null ? platformSpecificPipeTransportOptions.pipeTransport.quoteArgs : quoteArgs;
             }
 
-            let argList = RemoteAttachPicker.createArgumentList(pipeArgs);
-            let pipeCmd: string = `"${pipeProgram}" ${argList}`;
+            let pipeCmdList: string[] = [];
+            pipeCmdList.push(pipeProgram);
+            pipeCmdList = pipeCmdList.concat(pipeArgs);
+
+            const scriptShellCmdList: string[] = ["bash", "-s"];
```
gregg-miskelly: >bash [](start = 51, length = 4)

I am not sure it really matters much for .NET, but should we use 'sh' instead of 'bash' to make it easier to port code between this extension and C++?

gregg-miskelly: >scriptShellCmdList [](start = 18, length = 18)

Maybe I am remembering wrong, but I think we are doing something different here than what vsdbg-ui does.

In vsdbg-ui, if quoteArgs is enabled the command line would be something like:

   /mypipePath/myPipe -myPipeArg "/vsdbg/vsdbg --interpreter=vscode"

Here, it looks like we will wind up using:
   /mypipePath/myPipe -myPipeArg /vsdbg/vsdbg --interpreter=vscode

And we would only quote an argument if one of the individual arguments contained a space.


I think instead you want:
let scriptShellCmd : string = "bash -s";

if (quoteArgs) {
   scriptShellCmd = `"${scriptShellCmd}"`;
}

src/features/processPicker.ts
```diff
@@ -65,17 +67,27 @@ export class RemoteAttachPicker {
...
                 pipeProgram = platformSpecificPipeTransportOptions.pipeProgram || pipeProgram;
                 pipeArgs = platformSpecificPipeTransportOptions.pipeArgs || pipeArgs;
+                quoteArgs = platformSpecificPipeTransportOptions.pipeTransport.quoteArgs != null ? platformSpecificPipeTransportOptions.pipeTransport.quoteArgs : quoteArgs;
             }
 
-            let argList = RemoteAttachPicker.createArgumentList(pipeArgs);
-            let pipeCmd: string = `"${pipeProgram}" ${argList}`;
+            let pipeCmdList: string[] = [];
+            pipeCmdList.push(pipeProgram);
+            pipeCmdList = pipeCmdList.concat(pipeArgs);
```
gregg-miskelly: >pipeArgs [](start = 45, length = 8)

We need code to look for `${debuggerCommand}`

src/features/processPicker.ts
```diff
@@ -117,16 +129,19 @@ export class RemoteAttachPicker {
...
 
         // Commands to get OS and processes
-        const command = `bash -c 'uname && if [ "${remoteUnameCommand}" == "Linux" ] ; then ${RemoteAttachPicker.linuxPsCommand} ; elif [ "${remoteUnameCommand}" == "Darwin" ] ; ` +
-            `then ${RemoteAttachPicker.osxPsCommand}; fi'`;
+        const command = `uname && if [ "$(uname)" == "Linux" ] ; then ${RemoteAttachPicker.linuxPsCommand} ; elif [ "$(uname)" == "Darwin" ] ; ` +
+            `then ${RemoteAttachPicker.osxPsCommand}; fi`;
+
+        // Create a temp file to redirect commands to the pipeProgram to solve quoting issues.
+        const tempFile = tmp.fileSync();
+        fs.write(tempFile.fd, command);
```
gregg-miskelly: Instead of creating a temp file with this, should we should stick this in a file in our extension?

src/features/processPicker.ts
```diff
@@ -117,16 +129,19 @@ export class RemoteAttachPicker {
...
+
+        return execChildProcessAndOutputErrorToChannel(`${pipeCmd} < ${tempFile.name}`, null, RemoteAttachPicker._channel).then(output => {
+            // Remove temp file since the uname and ps commands have been executed.
+            tempFile.removeCallback();
```
gregg-miskelly: >tempFile [](start = 12, length = 8)

If we keep this, make sure to add a catch handler to clean it up in the case of failure as well.

src/features/processPicker.ts
```diff
@@ -65,17 +67,27 @@ export class RemoteAttachPicker {
...
-            let pipeCmd: string = `"${pipeProgram}" ${argList}`;
+            let pipeCmdList: string[] = [];
+            pipeCmdList.push(pipeProgram);
+            pipeCmdList = pipeCmdList.concat(pipeArgs);
+
+            const scriptShellCmdList: string[] = ["bash", "-s"];
+
+            pipeCmdList = pipeCmdList.concat(scriptShellCmdList);
+
+            let pipeCmd: string = quoteArgs ? this.createArgumentList(pipeCmdList) : pipeCmdList.join(' ');
```
WardenGnaw: Should I modify createArgumentList to only insert quotes if it has a space in the string?

scripts/remoteProcessPickerScript
```diff
@@ -0,0 +1 @@
+uname && if [ "$(uname)" = "Linux" ] ; then ps -axww -o pid=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= ; elif [ "$(uname)" = "Darwin" ] ; then ps -axww -o pid=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= -c; fi
```
gregg-miskelly: Now that we don't need to have this as one big line, would it make sense to separate this out?

gregg-miskelly: Do we need a .gitattributes on this to make sure that it keeps Unix line endings?

src/features/processPicker.ts
```diff
@@ -65,17 +67,29 @@ export class RemoteAttachPicker {
...
+                quoteArgs = platformSpecificPipeTransportOptions.pipeTransport.quoteArgs != null ? platformSpecificPipeTransportOptions.pipeTransport.quoteArgs : quoteArgs;
             }
 
-            let argList = RemoteAttachPicker.createArgumentList(pipeArgs);
-            let pipeCmd: string = `"${pipeProgram}" ${argList}`;
+            let pipeCmdList: string[] = [];
+            pipeCmdList.push(pipeProgram);
+
+            // Remove debuggerCommand for remoteProcessPicker
+            pipeCmdList = pipeCmdList.concat(pipeArgs.filter(arg => arg !== "${debuggerCommand}"));
```
gregg-miskelly: This doesn't work, as ${debuggerCommand} might be only part of one of the args, and because we want to preserve the order.

gregg-miskelly:
RE: Should I modify createArgumentList to only insert quotes if it has a space in the string?
A: I think so. The closer we can be to vsdbg-ui the better.

scripts/remoteProcessPickerScript
WardenGnaw: I believe there will be issues with \r\n vs \n if its separated on different lines.

gregg-miskelly: You are probably right. Oh line endings...

gregg-miskelly: I think you can 'Won't fix" this one since the script is a single line.

src/features/processPicker.ts
```diff
@@ -79,14 +79,26 @@ export class RemoteAttachPicker {
...
-            pipeCmdList = pipeCmdList.concat(pipeArgs.filter(arg => arg !== "${debuggerCommand}"));
-
             let scriptShellCmd: string = "sh -s";
+            pipeCmdList.push(pipeProgram);
 
-            pipeCmdList.push(scriptShellCmd);
+            if (pipeArgs.filter(arg => arg === "${debuggerCommand}").length > 0) {
+                for (let arg of pipeArgs)
+                {
+                    if (arg === "${debuggerCommand}")
```
gregg-miskelly: This should be:
```javascript
     while (message.indexOf("${debuggerCommand}") >= 0) {
            message = message.replace("${debuggerCommand}", scriptShellCmd);
        }
```

src/features/processPicker.ts
```diff
@@ -79,14 +79,26 @@ export class RemoteAttachPicker {
...
-            pipeCmdList.push(pipeProgram);
-
-            // Remove debuggerCommand for remoteProcessPicker
-            pipeCmdList = pipeCmdList.concat(pipeArgs.filter(arg => arg !== "${debuggerCommand}"));
-
             let scriptShellCmd: string = "sh -s";
+            pipeCmdList.push(pipeProgram);
 
-            pipeCmdList.push(scriptShellCmd);
+            if (pipeArgs.filter(arg => arg === "${debuggerCommand}").length > 0) {
```
gregg-miskelly: Replace: `arg === "${debuggerCommand}"`
With: `arg.indexOf("${debuggerCommand}") >= 0`

src/features/processPicker.ts
WardenGnaw: Did you mean? I believe array does not have a replace func.
```javascript
while (arg.indexOf("${debuggerCommand}") >= 0) {
   arg[arg.indexOf("${debuggerCommand}")] = "${debuggerCommand}";
}
```

src/features/processPicker.ts
```diff
@@ -65,17 +67,39 @@ export class RemoteAttachPicker {
...
 
-            let argList = RemoteAttachPicker.createArgumentList(pipeArgs);
-            let pipeCmd: string = `"${pipeProgram}" ${argList}`;
+            let pipeCmdList: string[] = [];
+            let scriptShellCmd: string = "sh -s";
+            pipeCmdList.push(pipeProgram);
+
+            const debuggerCommandString: string = "${debuggerCommand}";
+
+            if (pipeArgs.indexOf(debuggerCommandString) > 0) {
```
gregg-miskelly: `>=` not `>`

gregg-miskelly: And don't you still need your filter like you did before?

PR #1378, opened 2017-04-11
Don't create hidden diagnostics

### LINKED ISSUE #1231: Problems list dominated by unused usings
_From @davidfowl on February 12, 2017 18:13_

Makes it hard to see real problems. Can we downgrade this thing so it doesn't show up by default in the list of problems?

image

_Copied from original issue: OmniSharp/omnisharp-roslyn#769_

DustinCampbell:
Fixes #1231

This stops reporting "Remove unnecessary using directives" (and other Roslyn hidden diagnostics) as "info" diagnostics in the problems window. This has an impact of no longer showing a squiggle for unnecessary usings in the editor, but we will hopefully be able to address that by fading the usings in the editor (https://github.com/Microsoft/vscode/issues/20219).

### DIFF
--- a/src/features/codeActionProvider.ts
+++ b/src/features/codeActionProvider.ts
@@ -33,7 +33,6 @@ export default class OmnisharpCodeActionProvider extends AbstractProvider implem
     }
 
     public provideCodeActions(document: TextDocument, range: Range, context: CodeActionContext, token: CancellationToken): Promise<Command[]> {
-
         if (this._disabled) {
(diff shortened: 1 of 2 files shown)


guidobouman:
Microsoft/vscode#20219 got merged, but we still don't have unused usings highlighted with the new grayed out style. Should this still be done, or are we stuck with the setting "Suppress Hidden Diagnostics" for now?

PR #1379, opened 2017-04-12
Send events from the debugger to unit test code

gregg-miskelly:
This checkin completes the next chunk of work for unit test debugging in VS Code. The changes here are to create an event channel to receive events from the debugger.

Remaining work:
- Fire new events into OmniSharp
- Somehow build the project if it is out of date before running the test

### DIFF
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -129,6 +129,8 @@ export function debugDotnetTest(testMethod: string, fileName: string, testFramew
     // using VS Test. These require a different level of communication.
     let debugType: string;
     let debugEventListener: DebugEventListener = null;
+    let outputChannel = getTestOutputChannel();
+    outputChannel.appendLine(`Debugging method '${testMethod}'.`);
 
     return serverUtils.requestProjectInformation(server, { FileName: fileName} )
         .then(projectInfo => {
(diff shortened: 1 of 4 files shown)


src/features/dotnetTest.ts
```diff
@@ -166,4 +184,81 @@ export function updateCodeLensForTest(bucket: vscode.CodeLens[], fileName: strin
...
+            DebugEventListener.s_activeInstance = null;
+        }
+
+        if (this._serverSocket !== null) {
+            this._serverSocket.close();
+        }
+    }
+
+    private fireDebuggingStopped() : void {
+        // TODO: notify omniSharp
```
DustinCampbell: How were you thinking of doing this? Should `DebugEventListener` take the `OmniSharpServer` instance to notify it directly? Or, should `DebugEventListener` trigger events that we listen to elsewhere?

src/features/dotnetTest.ts
```diff
@@ -143,7 +156,12 @@ export function debugDotnetTest(testMethod: string, fileName: string, testFramew
                 serverUtils.debugTestRun(server, { FileName: fileName });
```
DustinCampbell: We'll want to trigger this after `processLaunched` event is received and pass the process ID to OmniSharp. I can update OmniSharp's protocol to take this.

src/features/dotnetTest.ts
gregg-miskelly: I hadn't thought too deeply on it since I wasn't sure what exactly the OmniSharp API needed to look like, but tentatively I was thinking we could just call to the OmniSharp code from there. It certainly also wouldn't be hard to take some delegates though if you think that makes things cleaner.

gregg-miskelly:
@DustinCampbell please have a finial look at these changes when you get back from vacation.

DustinCampbell:
Looks good. I'm working on the necessary OmniSharp changes right now.

PR #1405, opened 2017-04-18
Fixing processPicker for WSL

WardenGnaw:
Fixing error message for process picker.

Find and replace System32 for WSL bash because it spawns a 32 bit instance which bash.exe does not reside in System32 but rather sysnative.

### DIFF
--- a/src/features/processPicker.ts
+++ b/src/features/processPicker.ts
@@ -48,18 +48,23 @@ export class RemoteAttachPicker {
 
     public static ValidateAndFixPipeProgram(program: string): Promise<string> {
         return PlatformInformation.GetCurrent().then(platformInfo => {
-            // If its 64 bit windows and the program does not exist. Try to replace System32 with sysnative
-            if (platformInfo.isWindows && platformInfo.architecture === "x86_64" && !fs.existsSync(program) && program.toLowerCase().includes("system32")) {
-                let sysRoot: string = process.env.SystemRoot;
+            let sysRoot: string = process.env.SystemRoot;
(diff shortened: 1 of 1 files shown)


src/features/processPicker.ts
```diff
@@ -45,6 +46,27 @@ export class RemoteAttachPicker {
 
     private static _channel: vscode.OutputChannel = null;
 
+    public static ValidateAndFixPipeProgram(program: string): Promise<string> {
+        return PlatformInformation.GetCurrent().then(platformInfo => {
+            // If its 64 bit windows and the program does not exist. Try to replace System32 with sysnative
+            if (platformInfo.isWindows && platformInfo.architecture === "x86_64" && !fs.existsSync(program) && program.toLowerCase().includes("system32")) {
```
rajkumar42: It would be nice if we don't enter the if block if the following condition is not satisfied program.toLowerCase().startsWith(process.env.SystemRoot + "system32").

src/features/processPicker.ts
```diff
@@ -45,6 +46,27 @@ export class RemoteAttachPicker {
...
+                let newPath = path.join(sysRoot, 'sysnative');
+
+                // Regex to replace and ignore casing
+                let regex = RegExp(oldPath, "ig");
+
+                return program.replace(regex, newPath);
+            }
+
+            // Return original program and let it fall through
+            return program;
```
rajkumar42: It is possible the program does not exist in the sysnative path as well. Should we try and return the original program path if file wasn't found in sysnative path as well?

PR #1417, opened 2017-04-21
Be sure to exit from resolved promise deleteIfExists(...) helper

DustinCampbell:
Failing to exit when the specified file path doens't exist means that we try to delete that file even though it doesn't exist. This results in a timing issue because the deletion is asynchronous. Essentially, it might unintentionally delete the file springs into being before the promise is resolved.

The issue I'm fixing is where a Unix pipe file path might get deleted while we're trying to connect to it.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -78,7 +78,7 @@ export function deleteIfExists(filePath: string): Promise<void> {
     .then((exists: boolean) => {
         return new Promise<void>((resolve, reject) => {
             if (!exists) {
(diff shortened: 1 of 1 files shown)


DustinCampbell:
I'm going to go ahead and merge this and respin the 1.9.0 build.

gregg-miskelly:
Thanks Dustin!

DustinCampbell:
subtle 😄

Issue #1429, opened 2017-04-26
Add option to make 'unnecessary using directive' diagnostics to appear


cruz82:
## Environment data
`dotnet --info` output:

> .NET Command Line Tools (1.0.1)
> 
> Product Information:
>  Version:            1.0.1
>  Commit SHA-1 hash:  005db40cd1
> 
> Runtime Environment:
>  OS Name:     fedora
>  OS Version:  25
>  OS Platform: Linux
>  RID:         fedora.24-x64
>  Base Path:   /opt/dotnet/sdk/1.0.1

VS Code version: 1.11.2
C# Extension version: 1.9.0

## Steps to reproduce
Add an unnecessary using statement.

## Expected  behavior
A diagnostic info message about the unnecessary using statement.
![using](https://cloud.githubusercontent.com/assets/735452/25416885/679301de-2a94-11e7-8bc9-5487eefdfccf.png)

## Actual behavior
No diagnostic info message is shown and the using statement is not highlighted.

Leopotam:
vscode 1.12.0, c# extension 7439ef476f038e927604fe3f9753c7132c2cf42d commit (after 1.9.0 release) - same behaviour.

DustinCampbell:
This is by design. We suppressed this message in the PR to https://github.com/OmniSharp/omnisharp-vscode/issues/1231. Eventually, we hope to have faded code support in VS Code so we can provide this information a bit more unobtrusively (https://github.com/Microsoft/vscode/issues/20219 tracks this).

Note that the light bulb still appears and works if you move it to an unnecessary using, but we will no longer spam unnecessary using diagnostics.

cateyes99:
At least don't agree with removing the wavy underline in the editor for those unnecessary using directives! without those signs in the editor so we cannot spot them anymore, would u expect us to move the cursor each line by line for each .cs file to find out unnecessary using directives? this change makes the `Remove Unnecessary Usings` useless.

DustinCampbell:
I appreciate your passion @cateyes99! Unfortunately, VS Code doesn't make it possible to remove the item from the Problems list (which _many_ complained about) while keeping the squiggle in the editor.

If you like, you can file an issue (or submit a PR) to make this behavior optional. And, eventually, I hope we'll have the ability to fade code as I mentioned before to make this behave more like it does in Visual Studio (in VS, we fade unnecessary using directives but don't just the message in the Error List).

cruz82:
[...]

PR #1435, opened 2017-04-27
Allow suppression of hidden messages to be configurable with a default to supress

cruz82:
…t to supress (#1429)

### DIFF
--- a/package.json
+++ b/package.json
@@ -364,6 +364,11 @@
           "default": false,
           "description": "Suppress the notification window to perform a 'dotnet restore' when dependencies can't be resolved."
         },
+        "csharp.suppressHiddenMessages": {
+            "type": "boolean",
+            "default": true,
+            "description": "Suppress hidden messages from being shown."
+        },
         "omnisharp.path": {
           "type": [
             "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -364,6 +364,11 @@
           "default": false,
           "description": "Suppress the notification window to perform a 'dotnet restore' when dependencies can't be resolved."
         },
+        "csharp.suppressHiddenMessages": {
+            "type": "boolean",
+            "default": true,
+            "description": "Suppress hidden messages from being shown."
```
DustinCampbell: How about we go with "hidden diagnostics" rather than "hidden messages".

* setting name: `csharp.suppressHiddenDiagnostics`
* description: `Suppress 'hidden' diagnostics (such as 'unnecessary using directives') from appearing in the editor or the Problems pane`

cruz82:
@DustinCampbell, updated the setting name and the description.
Thanks.

DustinCampbell:
I'm planning to release a new C# for VS Code beta later today or early tomorrow. That will have this setting in it.

pgolebiowski:
@cruz82 You're awesome, man!

Issue #940, opened 2016-11-15
JustMyCode is always enabled when using "debug test" CodeLens annotation link


dmccaffery:
## Environment data
`dotnet --info` output:

.NET Command Line Tools (1.0.0-preview2-003148)

Product Information:
 Version:            1.0.0-preview2-003148
 Commit SHA-1 hash:  18ca244551

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.11-x64

VS Code version:

1.7.1
02611b40b24c9df2726ad8b33f5ef5f67ac30b44

C# Extension version:

1.5.1

## Steps to reproduce

1. Open any project containing unit tests
2. Add a [DebuggerStepThrough] attribute to a method called from the unit test
3. Add a breakpoint inside the [DebuggerStepThrough] method
4. Click the "debug test" CodeLens annotation link

## Expected  behavior

Disable _JustMyCode_ by default when debugging is activated from the "debug test" CodeLens annotation link

**OR**

Add an OmniSharp setting (user/workspace setting) that allows the _JustMyCode_ flag to be overridden for the "debug test" CodeLens annotation link:

```json
"omnisharp.codelens.justmycode": false
```

**OR**

Allow a uniquely identified *configuration* to be included in launch.json that OmniSharp will use when activating the debugger from CodeLens (if present) before falling through to the defaults.

```json
{
    "name": "OmniSharp CodeLens",
    "type": "coreclr",
    "request": "launch",
    "justMyCode": false
}
```

## Actual behavior

The configurations specified in launch.json are seemingly completely ignored, but now that _JustMyCode_ is enabled by default, it is impossible to step into external code or code decorated with the [DebuggerStepThrough] attribute when using the "debug test" CodeLens annotation link.

gregg-miskelly:
@dmccaffery The debugger will always respect DebuggerStepThrough - there is no way to hit breakpoints in those methods unless you remove that attribute. If you want stepping behavior that goes away when JMC is off, use DebuggerNonUserCode. So we can certainly leave this bug open for other reasons, but it will not actually help the exact scenario you described.

For reference, I believe this is the code that will need to be enhanced to enable this: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/features/dotnetTest.ts#L60

dmccaffery:
[...]

PR #1446, opened 2017-05-02
Add support for passing debug options when unit test debugging

gregg-miskelly:
Recently there have been a few cases where I wanted to have folks enable debugging options when debugging unit tests, but we didn't have any way of providing them. This checkin fixes that.

This will address https://github.com/OmniSharp/omnisharp-vscode/issues/940

### DIFF
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -85,18 +85,28 @@ export function runDotnetTest(testMethod: string, fileName: string, testFramewor
 }
 
 function createLaunchConfiguration(program: string, args: string, cwd: string, debuggerEventsPipeName: string) {
-    return {
-        // NOTE: uncomment this for vsdbg developement
-        // debugServer: 4711,
-        name: ".NET Test Launch",
-        type: "coreclr",
-        request: "launch",
-        debuggerEventsPipeName: debuggerEventsPipeName,
-        program,
-        args,
-        cwd,
-        stopAtEntry: true
-    };
+    let debugOptions = vscode.workspace.getConfiguration('csharp').get('unitTestDebugingOptions');
+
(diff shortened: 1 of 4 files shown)


WardenGnaw:
Would we want to have the generateOptionsSchema gulp task to also populate csharp.unitTestDebugingOptions? Or have a note somewhere to update the schema if we change OptionsSchema.json.

src/features/dotnetTest.ts
```diff
@@ -85,18 +85,28 @@ export function runDotnetTest(testMethod: string, fileName: string, testFramewor
...
+
+    // Now fill in the rest of the options
+    result.name = ".NET Test Launch";
+    result.type = "coreclr";
+    result.request = "launch";
+    result.debuggerEventsPipeName = debuggerEventsPipeName;
+    result.program = program;
+    result.args = args;
+    result.cwd = cwd;
+    result.stopAtEntry = true;
```
gregg-miskelly: Thinking about this more, I think we should stop setting `stopAtEntry`. It didn't really do anything before, but now that we are exposing justMyCode and symbolPaths, users could wind up stopping at the entry point of the unit test runner, which isn't what people would expect.

gregg-miskelly:
@WardenGnaw makes sense. I updated the [readme.md](https://github.com/gregg-miskelly/omnisharp-vscode/tree/unitTestDebugingOptions/src/tools#important-notes) with this note.

WardenGnaw:
LGTM

gregg-miskelly:
Here is a private with the fix: [csharp-1.10.0-beta1-priv.zip](https://github.com/OmniSharp/omnisharp-vscode/files/971886/csharp-1.10.0-beta1-priv.zip).

filipw:
very cool!

Issue #1459, opened 2017-05-06
ctor (and other expected) snippets are not available from the Insert Snippet command.


shaunluttin:
## Environment data
`dotnet --info` output:

.NET Command Line Tools (1.0.1) 
                                
Product Information:            
 Version:            1.0.1      
 Commit SHA-1 hash:  005db40cd1 

VS Code version: 1.12.1
C# Extension version: 1.9.0

## Steps to reproduce

1. `dotnet new console`
1. `dotnet restore`
1. `dotnet build`
1. `code .` 
1. Required assets to build and debug are missing from 'temp'. Add them? Yes.
1. Place the cursor inside the `Program` class definition. 
1. Open the command pallet and type: insert snippet.

## Expected  behavior

`ctor` appears in the list of snippets to insert, because ctor is one of the snippets listed in `~\.vscode\extensions\ms-vscode.csharp-1.9.0\snippets\csharp.json` on my local machine.

## Actual behavior

`ctor` does not appear in the list of snippets to insert. In the `c` section we have: 

* checked
* class
* cw

PR #1460, opened 2017-05-06
Replace quote-tab with quote

shaunluttin:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1459


shaunluttin:
Hmm. Need to think about this some more before requesting review.

PR #1461, opened 2017-05-06
Delete corrupt iterindex snippet

shaunluttin:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1459

### DIFF
--- a/snippets/csharp.json
+++ b/snippets/csharp.json
@@ -1,503 +1,553 @@
 {
-	"Attribute using recommended pattern": {
-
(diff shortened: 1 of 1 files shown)


shaunluttin:
There were two ways to fix the bug. One was to delete the iterindex snippet, the other was to convert from tabs to spaces. The tabs had corrupted iterindex, such that all subsequent snippets did not show up in the command pallet. My sense is that converting from tabs to spaces is the best option.

PR #1466, opened 2017-05-09
Ensure children of OmniSharp server process are killed on Unix

### LINKED ISSUE #1445: Omnisharp does not terminate it's existing process when it is restarted.
## Environment data
`dotnet --info` output:
.NET Command Line Tools (1.0.1)

Product Information:
 Version:            1.0.1
 Commit SHA-1 hash:  005db40cd1

Runtime Environment:
 OS Name:     fedora
 OS Version:  25
 OS Platform: Linux
 RID:         fedora.24-x64
 Base Path:   /opt/dotnet/sdk/1.0.1

VS Code version: 1.11.2
C# Extension version: 1.9.0

## Steps to reproduce

1. Open a C# project and let omnisharp do its thing.
1. Select restart omnisharp from the command palette.

## Expected  behavior
The old omnisharp process is terminated and a new one is created.

## Actual behavior
The old omnisharp process remains and a new one is created.
This causes multiple omnisharp process for the same window and chews up memory.
pasted image at 2017_05_02 11_39 am

DustinCampbell:
Fixes #1445

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -5,6 +5,7 @@
 
 import * as cp from 'child_process';
 import * as fs from 'fs';
+import * as os from 'os';
 import * as path from 'path';
 
 let extensionPath: string;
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
One suggestion, which you can take or leave: have OmniSharp exit automatically if its communication (stdin/stdout?) handles are closed. You can still certainly also make this fix as it will help if say mono got itself into a really weird state in initialization. But having it exit automatically can fix the case that our extension process crashed or was killed.

DustinCampbell:
Good idea!

DustinCampbell:
OK. I spent a little time working on this, and it's a bit more involved than I'd like, requiring a fair amount of refactoring of the `OmniSharpServer` code. It's all good and useful refactoring, but it's not important to dig into now. I filed #1488 to add that improvement in the future.

PR #1472, opened 2017-05-11
Updating completionItemProvider to use new icons for Event, Struct and EnumMember

dopare:
Changed temporary icons with new ones which were added in latest version of vscode.


dopare:
I have been waiting for this too long :)

PR #1567, opened 2017-06-13
Update OmniSharp and Mono downloads to use Azure CDN

DustinCampbell:
This should speed up acquisition of Mono and OmniSharp binaries worldwide.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     {
       "description": "Mono Runtime (Linux / x86)",
       "url": "https://omnisharpdownload.azureedge.net/ext/mono.linux-x86-5.2.0.104.zip",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
   "runtimeDependencies": [
     {
       "description": "Mono Runtime (Linux / x86)",
-      "url": "https://omnisharpdownload.blob.core.windows.net/ext/mono.linux-x86-5.2.0.104.zip",
+      "url": "https://omnisharpdownload.azureedge.net/ext/mono.linux-x86-5.2.0.104.zip",
       "installPath": "./bin",
```
gregg-miskelly: You probably also want a `fallbackUrl` with the original URL set. We found some folks couldn't connect to the CDN.

DustinCampbell: Good point. I'll do that.

PR #1576, opened 2017-06-16
Update debugger to 1.12.2 and support large set of Linux distros

gregg-miskelly:
This checkin updates the debugger to the 1.12.2 release. The major change here is that we are now running on .NET Core 2.0 preview 2, and so we can use a single runtime id for all glibc-based Linux distros. This means we can remove all the support we have for runtime id detection and filtering. Lastly, this means we no longer have an openssl dependency on OSX.


gregg-miskelly:
If anyone want to try this:

1. Download [csharp-1.12.0-debugger.zip](https://github.com/OmniSharp/omnisharp-vscode/files/1081591/csharp-1.12.0-debugger.zip)
2. Rename from .vsix to .zip
3. Follow instructions on https://github.com/OmniSharp/omnisharp-vscode/wiki/Installing-Beta-Releases

matachi:
I'm running Fedora 25 and gave VS Code a spin today. First I installed version v1.10.0 of this extension, but using the Fedora 24 version of the debugger didn't work; the debugger in VS Code failed with the error "Debug adapter process has terminated unexpectedly". However, I downloaded your v1.12.0 version of the extension, and with that I'm able to run, debug and step through the code. Haven't done any extensive testing yet, but at first glance it seems to work perfectly.

gregg-miskelly:
@matachi Thanks for trying and I am glad it worked for you!

src/packages.ts
```diff
@@ -21,7 +21,6 @@ export interface Package {
     fallbackUrl?: string;
     installPath?: string;
     platforms: string[];
-    runtimeIds: string[];
```
gregg-miskelly: Its certainly theoretically possible, but I felt like it was a lot of effectively dead code to keep around when we aren't sure how we would use it. I felt like it was better to remove it, and if we want it back we can go into the history and undo this change.

DustinCampbell: Sounds reasonable to me.

gregg-miskelly:
@starquake @NikkyAI @JBildstein @gegenschall: If you guys are still using Arch or Manjaro and .NET Core I was wondering if you would be willing to help validate this change as we have tried to, but our lack of knowledge of Arch is making this hard. If you can help, we would really appreciate it.

**Background**: For .NET Core 2.0, the .NET Team has tried to make the coreclr binaries much more compatible on Linux. They now have a single binary package that runs on all the officially supported distros. However, Arch is still not officially supported, it sounds like primarily due to https://github.com/dotnet/corefx/issues/9855, though maybe I have that wrong. I am hoping we can find a set of instructions we can give people to make the official package work (ex: install [openssl 1.0](https://www.archlinux.org/packages/core/x86_64/openssl-1.0/))?

**Instructions**:
1. Download [csharp-1.12.0-debugger.zip](https://github.com/OmniSharp/omnisharp-vscode/files/1082022/csharp-1.12.0-debugger.zip)
2. Rename from .zip to .vsix
3. Follow instructions on https://github.com/OmniSharp/omnisharp-vscode/wiki/Installing-Beta-Releases

gregg-miskelly:
@DustinCampbell  I pushed another [commit](https://github.com/OmniSharp/omnisharp-vscode/pull/1576/commits/708763dfcf7659d5b721ab6706ab8fc838bf18c8) with updated docs. Can you have a look when you get the chance.

Also, I would like to push a beta release. Anything you are about to checkin that I should wait for?

DustinCampbell:
> Also, I would like to push a beta release. Anything you are about to checkin that I should wait for?

Go right ahead. None of my changes need to go in without a new OmniSharp build.

starquake:
@gregg-miskelly Sorry for my late response. I see my installation of vscode is now using version ms-vscode.csharp-1.12.1 of the extension and everything is working great!

I love working with vs code and ASP.NET Core. Would work with it more often if we didn't  have a legacy application that is tied to Windows DLL's.

gregg-miskelly:
@starquake Thanks for letting us know that it still works for you. Glad to hear it!

Issue #1150, opened 2017-01-25
The C# extension does not start unless *.cs file is opened or the workspace contains a project.json


DustinCampbell:
It should also launch if the workspace contains an .sln or .csproj file.

DustinCampbell:
The simplest fix is blocked by https://github.com/Microsoft/vscode/issues/944.

JerryNixon:
Here's a repro to creating the error message in Visual Studio Code. Please notice that the entire scenario takes place in the Visual Studio Code terminal. There is no `cmd` shell. This means opening a `.cs` file is not a certainty in the developer's workflow. Also, notice a `.cs` file is NEVER opened before a debug is attempted.

> launch: launch.json must be configured. Change 'program' to the path to the executable file that you would like to debug

![netcore](https://cloud.githubusercontent.com/assets/1749983/22308859/faa12e1e-e305-11e6-90d3-1a968b0df6a7.gif)

If the user manually edits the value of `program` in `launch.json` then this error can be resolved. The error message clearly states that. However, this manual extra step is certainly a confusing one in light of how automated the rest of the workflow is. I would also argue that such a path is prone to error. 

**workaround**

Just switching into the project and also opening a `.cs` file will cause the tooling to handle this properly. This needs to be done prior to the first debugging session. Probably most developers won't realize the subtly of this requires step, but at least there's a way.

![netcore3](https://cloud.githubusercontent.com/assets/1749983/22309215/8aaba678-e307-11e6-865f-7d8530e0f4f5.gif)

DustinCampbell:
Unfortunately, this won't make it for our 1.7 release as we're blocked waiting for changes to VS Code in support of this scenario.

DustinCampbell:
OK. Thanks to @eamodio's PR for VS Code, we're able to improve the experience here a bit. However, this particular repro is still problematic. Essentially, we can now update the extension to activate when the workspace contains a *.csproj or *.sln (using wildcards). However, that only activates an extension when a folder is *opened*. In this particular scenario, the folder is already open, so the extension still won't activate until you open a .cs file.

eamodio:
@DustinCampbell That is only because there was no other projects or anything opened beforehand to have the extension activate correct? FYI, https://github.com/Microsoft/vscode/issues/28122 is a feature request to have it detect when new files are added.

DustinCampbell:
Thanks for the pointer @eamodio! I think that would address this.

akshita31:
[...]

Issue #1328, opened 2017-03-16
Feature Request: Disable MSBuild automatic creation of obj & bin folders


GuardRex:
Opened from [comment (omnisharp-vscode #1272)](https://github.com/OmniSharp/omnisharp-vscode/issues/1272#issuecomment-286928801)

VS Code version: 1.10.2
C# Extension version: 1.7.0-beta4

## Steps to reproduce

1. Open a folder that contains a *.csproj* file.
2. MSBuild detects the project file and automatically creates `obj` and `bin` folders.

## Desired behavior

I'd like a setting to disable this behavior. When using VSC to work on documentation samples, I'd like to be able to use VSC/OmniSharp without MSBuild creating `obj` and `bin` folders merely because I've opened a folder that contains a *.csproj* file.

As it stands today, I must stop for a few seconds to delete them each time MSBuild creates them. It's especially troublesome when I open a folder that contains several project files in several subfolders ... I must delete multiple pairs of these folders in each project folder.

GuardRex:
This is still a pain point. Opening a project folder in my local dotnet/docs repo to mod some documentation sample code fires off several dozen (or hundreds ... idk how many) project load attempts. I still need a way to disable this behavior. I'd like to have a setting that I can add to my current ...

```
"csharp.suppressDotnetInstallWarning": true,
"csharp.suppressDotnetRestoreNotification": true,
```

... settings that will disable this MSBuild behavior, too. Ideally, VSC/OmniSharp/MSBuild should do **nothing** when a project folder is opened for working on the docs sample projects. The Intellisense behaviors are great and welcome when working on these files (e.g., C#) ... it's just the behaviors that try to load projects, restore deps, create `obj`/`bin` folders, build the project, or anything along those lines that are bothersome.

@DustinCampbell Is there any way you can think of that I can workaround/hack to stop the MSBuild behavior short of actually moving to a different IDE? (I'll suffer this before changing IDE, because VSC/Omnisharp are so great otherwise :smile: :+1:). Thanks in advance for any tips or ideas.

![capture](https://user-images.githubusercontent.com/1622880/26853791-0fada46e-4ad9-11e7-8b32-24638fca224f.PNG)

pflannery:
[...]

PR #1592, opened 2017-06-26
Add activation events for *.csproj and *.sln

### LINKED ISSUE #1375: 'Go to Symbol in Workspace' doesn't work unless there is an open C# file
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-preview2-1-003177)

Product Information:
 Version:            1.0.0-preview2-1-003177
 Commit SHA-1 hash:  a2df9c2576

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.12-x64
```
VS Code version: `1.11.1`
C# Extension version: `1.8.1`

`Go to Symbol in Workspace` only seems to work if there is at least one C# file (editor) open. Originally opened here https://github.com/Microsoft/vscode/issues/23584, but was asked to open an issue here (https://github.com/Microsoft/vscode/issues/23584#issuecomment-291290695)

I often hit this when working in web projects -- for example I'm working in js/ts or html/css and then need to jump to a backend (C#) class and I try a symbol search and get nothing :)

## Steps to reproduce
1. Open a folder
2. With no editors open, do a symbol search for a known symbol
3. No results will be displayed

1. Open a folder
2. Open a C# file
3. Do a symbol search for a known Typescript symbol
4. Results will be displayed

## Expected  behavior

Results should be displayed in both scenarios

## Actual behavior

Results only displayed if there is an open C# file

DustinCampbell:
Should fix #1375 and improve #1150.

### DIFF
--- a/package.json
+++ b/package.json
@@ -340,7 +340,9 @@
     "onCommand:csharp.downloadDebugger",
     "onCommand:csharp.listProcess",
     "onCommand:csharp.listRemoteProcess",
-    "workspaceContains:project.json"
+    "workspaceContains:project.json",
+    "workspaceContains:*.csproj",
+    "workspaceContains:*.sln"
   ],
   "contributes": {
     "configuration": {
--- a/package.json
+++ b/package.json
@@ -343,7 +343,10 @@
     "workspaceContains:project.json",
     "workspaceContains:*.csproj",
     "workspaceContains:*.sln",
-    "workspaceContains:*.csx"
+    "workspaceContains:*.csx",
+    "workspaceContains:**/*.csproj",
+    "workspaceContains:**/*.sln",
+    "workspaceContains:**/*.csx"
   ],
   "contributes": {
     "configuration": {
--- a/package.json
+++ b/package.json
@@ -342,7 +342,8 @@
     "onCommand:csharp.listRemoteProcess",
     "workspaceContains:project.json",
     "workspaceContains:*.csproj",
-    "workspaceContains:*.sln"
+    "workspaceContains:*.sln",
+    "workspaceContains:*.csx"
   ],
   "contributes": {
     "configuration": {


DustinCampbell:
cc @eamodio

filipw:
can we also add `*.csx`? 
it's already included as a launch target https://github.com/OmniSharp/omnisharp-vscode/blob/dd8c7e188e6025b6ba9f0302f5474ab067a4eb8e/src/omnisharp/launcher.ts#L48 and https://github.com/OmniSharp/omnisharp-vscode/blob/dd8c7e188e6025b6ba9f0302f5474ab067a4eb8e/src/features/status.ts#L30

DustinCampbell:
Is this actually a problem for *.csx? Note that the extension already activates on *.csx files when they're opened because that's how the "csharp" language is defined in VS Code ([link](https://github.com/Microsoft/vscode/blob/master/extensions/csharp/package.json#L18)).

Note I'm not adding *.cs because that seems a bit *too* invasive. Could you explain how *.csx would be different.

DustinCampbell:
> Note I'm not adding *.cs because that seems a bit *too* invasive.

I'm totally willing to discuss this bit. That's my first gut feeling, but I'm not 100% certain it's something I should be concerned about. I definitely wouldn't want to add `**/*.cs` because a random C# file somewhere in a subfolder could cause OmniSharp to launch. However, if there's a .cs file in the root?

filipw:
I think `csx` case would be different than `cs`, because `csx` is a de facto project file - either as a fully self contained app or as an entry point that may contain references to other `csx` files (so other compilation items) as well as assembly references.

A standalone `cs` file could indeed have no meaning if it doesn't have a project file, as it would not know how to compile itself (plus, Omnisharp wouldn't even know how to treat such loose `cs` files). OTOH, a loose `csx` file (or a bunch of them) is always a valid start point for the server, and we can always provide proper language services.

It would be nice to open a folder with some `csx` scripts in there and have Omnisharp light up immediately so that:
 - you avoid the little delay to start the server when a `csx` is actually open
 - you can immediately do things like i.e. go to definition

DustinCampbell:
> you can immediately do things like i.e. go to definition

How does this help that? After all, you have to have a file open to go to definition, no?

DustinCampbell:
I suppose that's true. OK. I'll add *.csx as well. I still think *.cs is too much though. Do you disagree?

filipw:
no, I think you right about that 👍
thanks!

package.json
```diff
@@ -340,7 +340,9 @@
     "onCommand:csharp.downloadDebugger",
     "onCommand:csharp.listProcess",
     "onCommand:csharp.listRemoteProcess",
-    "workspaceContains:project.json"
+    "workspaceContains:project.json",
+    "workspaceContains:*.csproj",
```
eamodio: I'm not 100% sure, but should those be `workspaceContains:**/*.csproj` so that it could be in any folder? It *may* already be handled though

DustinCampbell: I intentionally didn't do that but I'm open to discussing it. I feared that this might be a bit invasive.

eamodio: Ah -- personally I would definitely want that. We typically open vscode to the root of the repository, and it doesn't have any projects or anything in it -- they are all nested under sub-folders. What are the cases you are thinking about where you wouldn't want that behavior?

DustinCampbell: I was thinking of scenarios like the one described by https://github.com/OmniSharp/omnisharp-vscode/issues/1328.

DustinCampbell:
OK. I've added patterns for *.csx and globs to search subfolders. I'm planning to ship 1.11 shortly and would like to hold this change for 1.12 to let folks play with it to ensure it feels good. Does that work for everybody?

filipw:
sounds like a plan 👍

DustinCampbell:
OK. I'm going to go ahead and merge this for 1.12.

PR #1599, opened 2017-06-27
Fix errors introduced with TypeScript 2.4

DustinCampbell:


DustinCampbell:
Two issues here:

1. In status.ts, we were incorrectly filling an array of `DocumentSelector` when it should have been `DocumentFilter`. TypeScript did not report this error until now.
2. In assets.test.ts, we passed `true` as an argument of type `ParseOptions`.

PR #1627, opened 2017-07-06
Fix IntelliSense

seunghwanh:
This is a fix I made to make the code completion faster.
I know that this might be against the whole `Omnisharp` architecture where the client should simply adopt the work done by the server to the specific client. But at least I give it a try :)

* Keywords that are unlikely to be a prefix for a user snippet come
  with a trailing white space.

* Keywords that require a parameter comes with a snippet where the
  single tabstop is for the parameter.

* For each CompletionItem, the label is a simplified signature.
  A simplified signature is a signature where the actual type
  parameters and parameters are stripped away so that only the
  <> and () are remaining. For example, the simplified signature
  of Something<T>(float f, int n) is Something<>().

* If an item contains type parameters or parameters, the item
  ships with a snippet where tabstops are the places for
  type parameters and parameters. Note that each parameter does
  not have its own tabstop. Instead, a tabstop is placed for
  the whole type parameters or parameters.

* Items are considered overloaded if they have the same label.
  If overloaded, the representing item has a snippet with the
  highest tabstop count, and the documentation can be any
  arbitrary one.

* For overriding methods, the completion text auto completes
  the whole method simply by calling the base method.

### Preview
Completing methods for two different types.
method_completion

Completing an overriding method.
overriding_method

### DIFF
--- a/src/features/completionItemProvider.ts
+++ b/src/features/completionItemProvider.ts
@@ -5,12 +5,13 @@
 
 'use strict';
 
-import {extractSummaryText} from './documentation';
+import { extractSummaryText } from './documentation';
 import AbstractSupport from './abstractProvider';
 import * as protocol from '../omnisharp/protocol';
 import * as serverUtils from '../omnisharp/utils';
-import {createRequest} from '../omnisharp/typeConvertion';
-import {CompletionItemProvider, CompletionItem, CompletionItemKind, CancellationToken, TextDocument, Range, Position} from 'vscode';
+import { createRequest } from '../omnisharp/typeConvertion';
+import { CompletionItemProvider, CompletionItem, CompletionItemKind, CancellationToken, TextDocument, Range, Position, SnippetString } from 'vscode';
+import { Delimiter } from "./utils";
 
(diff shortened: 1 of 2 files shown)


DustinCampbell:
Thanks very much for your contribution! Unfortunately, as I feared from the issue you filed the other day, I don't think we'll be able to take this pull request. This is super-cool functionality, but we're working to unify the C# completion list experiences between Visual Studio, Visual Studio Code, and Visual Studio for Mac. So, while this changr could be useful, it takes completion in a different direction than we're heading. In fact, we'll soon be moving completion to a new OmniSharp completion endpoint that doesn't have the snippet functionality you used here at all.

seunghwanh:
Thanks for the reply. I originally thought the best solution is to implement these features in omnisharp-roslyn, but it looked like out of my scope. So I made this as a temporary solution. 
Anyway, I am glad to hear that Omnisharp will have a new feature for the completion. I hope new cool features will arrive soon :)

DustinCampbell:
I can close the PR. Thanks very much for contributing! Please feel free to keep involved.

PR #1646, opened 2017-07-18
Switch to download.visualstudio.microsoft.com part 1

gregg-miskelly:
This checkin in the first part of the work to switch to using download.visualstudio.microsoft.com.

This checkin does two things:
1. It replaces the OmniSharp links with CDN links
2. It adds a new gulp task to update links in package.json. This will allow us to script the insertion process.


gregg-miskelly:
@DustinCampbell I will CC you on the debugger packaging script updates where I make use of `gulp updatePackageDependencies` so you can leverage something similar for OmniSharp if you would like.

PR #1657, opened 2017-07-21
Add nag message when project.json projects are opened

DustinCampbell:


src/omnisharp/extension.ts
```diff
@@ -88,6 +88,25 @@ export function activate(context: vscode.ExtensionContext, reporter: TelemetryRe
         }));
     }
 
+    // After server is started (and projects are loaded), check to see if there are
+    // any project.json projects. If so, notify the user about migration.
+    disposables.push(server.onServerStart(() => {
+        utils.requestWorkspaceInformation(server)
+            .then(workspaceInfo => {
+                if (workspaceInfo.DotNet && workspaceInfo.DotNet.Projects.length > 0) {
+                    const shortMessage = 'project.json is no longer a supported project format for .NET Core applications.';
```
gregg-miskelly: You have MSWord style quotes in this string. Expected?

src/omnisharp/extension.ts
```diff
@@ -88,6 +88,25 @@ export function activate(context: vscode.ExtensionContext, reporter: TelemetryRe
         }));
     }
 
+    // After server is started (and projects are loaded), check to see if there are
+    // any project.json projects. If so, notify the user about migration.
+    disposables.push(server.onServerStart(() => {
+        utils.requestWorkspaceInformation(server)
+            .then(workspaceInfo => {
+                if (workspaceInfo.DotNet && workspaceInfo.DotNet.Projects.length > 0) {
```
gregg-miskelly: Will this fire for the old PCL projects that use project.json for dependencies but still have a .csproj? Do you want to deprecate those as well?

src/omnisharp/extension.ts
DustinCampbell: Hmm... I'm not spotting them

gregg-miskelly: Sorry, somehow got this comment on the wrong line. See `‘dotnet migrate’`.

src/omnisharp/extension.ts
DustinCampbell: No. Just for project.json files that are handled as projects.

src/omnisharp/extension.ts
DustinCampbell: FWIW, my phone was hiding them and printing normal quotes. Good catch!

filipw:
The one thing that is a little problematic with this is if you use VS Code to develop Azure Functions. 
CSX Azure Functions use `project.json` for dependencies.

It is somewhat of a very niche use case I guess.

DustinCampbell:
Yeah, I realize this is going to be a bit ugly to roll out. Do Azure Functions actually use project.json for a 'project'? That is, does the OmniSharp DotNetProjectSystem create a project in the OmniSharpWorkspace for them? If not, this won't affect that.

filipw:
yeah so technically we don't even officially support Azure Functions CSX *yet* (see https://github.com/OmniSharp/omnisharp-roslyn/issues/779)
However, since Azure Functions CSX are very, veryclose to what we have built-in CSX support (which can work with standard CSI scripts, but can also read `project.json`) you *could* use the existing OmniSharp to support your Azure Functions development. 

From the technical standpoint, on the OmniSharp side, it would mean that actually both Dotnet (since that kicks in autoamtically for `project.json`) and Script (since that kicks in for `.csx`) project systems would be set up for you in the workspace, but in the editor-server communication, only the Script project system would be responding, since the Dotnet project system ignores `.csx` files.

DustinCampbell:
I wouldn't expend a ton of effort trying to support Azure Functions based on project.json just yet. Note that there are good reasons to stop proliferating project.json and the latest Azure Functions support for VS 2017 doesn't use them (https://blogs.msdn.microsoft.com/webdev/2017/05/10/azure-function-tools-for-visual-studio-2017/).

PR #1663, opened 2017-07-25
Allow 'type' to be customized for unit test debugging

### LINKED ISSUE #1586: Debugging unit tests always uses coreclr launch, even for non-coreclr project
## Environment data
`dotnet --info` output:

> .NET Command Line Tools (1.0.0-preview2-003131)
> 
> Product Information:
>  Version:            1.0.0-preview2-003131
>  Commit SHA-1 hash:  635cf40e58
> 
> Runtime Environment:
>  OS Name:     Windows
>  OS Version:  6.1.7601
>  OS Platform: Windows
>  RID:         win7-x64

VS Code version: `1.13.1`
C# Extension version: `1.10.0`

## Steps to reproduce
Using [example.zip](https://github.com/OmniSharp/omnisharp-vscode/files/1098983/example.zip) as base net46 xunit workspace:
Set a breakpoint on Console.WriteLine line in Test method
Click `debug test` on Test method

## Expected  behavior
Breakpoint should be hit

## Actual behavior
Breakpoint is missed, the debug launch uses a coreclr configuration and vsdbg outputs

> The target process exited without raising a CoreCLR started event. Ensure that the target process is configured to use .NET Core. This may be expected if the target process did not run on .NET Core.

Looking at https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/features/dotnetTest.ts#L116 the launch configuration generated has the type hard-coded to "coreclr" while this assumption may be (and in this case is) invalid.  It appears that the csharp.unitTestDebuggingOptions setting is used as a basis for the launch configuration, but the type is set after the fact.

Manually setting the type to "clr" in the extension allows the debugger to properly attach to the process and debugging works as expected.

I'm not sure about the effort required to auto-detect the launch type, however a suitable workaround would be to allow type to be specified in unitTestDebuggingOptions.

gregg-miskelly:
This checkin allows the unit test debugging options to set 'type'. This resolves #1586.

### DIFF
--- a/package.json
+++ b/package.json
@@ -315,6 +315,15 @@
                   "default": true
                 }
               }
+            },
+            "type": {
+              "type": "string",
+              "enum": [
+                "coreclr",
+                "clr"
+              ],
+              "description": "Type type of code to debug. Can be either 'coreclr' for .NET Core debugging, or 'clr' for Desktop .NET Framework. 'clr' only works on Windows as the Desktop framework is Windows-only.",
+              "default": "coreclr"
             }
           }
         },
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -111,9 +111,12 @@ function createLaunchConfiguration(program: string, args: string, cwd: string, d
         result = {};
     }
 
+    if (!result.type) {
+        result.type = "coreclr";
+    }
+
     // Now fill in the rest of the options
     result.name = ".NET Test Launch";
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -451,13 +460,11 @@
             "razor"
           ]
         },
-        "runtime": "node",
```
gregg-miskelly: Opps.

PR #1673, opened 2017-07-28
Launch json program doc fix

### LINKED ISSUE #1668: launch.json: default value and help text for 'program' is confusing
## Environment data
`dotnet --info` output: 2.0.0-preview2-006497
VS Code version: 1.14.2
C# Extension version: 1.11.0

## Steps to reproduce
Generate and open a new dotnet core project
Select .NET Core as debugging environment
Select .NET Core tasks

~~tasks.json contains ```"command": "dotnet",```, however since task versions 2.0.0 the task name does not get appended automatically, so by default it should be ```"command": "dotnet build",```~~ (coming directly from VS Code)

launch.json contains ```"program": "${workspaceRoot}/bin/Debug/<target-framework>/<project-name.dll>",```, however according to the tooltip (which shows <project\-name>, containing a backslash) it should be ```<project-name>.dll``` instead of ```<project-name.dll>```.

However even when using ```<project-name.dll>``` debugging is not working, one needs to manually replace both placeholders (the tooltip mentions that the project.json file, not the .csproj file, gets read).

gregg-miskelly:
This checkin makes tweaks to the docs/default value/template value for 'program' to hopefully make it really clear what one should use, and that you can't just leave it at some weird thing with '<' characters in it.

This resolves #1668.

### DIFF
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -18,12 +18,16 @@ The `preLaunchTask` field runs the associated taskName in tasks.json before debu
 This will create a task that runs `dotnet build`. You can read more about tasks at [https://code.visualstudio.com/docs/editor/tasks](https://code.visualstudio.com/docs/editor/tasks).
 
 ## Program
-The program field is set to the path to the application dll or .NET Core host executable to launch. 
+The program field is set to the path of the application dll or .NET Core host executable to launch.
 
-Example: "${workspaceRoot}/bin/Debug/\<target-framework\>/\<project-name.dll\>" where:
+This property normally takes the form: "${workspaceRoot}/bin/Debug/\<target-framework\>/\<project-name.dll\>".
 
-* \<target-framework\>: (example: 'netstandard1.5') This is the framework that the app is being built for. It is set in the project.json file.
-* \<project-name\>: (example: 'MyApp') The name of the project being debugged.",
+Example: "${workspaceRoot}/bin/Debug/netcoreapp1.1/MyProject.dll"
+
+Where:
+
+* \<target-framework\> is the framework that the debugged project is being built for. This is normally found in the project file as the 'TargetFramework' property.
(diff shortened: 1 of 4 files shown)


gregg-miskelly:
Here is a link to the new md file: https://github.com/gregg-miskelly/omnisharp-vscode/blob/LaunchJsonProgramDocFix/debugger-launchjson.md#program

debugger-launchjson.md
```diff
@@ -18,12 +18,16 @@ The `preLaunchTask` field runs the associated taskName in tasks.json before debu
 This will create a task that runs `dotnet build`. You can read more about tasks at [https://code.visualstudio.com/docs/editor/tasks](https://code.visualstudio.com/docs/editor/tasks).
 
 ## Program
-The program field is set to the path to the application dll or .NET Core host executable to launch. 
+The program field is set to the path of the application dll or .NET Core host executable to launch.
 
-Example: "${workspaceRoot}/bin/Debug/\<target-framework\>/\<project-name.dll\>" where:
+This property normally takes the form: "${workspaceRoot}/bin/Debug/\<target-framework\>/\<project-name.dll\>".
```
stephtr: Why not write `\<project-name\>.dll` (which would also slightly simplify the description)? Are there any cases in which the dll gets placed into the framework folder but with a different name?

gregg-miskelly: I put the '.dll' within brackets as there are some cases where you don't want `program` to point at a .dll. Does that answer your question?

stephtr: Yes, thank you. I asked because the old description once mentioned <project-name.dll> and once <project-name> without dll.

gregg-miskelly: FYI my thinking was this: is the default values - stick the `.dll` at the end since that is the formula that 98% of people should probably use, so hopefully it is the most clear way of telling people what they need to fill out. In the documentation, stick is as part of the name since there are cases for non-dll extensions.

PR #1681, opened 2017-08-02
[WIP] Cake support

mholo65:
Adding support for Cake (I think). Honestly, I have no idea what I'm doing 😄 @gep13 This needs some TypeScript love from you.

OmniSharp-Roslyn PR here https://github.com/OmniSharp/omnisharp-roslyn/pull/932

NB! For some reason, you'll have to switch language mode from `Cake` to `C#` (If you have Cake VS Code Extension installed) to get intellisense.

### DIFF
--- a/package.json
+++ b/package.json
@@ -206,9 +206,11 @@
     "workspaceContains:*.csproj",
     "workspaceContains:*.sln",
     "workspaceContains:*.csx",
+    "workspaceContains:*.cake",
     "workspaceContains:**/*.csproj",
     "workspaceContains:**/*.sln",
-    "workspaceContains:**/*.csx"
+    "workspaceContains:**/*.csx",
+    "workspaceContains:**/*.cake"
   ],
   "contributes": {
     "configuration": {
--- a/src/features/status.ts
+++ b/src/features/status.ts
@@ -27,7 +27,8 @@ let defaultSelector: vscode.DocumentSelector = [
     { pattern: '**/project.json' }, // project.json-files OR
     { pattern: '**/*.sln' }, // any solution file OR
     { pattern: '**/*.csproj' }, // an csproj file
-    { pattern: '**/*.csx' } // C# script
+    { pattern: '**/*.csx' }, // C# script
+    { pattern: '**/*.cake' } // Cake script
 ];
 
 class Status {
--- a/src/omnisharp/launcher.ts
+++ b/src/omnisharp/launcher.ts
@@ -17,7 +17,8 @@ export enum LaunchTargetKind {
     Solution,
     ProjectJson,
     Folder,
-    Csx
+    Csx,
+    Cake
 }
 
 /**
@@ -45,7 +46,7 @@ export function findLaunchTargets(): Thenable<LaunchTarget[]> {
     const options = Options.Read();
 
     return vscode.workspace.findFiles(
-        /*include*/ '{**/*.sln,**/*.csproj,**/project.json,**/*.csx}', 
+        /*include*/ '{**/*.sln,**/*.csproj,**/project.json,**/*.csx,**/*.cake}', 
         /*exclude*/ '{**/node_modules/**,**/.git/**,**/bower_components/**}',
         /*maxResults*/ options.maxProjectResults)
     .then(resources => {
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -460,13 +462,11 @@
             "razor"
           ]
         },
-        "runtime": "node",
```
DustinCampbell: You don't want this change

package.json
```diff
@@ -460,13 +462,11 @@
...
         "runtimeArgs": [],
         "variables": {
           "pickProcess": "csharp.listProcess",
           "pickRemoteProcess": "csharp.listRemoteProcess"
         },
-        "program": "./out/src/coreclr-debug/proxy.js",
```
DustinCampbell: Or this one

package.json
```diff
@@ -1376,7 +1376,16 @@
...
+        "windows": {
+          "program": "./.debugger/vsdbg-ui.exe"
+        },
+        "osx": {
+          "program": "./.debugger/vsdbg-ui"
+        },
+        "linux": {
+          "program": "./.debugger/vsdbg-ui"
+        }
       },
```
DustinCampbell: Or these

package.json
```diff
@@ -1387,13 +1396,11 @@
             "razor"
           ]
         },
-        "runtime": "node",
```
DustinCampbell: Not this

package.json
```diff
@@ -1387,13 +1396,11 @@
...
         "runtimeArgs": [],
         "variables": {
           "pickProcess": "csharp.listProcess",
           "pickRemoteProcess": "csharp.listRemoteProcess"
         },
-        "program": "./out/src/coreclr-debug/proxy.js",
```
DustinCampbell: nor this

package.json
```diff
@@ -2156,6 +2163,15 @@
...
           }
+        },
+        "windows": {
+          "program": "./.debugger/vsdbg-ui.exe"
+        },
+        "osx": {
+          "program": "./.debugger/vsdbg-ui"
+        },
+        "linux": {
+          "program": "./.debugger/vsdbg-ui"
```
DustinCampbell: Or this

package.json
mholo65: Oh, are VSCode adding these changes automatically? Can't remember changing these.

DustinCampbell: Yes, this happens when you debug the extension.

gep13:
@mholo65 was there something in particular that you thought was "wrong" with this PR?  The changes look good to me. :+1:

mholo65:
@gep13 well if we could figure out why we have to change language mode from `Cake` to `C#` before this extension wants to send any requests to `omnisharp.exe`, that would be great :wink:

`omnisharp.exe` is started when a`.cake` file is opened, but no requests are sent. Current fix is to switch language mode

DustinCampbell:
> well if we could figure out why we have to change language mode from `Cake` to `C#` before this extension wants to send any requests to `omnisharp.exe`, that would be great :wink:

I posted how that happens in the #cake channel on Slack. The file types that trigger "csharp" as the language is defined in VS Code itself right [here](https://github.com/Microsoft/vscode/blob/master/extensions/csharp/package.json#L16-L19). I'll submit a change to add "*.cake" to this list on your behalf today.

gep13:
>@DustinCampbell said...
The file types that trigger "csharp" as the language is defined in VS Code itself right here. I'll submit a change to add "*.cake" to this list on your behalf today.

Yes, I saw that in the Slack channel, and that is why I raised this issue https://github.com/OmniSharp/omnisharp-vscode/issues/1684, I was trying to get clarity on whether the syntax highlighting that we provide for .cake files would be affected by making that change.

Do you know if this will be the case?

DustinCampbell:
Yes, the syntax highlighting for .cake files would be affected. What differences does Cake add to C# syntax highlighting? Note that the official C# support for cake highlighting is here: https://github.com/dotnet/csharp-tmLanguage.

DustinCampbell:
I think this is generally a problem of trying to spread Cake support across two extensions. If you're trying to have a Cake Support extension provide syntax highlighting *and* the C# extension, I'm less comfortable with taking this change.

DustinCampbell:
Cake is really just C# script with some semantic differences, correct? If that's the case, I would proposal that the Cake Support extension doesn't provide it's own syntax highlighting, but takes advantage of C#'s. Then, the Cake Support extension would be modified to operate on the "csharp" language but only when the extension is ".cake". Thoughts?

DustinCampbell:
I'm going to go ahead and move this conversation over to #1684.

DustinCampbell:
This looks good. Sorry for the delay!

gep13:
WOO HOO! Thanks for merging this!

PR #1688, opened 2017-08-08
Debugger: Make changes to detect and require macOS Sierra

gregg-miskelly:
.NET Core 2.0 only supports Sierra, and it will fall over on older versions of macOS. This adds detection code to let users know what is going on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ See our [change log](https://github.com/OmniSharp/omnisharp-vscode/blob/v1.10.0/
 * Currently, the C# debugger officially supports the following operating systems:
 
   * Windows (64-bit only)
-  * macOS
+  * macOS 10.12 (Sierra) and newer
   * Ubuntu 14.04+ (and distros based on it)
   * Debian 8.7+
   * Red Hat Enterprise Linux (RHEL) / CentOS / Oracle Linux 7.3+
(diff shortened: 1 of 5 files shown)


gregg-miskelly:
Correct, this will just block debugging.

Micha-Richter:
I'm using .Net Core 1.x on macOS 10.11. Not installed .Net Core 2.0 yet.
Now I'm getting your message and debugger won't start..

Any hints? :|

gregg-miskelly:
Can you upgrade to Sierra? There isn't really a work around - regardless of what the target app is running on, the debugger itself now runs .NET Core 2.0, and the .NET team decided to cut support for pre-Serra (it will crash on startup).

Micha-Richter:
This went by way too fast imho. .Net Core 2.0 final release was made available TODAY!?!?
Really disappointing :(

DustinCampbell:
Note that you don't need to install .NET Core 2.0 to make debugging work. The problem is that the _debugger_ now runs on .NET Core 2.0.

Micha-Richter:
How can I make debugging work? TIA!
(can't install the debugger: "[ERROR] The debugger cannot be installed. The debugger requires macOS 10.12 (Sierra) or newer.")

DustinCampbell:
You need to update to macOS Sierra. That is, you must update OSX on your Mac. It will not run on El Capitan or lower.

RobStorm:
Is there a way to downgrade to older .net and still run on El Capitan?

Andergraw:
@RobStorm yes, use a previous version of C# Extension. Look at #1742

PR #1692, opened 2017-08-09
Add telemetry reporting the number of times tests are run or debugged

### LINKED ISSUE #1594: Need telemetry for test execution triggered from VSCode
Need telemetry support to track how many times test execution of .NET Core based tests is triggered from VS Code.

DustinCampbell:
Fixes #1594

This change looks large, but it's mostly mechanical refactoring to thread the TelemetryReporter through various OmniSharp features. In addition, I took the opportunity to introduce a `TestManager` class that simply encapsulates all of the functionality in dotnetTest.


src/features/dotnetTest.ts
```diff
@@ -13,236 +12,284 @@ import * as utils from '../common';
...
+        // register commands
+        let d1 = vscode.commands.registerCommand(
+            'dotnet.test.run',
+            (testMethod, fileName, testFrameworkName) => this._runDotnetTest(testMethod, fileName, testFrameworkName));
+
+        let d2 = vscode.commands.registerCommand(
+            'dotnet.test.debug',
+            (testMethod, fileName, testFrameworkName) => this._debugDotnetTest(testMethod, fileName, testFrameworkName));
+
+        this._telemetryIntervalId = setInterval(() =>
```
gregg-miskelly: I wonder if we should defer this until we first run a test so we don't have something polling for no reason?

DustinCampbell: Potentially, though I don't think it's huge deal.

I think what I really want to do here is have a shared component that sends telemetry like this. This is *really* similar to the telemetry reporting in the OmniSharpServer class, and I'd like to unify them. Are you OK if I leave this for now and unify later? I've filed #1693 to track unifying these in either case.

DustinCampbell: Going to go ahead and merge.

PR #1735, opened 2017-08-24
Support for optional formatting.

Leopotam:
* New option: "omnisharp.format.enable".
* Omnisharp respects "omnisharp.format.enable" option for self-registering as code formatter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -370,7 +370,7 @@
         "omnisharp.format.enable": {
           "type": "boolean",
           "default": true,
-          "description": "Specifes whether OmniSharp should be used for C# code formatting."
+          "description": "Enable/disable default C# formatter (requires restart)."
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -366,6 +366,11 @@
           "type": "boolean",
           "default": true,
           "description": "Specifes whether OmniSharp should use VS Code editor settings for C# code formatting (use of tabs, indentation size)."
+        },
+        "omnisharp.format.enable": {
+          "type": "boolean",
+          "default": true,
+          "description": "Specifes whether OmniSharp should be used for C# code formatting."
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -214,6 +214,11 @@
     "configuration": {
       "title": "C# configuration",
       "properties": {
+        "csharp.format.enable": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable default C# formatter (requires restart)."
+        },
         "csharp.suppressDotnetInstallWarning": {
           "type": "boolean",
           "default": false,
@@ -366,11 +371,6 @@
           "type": "boolean",
           "default": true,
           "description": "Specifes whether OmniSharp should use VS Code editor settings for C# code formatting (use of tabs, indentation size)."
-        },
-        "omnisharp.format.enable": {
-          "type": "boolean",
-          "default": true,
-          "description": "Enable/disable default C# formatter (requires restart)."
         }
       }
     },
--- a/src/omnisharp/extension.ts
+++ b/src/omnisharp/extension.ts
@@ -37,6 +37,8 @@ export function activate(context: vscode.ExtensionContext, reporter: TelemetryRe
         scheme: 'file' // only files from disk
     };
 
+    const options = Options.Read();
+
     const server = new OmniSharpServer(reporter);
     const advisor = new Advisor(server); // create before server is started
     const disposables: vscode.Disposable[] = [];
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -366,6 +366,11 @@
           "type": "boolean",
           "default": true,
           "description": "Specifes whether OmniSharp should use VS Code editor settings for C# code formatting (use of tabs, indentation size)."
+        },
+        "omnisharp.format.enable": {
```
DustinCampbell: This should be "csharp.format.enable".

package.json
```diff
@@ -366,6 +366,11 @@
...
+          "type": "boolean",
+          "default": true,
+          "description": "Specifes whether OmniSharp should be used for C# code formatting."
```
DustinCampbell: How about we match the description of other "*.format.enable" options and go with:

"Enable/disable default C# formatter (requires restart)"

package.json
Leopotam: Bad idea about "csharp." prefix because each non-builtin plugin should use own namespace prefix for custom settings on it. For example, vscode has builtin typescript support - its why there are "typescript.XXX" options, specified for builtin language. What does it means - "csharp.formatting.enable"? Is it for omnisharp plugin, or any other? Who knows, i suggest to move all options to "omnisharp.XXX" namespace.

package.json
Leopotam: Now all options "requires restart", why only this option should be declared in this way?

package.json
DustinCampbell: Do note that, while this project lives in the OmniSharp organization on GitHub, it is the official C# for VS Code that is owned and built by the C# team at Microsoft.

DustinCampbell: If we want this option to be shared among other extensions that target C#, we should submit it to our barebones csharp extension that is included in VS Code here: https://github.com/Microsoft/vscode/blob/master/extensions/csharp/package.json

DustinCampbell: To be super-clear, this is a Microsoft product, not an OmniSharp product. I am a member of the C# team and work on the C# language design.

Leopotam: So, it means that "csharp.XXX" is main namespace, right? And all settings from "omnisharp.XXX" will be migrated to main namespace later.

DustinCampbell: The "omnisharp.XXX" settings are all about configuring OmniSharp rather than configuring the C# experience in VS Code.

* "omnisharp.path" can be used to specify a path to a different OmniSharp executable than the one that C# for VS Code downloads locally.
* "omnisharp.useMono" can be set to tell C# for VS Code to use the "mono" on the PATH when launch the OmniSharp executable specified by "omnisharp.path".
* "omnisharp.waitForDebugger" is set to pass the "--debug" flag when launching OmniSharp.
* "omnisharp.loggingLevel" is about the level of logging detail that is specified with the "--logLevel" flag when launching OmniSharp.
* "omnisharp.autoStart" can be set to tell C# for VS Code to automatically launch OmniSharp when the extension is started.
* "omnisharp.projectLoadTimeout" allows the user to specify a custom timeout to wait for before C# for VS Code abandons launching OmniSharp.
* "omnisharp.useEditorFormattingSettings" can be unset to control whether VS Code's insert tabs/spaces configuration are passed as arguments when OmniSharp is launched.

There is one "omnisharp.XXX" that is a bug 😦:

* "omnisharp.maxProjectResults" the number of project results to display in the project picker. 

Really, this should be "csharp.maxProjectResults" because it's controlling behavior of the C# extension rather than OmniSharp.

package.json
DustinCampbell: Yes, many probably should -- at least those that require OmniSharp to be relaunched (which technically doesn't require VS Core to be relaunched).

However, there are other settings like "csharp.suppressDotnetInstallWarning", "csharp.unitTestDebuggingOptions", "csharp.suppressDotnetRestoreNotification" and "csharp.suppressHiddenDiagnostics" that do not require a restart.

Leopotam: > However, there are other settings
I mean - current implementation requires reload window / restart environment (even without full restart vscode) - options wrapper load preferences only once.

package.json
Leopotam: ok, what about omnisharp.useEditorFormattingSettings? Why format.enable should be at csharp, but format options about tabs / new lines - at omnisharp namespace?

package.json
DustinCampbell: You should be able to run the "Restart OmniSharp" command in VS Code to avoid restarting.

package.json
DustinCampbell: It's at the "omnisharp" namespace because it ends up affecting whether particular flags are passed to OmniSharp on launch. With the advantage of hindsight, I don't think "omnisharp.useEditorFormattingSettings" should have been added. This should have been automatic. @filipw, do you agree?

package.json
Leopotam: Well, it takes same (very close on my machine) time: reload window / restart omnisharp.

DustinCampbell: Sure, but I'm not sure how that's relevant. The point is that it *doesn't* require a restart. :smile:

Leopotam: I mean - extension can autoupdate internal options state through vscode event without any user interaction with reloading window / extension.

DustinCampbell: I think we're in agreement. Almost all of the "omnisharp.XXX" would require the suffix because OmniSharp would need to be reloaded for the option to have an effect. The "csharp.XXX" options that I listed above don't need the suffix.

This is really fit and finish outside of the scope of your PR though. Are you comfortable with your change?

Leopotam: Sure, waiting for public release with this fix :)

DustinCampbell:
Thanks for your contribution!

package.json
filipw: I think it could be phased out now, yes. If I recall correctly, it was added because it wasn't entirely understood how invasive it would be if the extension at one release suddenly started picking up VS Code indentation sizes, after previously ignoring them.

filipw:
One general comment - this can potentially create awkward behavior for code fixes. 
When code fixes are applied, OmniSharp will use it's own formatting rules to format the code. Normally that would be consistent with the rest of the formatting behavior of the C# extension, as the two work hand in hand in terms of indentiation rules and what have you.

With a different formatter enabled in VS Code, rather then the one built into the C# extension, you will potentially see differently formatted code after a code fix is invoked, and you will need to reformat in VS Code manually again.

PR #1772, opened 2017-10-02
Fix "go to definition" from metadata within the same metadata file

filipw:
At the moment we have a nice story of going from `source -> metadata` and from `metadata -> other metadata`.
However, when trying to go from metadata as source file to **the same** metadata as source file, we get an error.

Albeit not a huge issue, it can be a little annoying. An example is below. 

Given the following metadata as source file (shortened for brevity), where we position the caret at `$$`:

```csharp
namespace System.Text
{
    //
    // Summary:
    //     Represents a character encoding.To browse the .NET Framework source code for
    //     this type, see the Reference Source.
    public abstract class Encoding
    {
        //
        // Summary:
        //     Gets an encoding for the UTF-7 format.
        //
        // Returns:
        //     An encoding for the UTF-7 format.
        public static Encodin$$g UTF7 { get; }
     }
}
```

When I invoke "go to definition" on `$$`, I should be taken back to `Encoding` definition within the same metadata file. Instead I get:

screenshot 2017-10-02 20 47 03

This PR fixes this.

### DIFF
--- a/src/features/definitionMetadataDocumentProvider.ts
+++ b/src/features/definitionMetadataDocumentProvider.ts
@@ -30,8 +30,7 @@ export default class DefinitionMetadataDocumentProvider implements TextDocumentC
     }
 
     public getExistingMetadataResponseUri(sourceName:string) : Uri {
-        const uri = this.createUri(sourceName);
-        return uri;
+        return this.createUri(sourceName);
     }
 
     public register() : void {
@@ -44,7 +43,6 @@ export default class DefinitionMetadataDocumentProvider implements TextDocumentC
 
     private createUri(sourceName:string) : Uri {
         return Uri.parse(this.scheme + "://" +
-                         sourceName.replace(/\\/g, "/")
-                                                    .replace(/(.*)\/(.*)/g, "$1/[metadata] $2"));
+            sourceName.replace(/\\/g, "/").replace(/(.*)\/(.*)/g, "$1/[metadata] $2"));
     }
 }
--- a/src/features/definitionMetadataDocumentProvider.ts
+++ b/src/features/definitionMetadataDocumentProvider.ts
@@ -29,7 +29,7 @@ export default class DefinitionMetadataDocumentProvider implements TextDocumentC
         return uri;
     }
 
-    public getExistingMetadataResponseUri(sourceName:string) : Uri {
+    public getExistingMetadataResponseUri(sourceName: string) : Uri {
         return this.createUri(sourceName);
     }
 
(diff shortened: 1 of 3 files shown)


src/features/definitionMetadataDocumentProvider.ts
```diff
@@ -38,9 +41,8 @@ export default class DefinitionMetadataDocumentProvider implements TextDocumentC
         return this._documents.get(uri.toString()).Source;
     }
 
-    private createUri(metadataResponse: MetadataResponse) : Uri {
+    private createUri(sourceName:string) : Uri {
```
DustinCampbell: nit: space after `:`

src/features/definitionMetadataDocumentProvider.ts
```diff
@@ -23,13 +23,16 @@ export default class DefinitionMetadataDocumentProvider implements TextDocumentC
...
     public addMetadataResponse(metadataResponse: MetadataResponse) : Uri {
-        const uri = this.createUri(metadataResponse);
-
+        const uri = this.createUri(metadataResponse.SourceName);
         this._documents.set(uri.toString(), metadataResponse);
 
         return uri;
     }
 
+    public getExistingMetadataResponseUri(sourceName:string) : Uri {
```
DustinCampbell: nit: space after `:`

src/features/definitionProvider.ts
```diff
@@ -29,11 +29,23 @@ export default class CSharpDefinitionProvider extends AbstractSupport implements
 
         return serverUtils.goToDefinition(this._server, req, token).then(gotoDefinitionResponse => {
 
+            // the defintion is in source
             if (gotoDefinitionResponse && gotoDefinitionResponse.FileName) {
+
+                // if it is part of an already used metadata file, retrieve its uri instead of going to the physical file
+                if (gotoDefinitionResponse.FileName.startsWith("$metadata$")) {
```
DustinCampbell: We should add something to the OmniSharp response to indicate that the result is in metadata, rather than checking the file name. Not a blocking a issue, but I wanted to capture it.

src/omnisharp/typeConvertion.ts
```diff
@@ -9,17 +9,21 @@ import * as vscode from 'vscode';
 
 export function toLocation(location: protocol.ResourceLocation | protocol.QuickFix): vscode.Location {
     const fileName = vscode.Uri.file(location.FileName);
+    return toLocationFromUri(fileName, location);
+}
+
+export function toLocationFromUri(uri:vscode.Uri, location: protocol.ResourceLocation | protocol.QuickFix): vscode.Location {
```
DustinCampbell: nit: space after `:`

src/features/definitionProvider.ts
```diff
@@ -29,11 +29,23 @@ export default class CSharpDefinitionProvider extends AbstractSupport implements
...
+                    let uri = this._definitionMetadataDocumentProvider.getExistingMetadataResponseUri(gotoDefinitionResponse.FileName);
```
DustinCampbell: nit: `const` rather than `let`

src/features/definitionProvider.ts
filipw: yes I agree, good point

DustinCampbell:
Looks good! Anything else?

filipw:
> Looks good! Anything else?

No, I think this can go as is. We will do something smarter in Omnisharp in the future and then we can tweak this code to be a bit cleaner. ✨

DustinCampbell:
Sounds great.

Issue #146, opened 2016-04-06
No suggestions after typing new keyword


pflannery:
When typing the `new` keyword no suggestions are shown. i.e. `List<string> myList = new` at this point visual studio would suggestion completions.

pflannery:
Is there a block on this? anything I can do to fix this?

pflannery:
@DustinCampbell at'ing you just in case you missed my comments

DustinCampbell:
Unfortunately, it'll require some significant tweaks to both OmniSharp and probably a change to VS Code to make it work. It's definitely on my list to get working, but triggering completion on <SPACE> in specific positions is tricky.

Novack-:
This in an old one! Following.

pflannery:
@DustinCampbell thanks for the progress on this

However I'm seeing this in v1.13.1 

![image](https://user-images.githubusercontent.com/1727302/32752151-07f04668-c8c0-11e7-81e7-735ddbb74856.png)

But I expected something like this

![image](https://user-images.githubusercontent.com/1727302/32752206-4308e3f4-c8c0-11e7-988b-d37fc275acc6.png)

DustinCampbell:
Yes, there's is some work yet to handle sorting. It's getting much closer though! :smile: cc @rchande, who's working on this as part of https://github.com/OmniSharp/omnisharp-vscode/issues/1765.

Novack-:
The last missing piece before @DustinCampbell can do his magic seems to be happening on #35551.

PR #1776, opened 2017-10-04
Consume the completion trigger character API

rchande:
Also take a dependency on version 1.17 where it was added and request to trigger completion on space.

### DIFF
--- a/package.json
+++ b/package.json
@@ -444,11 +444,13 @@
             "razor"
           ]
         },
+        "runtime": "node",
         "runtimeArgs": [],
         "variables": {
           "pickProcess": "csharp.listProcess",
           "pickRemoteProcess": "csharp.listRemoteProcess"
         },
+        "program": "./out/src/coreclr-debug/proxy.js",
         "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1360,7 +1358,16 @@
             "request": "attach",
             "processId": "${command:pickProcess}"
           }
-        ]
+        ],
+        "windows": {
```
rchande: I'm not sure why these got added. Presumably they can be removed?

DustinCampbell: Remove these. Otherwise, they'll break the debugger.

rchande:
@DustinCampbell Right--I was going to ask you if you if we had a different branch that tracked the latest insiders build (or something).

rchande:
Also cc @piotrpMSFT and @jonsequitur in case they're interested.

DustinCampbell:
We don't have a different branch as we haven't been on the bleeding edge with VS Code APIs yet.

DustinCampbell:
Fixes #146

rchande:
Thanks @DustinCampbell

DustinCampbell:
For some reason, CI is hung for this PR. So, I went ahead and built and ran tests on my machine.

Issue #1570, opened 2017-06-13
Disable references code lens


andy-hanson:
## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (2.0.0-preview1-005977)

Product Information:
 Version:            2.0.0-preview1-005977
 Commit SHA-1 hash:  414cab8a0b

Runtime Environment:
 OS Name:     linuxmint
 OS Version:  18.1
 OS Platform: Linux
 RID:         linuxmint.18.1-x64
 Base Path:   /usr/share/dotnet/sdk/2.0.0-preview1-005977/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview1-002111-00
  Build    : 1ff021936263d492539399688f46fd3827169983
```

VS Code version: 1.14.0-insider
C# Extension version: 1.10.0

## Steps to reproduce

Open any file, see references code lens

## Expected  behavior

I would like a way to disable the code lens.

## Actual behavior

According to Microsoft/vscode#10238 there is no way to disable references from within vscode. Could this be an option in omnisharp.json so I can use the xunit codelens but not the references codelens?

DustinCampbell:
Good request. Thanks!

anfomin:
+1

teize001:
+1

pnloyd:
+1

PR #1781, opened 2017-10-06
Add an option to disable the references CodeLens

rchande:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1570

### DIFF
--- a/package.json
+++ b/package.json
@@ -373,6 +373,11 @@
           "type": "boolean",
           "default": true,
           "description": "Specifes whether OmniSharp should use VS Code editor settings for C# code formatting (use of tabs, indentation size)."
+        },
+        "omnisharp.showReferencesCodeLens": {
+          "type": "boolean",
+          "default": true,
+          "description": "Specifies whether Omnisharp should show the references CodeLens"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -323,6 +323,11 @@
           "default": true,
           "description": "Suppress 'hidden' diagnostics (such as 'unnecessary using directives') from appearing in the editor or the Problems pane."
         },
+        "csharp.showReferencesCodeLens": {
+          "type": "boolean",
+          "default": true,
+          "description": "Specifies whether the references CodeLens show be shown."
+        },
         "omnisharp.path": {
           "type": [
             "string",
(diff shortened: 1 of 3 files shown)


rchande:
@DustinCampbell I didn't find any tests for the references indicator. Can you confirm whether or not there are any?

DustinCampbell:
There are no tests for the references indicator. You're are 100% welcome to explore writing one.

package.json
```diff
@@ -373,6 +373,11 @@
           "type": "boolean",
           "default": true,
           "description": "Specifes whether OmniSharp should use VS Code editor settings for C# code formatting (use of tabs, indentation size)."
+        },
+        "omnisharp.showReferencesCodeLens": {
```
DustinCampbell: This should be "csharp.showReferencesCodeLens". The "omnisharp.*" prefixed settings are intended to be for OmniSharp itself (though we diverged in one or two cases).

src/features/codeLensProvider.ts
```diff
@@ -43,6 +44,13 @@ export default class OmniSharpCodeLensProvider extends AbstractProvider implemen
     };
 
     provideCodeLenses(document: vscode.TextDocument, token: vscode.CancellationToken): vscode.CodeLens[] | Thenable<vscode.CodeLens[]> {
+        const options = Options.Read();
+        if (!options.showReferencesCodeLens)
+        {
+            let arr: vscode.CodeLens[] = [];
+            return arr;
```
DustinCampbell: Can't use just return `[]` here? Or, does TypeScript complain.

package.json
```diff
@@ -373,6 +373,11 @@
...
+          "type": "boolean",
+          "default": true,
+          "description": "Specifies whether Omnisharp should show the references CodeLens"
```
DustinCampbell: How about: "Specifies whether the references CodeLens should be shown".

Remember that the extension isn't "OmniSharp".

rchande:
Thanks @DustinCampbell -- incorporated your feedback.
Regarding tests, I think maybe there's a larger workitem to go and add tests for all the features that depend on the OmniSharp server (would require some kind of mock of the service and so forth).

DustinCampbell:
@rchande: I'm not sure a mock service is actually needed. VS Code extension tests run by launching an instance of VS Code and running the extension. If we could add the infrastructure to allow our platform specific components to download during test runs, I think it'd work fine.

DustinCampbell:
Of course, that'd be more of an integration test. Adding a mock for the omnisharp service endpoints wouldn't be a big deal. Agree that it's a bigger work item.

pnloyd:
@rchanda Thanks for this!

PR #1782, opened 2017-10-06
Fixed disabling of code lens references

filipw:
This is a follow up to #1781.
The setting to disable references code lens inadvertently also disabled the run/debug test code lens. 

This PR fixes that - so setting `csharp.showReferencesCodeLens` doesn't affect the test code lens.
It also introduces a sibling setting - `csharp.showTestsCodeLens`, which controls toggling the test code lens, without affecting the references code lens.

### DIFF
--- a/package.json
+++ b/package.json
@@ -328,6 +328,11 @@
           "default": true,
           "description": "Specifies whether the references CodeLens show be shown."
         },
+        "csharp.showTestsCodeLens": {
+          "type": "boolean",
+          "default": true,
+          "description": "Specifies whether the run and debug test CodeLens show be shown."
+        },
         "omnisharp.path": {
           "type": [
             "string",
--- a/package.json
+++ b/package.json
@@ -326,12 +326,12 @@
         "csharp.showReferencesCodeLens": {
           "type": "boolean",
           "default": true,
-          "description": "Specifies whether the references CodeLens show be shown."
+          "description": "Specifies whether the references CodeLens should be show be shown."
         },
         "csharp.showTestsCodeLens": {
           "type": "boolean",
           "default": true,
-          "description": "Specifies whether the run and debug test CodeLens show be shown."
+          "description": "Specifies whether the run and debug test CodeLens should be show be shown."
         },
         "omnisharp.path": {
           "type": [
--- a/src/features/codeLensProvider.ts
+++ b/src/features/codeLensProvider.ts
@@ -28,12 +28,14 @@ class OmniSharpCodeLens extends vscode.CodeLens {
 export default class OmniSharpCodeLensProvider extends AbstractProvider implements vscode.CodeLensProvider {
 
     private _testManager: TestManager;
+    private _options: Options;
 
     constructor(server: OmniSharpServer, reporter: TelemetryReporter, testManager: TestManager)
     {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -328,6 +328,11 @@
           "default": true,
           "description": "Specifies whether the references CodeLens show be shown."
         },
+        "csharp.showTestsCodeLens": {
+          "type": "boolean",
+          "default": true,
+          "description": "Specifies whether the run and debug test CodeLens show be shown."
```
DustinCampbell: Typo: "should be should". It looks like the same typo appears above.

filipw: fixed!

DustinCampbell:
@rchande: Does this look good to you?

src/features/codeLensProvider.ts
```diff
@@ -28,12 +28,14 @@ class OmniSharpCodeLens extends vscode.CodeLens {
...
 
     private _testManager: TestManager;
+    private _options: Options;
 
     constructor(server: OmniSharpServer, reporter: TelemetryReporter, testManager: TestManager)
     {
         super(server, reporter);
 
         this._testManager = testManager;
+        this._options = Options.Read();
```
rchande: With this change, do we still turn codelens on/off as you switch between the prefernces tab and file?

DustinCampbell: I don't think so. I would expect that this will require a restart of the extension to take effect.

filipw: yeah indeed. Though I think this is consistent with the rest of the extension at the moment - we don't read any options in "real time" I think

DustinCampbell: I think we're just inconsistent. For example, setting `csharp.suppressDotnetRestoreNotification` will take effect immediately. So will `csharp.disableCodeActions`. Maybe it's worth adding the extra logic to listen for `vscode.workspace.onDidChangeConfiguration`? Check out what we do in `CodeActionProvider` [here](https://github.com/OmniSharp/omnisharp-vscode/blob/2adadbd622dd5fe11035a380ef7c56dca1464562/src/features/codeActionProvider.ts#L16-L41)

filipw: thanks for the tip - I did not see that before. I pushed the appropriate change. Indeed, the experience is much nicer now 👍

src/features/codeLensProvider.ts
```diff
@@ -28,12 +28,21 @@ class OmniSharpCodeLens extends vscode.CodeLens {
...
         super(server, reporter);
 
         this._testManager = testManager;
+        this._checkOptions();
+
+        let configChangedDisposable = vscode.workspace.onDidChangeConfiguration(this._checkOptions, this);
+        this.addDisposables(configChangedDisposable);
+    }
+
+    private _checkOptions(): void {
```
rchande: Nit--this method doesn't actually check anything. Maybe call it "_updatedCachedOptions" or something?

DustinCampbell: Let's go back and fix that up with `codeActionProvider` at the same time, so that we're consistent. I'll send a follow up PR with that change.

filipw:
thanks guys! 👍

PR #1789, opened 2017-10-13
Revisit launch on Mono

DustinCampbell:
This change causes OmniSharp to be with the 'mono' on the path, if it is available and at least version 5.2.0. If 'mono' can't be found, we fall back to our local Mono.

In addition, I fixed a few bugs I noticed while testing:

1. STDOUT lines received from OmniSharp are trimmed before processing to ensure that they are properly handled if they start with non-printable characters.
2. Allow promise rejections from OmniSharp launch failures to propagate. If they aren't, any error messages will not be printed to the output.
3. Ensure that the `--assembly-loader=strict` argument is passed to Mono in the proper order. Since it's an argument to Mono, it needs to appear before any OmniSharp arguments.

### DIFF
--- a/src/omnisharp/launcher.ts
+++ b/src/omnisharp/launcher.ts
@@ -197,7 +197,8 @@ export function launchOmniSharp(cwd: string, args: string[]): Promise<LaunchResu
                 setTimeout(function () {
                     resolve(result);
                 }, 0);
-            });
+            })
+            .catch(reason => reject(reason));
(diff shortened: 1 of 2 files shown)


DustinCampbell:
One nice benefit of this change is that OmniSharp will run faster on OSX when running on a globally-installed Mono because of the AOT images that Mono installs for core framework assemblies.

PR #1801, opened 2017-10-23
Add ConfigurationProvider API, ${workspaceRoot} -> ${workspaceFolder}

WardenGnaw:
Adding ConfigurationProvider API

Changing ${workspaceRoot} to ${workspaceFolder}

tasks.json to version 2.0.0 and updating associated tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -474,13 +474,13 @@
             "properties": {
               "program": {
                 "type": "string",
-                "description": "Path to the application dll or .NET Core host executable to launch.\nThis property normally takes the form: '${workspaceRoot}/bin/Debug/(target-framework)/(project-name.dll)'\nExample: '${workspaceRoot}/bin/Debug/netcoreapp1.1/MyProject.dll'\n\nWhere:\n(target-framework) is the framework that the debugged project is being built for. This is normally found in the project file as the 'TargetFramework' property.\n(project-name.dll) is the name of debugged project's build output dll. This is normally the same as the project file name but with a '.dll' extension.",
-                "default": "${workspaceRoot}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll"
+                "description": "Path to the application dll or .NET Core host executable to launch.\nThis property normally takes the form: '${workspaceFolder}/bin/Debug/(target-framework)/(project-name.dll)'\nExample: '${workspaceFolder}/bin/Debug/netcoreapp1.1/MyProject.dll'\n\nWhere:\n(target-framework) is the framework that the debugged project is being built for. This is normally found in the project file as the 'TargetFramework' property.\n(project-name.dll) is the name of debugged project's build output dll. This is normally the same as the project file name but with a '.dll' extension.",
+                "default": "${workspaceFolder}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll"
               },
               "cwd": {
                 "type": "string",
                 "description": "Path to the working directory of the program being debugged. Default is the current workspace.",
-                "default": "${workspaceRoot}"
+                "default": "${workspaceFolder}"
               },
               "args": {
                 "anyOf": [
@@ -706,7 +706,7 @@
                   "debuggerPath"
                 ],
                 "default": {
-                  "pipeCwd": "${workspaceRoot}",
+                  "pipeCwd": "${workspaceFolder}",
                   "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
                   "pipeArgs": [],
                   "debuggerPath": "enter the path for the debugger on the target machine, for example ~/vsdbg/vsdbg"
@@ -715,7 +715,7 @@
                   "pipeCwd": {
                     "type": "string",
                     "description": "The fully qualified path to the working directory for the pipe program.",
-                    "default": "${workspaceRoot}"
+                    "default": "${workspaceFolder}"
                   },
                   "pipeProgram": {
                     "type": "string",
@@ -761,7 +761,7 @@
                   "windows": {
                     "description": "Windows-specific pipe launch configuration options",
                     "default": {
-                      "pipeCwd": "${workspaceRoot}",
+                      "pipeCwd": "${workspaceFolder}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
                       "pipeArgs": []
                     },
@@ -770,7 +770,7 @@
                       "pipeCwd": {
                         "type": "string",
                         "description": "The fully qualified path to the working directory for the pipe program.",
-                        "default": "${workspaceRoot}"
+                        "default": "${workspaceFolder}"
                       },
                       "pipeProgram": {
                         "type": "string",
@@ -813,7 +813,7 @@
                   "osx": {
                     "description": "OSX-specific pipe launch configuration options",
                     "default": {
-                      "pipeCwd": "${workspaceRoot}",
+                      "pipeCwd": "${workspaceFolder}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
                       "pipeArgs": []
                     },
@@ -822,7 +822,7 @@
                       "pipeCwd": {
                         "type": "string",
                         "description": "The fully qualified path to the working directory for the pipe program.",
-                        "default": "${workspaceRoot}"
+                        "default": "${workspaceFolder}"
                       },
                       "pipeProgram": {
                         "type": "string",
@@ -865,7 +865,7 @@
                   "linux": {
                     "description": "Linux-specific pipe launch configuration options",
                     "default": {
-                      "pipeCwd": "${workspaceRoot}",
+                      "pipeCwd": "${workspaceFolder}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
                       "pipeArgs": []
                     },
@@ -874,7 +874,7 @@
                       "pipeCwd": {
                         "type": "string",
                         "description": "The fully qualified path to the working directory for the pipe program.",
-                        "default": "${workspaceRoot}"
+                        "default": "${workspaceFolder}"
                       },
                       "pipeProgram": {
                         "type": "string",
@@ -1014,7 +1014,7 @@
                   "debuggerPath"
                 ],
                 "default": {
-                  "pipeCwd": "${workspaceRoot}",
+                  "pipeCwd": "${workspaceFolder}",
                   "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
                   "pipeArgs": [],
                   "debuggerPath": "enter the path for the debugger on the target machine, for example ~/vsdbg/vsdbg"
@@ -1023,7 +1023,7 @@
                   "pipeCwd": {
                     "type": "string",
                     "description": "The fully qualified path to the working directory for the pipe program.",
-                    "default": "${workspaceRoot}"
+                    "default": "${workspaceFolder}"
                   },
                   "pipeProgram": {
                     "type": "string",
@@ -1069,7 +1069,7 @@
                   "windows": {
                     "description": "Windows-specific pipe launch configuration options",
                     "default": {
-                      "pipeCwd": "${workspaceRoot}",
+                      "pipeCwd": "${workspaceFolder}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example 'c:\\tools\\plink.exe'",
                       "pipeArgs": []
                     },
@@ -1078,7 +1078,7 @@
                       "pipeCwd": {
                         "type": "string",
                         "description": "The fully qualified path to the working directory for the pipe program.",
-                        "default": "${workspaceRoot}"
+                        "default": "${workspaceFolder}"
                       },
                       "pipeProgram": {
                         "type": "string",
@@ -1121,7 +1121,7 @@
                   "osx": {
                     "description": "OSX-specific pipe launch configuration options",
                     "default": {
-                      "pipeCwd": "${workspaceRoot}",
+                      "pipeCwd": "${workspaceFolder}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
                       "pipeArgs": []
                     },
@@ -1130,7 +1130,7 @@
                       "pipeCwd": {
                         "type": "string",
                         "description": "The fully qualified path to the working directory for the pipe program.",
-                        "default": "${workspaceRoot}"
+                        "default": "${workspaceFolder}"
                       },
                       "pipeProgram": {
                         "type": "string",
@@ -1173,7 +1173,7 @@
                   "linux": {
                     "description": "Linux-specific pipe launch configuration options",
                     "default": {
-                      "pipeCwd": "${workspaceRoot}",
+                      "pipeCwd": "${workspaceFolder}",
                       "pipeProgram": "enter the fully qualified path for the pipe program name, for example '/usr/bin/ssh'",
                       "pipeArgs": []
                     },
@@ -1182,7 +1182,7 @@
                       "pipeCwd": {
                         "type": "string",
                         "description": "The fully qualified path to the working directory for the pipe program.",
-                        "default": "${workspaceRoot}"
+                        "default": "${workspaceFolder}"
                       },
                       "pipeProgram": {
                         "type": "string",
@@ -1352,13 +1352,13 @@
             "properties": {
               "program": {
                 "type": "string",
-                "description": "Path to the application dll or .NET Core host executable to launch.\nThis property normally takes the form: '${workspaceRoot}/bin/Debug/(target-framework)/(project-name.dll)'\nExample: '${workspaceRoot}/bin/Debug/netcoreapp1.1/MyProject.dll'\n\nWhere:\n(target-framework) is the framework that the debugged project is being built for. This is normally found in the project file as the 'TargetFramework' property.\n(project-name.dll) is the name of debugged project's build output dll. This is normally the same as the project file name but with a '.dll' extension.",
-                "default": "${workspaceRoot}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll"
+                "description": "Path to the application dll or .NET Core host executable to launch.\nThis property normally takes the form: '${workspaceFolder}/bin/Debug/(target-framework)/(project-name.dll)'\nExample: '${workspaceFolder}/bin/Debug/netcoreapp1.1/MyProject.dll'\n\nWhere:\n(target-framework) is the framework that the debugged project is being built for. This is normally found in the project file as the 'TargetFramework' property.\n(project-name.dll) is the name of debugged project's build output dll. This is normally the same as the project file name but with a '.dll' extension.",
+                "default": "${workspaceFolder}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll"
               },
               "cwd": {
                 "type": "string",
                 "description": "Path to the working directory of the program being debugged. Default is the current workspace.",
-                "default": "${workspaceRoot}"
+                "default": "${workspaceFolder}"
               },
               "args": {
                 "anyOf": [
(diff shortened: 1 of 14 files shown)


src/configurationProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * TODO: Remove function. 
+     * 
+     * Note: serverUtils.requestWorkspaceInformation only retrieves one project. Therefore, generator will be incorrect for all folders
+     * except the first in a workspace. 
+     */
+    private checkWorkspaceInformationMatchesWorkspaceFolder(folder: vscode.WorkspaceFolder, info: WorkspaceInformationResponse): boolean {
+        return info.MsBuild.SolutionPath === folder.uri.fsPath;
```
DustinCampbell: Note: It is entirely possible for `MsBuild` to be null.

src/configurationProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+    private server: OmniSharpServer;
+
+    public constructor(server: OmniSharpServer) {
+        this.server = server;
+    }
+
+    /**
+     * TODO: Remove function. 
+     * 
+     * Note: serverUtils.requestWorkspaceInformation only retrieves one project. Therefore, generator will be incorrect for all folders
```
DustinCampbell: > serverUtils.requestWorkspaceInformation only retrieves one project.

This isn't true.

WardenGnaw: Is there a way to make it retrieve all projects in the current workspace for VsCode?

DustinCampbell: It will retrieve all projects loaded by OmniSharp. Currently, OmniSharp can only target a single folder. However, that folder contains multiple `.csproj` projects or an `.sln` with multiple projects, it will return them. We're going to be doing some work in omnisharp-roslyn to make OmniSharp support multiple folders: https://github.com/OmniSharp/omnisharp-roslyn/issues/909

src/configurationProvider.ts
DustinCampbell: Other than this, the change looks good to me. Note that this can be null when the project is a `project.json` file.

DustinCampbell:
If this requires 1.18, we have to wait until the next VS Code is out before shipping, correct?

gregg-miskelly:
Correct.

DustinCampbell:
Do we want to get this in for 1.13 then? IIRC, the PR you just merged works around the issue that the `debug.start` command is being removed in 1.18. So, if we hold 1.13 that for this change, there'll be a window of time where things won't really work.

DustinCampbell:
Also, what API are we talking about? Debug configuration providers have been present for a little while.

WardenGnaw:
During my testing, the current stable version of VsCode (1.17.2) was not providing initial configurations with the DebugConfigurationProvider.

DustinCampbell:
> During my testing, the current stable version of VsCode (1.17.2) was not providing initial configurations with the DebugConfigurationProvider.

Got it -- thanks! In that case, how do we want to coordinate this change? Should we hold it until after 1.13?

WardenGnaw:
@gregg-miskelly Yep. We would only need to put back initialConfigurations in launch.json.

gregg-miskelly:
@WardenGnaw would that still retain the benefits in 1.18? (Will they ignore `initialConfigurations` if we have a provider)

WardenGnaw:
@gregg-miskelly No. We are using it as a JSON object literal, so it will not be ignored.

WardenGnaw:
@DustinCampbell @gregg-miskelly I lied. This feature should work for 1.17.2. I'll be reverting the package.json to 1.17.0.

I was testing the C/C++ extension at the same time as this one. They do some overwriting of the activation events which made me lose the "onDebug" event which is required for DebugConfigurationProvider.

DustinCampbell:
You shouldn't lie. :smile:

gregg-miskelly:
>                 const generator = new AssetGenerator(info);

This will currently just check the workspace root for a launch.json/tasks.json. Do we need to check all the root folders instead?

---
Refers to: src/assets.ts:500 in b3b56af. [](commit_id = b3b56af386fbd5c3aa397766fd7936bb909b3d7a, deletion_comment = False)

src/configurationProvider.ts
```diff
@@ -0,0 +1,74 @@
+import * as vscode from 'vscode';
```
gregg-miskelly: You are missing a copyright header

src/configurationProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+            }
+            else {
+                // Error to write default C# configurations.
+                throw new Error("Does not contain dotnet core projects.");
+            }
+        }).catch((err) => {
+            // Provider will always create an launch.json file. Providing default C# configurations.
+            // jsonc-parser's parse to convert to JSON object without comments. 
+            return [
+                parse(createLaunchConfiguration(
```
gregg-miskelly: >createLaunchConfiguration [](start = 22, length = 25)

In the fall back we probably should create both web and console configurations.

src/configurationProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+     * except the first in a workspace. 
+     */
+    private checkWorkspaceInformationMatchesWorkspaceFolder(folder: vscode.WorkspaceFolder, info: WorkspaceInformationResponse): boolean {
+        return info && ((info.MsBuild && info.MsBuild.SolutionPath === folder.uri.fsPath) || (info.DotNet && info.DotNet.RuntimePath === folder.uri.fsPath));
+    }
+
+    /**
+	 * Returns a list of initial debug configurations based on contextual information, e.g. package.json or folder.
+	 */
+    provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration[]> {
```
gregg-miskelly: >undefined [](start = 64, length = 9)

Can this really be undefined? If so, we don't seem to handle it.

src/configurationProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export class CSharpConfigurationProvider implements vscode.DebugConfigurationProvider {
+    private server: OmniSharpServer;
+
+    public constructor(server: OmniSharpServer) {
+        this.server = server;
+    }
+
+    /**
+     * TODO: Remove function.
```
gregg-miskelly: >TODO [](start = 7, length = 4)

Can we remove this now?

WardenGnaw: Can I retrieve other workspace folder information with the change in #1806?

gregg-miskelly: @DustinCampbell should this work now?

DustinCampbell: You're waiting on https://github.com/OmniSharp/omnisharp-roslyn/issues/909. OmniSharp is not being launched on multiple workspace folders. Instead, the user can now launch OmniSharp on targets within multiple workspace folders.

gregg-miskelly: In that case I would suggest:
* Add a link the bug
* Instead of checking if the solution path is the same, seems like we should be checking if the folder is what we launched OmniSharp against.

DustinCampbell: Yes, I hadn't spotted that we were checking the SolutionPath against the folder name. That's going to be wrong. Note that you *can* have an `.sln` file, and in that case SolutionPath will point to the `.sln`.

src/assets.ts
```diff
@@ -327,7 +325,7 @@ function getBuildTasks(tasksConfiguration: tasks.TaskConfiguration): tasks.TaskD
 
     function findBuildTask(tasksDescriptions: tasks.TaskDescription[]) {
```
gregg-miskelly: What is VSCode's behavior if the user has an old style tasks.json? Should we be ignoring their tasks.json, replacing it, or is it really okay to add a new style task to an old style file?

src/configurationProvider.ts
WardenGnaw: Whoops I did see the question, VsCode had the function defined as that in their interface and I copied it over. So I added a check for ```folder``` but it got removed during other changes.

src/assets.ts
WardenGnaw: @gregg-miskelly The current behavior if a user has 0.1.0 in the version field of tasks.json is only an issue if they decide to use workspace folders.
image

Therefore, if we see an old tasks.json that is nonempty, we should be ignoring it. We could add a warning message saying to upgrade it if they wish to use workspace folders?

typings/vscode-tasks.d.ts
```diff
@@ -32,25 +32,21 @@ declare module "vscode-tasks" {
     export interface BaseTaskConfiguration {
```
TheRealPiotrP: Why do we have a copy of these typings? In the 1.1.6 version of the vscode package I see that these types are already defined in vscode.d.ts.

src/configurationProvider.ts
```diff
@@ -0,0 +1,101 @@
...
+            const generator = new AssetGenerator(info);
+            if (this.checkWorkspaceInformationMatchesWorkspaceFolder(folder) && containsDotNetCoreProjects(info)){
+                const dotVscodeFolder: string = path.join(folder.uri.fsPath, '.vscode');
+                const tasksJsonPath: string = path.join(dotVscodeFolder, 'tasks.json');
+                
+                // Make sure .vscode folder exists, addTasksJsonIfNecessary will fail to create tasks.json if the folder does not exist. 
+                fs.ensureDirSync(dotVscodeFolder);
+    
+                // if the file does not exist, addTasksJson
+                const addTasksJson: boolean = !fs.existsSync(tasksJsonPath);
```
TheRealPiotrP: I'm preparing a PR to take advantage of the taskProvider API. We'll want to keep these two implementations in-sync. I'm also adding quite a bit of test infrastructure that we can use to ensure that they are in sync.

typings/vscode-tasks.d.ts
WardenGnaw: Not sure why we originally had a copy of this. @DustinCampbell may be able to answer that question. However, the current typings do not have the old Tasks schema, specifically isBuildCommand.

DustinCampbell: Back in the day, it wasn't included in vscode.d.ts.

WardenGnaw:
@DustinCampbell, when you have the time. Can you review this PR again?

package.json
```diff
@@ -193,6 +193,7 @@
     "vscode": "^1.17.0"
   },
   "activationEvents": [
+    "onDebug",
```
DustinCampbell: Does this mean that the C# extension will activate whenever the user starts debugging any non-CS file? If so, wouldn't that be a bad experience if the user installs the C# extension, tries to debug JS and the C# extension starts downloading OmniSharp and the debugger?

src/assets.ts
```diff
@@ -244,28 +179,91 @@ export class AssetGenerator {
...
+            version: "2.0.0",
             tasks: [this.createBuildTaskDescription()]
         };
     }
 }
 
+export function createWebLaunchConfiguration(programPath: string, workingDirectory: string): string {
+    return `
+{
+"name": ".NET Core Launch (web)",
```
DustinCampbell: Any reason not to indent this like you did below to make it a bit more readable?

src/assets.ts
```diff
@@ -294,7 +292,7 @@ function findExecutableProjectJsonProjects(projects: protocol.DotNetProject[], c
     return result;
 }
 
-function containsDotNetCoreProjects(workspaceInfo: protocol.WorkspaceInformationResponse) {
+export function containsDotNetCoreProjects(workspaceInfo: protocol.WorkspaceInformationResponse) {
```
DustinCampbell: It feels like this helper should move someplace else if it needs to be accessed by more than just the assets generation. Perhaps `src/omnisharp/protocol.ts`? There's already other helpers like this there: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/protocol.ts#L589-L592.

src/common.ts
```diff
@@ -167,4 +167,20 @@ export function deleteInstallFile(type: InstallFileType): Promise<void> {
...
+
+/**
+ * This function checks to see if a subfolder is part of folder.
+ * 
+ * Assumes subfolder and folder are absolute paths.
+ * 
+ * @param subfolder subfolder to check if it is part of the folder parameter
+ * @param folder folder to check aganist
+ */
+export function isSubfolderOf(subfolder: string, folder:string): boolean {
```
DustinCampbell: Does this work if the paths differ by case? Does it matter?

DustinCampbell: Nit: space after `folder:`

src/configurationProvider.ts
```diff
@@ -0,0 +1,101 @@
...
+	 */
+    provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration[]> {
+        return serverUtils.requestWorkspaceInformation(this.server).then(info => {
+            const generator = new AssetGenerator(info);
+            if (this.checkWorkspaceInformationMatchesWorkspaceFolder(folder) && containsDotNetCoreProjects(info)){
+                const dotVscodeFolder: string = path.join(folder.uri.fsPath, '.vscode');
+                const tasksJsonPath: string = path.join(dotVscodeFolder, 'tasks.json');
+                
+                // Make sure .vscode folder exists, addTasksJsonIfNecessary will fail to create tasks.json if the folder does not exist. 
+                fs.ensureDirSync(dotVscodeFolder);
```
DustinCampbell: I'm seeing a lot of synchronous file I/O mixed with async code. Any concerns with that?

src/configurationProvider.ts
```diff
@@ -0,0 +1,101 @@
...
+                    const isWebProject = generator.hasWebServerDependency();
+                    const launchJson: string = generator.createLaunchJson(isWebProject);
+    
+                    // jsonc-parser's parse function parses a JSON string with comments into a JSON object. However, this removes the comments. 
+                    return parse(launchJson);
+                });
+            }
+            else {
+                // Error to write default C# configurations.
+                throw new Error("Does not contain dotnet core projects.");
```
DustinCampbell: If this message gets surfaced to the user, that should be ".NET Core", not "dotnet core"

WardenGnaw: It does not get surfaced to the user since its caught by the .catch(), but I will still change it.

src/common.ts
WardenGnaw: @DustinCampbell  Case sensitivity matters for Linux OS. Should I make it insensitive for Windows and Mac?

src/assets.ts
WardenGnaw: Oops. Lost indentation when copying.

package.json
WardenGnaw: This activation event is required for DebugConfigurationProviders. 

https://code.visualstudio.com/updates/v1_17#_debug-contributions-in-packagejson

DustinCampbell: Sure, but that's not what I asked. :smile:

src/common.ts
DustinCampbell: Not necessarily. If both values are expected to have the correct casing, it shouldn't matter. However, if they come from different sources, it might be a problem.

package.json
WardenGnaw: Sorry, I forgot to mention that I will still look into if the C# extension is activated in a node project. That was an FYI of why its there. :)

DustinCampbell: Thanks!

src/configurationProvider.ts
WardenGnaw: I'm not too experienced with async programming patterns. I used the synchronous version of ensureDir and exist to make sure the .vscode folder exists and to make sure the .vscode/tasks.json does not exist before calling addTasksJsonIfNecessary. Is there a better way to do this?

package.json
WardenGnaw: @DustinCampbell You are correct. It will activate on any F5 on any file type. 😞 

image

src/configurationProvider.ts
DustinCampbell: I don't think there's any concern for these small checks. I just wanted to make note of it.

DustinCampbell: FWIW, NodeJS async isn't really challenging. Someday, maybe you can spend some time digging into it since the platform really hangs on asynchronous patterns. It's not necessary now.

package.json
DustinCampbell: That's what I guessed. @gregg-miskelly: What do you think about this? I'm not sure there's much we can do about it, but I'm betting we'll hear about this.

gregg-miskelly: It certainly isn't ideal. I am assuming VS Code will activate us only once is a given VS Code session, so there isn't much we could do about this by trying to filter the activations. Right?

DustinCampbell: I think that's right. AFAIK, we'd essentially need to test all of our activation events manually when the extension activates.

gregg-miskelly: Do you think there is anything we should improve for 1.13? 

Here is one idea I had, but I am not sure if I like it - we could ship with an activation handler of '*' (see [docs](https://code.visualstudio.com/docs/extensionAPI/activation-events#_activationevents)). This will make sure that our extension activates right after the user installs the extension so hopefully it will feel connected with installing it. Then in our activate handler we update the package.json to the set of things currently in there.

DustinCampbell: That's not a bad idea, though you're right that it's pretty gross. :smile: Do you know if VS Code caches activation events from installed extensions?

Personally, I'm OK with living with this for 1.13 and then improve it later.

gregg-miskelly: The C++ extension rewrites several section of their package.json, but I am not sure if this is one of them. @WardenGnaw @pieandcakes do you know?

WardenGnaw: @DustinCampbell VsCode rereads the package.json to see what the activation events are. If we originally had ```*``` and replaced it with something else. On the next session of VSCode, it will reread the activationEvents array.

The C++ extension currently does this.

DustinCampbell: Clever! In that case, this seems like a reasonable approach.

DustinCampbell: @gregg-miskelly: Do you think we should do this for 1.13, or are you OK with doing it in 1.14, Same question for @TheRealPiotrP

gregg-miskelly: @DustinCampbell do you mean the change all up or the switch to use the '*' activation event?

If change all up - @WardenGnaw how broken are we in the VS Code 1.18 (the one that is just about to come out in stable channel) without this? Are the things we are replacing just deprecated or are the broken?

If the '*' activation - I don't have a strong opinion. I am happy to wait for 1.14 on that part if you would like.

src/common.ts
```diff
@@ -167,4 +167,25 @@ export function deleteInstallFile(type: InstallFileType): Promise<void> {
...
+/**
+ * This function checks to see if a subfolder is part of folder.
+ * 
+ * Assumes subfolder and folder are absolute paths.
+ * 
+ * @param subfolder subfolder to check if it is part of the folder parameter
+ * @param folder folder to check aganist
+ */
+export function isSubfolderOf(subfolder: string, folder: string): boolean {
+    if (os.platform() !== 'win32')
```
gregg-miskelly: Did you mean to do something here?

package.json
DustinCampbell: I meant using the '*' activation event. I'm OK with waiting for 1.14 for this.

src/common.ts
gregg-miskelly: Assuming your idea was to try and solve the case sensitivity problem, here is my thought - I don't think we should try to solve it. From what I can tell, I think the paths we are dealing with are all from the VS Code search APIs using the same roots they are providing to us, so it seems like it would be very surprising if we saw the same path but with different case. Solving the casing problem in java script is likely a hard problem. For one because on both OSX and Linux its possible to have either case sensitivity or insensitivity (though OSX is generally insensitive and Linux is sensitive). For another because I don't know how closely the java script casing tables are actually going to match what the OS uses.

My suggestion: Just add a note to the header that the two paths must have consistent casing.

src/coreclr-debug/install.ts
```diff
@@ -71,6 +71,28 @@ export class DebugInstaller {
         let manifestString = fs.readFileSync(manifestPath, 'utf8');
         let manifestObject = JSON.parse(manifestString);
 
+        manifestObject.activationEvents = [
```
gregg-miskelly: Per Dustin's comment, this can wait for the next PR, but this needs to go to a different spot - I think we need to rewrite `activationEvents` at the very start of our activation handler if it is still set to `[ "*" ]` to make sure that we will not continuously try to download things in the case that we don't succeed in downloading the debugger.

package.json
WardenGnaw: Without the '*' activation event, we are not broken. 

This was to fix the rare scenarios of people installing the C# extension, not opening a C# project for a while, then going to a non-C# project and then F5'ing would case the C# dependencies to start downloading.

src/common.ts
DustinCampbell: > My suggestion: Just add a note to the header that the two paths must have consistent casing.

yup. That's what I was thinking too.

PR #1805, opened 2017-10-24
Provide more granular file change events based on VS Code file create/delete events

rchande:
…/delete events

### DIFF
--- a/src/features/changeForwarding.ts
+++ b/src/features/changeForwarding.ts
@@ -45,17 +45,17 @@ function forwardFileChanges(server: OmniSharpServer): Disposable {
                 console.warn(`[o] failed to forward file change event for ${uri.fsPath}`, err);
                 return err;
             });
-        }        
+        };  
     }
 
     const watcher = workspace.createFileSystemWatcher('**/*.*');
     let d1 = watcher.onDidCreate(onFileSystemEvent(FileChangeType.Create));
+    let d2 = watcher.onDidDelete(onFileSystemEvent(FileChangeType.Delete));
+
     // In theory we don't need to subscribe to "change" notifications
     // because we already get them through the buffer update request
-    //let d2 = watcher.onDidChange(onFileChange);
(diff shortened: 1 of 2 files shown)


src/features/changeForwarding.ts
```diff
@@ -31,25 +32,30 @@ function forwardDocumentChanges(server: OmniSharpServer): Disposable {
...
     }
 
     const watcher = workspace.createFileSystemWatcher('**/*.*');
-    let d1 = watcher.onDidCreate(onFileSystemEvent);
-    let d2 = watcher.onDidChange(onFileSystemEvent);
-    let d3 = watcher.onDidDelete(onFileSystemEvent);
+    let d1 = watcher.onDidCreate(onFileSystemEvent(FileChangeType.Create));
+    // In theory we don't need to subscribe to "change" notifications
+    // because we already get them through the buffer update request
+    //let d2 = watcher.onDidChange(onFileChange);
```
DustinCampbell: Maybe move this to be last so that there isn't a gap between `d1` and `d3`?

src/omnisharp/protocol.ts
```diff
@@ -417,6 +417,17 @@ export interface PackageDependency {
     Name: string;
     Version: string;
 }
+    
+export interface FilesChangedRequest extends Request{
+    ChangeType: FileChangeType
```
DustinCampbell: You called this `FileChangeType` in your `omnisharp-roslyn` PR. So this probably doesn't actually work. :smile:

rchande: Updated the member name in the other PR.

DustinCampbell:
@rchande: Out of curiosity, have you tested that this works? I'm curious since the protocol didn't match omnisharp-roslyn at first.

rchande:
@DustinCampbell I did test this--verified in the debugger that we were getting hit for both create and delete events.

PR #1813, opened 2017-10-27
for single character completions, treat the completion list as incomplete

filipw:
Since we now have https://github.com/OmniSharp/omnisharp-roslyn/pull/990, which will offer good subsequence matching, we should treat completion lists for short words (empty and single letters) as incomplete. At the moment every completion list is considered complete and VS Code doesn't go back to the server for better matches as we type more characters.

This will force the editor to go again to the server with 2 letters, at which point the server will offer a much better result (as it will include the subsequence matches originally withheld for performance reasons). We may want to change that to 3 letters or more if the length of 2 is still too aggressive.

### DIFF
--- a/src/features/completionItemProvider.ts
+++ b/src/features/completionItemProvider.ts
@@ -10,7 +10,7 @@ import AbstractSupport from './abstractProvider';
 import * as protocol from '../omnisharp/protocol';
 import * as serverUtils from '../omnisharp/utils';
 import {createRequest} from '../omnisharp/typeConvertion';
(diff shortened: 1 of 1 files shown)


DustinCampbell:
I'm OK with doing this at 2 letters. In Visual Studio, note that we don't apply any substring filtering until at least two characters are typed.

DustinCampbell:
We do not have tests that specifically test `OmniSharpCompletionItemProvider` yet.

filipw:
> I'm OK with doing this at 2 letters. In Visual Studio, note that we don't apply any substring filtering until at least two characters are typed.

great - it *appeared* so to me but I wasn't 100% sure. 
Should have just asked you straight away 😊

Thanks for merging!

PR #1824, opened 2017-10-30
Ensure that we listen for file changes

DustinCampbell:
It's entirely possible for files to change without being opened and edited by VS Code. We need to watch for such changes.


rchande:
Thanks!

Issue #1820, opened 2017-10-30
Build system for writing integration tests


DustinCampbell:
@TheRealPiotrP : I'm adding this issue per our conversations last week. Feel free to adjust it as needed.

TheRealPiotrP:
Fixed in #1834

PR #1826, opened 2017-10-31
Add TaskProvider + integration test infrastructure

### LINKED ISSUE #1608: Make use of tasks-api
Since [today](https://github.com/Microsoft/vscode/issues/29835) and with the June release, VS Code will have API to programmatically contribute and refine tasks. With that API parts of the asset-generator should become obsolete. Feel free to ping me or @dbaeumer or drop some details about how the task.json is generated today and we might be able to do a PR for that.

TheRealPiotrP:
This PR aims to address #1820 and #1608.

I originally set out to fix #1608 by adding support for VS Code's Tasks API. However, I noticed that the existing build / test infrastructure did not enable me to write meaningful integration tests for this feature. Given that much of the code here is only verifiable in the context of VS Code I decided to make an infrastructure update prior to addressing the task at hand.

### Test Infrastructure Changes
My strategy in updating the test infrastructure is two-fold. First, I wanted to be able to run integration tests in the context of representative workspaces. Second, I wanted to be able to access data shared with VS Code which is not subsequently queryable via VS Code API's. Here are the key components of the change:

#### Test Assets
I introduced a new top-level folder called `TestAssets`. Within this folder you will find three representative workspaces:
  - A workspace with a single csproj-based project
  - A workspace with a sln file and several related projects
  - A workspace with a single project.json-based project

These workspaces are accompanied by matching .ts files which describe, from a test expectations perspective, the contents of the workspaces.

#### launch.json updates
The launch.json has been updated to have test execution configurations for each of these workspaces. Selecting a given configuration will open the VS Code Test Host targetted at the specified TestAsset workspace. All of the tests running within that Test Host will then be interacting directly with the target workspace, allowing them to produce predictable results while exhibiting their ability to reason about that wokspace type. Note that in order to run through all test scenarios you must iterate through all available Test Configurations.

#### package.json updates
package.json was updated to allow cmd-line execution of each of the new Test Configurations. The base `npm test` command was updated to iterate through all three configurations, thereby ensuring that a single cmd-line invocation will execute all tests.

### Product Updates
Atop this test infrastructure I have implemented a basic task provider which intends to replicate the behavior of the original tasks.json implementation. Namely, it generates a `build` task for each identified project. The feature itself is fairly straightforward, though it does demand that we answer several questions:

  [ ] should we simultaneously disable tasks.json generation?
  [ ] How do we want the tasks to appear in the VS Code UI? Currently they read `dotnet: build {project name}`. I would prefer to make these `dotnet: build {relative path to project file}`. However, I'd like to hear opinions on whether that relative path should be calculated by o#-vscode [by removing the workspace folder path from the fully qualified project path] or by o#-roslyn. @DustinCampbell looking for your insights here.
  [ ] How do we coordinate this PR with @WardenGnaw's #1801 work? I'm working with @WardenGnaw offline to figure this out.
  [ ] Should we block on https://github.com/Microsoft/vscode/issues/33523 being available? Without that change we don't have a mechanism for users to provide additional task configuration data via tasks.json.


TheRealPiotrP:
@DustinCampbell @rchande the build failure's output is fairly sparse. A full test pass, however, does require that a project.json-supporting CLI be installed on the build agent. Do we know if 1.0.0-preview2-003231 is available on these agents?

src/omnisharp/extension.ts
```diff
@@ -74,6 +76,8 @@ export function activate(context: vscode.ExtensionContext, reporter: TelemetryRe
         localDisposables.push(vscode.languages.registerCodeActionsProvider(documentSelector, codeActionProvider));
         localDisposables.push(reportDiagnostics(server, reporter, advisor));
         localDisposables.push(forwardChanges(server));
+
+        taskProvider.activate(context, server);
```
rchande: taskProvider is disposable, should it get added to `localDisposables`?

DustinCampbell:
@TheRealPiotrP : That would not be available on these build agents and would need to be installed separately.

DustinCampbell:
Also, it looks like you need to sign the CLA? That's a bit weird.

test/tasks.test.ts
```diff
@@ -3,18 +3,49 @@
...
+             .should.have.length(workspaceData.projects.length);
+    });
+    
+    test("build tasks should produce non-empty bin and obj directories", async () => {        
+        await workspaceData.deleteBuildArtifacts();
+        
+        for (let task of buildTasks) {
+            await vscode
+                .commands
+                .executeCommand('workbench.action.tasks.runTask', task.name);
```
rchande: Nice!

src/taskProvider.ts
```diff
@@ -0,0 +1,76 @@
...
+import { TaskRevealKind } from 'vscode';
+
+let taskProvider: vscode.Disposable | undefined;
+
+export function activate(_context: vscode.ExtensionContext, server: OmniSharpServer): void {
+	if (!vscode.workspace.workspaceFolders) {
+		return;
+	}
+
+	taskProvider = vscode.workspace.registerTaskProvider('dotnet', {
```
rchande: Where did the 'dotnet' constant come from?

src/taskProvider.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import * as serverUtils from './omnisharp/utils';
+import * as vscode from './vscodeAdapter';
+
+import { OmniSharpServer } from './omnisharp/server';
+import { TaskRevealKind } from 'vscode';
+
+let taskProvider: vscode.Disposable | undefined;
+
+export function activate(_context: vscode.ExtensionContext, server: OmniSharpServer): void {
```
rchande: Most of the other features are registered by the extension defined in `extension.ts`. Does the adapter pattern mean that we will have a separate extension class for every feature we want to test through the adapter?

TheRealPiotrP: I'm open for comment on this one. I prefer smaller, task-oriented units of code as I find them easier to understand, test, and modify.

src/taskProvider.ts
TheRealPiotrP: It's defined in package.json. I chose this constant fairly arbitrarily. Would we prefer `c#`, `csharp`, `.net`?

test/tasks.test.ts
TheRealPiotrP: Thanks @DustinCampbell for this little nugget :)

WardenGnaw:
Is the generated task's type field correct? 

image

dnfgituser:
Hi, I am closing and re-opening this PR to bump the CLA service. Sorry for the inconvenience!

TheRealPiotrP:
@WardenGnaw I believe this is due to the default tasks.json schema validator looking at the 0.1.0 schema instead of 2.0.0. The configuration here is the same as in the example: https://github.com/Microsoft/vscode-extension-samples/blob/master/task-provider-sample/src/extension.ts#L125.

src/omnisharp/extension.ts
TheRealPiotrP: done

TheRealPiotrP:
@rchande @DustinCampbell any guesses here? I finally got myself into a state where tests are running on CI, but O#-Roslyn is discovering 0 projects where I expected to see 1. Could this be a missing dependency issue?

DustinCampbell:
Out of curiosity, why are we working so hard to add support for project.json? Project.json projects are officially deprecated in C# for VS Code and the extension displays a warning to the user about migration whenever OmniSharp is launched in such a way that it loads a project.json project. I would propose that we don't bother with project.json and just target csproj.

Also, note that project.json simply does not work on macOS 10.13 (High Sierra). Old .NET Core SDKs from the project.json error just segfault on 10.13. So, any project.json tests that require restore or build will fail on newer operating systems. I'm on High Sierra already. So, I wouldn't even be able to run tests like that.

package.json
```diff
@@ -2154,6 +2163,22 @@
...
+					"type": "string",
+					"description": "The Rake task to customize"
+					},
+					"file": {
+					"type": "string",
+					"description": "The Rake file that provides the task. Can be omitted."
+					}
+				}
+			}
+		]
```
DustinCampbell: indent level looks different here. Was that intentional?

src/taskProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+		for (let project of workspaceInformation.MsBuild.Projects) {
+			tasks.push(provideBuildTask(project.AssemblyName, project.Path));
+		}
+	}
+
+	if (workspaceInformation.DotNet && workspaceInformation.DotNet.Projects) {
+		for (let project of workspaceInformation.DotNet.Projects) {
+			tasks.push(provideBuildTask(project.Path, project.Path));
+		}
+	}
```
DustinCampbell: FWIW, I wouldn't bother adding this feature for project.json, since the support is officially deprecated.

src/taskProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+'use strict';
+
+import * as serverUtils from './omnisharp/utils';
+import * as vscode from './vscodeAdapter';
+
+import { OmniSharpServer } from './omnisharp/server';
+import { TaskRevealKind } from 'vscode';
+
+export function activate(_context: vscode.ExtensionContext, server: OmniSharpServer): vscode.Disposable {
+	if (!vscode.workspace.workspaceFolders) {
```
DustinCampbell: Please use spaces rather than tabs. This is defined the repro's .vscode/settings.json file.

src/taskProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+		resolveTask(_task: vscode.Task): vscode.Task | undefined {
+			return undefined;
+		}
+	});
+}
+
+async function provideDotnetTasks(server: OmniSharpServer): Promise<vscode.Task[]> {
+	let tasks: vscode.Task[] = [];
+	
+	let workspaceInformation = await serverUtils.requestWorkspaceInformation(server);
```
DustinCampbell: Thank you for the first usage of `await`! :smile:

src/vscodeAdapter.ts
```diff
@@ -0,0 +1,62 @@
+import * as vscode from "vscode";
```
DustinCampbell: It feels like this test infrastructure stuff should be in a separate folder rather than at the top-level `src`. Should it be somewhere in the `test` folder? Or, does this need to be included with the extension?

src/taskProvider.ts
TheRealPiotrP: not sure why that wasn't respected... fixed.

TheRealPiotrP: also added a tslint rule

src/vscodeAdapter.ts
TheRealPiotrP: fixed.

src/vscodeTasksAdapter.ts
```diff
@@ -0,0 +1,40 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
```
DustinCampbell: Should this be in the 'src' folder or the 'test' folder?

src/taskProvider.ts
DustinCampbell: Thanks!

src/vscodeTasksAdapter.ts
TheRealPiotrP: fixed. I created this directory structure:
- test
  - unitTests
  - integrationTests
    - testAssets

.vscode/launch.json
```diff
@@ -7,6 +7,7 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
+                "${workspaceRoot}/test/integrationTests/testAssets/slnWithCsproj",
```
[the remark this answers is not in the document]
TheRealPiotrP: Yes I did. The extension typically opens in an arbitrary directory and I wanted to have it open in a directory that represents a 'kitchen sink'. I'm fine with reverting this, but found it useful for my own development.

src/features/definitionMetadataDocumentProvider.ts
```diff
@@ -1,4 +1,10 @@
-import { workspace, Uri, TextDocument, Disposable, TextDocumentContentProvider} from 'vscode';
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
```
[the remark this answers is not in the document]
TheRealPiotrP: yay tslint!

.vscode/launch.json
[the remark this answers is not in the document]
DustinCampbell: Often, when reproducing an issue, I open the debuggee on a particular folder. Then, subsequent F5 will launch on the same directory allowing me to continue debugging. With this change, I'd need to either update launch.json first or open the folder on every debug session. I would definitely prefer not having this.

PR #1836, opened 2017-11-06
Update to latest OmniSharp and prepare for 1.3.0-beta5

DustinCampbell:


DustinCampbell:
@gregg-miskelly are there any updates to the changelog for the debugger? I did not add anything for the DebugConfiguraitonProvider, but that should probably be called out.

gregg-miskelly:
In addition to the configuration provider, the only other debugger related change to call out is that we fixed the breaking change that VS Code is about to roll out with unit test debugging.

gregg-miskelly:
I will work on updating the changelog for debugger items once you merge this.

DustinCampbell:
Sounds good!

PR #1839, opened 2017-11-06
Add Debug Launch integration test

TheRealPiotrP:
The integration test:
 - deletes the .vscode directory
 - deletes existing build artifacts
 - issues the `dotnet.generateAssets` command
 - executes the `.NET Core Launch (console)` debug configuration
 - validates that a debugSession becomes active
 - waits for debugging to stop

To enable this, and other integration tests, I exposed an export on the csharp extension which can be awaited until the extension is initialized. Initialized is defined as omnisharp and the debugger being ready for invocation. Without awaiting this promise, commands issues to the extension have unpredictable behaviors since they are registered but not yet able to be handled.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -3,19 +3,19 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as vscode from 'vscode';
-import TelemetryReporter from 'vscode-extension-telemetry';
-
-import * as coreclrdebug from './coreclr-debug/activate';
 import * as OmniSharp from './omnisharp/extension';
+import * as coreclrdebug from './coreclr-debug/activate';
 import * as util from './common';
-import { Logger } from './logger';
+import * as vscode from 'vscode';
+
 import { CSharpExtDownloader } from './CSharpExtDownloader';
+import { Logger } from './logger';
+import TelemetryReporter from 'vscode-extension-telemetry';
 import { addJSONProviders } from './features/json/jsonContributions';
 
 let _channel: vscode.OutputChannel = null;
 
-export function activate(context: vscode.ExtensionContext): any {
+export async function activate(context: vscode.ExtensionContext): Promise<{ initializationFinished: Promise<void> }> {
 
     const extensionId = 'ms-vscode.csharp';
     const extension = vscode.extensions.getExtension(extensionId);
@@ -29,19 +29,25 @@ export function activate(context: vscode.ExtensionContext): any {
 
     let logger = new Logger(text => _channel.append(text));
 
-    ensureRuntimeDependencies(extension, logger, reporter)
-        .then((success : boolean) => {
-            // activate language services
-            OmniSharp.activate(context, reporter, _channel);
+    let runtimeDependenciesExist = await ensureRuntimeDependencies(extension, logger, reporter);
 
-            // register JSON completion & hover providers for project.json
-            context.subscriptions.push(addJSONProviders());
-            
-            if (success) {
-                // activate coreclr-debug
-                coreclrdebug.activate(extension, context, reporter, logger, _channel);
-            }
-        });
+    if (!runtimeDependenciesExist) {
+        //do something
+    }
+    
+    // activate language services
+    let omniSharpPromise = OmniSharp.activate(context, reporter, _channel);
+
+    // register JSON completion & hover providers for project.json
+    context.subscriptions.push(addJSONProviders());
+    
+    // activate coreclr-debug
+    let coreClrDebugPromise = coreclrdebug.activate(extension, context, reporter, logger, _channel);
+
+    return {
+        initializationFinished: Promise.all([omniSharpPromise, coreClrDebugPromise])
+        .then(a => {})
+    };
 }
 
 function ensureRuntimeDependencies(extension: vscode.Extension<any>, logger: Logger, reporter: TelemetryReporter): Promise<boolean> {
--- a/src/main.ts
+++ b/src/main.ts
@@ -30,10 +30,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<{ init
     let logger = new Logger(text => _channel.append(text));
 
     let runtimeDependenciesExist = await ensureRuntimeDependencies(extension, logger, reporter);
-
-    if (!runtimeDependenciesExist) {
-        //do something
-    }
     
     // activate language services
     let omniSharpPromise = OmniSharp.activate(context, reporter, _channel);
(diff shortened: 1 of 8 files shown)


src/main.ts
```diff
@@ -29,19 +29,25 @@ export function activate(context: vscode.ExtensionContext): any {
...
-            
-            if (success) {
-                // activate coreclr-debug
-                coreclrdebug.activate(extension, context, reporter, logger, _channel);
-            }
-        });
+    let runtimeDependenciesExist = await ensureRuntimeDependencies(extension, logger, reporter);
+
+    if (!runtimeDependenciesExist) {
+        //do something
```
gregg-miskelly: I don't think you want to do anything else in this function if `runtimeDependenciesExist` is false.

src/main.ts
```diff
@@ -29,19 +29,25 @@ export function activate(context: vscode.ExtensionContext): any {
...
+
+    // register JSON completion & hover providers for project.json
+    context.subscriptions.push(addJSONProviders());
+    
+    // activate coreclr-debug
+    let coreClrDebugPromise = coreclrdebug.activate(extension, context, reporter, logger, _channel);
+
+    return {
+        initializationFinished: Promise.all([omniSharpPromise, coreClrDebugPromise])
+        .then(a => {})
```
gregg-miskelly: Is the `.then(a => {})` part required? If so, out of curiosity -- why?

test/integrationTests/launchConfiguration.integration.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { should } from 'chai';
+import testAssetWorkspace from './testAssets/testAssetWorkspace';
+
+const chai = require('chai'); 
+chai.use(require('chai-arrays')); 
+chai.use(require('chai-fs')); 
+ 
+suite(`Tasks generation: ${testAssetWorkspace.description}`, function() {
+    suiteSetup(async function() { 
+        should();
```
gregg-miskelly: Did you want to call `should()` by itself?

src/main.ts
TheRealPiotrP: throw new Error("dependencies don't exist");
?

Since we're just starting to use async/await in this repo, I wanted to hear what folks think...

test/integrationTests/testAssets/testAssets.ts
```diff
@@ -45,6 +45,18 @@ export class TestAssetWorkspace {
     async deleteBuildArtifacts(): Promise<void> {
         this.projects.forEach(async p => await p.deleteBuildArtifacts());
     }
+    
+    get vsCodeDirectoryPath(): string {
+        return path.join(vscode.workspace.rootPath, ".vscode");;
```
gregg-miskelly: You have an extra semicolon here.

src/main.ts
gregg-miskelly: If you `throw`, what will VS Code do?

src/main.ts
TheRealPiotrP: the effect of `.then(a => {})` is to hide the value of `a`. 

I wanted to expose a Thenable which resolves once our extension is fully initialized [in addition to activated]. I did not intend to leak the internal implementation of that Thenable, so I'm using this statement to hide Promise.all's default result value.

DustinCampbell: That sounds very commentable :smile:

test/integrationTests/launchConfiguration.integration.test.ts
TheRealPiotrP: This is the same pattern as `common.test.ts`, `json.test.ts`, etc. I'm just happening to use a function instead of a lambda here, right?

TheRealPiotrP:
@WardenGnaw @gregg-miskelly are there better validations for "we are debugging" than what I'm doing?

src/main.ts
DustinCampbell: I'm pretty sure VS Code handles it and displays a message. https://github.com/Microsoft/vscode/blob/27c07934d7d538f6799dc78dc19a174f8e31aaa7/src/vs/workbench/api/node/extHostExtensionActivator.ts#L314-L320

src/main.ts
gregg-miskelly: Ah, okay.

src/main.ts
gregg-miskelly: In this case I think you do NOT want to throw as we already tell the user what failed. Maybe expose a bool back to the test that the test can check?

test/integrationTests/launchConfiguration.integration.test.ts
gregg-miskelly: Oh, got it.

gregg-miskelly:
We have never looked at the debugging related API that VS Code exposes to extensions, so we are definitely not experts, but what you did looks reasonable to me.

test/integrationTests/testAssets/testAssets.ts
TheRealPiotrP: fixed.

src/main.ts
TheRealPiotrP: Added comment.

src/main.ts
TheRealPiotrP: This is not present for test purposes. Once I de-thenefied this block of code it became apparent that we have forked behaviors. Prior to this change:
 - IF ensureRuntimeDependencies == false
  - Omnisharp tries to activate anyways. Is that legal?
  - JsonProviders are registered. Again, is that legal?
  - coreclrdebug does not activate

This seemed funny to me. Is it the intent that Omnisharp starts happily but the debugger is not activated? Shouldn't we behave like a single extension?

gregg-miskelly: Registering the Json providers is fine.
Starting OmniSharp could be fine if the issue was that the debugger failed to download. I think that is why that code always runs today.
We definitely do NOT want to activate coreclrdebug, as that code is there to finalize the debugger install, but it doesn't make sense to do so if the download failed.

TheRealPiotrP: So are we saying that `ensureRuntimeDependencies` only pulls down debugger dependencies and that it has no effect on omnisharp and json provider?

While writing these integration tests I've already discovered an undefined state in the control [The control claims to be active even though it's not ready to receive commands]. This feels like another place where we end up in an undefined state. If we're confident that `ensureRuntimeDependencies` only concerns the debugger then we can design/test for that scenario. Otherwise, we are missing an opportunity to direct the user to fix their installation and are instead providing them with an undefined experience...

TheRealPiotrP:
Note, running these first two integration tests costs about 20s per configuration. My hope is that much of this time is spent downloading & installing required components, and warming up the CLI. If we see a 20s growth per-test in the future we will need to reconsider this approach.

src/main.ts
gregg-miskelly: `ensureRuntimeDependencies` will download OmniSharp and the Debugger.

What was the case you ran into?

I know that `completeDebuggerInstall` is really async so if we want to wait for the debugger install to be complete, there should be an `await` on src/coreclr/activate.ts line 51.

TheRealPiotrP: @gregg-miskelly the issue I ran into is that I issued a command registered by our extension before omnisharp initialized and that command invocation resulted in an error.

I opened https://github.com/OmniSharp/omnisharp-vscode/issues/1840 to track startup behavior. Let's decide what we want to do in that issue. I'll remove the current `do something` block for this PR.

TheRealPiotrP:
I addressed the PR feedback so I'm merging the PR. I especially want to get the first integration test in so that folks can give feedback on the current integration test infrastructure.

PR #1841, opened 2017-11-07
Updates for 1.13 release

DustinCampbell:


DustinCampbell:
@gregg-miskelly : I noticed that the readme states, "NOTE: Mono and Desktop CLR debugging is not supported." Is this still true? I though Desktop CLR debugging was there now.

gregg-miskelly:
Desktop CLR support is present, but only partial - it will only work for 64-bit processes, and only portable PDBs are supported.

DustinCampbell:
Got it, so "not supported' is probably OK then.

PR #1842, opened 2017-11-07
Fix asset generation for multi-root workspaces

DustinCampbell:


src/assets.ts
```diff
@@ -379,13 +399,13 @@ interface PromptItem extends vscode.MessageItem {
     result: PromptResult;
 }
 
-function promptToAddAssets() {
+function promptToAddAssets(workspaceFolder: vscode.WorkspaceFolder) {
```
WardenGnaw: Does this create the tasks.json and launch.json for the multi-root workspace?

DustinCampbell: No, this causes us to prompt for the correct workspace folder on which OmniSharp is launched. OmniSharp itself does not support multiple roots. It can only be launched on one thing at a time, which can be switched by clicking the OmniSharp status bar item in VS Code. So, if OmniSharp is launched in a different folder other than the first one, we need to ensure that we generate in that workspace folder. Make sense?

DustinCampbell:
Anybody interested in signing off? I've tested asset generation in multi-root scenarios when OmniSharp is launched on different folders and it is working fine.

(FWIW, the debug configuration stuff has some issues specifically when generating configurations when F5 is pressed. In that case, it seems the VS Code will only generate for the first folder.)

DustinCampbell:
I'm going to force merge this PR so I can push it to the release branch. Let me know if you have any concerns.

gregg-miskelly:
LGTM

src/assets.ts
WardenGnaw: Sounds good!

PR #1853, opened 2017-11-10
Enable Integration Tests to expose Logger output

TheRealPiotrP:
Enables subscribing a logging callback that will be invoked by all instances of Logger.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -2,6 +2,11 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
+let Subscriber: (message: string) => void;
+
+export function SubscribeToAllLoggers(subscriber: (message:string) => void) {
+    Subscriber = subscriber;
+}
 
 export class Logger {
     private _writer: (message: string) => void;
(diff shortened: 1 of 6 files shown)


src/logger.ts
```diff
@@ -2,6 +2,11 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
+let Subscriber: (message: string) => void;
```
DustinCampbell: minor, minor nit: blank line before

DustinCampbell:
CI failed with the same weird array of nulls we saw in my PR. Maybe that was unrelated to issue I needed to fix? I'm wondering if the problem is related to the fact that VS Code 1.18 just shipped, so the stable build used for test was updated.

TheRealPiotrP:
@DustinCampbell @rchande @akshita31 this change ended up being super noisy in the logs. However, I was able to make the UX quite nice anyways :)

The latest revision of this change:
* writes logs to log files in a `.logs` directory
* after the test script runs it `cat`'s the log files to the Travis build log
* It employs a loosely documented Travis trick to enable log folding around the printed log files

Check it out!
screen shot 2017-11-17 at 12 28 32 am

PR #1854, opened 2017-11-10
Handle Rename and OpenDocument responses from Omnisharp

rchande:
Also adds an integration test that validates the behavior of the "rename
file to match type" refactoring:
* File is renamed on disk
* Old buffer is closed
* New buffer is opened


rchande:
Omnisharp side: https://github.com/OmniSharp/omnisharp-roslyn/pull/1023

rchande:
cc @akshita31

src/features/codeActionProvider.ts
```diff
@@ -112,20 +114,62 @@ export default class CodeActionProvider extends AbstractProvider implements vsco
...
+                        // Record that file name and keep processing CodeActions.
+                        // If a CodeAction requests that we open multiple files 
+                        // we only open the last one (what would it mean to open multiple files?)
+                        fileToOpen = vscode.Uri.file(change.FileName);
                     }
 
-                    edit.set(uri, edits);
+                    if (change.ModificationType == FileModificationType.Modified)
+                    {
+                        let uri = vscode.Uri.file(change.FileName);
```
DustinCampbell: Based on your comment when `FileModificationType.Renamed` is true, should we check to see if this file is included in `renamedFiles`? And if so, does that change the order in which we loop through the changes?

src/features/codeActionProvider.ts
```diff
@@ -112,20 +114,62 @@ export default class CodeActionProvider extends AbstractProvider implements vsco
...
                 }
 
-                return vscode.workspace.applyEdit(edit);
-            }
+                let applyEditPromise = vscode.workspace.applyEdit(edit);
+
+                // Unfortunately, the textEditor.Close() API has been deprecated
+                // and replaced with a command that can only close the active editor.
+                // If files were renamed that weren't the active editor, their tabs will
+                // be left open and marked as "deleted" by VS Code
```
DustinCampbell: Do we have a bug on VS Code about this?

test/providertest.ts
```diff
@@ -0,0 +1,12 @@
+import { should } from 'chai';
```
DustinCampbell: What is this file? Did it get added accidentally?

src/features/codeActionProvider.ts
rchande: Right. We should probably collect all the renames first, so we don't try to edit files that got renamed.

DustinCampbell:
OK. CI should *properly* fail for this now, since this PR requires a changes in omnisharp-roslyn

src/features/codeActionProvider.ts
rchande: Filed https://github.com/Microsoft/vscode/issues/38239

test/integrationTests/codeActionRename.integration.test.ts
```diff
@@ -0,0 +1,42 @@
...
+suite(`Code Action Rename ${testAssetWorkspace.description}`, function() {
+    suiteSetup(async function() { 
+        should();
+
+        let csharpExtension = vscode.extensions.getExtension("ms-vscode.csharp"); 
+        if (!csharpExtension.isActive) { 
+            await csharpExtension.activate(); 
+        }
+
+        testAssetWorkspace.deleteBuildArtifacts();
```
TheRealPiotrP: do you need to delete build artifacts?

test/integrationTests/codeActionRename.integration.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        await fs.rimraf(testAssetWorkspace.vsCodeDirectoryPath);
```
TheRealPiotrP: do you need to delete the `.vscode` directory?

test/integrationTests/codeActionRename.integration.test.ts
rchande: Copypasta, will remove.

test/integrationTests/codeActionRename.integration.test.ts
rchande: Copypasta, will remove.

rchande:
@DustinCampbell Do these tests automatically consume the latest bits from omnisharp?

test/integrationTests/codeActionRename.integration.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        await csharpExtension.exports.initializationFinished;
+
+    });
+    
+    test("Code actions can remame and open files", async () => {                
+        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("test.cs", "class C {}");
```
TheRealPiotrP: running this test more than once will give us false positives. The first execution will put C.cs in the workspace, the second will simply find it.

My first thought was to add a line that deletes both `test.cs` and `C.cs`... but this feels like we will have lots of this sort of cleanup code running around. What if we added a method on e.g. testAssetWorkspace that performed a `git clean -xfd`? This would allow tests to easily put the asset back into a known state [we could eliminate the custom methods for deleting bin/obj, as well as for `.vscode`]. Thoughts?

TheRealPiotrP:
@rchande the Code version gets pinned in a few places:
https://github.com/OmniSharp/omnisharp-vscode/search?utf8=%E2%9C%93&q=1.17.2&type=

We pinned it because 1.18 broke our integration tests. In general, I'd prefer we stick to this model so that we can deal with updates to VSCode separately from other feature work. @DustinCampbell got blocked on a feature recently, for example, because 1.18 happened to ship while he was working on his PR.

test/integrationTests/codeActionRename.integration.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        await vscode.commands.executeCommand("vscode.open", fileUri);
+        let c = await vscode.commands.executeCommand("vscode.executeCodeActionProvider", fileUri, new vscode.Range(0, 7, 0, 7)) as {command: string, arguments: string[]}[];
+        expect(c.length).to.equal(3);
```
TheRealPiotrP: 3 seems like a magic number. Can we provide some text [`expect` should take a second parameter that is used to create the failure output string] that tells us why 3 is the right answer?

test/integrationTests/codeActionRename.integration.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        await vscode.commands.executeCommand(c[1].command, ...c[1].arguments)
+        expect(vscode.window.activeTextEditor.document.fileName).contains("C.cs");
```
TheRealPiotrP: here as well... this might not be obvious... "Because the active file was renamed and its new name should be reflected" or something?

TheRealPiotrP:
I suspect tests are failing because `3` is the right magic context but not in the other [sln vs. csproj]. Do we need to run this test in both contexts? Perhaps we should just limit it to the csproj context?

DustinCampbell:
@DustinCampbell Do these tests automatically consume the latest bits from omnisharp?

Only once C# for VS Code is updated with new bits from OmniSharp. A release from omnisharp-roslyn has not yet been built with your changes. Once that happens, C# for VS Code will need to be updated with them (after using the tool you were building for uploading them to the CDNs).

rchande:
@TheRealPiotrP I think tests are failing because as, @DustinCampbell mentioned, we haven't consumed the new omnisharp bits that would provide the behavior being tested. That said, I will incorporate the changes to the test that you suggested.

test/integrationTests/codeActionRename.integration.test.ts
rchande: I somewhat dislike having the projects manipulated by tests under the repo because they show up in my git status, so it would be nice if we could do something to get them out of there.

Relatedly, Omnisharp-roslyn has a pretty nice system where you can request a real on-disk project from a test helper, and one is created for you under a random folder under %temp%. If we implemented that, we wouldn't have to worry about cleaning up test projects after we're done testing because a test would never see the same instance of a project.

O#r's build knows about the test projects so it can build/restore them as part of the solution build. That means deploying a project for test simply copies all the files into a temp directory. 

I think that might be a good solution for these tests. What do you think?

DustinCampbell:
FWIW, I *did* state that very fact above:

> OK. CI should properly fail for this now, since this PR requires a changes in omnisharp-roslyn

test/integrationTests/codeActionRename.integration.test.ts
DustinCampbell: Agreed 100%. This is why the tests in omnisharp-roslyn that operate on test projects go to an effort to copy test projects to a temporary folder and delete them when the test is finished.

david-driscoll:
We could deploy the latest omnisharp bits to a cdn / azure storage / myget / <insert thing here> and have it act as a beta feed.

test/integrationTests/codeActionRename.integration.test.ts
TheRealPiotrP: CLI does the same, @DustinCampbell. However, in this scenario we have a non-trivial startup cost for opening a new folder in VS Code. We also don't have a known model for plugging in an `outer loop` test orchestrator. Such a thing would need to enumerate all the tests, understand which test asset is needed by which test, make a copy of the asset & launch VS Code in that copy's folder, and then get vscode to execute just that test. I agree that this would be cool to have [if we can make it fast] but it feels like a considerable amount of work...

PR #1861, opened 2017-11-13
Set CODE_VERSION environment variable to 1.17.2 for tests

DustinCampbell:
Note: This is not a long-term solution, as we'll eventually need to update the VS Code engine version that C# for VS Code targets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: node_js
 node_js:
 - "6"
 
+env:
+- CODE_VERSION=1.17.2
+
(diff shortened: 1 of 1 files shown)


DustinCampbell:
@TheRealPiotrP : I'm going to go ahead and merge this. Please let me know if you have any concerns. We'll need to sort out the problem with our integration tests on VS Code 1.18 pretty soon. I've filed https://github.com/OmniSharp/omnisharp-vscode/issues/1862 to track this.

PR #1869, opened 2017-11-16
Handling Newlines in XML Documentation

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1057

The XML Documentation being returned by omnisharp-roslyn contained newlines properly, however vscode is displaying a single newline when there are two consecutive new lines are provided. Hence a change has been made to the hoverProvider to replace a single newline with two new lines. An integration test has been added for the same. However on adding this test, the "Launch singleCsproj Workspace Tests" and "Launch slnWithCsproj Workspace Tests" are failing.

Please review @DustinCampbell  @TheRealPiotrP @rchande

### DIFF
--- a/src/features/hoverProvider.ts
+++ b/src/features/hoverProvider.ts
@@ -21,7 +21,8 @@ export default class OmniSharpHoverProvider extends AbstractSupport implements H
 
         return serverUtils.typeLookup(this._server, req, token).then(value => {
             if (value && value.Type) {
-                let contents = [extractSummaryText(value.Documentation), { language: 'csharp', value: value.Type }];
+                let addLine = value.Documentation.split("\n").join("\n\n");
+                let contents = [extractSummaryText(addLine), { language: 'csharp', value: value.Type }];
                 return new Hover(contents);
             }
         });
--- a/test/integrationTests/hoverProvider.integration.test.ts
+++ b/test/integrationTests/hoverProvider.integration.test.ts
@@ -1,0 +1,63 @@
+
+import * as fs from 'async-file';
+import * as vscode from 'vscode';
+import * as path from 'path';
+
+import poll from './poll';
+import { should, expect } from 'chai';
+import testAssetWorkspace from './testAssets/testAssetWorkspace';
+
+const chai = require('chai'); 
+chai.use(require('chai-arrays')); 
+chai.use(require('chai-fs')); 
+
+suite(`Test Hover Behavior ${testAssetWorkspace.description}`, function() {
+    suiteSetup(async function() { 
+        should();
+
+        let csharpExtension = vscode.extensions.getExtension("ms-vscode.csharp"); 
+        if (!csharpExtension.isActive) { 
+            await csharpExtension.activate();
+        }
+
+        testAssetWorkspace.deleteBuildArtifacts();
+        
+        await fs.rimraf(testAssetWorkspace.vsCodeDirectoryPath);
+        
+        await csharpExtension.exports.initializationFinished;
+                
+        await vscode.commands.executeCommand("dotnet.generateAssets");
+        
+        await poll(async () => await fs.exists(testAssetWorkspace.launchJsonPath), 10000, 100);
+
+    });
+
+
+    test("Hover returns the correct XML", async () => {                
+        
+        let program : string = "using System;\
+        \nnamespace hoverXmlDoc \
+        \n{\
+        \n    class testissue\
+        \n    {\
+        \n    ///<summary>Checks if object is tagged with the tag.</summary>\
+        \n    /// <param name=\"gameObject\">The game object.</param> \
+        \n    /// <param name=\"tagName\">Name of the tag </param>\
+        \n    /// <returns>Returns <c> true</c>if object is tagged with tag.</returns>\
+        \n\
+        \n        public static bool Compare(int gameObject,string tagName)\
+        \n        {\
+        \n            return gameObject.TagifyCompareTag(tagName);\
+        \n        }\
+        \n    }\
+        }";
+        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("test.cs", program); 
+        await vscode.commands.executeCommand("vscode.open", fileUri);
+
+        let c = await vscode.commands.executeCommand("vscode.executeHoverProvider", fileUri,new vscode.Position(10,30));
+        let answer:string = "Checks if object is tagged with the tag.\n\ngameObject: The game object.\n\ntagName: Name of the tag \n\nReturns: Returns trueif object is tagged with tag.";
+        expect(c[0].contents[0].value).to.equal(answer);
+        
+    });
+
+});
--- a/test/integrationTests/hoverProvider.integration.test.ts
+++ b/test/integrationTests/hoverProvider.integration.test.ts
@@ -1,3 +1,7 @@
+/*--------------------------------------------------------------------------------------------- 
+ *  Copyright (c) Microsoft Corporation. All rights reserved. 
+ *  Licensed under the MIT License. See License.txt in the project root for license information. 
+ *--------------------------------------------------------------------------------------------*/
 
 import * as fs from 'async-file';
 import * as vscode from 'vscode';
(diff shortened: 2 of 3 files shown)


test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+suite(`Test Hover Behavior ${testAssetWorkspace.description}`, function() {
+    suiteSetup(async function() { 
+        should();
+
+        let csharpExtension = vscode.extensions.getExtension("ms-vscode.csharp"); 
+        if (!csharpExtension.isActive) { 
+            await csharpExtension.activate();
+        }
+
+        testAssetWorkspace.deleteBuildArtifacts();
```
TheRealPiotrP: do we need this?

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        
+        await fs.rimraf(testAssetWorkspace.vsCodeDirectoryPath);
```
TheRealPiotrP: or this?

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        
+        await csharpExtension.exports.initializationFinished;
+                
+        await vscode.commands.executeCommand("dotnet.generateAssets");
```
TheRealPiotrP: or this?

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        
+        await poll(async () => await fs.exists(testAssetWorkspace.launchJsonPath), 10000, 100);
```
TheRealPiotrP: or this?

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        \n        {\
+        \n            return gameObject.TagifyCompareTag(tagName);\
+        \n        }\
+        \n    }\
+        }";
+        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("test.cs", program); 
+        await vscode.commands.executeCommand("vscode.open", fileUri);
+
+        let c = await vscode.commands.executeCommand("vscode.executeHoverProvider", fileUri,new vscode.Position(10,30));
+        let answer:string = "Checks if object is tagged with the tag.\n\ngameObject: The game object.\n\ntagName: Name of the tag \n\nReturns: Returns trueif object is tagged with tag.";
```
TheRealPiotrP: the `:string` type specifier should not be necessary here.

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        expect(c[0].contents[0].value).to.equal(answer);
```
TheRealPiotrP: for my own education, why might `contents` have multiple values?

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        await vscode.commands.executeCommand("dotnet.generateAssets");
+        
+        await poll(async () => await fs.exists(testAssetWorkspace.launchJsonPath), 10000, 100);
+
+    });
+
+
+    test("Hover returns the correct XML", async () => {                
+        
+        let program : string = "using System;\
```
TheRealPiotrP: Can we use a [multi-line string](https://basarat.gitbooks.io/typescript/content/docs/template-strings.html) instead? it should be much easier to maintain since we won't be  in the business of moving around `\n`'s and `\`'s.

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        }";
+        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("test.cs", program); 
+        await vscode.commands.executeCommand("vscode.open", fileUri);
+
+        let c = await vscode.commands.executeCommand("vscode.executeHoverProvider", fileUri,new vscode.Position(10,30));
+        let answer:string = "Checks if object is tagged with the tag.\n\ngameObject: The game object.\n\ntagName: Name of the tag \n\nReturns: Returns trueif object is tagged with tag.";
+        expect(c[0].contents[0].value).to.equal(answer);
+        
+    });
+
```
DustinCampbell: extra blank line

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
```
DustinCampbell: Missing copyright header

src/features/hoverProvider.ts
```diff
@@ -21,7 +21,8 @@ export default class OmniSharpHoverProvider extends AbstractSupport implements H
 
         return serverUtils.typeLookup(this._server, req, token).then(value => {
             if (value && value.Type) {
-                let contents = [extractSummaryText(value.Documentation), { language: 'csharp', value: value.Type }];
+                let addLine = value.Documentation.split("\n").join("\n\n");
```
rchande: I'm not convinced that this is correct. The bug wants us to have line breaks between sections (summary, parameter, etc), but this seems like it's going to add them anywhere a newline appears in the doc comment. Can you post a screenshot with your changes, with a doc comment with a multline summary section?

test/integrationTests/hoverProvider.integration.test.ts
rchande: +1

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    test("Hover returns the correct XML", async () => {                
+        
+        let program : string = "using System;\
+        \nnamespace hoverXmlDoc \
+        \n{\
+        \n    class testissue\
+        \n    {\
+        \n    ///<summary>Checks if object is tagged with the tag.</summary>\
```
rchande: Put a newline inside this tag.

test/integrationTests/hoverProvider.integration.test.ts
rchande: There are 2 parts to the hover: the type info, and the documentation.

src/features/hoverProvider.ts
akshita31: Here is the output on adding a newline in the summary text. Isn't this what we are expecting?

rchande: I don't think there should be a visible line break between "the" and "tag".

akshita31: An interesting observation here is that even if there are multiple newlines together in the text, vscode displays it as a single newline only. But I am not really sure whether this is ok or we need to do something about it.

rchande: There are multiple "kinds" of newlines at play. 
* There are newlines that have to be in doc comments if a section spans more than one line. I contend that we should not cause a visible line break for those
* There might be multiple consecutive newlines in a doc comment section where the user was trying to separate paragraphs. We should make those visible.
* There are visible line breaks between sections (like between summary and parameters, or between parameters). We should always show those, even if the user wrote all their parameter documentation sections on one line.

akshita31: This is the output from the master branch(where no changes have been made). As noticeable, the  problem is that vscode doesn't show a newline for a single newline.It shows a "single" newline for 2 or more newlines(irrespective of whether they were part of the text originally or added by roslyn to separate the tags). Hence a quick solution was to double the '\n' character before being passed to the Hover object.

If we have to deal with single newlines and double newlines already present in the text separately, we will have to make changes on the roslyn side.Is that what you are suggesting?

rchande: Yes, it looks like we will need to make a change that spans both repos. I propose that we:
* Augment Roslyn to be able to return some data structure that contains all of the "sections"
* Consume that on the VSCode side and properly insert newlines between sections.

test/integrationTests/hoverProvider.integration.test.ts
TheRealPiotrP: @akshita31 if you install the `tslint` extension for vs code you will get notification for this and similar issues.

src/features/hoverProvider.ts
TheRealPiotrP: @rchande can we split this up into two features? @akshita31's change here makes the text much more readable than before, though it does not replicate VS's behavior. I'd start by finishing off this PR, then making it better once we have a chance to close on and implement the relevant change in omnisharp-roslyn. Thoughts?

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        let csharpExtension = vscode.extensions.getExtension("ms-vscode.csharp"); 
+        if (!csharpExtension.isActive) { 
+            await csharpExtension.activate();
+        }
+        
+        await csharpExtension.exports.initializationFinished;
+    });
+
+
+    test("Hover returns the correct XML", async () => {
```
TheRealPiotrP: Can we make this more specific? Perhaps `HoverProvider replaces \n with \n\n`? Hover does not return XML, and this test is validating our newline replacement strategy.

DustinCampbell:
> One remaining comment about test name, and a question to @rchande. LGTM otherwise.

Except that, looking at the new hover tooltip, this change doesn't solve the problem. :smile:

src/features/hoverProvider.ts
rchande: @TheRealPiotrP I pulled @akshita31's branch to play with the behavior. Given the doc comment in the first screenshot, this is what it looks like before and after the change, respectively. I feel that the second example is enough of a regression to the readability of longer documentation comments that we should wait until we have an appropriate change on the roslyn side. @DustinCampbell what do you think?

image
image

DustinCampbell: I agree. Let's not merge changes that regress the user experience.

DustinCampbell:
@akshita31 : Should this be closed in favor of #1918?

akshita31:
Yes .Closing this in favor of #1918

PR #1873, opened 2017-11-17
Add a git-based mechanism to clean up the workspace

rchande:
cc @akshita31


rchande:
Note that there are some node-based native packages that you can consume but there's some complicated configuration to make the electron version of the package and vscode work. VSCode itself just uses the cp library to shell out to git.exe (rather than using libgit2).

rchande:
This change deletes all files not tracked by git and also resets the contents of modified test assets under the workspace being tested.

test/integrationTests/testAssets/testAssets.ts
```diff
@@ -58,6 +59,28 @@ export class TestAssetWorkspace {
         return path.join(this.vsCodeDirectoryPath, "tasks.json");
     }
 
+    async cleanupWorkspace() : Promise<void>
```
TheRealPiotrP: can we make the name more specific? `gitCleanWorkspace`?

PR #1874, opened 2017-11-17
Update devDependencies to latest

TheRealPiotrP:
Update to tslint better enforces linter rules during build. I had to clean up some source content to permit this update.

I don't have sufficient confidence in existing tests to update runtime dependencies. This is still a todo.


tslint.json
```diff
@@ -2,7 +2,6 @@
     "rules": {
         "no-unused-expression": true,
         "no-unused-variable": true,
-        "no-unreachable": true,
```
TheRealPiotrP: Unreachable code is a on-by-default build breaker in tsc, controlled by [--allowUnreachableCode](https://www.typescriptlang.org/docs/handbook/compiler-options.html). We are not specifying a value for this parameter, so it defaults to false.

The tslint rule was deprecated, requiring that I remove it.

src/features/processPicker.ts
```diff
@@ -49,8 +50,8 @@ export class RemoteAttachPicker {
     public static get commColumnTitle() { return Array(PsOutputParser.secondColumnCharacters).join("a"); }
     public static get linuxPsCommand() { return `ps -axww -o pid=,comm=${RemoteAttachPicker.commColumnTitle},args=`; }
     public static get osxPsCommand() { return `ps -axww -o pid=,comm=${RemoteAttachPicker.commColumnTitle},args= -c`; }
-    public static get debuggerCommand() { return "${debuggerCommand}"; };
-    public static get scriptShellCmd() { return "sh -s"; };
+    public static get debuggerCommand() { return "${debuggerCommand}"; }
+    public static get scriptShellCmd() { return "sh -s"; }
```
TheRealPiotrP: TSLint requires we remove these meaningless `;`'s

src/features/codeLensProvider.ts
```diff
@@ -27,14 +29,12 @@ class OmniSharpCodeLens extends vscode.CodeLens {
 
 export default class OmniSharpCodeLensProvider extends AbstractProvider implements vscode.CodeLensProvider {
 
-    private _testManager: TestManager;
```
TheRealPiotrP: Unused field.

src/coreclr-debug/activate.ts
```diff
@@ -15,12 +15,10 @@ import { PlatformInformation } from './../platform';
 import TelemetryReporter from 'vscode-extension-telemetry';
 
 let _debugUtil: CoreClrDebugUtil = null;
-let _reporter: TelemetryReporter = null;
```
TheRealPiotrP: unused field.

package-lock.json
```diff
@@ -0,0 +1,5261 @@
+{
```
DustinCampbell: Is there a reason you're including this? We've not included a package-lock.json file before.

src/features/codeLensProvider.ts
```diff
@@ -5,15 +5,17 @@
...
-import TelemetryReporter from 'vscode-extension-telemetry';
-import TestManager from './dotnetTest';
-import * as vscode from 'vscode';
-import { toRange, toLocation } from '../omnisharp/typeConvertion';
-import AbstractProvider from './abstractProvider';
 import * as protocol from '../omnisharp/protocol';
 import * as serverUtils from '../omnisharp/utils';
+import * as vscode from 'vscode';
+
+import { toLocation, toRange } from '../omnisharp/typeConvertion';
```
rchande: What's the sorting order here? Alphabetical by file path?

package-lock.json
TheRealPiotrP: package-lock.json is [recommended](https://docs.npmjs.com/files/package-lock.json) to be maintained in source control. Benefits:
* Determinism. The same package version will be restored for all users, regardless of updates on npm, until this file is updated.
* Time Travel. Moving to an earlier commit will cause `npm i` to restore packages as they existed at the time of that commit. 

I looked through user complaints, and there are two classes:
* LIBRARIES can get into trouble since they don't have control over their dependencies as resolved in target apps. Folks believe that by sacrificing build-time determinism they increase the likelyhood of finding bugs in dependency packages. We ship an app, so this does not apply. In fact, the determinism increases the likelyhood that what we ship is what we tested.
* Noisy diffs. Time will tell...

src/features/codeLensProvider.ts
TheRealPiotrP: I use an extension called `Sort Typescript Imports`. It's fairly configurable, so we can decide on a sort order that we're happiest with. By default:

`sorts by the type and name of the import. Namespace imports are first, followed by default imports, named imports, and unnamed imports.`

Issue #1525, opened 2017-05-30
Only open files show problems in Problems tab


tverboon:
## Environment data
`dotnet --info` output:
.NET Command Line Tools (2.0.0-preview1-005977)

Product Information:
 Version:            2.0.0-preview1-005977
 Commit SHA-1 hash:  414cab8a0b

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.0.0-preview1-005977/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview1-002111-00
  Build    : 1ff021936263d492539399688f46fd3827169983

VS Code version: 1.12.2

C# Extension version: 1.10.0

## Steps to reproduce
We just migrated from project.json to csproj. With project.json compile errors were found in the entire workspace. With the new MSBuild projects, only problems are detected when the file is open. To reproduce I could just remove a parameter from a method in an interface.

## Expected  behavior
See all errors, warnings and info in my workspace when I change something that introduces problems.

## Actual behavior
No problems are detected, when the changed interface file is the only open file. When I open the implementation class, a compile error will be detected.

DustinCampbell:
Thanks for the report @tverboon.

tverboon:
I was looking in the code and I saw a _isHugeProject() check for 1000 source files, which is determined by Compile Item in MSBuild. So I added this to our common.props

```
<Target Name="Counter">
    <Message Text="Items: '@(Compile->Count())'" />
</Target>
```

Adding all counts in our projects and the total is 1325 files. I guess our project is 'huge' 😄
 #838 would solve this issue. Guess this one can be closed then.

DustinCampbell:
Ah, that makes sense. Yes, this is essentially a duplicate of that.

I think I'll repurpose this to add some sort of easily dismiss-able message for largish projects to let helpful users (like yourself!) know what the heck is going on.

Thanks for the additional information!

tverboon:
I just increased the file limit to 2000 and did some tests with our project, but I am happy with the file limit as it is now. It's just not fast enough on my 4GHz Core i7 iMac. How is this handled in Visual Studio, as I guess it's mostly Roslyn doing the work?

[...]

PR #1877, opened 2017-11-17
Allow to change huge project file limit

### LINKED ISSUE #838: Add Option to Force Full Diagnostics on Large Projects
## Environment data

`dotnet --info` output:
.NET Command Line Tools (1.0.0-preview2-003148)

Product Information:
 Version:            1.0.0-preview2-003148
 Commit SHA-1 hash:  18ca244551

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.11-x64

VS Code version: 1.6.3

C# Extension version: 1.5.0-beta4
## Suggestion

In version 1.4, large projects will not have full diagnostics computed except for opened files. I wonder if it is possible to have a new setting that can force the plugin to run full diagnostics regardless of the size of the project.

jakubfijalkowski:
Hello!

I've recently stumbled upon the huge project file limit. On project with ~1.1k files (~100k LOC) and i7-7700HQ it works reasonably well, so configuring it would be good. This PR adds the `omnisharp.hugeProjectLimit` option.

Fixes #838 and #1525.

### DIFF
--- a/package.json
+++ b/package.json
@@ -566,6 +566,11 @@
           "default": true,
           "description": "Specifies whether the run and debug test CodeLens should be show be shown."
         },
+        "csharp.maxProjectFileCountForDiagnosticAnalysis": {
+          "type": "number",
+          "default": 1000,
+          "description": "Specifies the maximum number of files for which diagnostics are reported for the whole workspace. If this limit is exceeded, diagnostics will be shown for currently opened files only. Specify 0 or less to disable the limit completely."
+        },
         "omnisharp.path": {
           "type": [
             "string",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Added preview Razor (cshtml) language service with support for C# completions and diagnostics in ASP.NET Core projects. Please report issues with the preview Razor tooling on the [aspnet/Razor.VSCode](https://github.com/aspnet/Razor.VSCode/issues) repo. To disable the preview Razor tooling set the "razor.disabled" setting to `true`. (PR: [2554](https://github.com/OmniSharp/omnisharp-vscode/pull/2554))
 * Added omnisharp.minFindSymbolsFilterLength setting to configure the number of characters a user must type in for "Go to Symbol in Workspace" command to return any results (default is 0 to preserve existing behavior). Additionally added omnisharp.maxFindSymbolsItems for configuring maximum number of items returned by "Go to Symbol in Workspace" command. The default is 1000. (PR: [#2487](https://github.com/OmniSharp/omnisharp-vscode/pull/2487))
 * Added a command - "CSharp: Start authoring a new issue on GitHub" to enable the users to file issues on github from within the extension with helpful config information from their system.(PR: [#2503](https://github.com/OmniSharp/omnisharp-vscode/pull/2503))
+* Added a `csharp.maxProjectFileCountForDiagnosticAnalysis` setting to configure the file limit when the extension stops reporting errors for whole workspace. When this threshold is reached, the diagnostics are reported for currently opened files only. This mechanism was available in previous versions, but now can be configured. The default is 1000. (PR: [#1877](https://github.com/OmniSharp/omnisharp-vscode/pull/1877))
 
 #### Debugger
 * Fixed crash at the end of debug sessions on Linux ([#2439](https://github.com/OmniSharp/omnisharp-vscode/issues/2439))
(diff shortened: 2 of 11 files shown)


TheRealPiotrP:
@dustincampbell how do you feel about the configuration name? Did we intend to use `huge` as the descriptor in public surface area?

jakubfijalkowski:
Also, I'm not quite sure if using `null` to disable the feature and ignoring values less than zero (setting it to zero or less will treat every project huge) is good - wouldn't it be better to remove the `null` and using non-positive values instead?

package.json
```diff
@@ -399,6 +399,11 @@
           "type": "boolean",
           "default": true,
           "description": "Specifes whether OmniSharp should use VS Code editor settings for C# code formatting (use of tabs, indentation size)."
+        },
+        "omnisharp.hugeProjectLimit": {
```
DustinCampbell: This probably isn't a good name for a user-facing Option. In addition, because the option controls the extension and not OmniSharp, it should be prefixed with `csharp` rather than `omnisharp`. Maybe something like `csharp.maxProjectFileCountForDiagnosticAnalysis`?

package.json
```diff
@@ -399,6 +399,11 @@
...
+          "type": "number",
+          "default": 1000,
+          "description": "The maximum number of files in the workspace before the project is treated like huge project. For huge projects, OmniSharp reports diagnostics for opened files only. Specify 0 or less to disable the limit completly."
```
DustinCampbell: Typo: 'completely'

jakubfijalkowski:
Sorry for the delay. Yes, it might *huge project* might be too internal (although I got used to the name - it makes me feel like my project really is big :) ) I've updated the name to the one you have proposed.

gregg-miskelly:
@DustinCampbell should this be reviewed again?

jakubfijalkowski:
Hey,

This has bitten me once again, so I decided to review the PR. I've re-applied the changes (after the comments from @DustinCampbell) on the current master. May I ask for re-review?

I've tried to add some test suite, but my TS/vscode-fu is less than optimal and failed miserably. I'm willing to add them, but need some guidance (e.g. how to properly check for diagnostics that, AFAIK, are populated asynchronously).

akshita31:
@jakubfijalkowski Apologies for the delayed response here. We will have to add an integrtaion test for the diagnostics provider. You can look up similar files in the integration test folder to see how they wait on the appropriate response from vscode. You can read more about writing the integration tests here : https://github.com/OmniSharp/omnisharp-vscode/wiki/Writing-tests#integration-tests.

jakubfijalkowski:
@akshita31 Thank you for the links! I'll try to write them after the weekend.

jakubfijalkowski:
I've tried to write the tests, but I don't think it is easily possible with current `Advisor`/`DiagnosticsProvider` design. None of these is published outside of `activate` in `omnisharp/extension.ts` and, as far as I know, vscode does not provide "wait for diagnostics to be generated" command. "Diagnostics providers" are not first-class citizens, but just background processes (or at least they are in omnisharp-vscode) that, in omnisharp-vscode case, do not report completion (`setTimeout` handlers in both `_validateDocument` and `_validateProject` do not raise any events). This makes it terrible to test - we would need to wait for the diagnostics to show for some undefined time (i.e. the tests would be non-deterministic).

My "simple" idea was to just test `Advisor` (as this PR changes just that class) by `export`ing it from `omnisharp/extension.ts` using global variable, but as you can see this fails in the Travis build (and I don't know why yet - I assume this has something to do with node version, as on my machine this works and I have node 11). This isn't pretty, but would have gotten the job done.

I think that the better approach would be to return the `Advisor` and/or `DiagosticsProvider` from `activate` in `omnisharp/extension` the same way `OmniSharpServer` is returned, and then expose both on `CSharpExtensionExports`. This would make the code cleaner and would not require hacky global variables, but is broader in scope.

What do you think, which way would be more appropriate?

akshita31:
@jakubfijalkowski Should we be more specific with the tests and call out that we are testing the "Advisor" and not the "Diagnostic Provider" itself ?

Something like Advisor.ShouldValidateDocument returns true when...

jakubfijalkowski:
@akshita31 You're right. Fixed!

akshita31:
We would need a changelog update for this one. @jakubfijalkowski If you are interested you might do that in this or a separate PR.

jakubfijalkowski:
@akshita31 Done, I've added the entry in this PR.

PR #1883, opened 2017-11-20
Users/piel/release 1.14.1 beta

pieandcakes:
Updates to scripts for 1.14.1 beta.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.14.0-beta1",
+  "version": "1.14.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.13.1",
+  "version": "1.14.1",
```
gregg-miskelly: This should be `1.14.0-beta1`

pieandcakes: @gregg-miskelly Since we uploaded the vsdbg as 1.14.1, will making this 1.14.0-beta2 be a problem?

Issue #1907, opened 2017-12-05
executeWorkspaceSymbolProvider returns SymbolKind.Class instead of SymbolKind.Interface for C# interfaces


craig006:
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (2.0.0)

Product Information:
 Version:            2.0.0
 Commit SHA-1 hash:  cdcd1928c9

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.0.0/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d
```
VS Code version: 1.18.1
C# Extension version: 1.13.0

## Steps to reproduce
Execute 

Run this code from an extension in a C# project in VS Code:
```
commands.executeCommand('vscode.executeWorkspaceSymbolProvider', "ISomeInterface").then((result:SymbolInformation[]) => {
     console.log(result)           
})
```

## Expected  behavior
The SymbolInformation returned for symbol `ISomeInterface` has its kind property set to SymbolKind.Interface

i.e. `result[0].kind should equal SymbolKind.Interface`

## Actual behavior
The SymbolInformation returned for symbol `ISomeInterface` has its kind property set to SymbolKind.Class

i.e. `result[0].kind is set to SymbolKind.Class`

Definition of ISomeInterface
```
public interface ISomeInterface
{
    void SomeMethod();
}
```

rchande:
@craig006 Good find--thanks for filing this.

rchande:
@akshita31 Here's another one for you.

craig006:
@akshita31 @rchande I created a fix here:
https://github.com/craig006/omnisharp-vscode/commit/82fd0cd90881ed3d5450dec4a70829874abbda72#diff-08c3f24b351caf5705429053d550e08a

These are all the vscode.SymbolKinds that I could reproduce. Omnisharp never returned types like "TypeParameter", "Variable" etc.

Let me know if you would like me to create a PR somewhere

craig006:
https://github.com/OmniSharp/omnisharp-vscode/pull/1911

rchande:
Thanks @craig006. We'll take a look.

DustinCampbell:
[...]

PR #1911, opened 2017-12-06
Map all symbol kinds to vscode.SymbolKinds

craig006:
Fixes this issue:
https://github.com/OmniSharp/omnisharp-vscode/issues/1907


rchande:
@akshita31 Before we merge this, do you mind pulling the branch and verifying the behavior?

TheRealPiotrP:
@rchande how do we add a test for this behavior?

rchande:
@TheRealPiotrP Well, if we had per-feature unit tests, that would be a good place to add these tests. However, we're doing approximately the same mapping in `completionItemProvider.ts`, so this might be a good time to extract out a common helper and add tests for it in a new file under unitTests\. @craig006 Any interest in doing that work, too?

akshita31:
@rchande  The type "Delegate" is missing here. It is not present in SymbolKind as well. Should we go ahead and add that ?

rchande:
@akshita31 Sure. What icon does it show up with now? Note that there's no VSCode icon for "Delegate" https://code.visualstudio.com/docs/extensionAPI/vscode-api#SymbolKind

akshita31:
symbolkinds

Only 3 icons are being used to represent all the types.

akshita31:
Currently it is defaulting to a class type, hence we are getting a symbol for the class only.

akshita31:
The change does make the functionality a lot better. @craig006 could you add a case for `Delegate` as well with the return type as `SymbolKind.Method` ?

DustinCampbell:
Assuming @craig006 doesn't have the time at the moment, can we take this PR and make the change @akshita31 suggested separately?

rchande:
Thanks for your contribution @craig006! This has been merged.

PR #1918, opened 2017-12-11
Typelookup changed

akshita31:
Fixes : #1057 

Added Code to allow VS Code to consume the "StructuredDocumentation" object returned by omnisharp-roslyn in typelookup and append newlines appropriately. Also added an integration test for the same.

Please review : @DustinCampbell @TheRealPiotrP @rchande 

Omnisharp-roslyn side : https://github.com/OmniSharp/omnisharp-roslyn/pull/1038


DustinCampbell:
It looks like there's something wrong with your branch in that it's picking up several changes from @rchande

rchande:
@DustinCampbell That was intentional. We needed to make a few tweaks to make integration tests more reliable.

DustinCampbell:
Got it.

src/features/hoverProvider.ts
```diff
@@ -21,9 +21,34 @@ export default class OmniSharpHoverProvider extends AbstractSupport implements H
...
+                                    documentation += "Exceptions:";
+                                }
+                                documentation += newline + structDoc[val].join(newline) + newline;
+                             }
+                         }
+                        else if(structDoc[val]){
+                            documentation += structDoc[val] + newline;
+                        } 
+                    }
+                );
```
DustinCampbell: I'm not sure I'm sold on the property loop.  This will cause the order of the tooltip to change if the properties are ever returned in a different order. Also, if any other element that becomes an array would get the heading "Exceptions:". Should we just check the properties we care about and construct the tooltip text in a deterministic way? E.g.

```TypeScript
if (structDoc.SummaryText) {
    documentation += structDoc.SummaryText;
}

if (structDoc.ParamElements && structDoc.ParamElements.length > 0)) {
    documentation += "Parameters:" + newLine;
    documentation += structDoc.ParamElements.join(newLine) + newLine;
}

etc...
```

DustinCampbell:
@akshita31 : Could include a screenshot of the hover tooltip with your change?

akshita31:
Here is the look after the change.
final_shot

rchande:
I suspect tests are going to fail because we haven't shipped a new Omnisharp into VS Code in a while. @DustinCampbell mind if I do that this week? (Likely tomorrow)

DustinCampbell:
@rchande: We should make this a topic during the OmniSharp meeting tomorrow.

DustinCampbell:
@akshita31 : Do we want to include all of the tags in the hover tooltip? Visual Studio specifically doesn't do that.

image

Also, it doesn't include a "Summary" heading or the "Returns".

DustinCampbell:
If we do want to include the parameters and other items like that, I'd recommend spending some time organizing the tooltip to look a bit better organized, rather than just putting a blank line in between everything. For example:

```
Checks if the object is tagged with the tag.

Parameters:
    gameObject: The game object.
    tagName: Name of the tag.

Returns: Returns true if the object is tagged with tag.
```

Also, note that there appears to be a bug in your screenshot where there is no space between "true" and "if".

akshita31:
I tried indenting the parameters but seems that vscode doesn't display any "\t" that I added.
adding_spaces

However on adding "\t", the text color changes. I am not sure what is happening around here.

rchande:
That looks like a classification bug for @DustinCampbell to discuss.
Regarding the spacing, I think we discovered earlier that the VS Code popup turns all whitespace characters into tabs. You'll probably want to add 4 spaces (or whatever you think looks right) instead of a \t.

akshita31:
Replacing "\t" with 4 spaces also doesn't work.

DustinCampbell:
I wouldn't worry about classifying. However, a Hover doesn't have to be simply a string: https://code.visualstudio.com/docs/extensionAPI/vscode-api#Hover. There's also MarkdownString.

akshita31:
This is how it looks now. To remove the "summary" tag, we will have to make changes to omnisharp-roslyn.
aligned_parameters

akshita31:
Also the missing space that you mentioned is because we are trimming the whitespaces from the string before appending it, but I think that it is a required functionality isn't it ?

DustinCampbell:
@akshita31 : The "Summary" text is extra noise and a regression from previous behavior. Here's what it looks like today:

image

IMO, none of the properties on StructuredDocumentation should include a heading. Parameters, type parameters and exceptions should probably be arrays of objects with "Name" and "Documentation" properties. The text should just be the documentation string. If the host wants to add a heading during display, that's fine, but it seems strange that omnisharp-roslyn would do it. @rchande: What do you think?

rchande:
@DustinCampbell I agree. Even if we wanted to present a heading, that's something that should be up to clients, not forced on them by OmniSharp. @akshita31 Can you help work on a change to Omnisharp that so we don't add "Summary" and so forth to the section text returned from OmniSharp? Sorry we didn't catch that during your code review!

akshita31:
Yes, I'll do that.

rchande:
Also @DustinCampbell, should we wait for that change before we do the release we're planning this week?

DustinCampbell:
It wouldn't take long to make the tweak. I'd just submit the PR to omnisharp-roslyn and we'll take it. After all, we're still waiting on the three Cake PRs.

DustinCampbell:
And if we ship omnisharp-roslyn before the PR goes in, c'est la vie! I'm assuming that we'll be taking another omnisharp-roslyn before shipping omnisharp-vscode.

DustinCampbell:
Oh yeah, that's weird. I wonder if VS Code's markdown grammar gets run over this.

rchande:
@rchande We need to publish omnisharp to the CDN in order to merge this.

src/features/documentation.ts
```diff
@@ -29,3 +30,53 @@ export function extractSummaryText(xmlDocComment: string): string {
...
+        documentation += "Parameters:" + newLine;
+        documentation += indentSpaces + structDoc.ParamElements.map(displayDocumentationObject).join("\n" + indentSpaces) + newLine;
+    }
+
+    if (structDoc.ReturnsText) {
+        documentation += structDoc.ReturnsText + newLine;
+    }
+
+    if (structDoc.RemarksText) {
+        documentation += structDoc.RemarksText + newLine;
```
rchande: Might be worth extracting out a little helper method that adds the newline for you.

src/features/hoverProvider.ts
```diff
@@ -21,9 +21,11 @@ export default class OmniSharpHoverProvider extends AbstractSupport implements H
 
         return serverUtils.typeLookup(this._server, req, token).then(value => {
             if (value && value.Type) {
-                let contents = [extractSummaryText(value.Documentation), { language: 'csharp', value: value.Type }];
+                let structDoc = value.StructuredDocumentation;
```
rchande: structDoc unused?

rchande:
@akshita31 That unused variable causing your Travis build to fail.

PR #1919, opened 2017-12-11
upgrade node on travis

colombod:
upgrade travis node settings to latest 8.x


DustinCampbell:
Hmmm... why isn't GitHub showing a proper diff of your change? Line endings?

colombod:
It seems so, the only real change is the version 8 for node at the top. I need to check the settings on my windows machine.

colombod:
Fixed line endings, now is clear the massive code change :)

Issue #1929, opened 2017-12-19
Remove debugger proxy code


chuckries:
Starting in 1.9, VS Code allows an extension command to be registered that will dynamically produce the path and args to the debugger executable. [Summary](https://code.visualstudio.com/updates/v1_9#_debugger-extension-authoring). 

With this, we should be able to remove our proxy code and do the equivalent in a command. This has some benefits as we will be running in the VS Code process, have access to the VS Code APIs, and we will not need to rewrite the manifest to move from the proxy to vsdbg-ui.

DustinCampbell:
@WardenGnaw : Should this issue be resolved as fixed and closed now?

WardenGnaw:
Yep! Forgot to close this issue.

DustinCampbell:
Is there a changelog update for this issue? Or, does it not affect the experience meaningfully?

chuckries:
This _**should**_ be invisible to the user.

DustinCampbell:
That's what I thought. I just wanted to check.

WardenGnaw:
I believe the only issue would be if someone decides to F5 before the extension is activated and registers the command, VsCode will complain that the command is not found.

DustinCampbell:
sure. That seems low-impact enough to leave out of the changes.

PR #1932, opened 2017-12-21
Fixing Issue #1929: Removing proxy.ts and install.ts

WardenGnaw:
Removing proxy.ts and install.ts

Addressing Issue #1929
With adapterExecutableCommand there is no need to rewrite the manifest.
Instead we register a command to return the correct executable.


gulpfile.js
```diff
@@ -54,17 +53,13 @@ function install(platformInfo, packageJSON) {
...
 
     return packageManager.DownloadPackages(logger)
         .then(() => {
             return packageManager.InstallPackages(logger);
         })
         .then(() => {
             return util.touchInstallFile(util.InstallFileType.Lock)
-        })
-        .then(() => {
-            return debugInstaller.finishInstall();
```
gregg-miskelly: Do we still need to write the install complete file?

gregg-miskelly:
Question: Have you tried doing some per measurements to see how much (if any) slower this is?

src/coreclr-debug/activate.ts
```diff
@@ -113,3 +111,39 @@ function showDotnetToolsWarning(message: string) : void
...
+
+    if (!CoreClrDebugUtil.existsSync(util.installCompleteFilePath())) {
+        // our install.complete file does not exist yet, meaning we have not completed the installation. Try to figure out what if anything the package manager is doing
+        // the order in which files are dealt with is this:
+        // 1. install.Begin is created
+        // 2. install.Lock is created
+        // 3. install.Begin is deleted
+        // 4. install.complete is created
+
+        success = await completeDebuggerInstall(logger, channel);
```
gregg-miskelly: This doesn't look right. Its going to write out the install successful file without actually waiting for any downloads to finish.

src/coreclr-debug/activate.ts
```diff
@@ -113,3 +111,39 @@ function showDotnetToolsWarning(message: string) : void
...
+    if (success)   
+    {
+        // debugger has finished install and manifest has been rewritten, kick off our debugger process
+        return {
+            command: path.join(common.getExtensionPath(), ".debugger", "vsdbg-ui" + CoreClrDebugUtil.getPlatformExeExtension())
+        };
+    }
+    else
+    {
+        return null;
```
gregg-miskelly: What does VS Code do if we return null? Does it do something more sensible if we throw an error?

WardenGnaw: At the moment it will error with the error message if we throw, or if it is invalid it will be ```it does not exist``` or ```cannot read property of 'command' null```.

This will be fixed in https://github.com/Microsoft/vscode/issues/33801

WardenGnaw:
Do I time a fresh install of the extension with all the package downloads or just time an already built package and time f5 to when the debugger shows up?

gulpfile.js
WardenGnaw: Yes. We should still need it.

WardenGnaw:
Old:
1239 ms
1346 ms
1339 ms
1173 ms
1214 ms
Avg: 1262.2 ms

w/ Command:
848 ms
1282 ms
1139 ms
901 ms
821 ms
Avg: 998.2 ms

src/coreclr-debug/activate.ts
```diff
@@ -124,14 +124,20 @@ export async function registerAdapterExecutionCommand(channel: vscode.OutputChan
...
-
+    // our install.complete file does not exist yet, meaning we have not completed the installation. Try to figure out what if anything the package manager is doing
+    // the order in which files are dealt with is this:
+    // 1. install.Begin is created
+    // 2. install.Lock is created
+    // 3. install.Begin is deleted
+    // 4. install.complete is created
+
+    // install.Lock does not exist, need to wait for packages to finish downloading.
+    if (!common.installFileExists(common.InstallFileType.Lock)) {
```
gregg-miskelly: This returns a promise. Don't you need to await it?

src/coreclr-debug/activate.ts
```diff
@@ -124,14 +124,20 @@ export async function registerAdapterExecutionCommand(channel: vscode.OutputChan
     let util = new CoreClrDebugUtil(common.getExtensionPath(), logger);
     let success = true;
 
-    if (!CoreClrDebugUtil.existsSync(util.installCompleteFilePath())) {
```
gregg-miskelly: I think you want to first check if the debugger file exists. This way in the warm case, you don't need to check if two files exist.

src/coreclr-debug/activate.ts
gregg-miskelly: As we discussed in person, I think all these error should be thrown, not logged.

src/features/commands.ts
```diff
@@ -33,11 +34,18 @@ export default function registerCommands(server: OmniSharpServer, reporter: Tele
...
     // Register command for generating tasks.json and launch.json assets.
     let d7 = vscode.commands.registerCommand('dotnet.generateAssets', () => generateAssets(server));
+
+    // Register command for remote process picker for attach
     let d8 = vscode.commands.registerCommand('csharp.listRemoteProcess', (args) => RemoteAttachPicker.ShowAttachEntries(args));
 
-    return vscode.Disposable.from(d1, d2, d3, d4, d5, d6, d7, d8);
+    // Register command for adapter executable command.
+    let d9 = vscode.commands.registerCommand('csharp.coreclrAdapterExecutableCommand', (args) => registerAdapterExecutionCommand(channel).then(adapterExecutableCommand => adapterExecutableCommand));
+    let d10 = vscode.commands.registerCommand('csharp.clrAdapterExecutableCommand', (args) => registerAdapterExecutionCommand(channel).then(adapterExecutableCommand => adapterExecutableCommand));
```
gregg-miskelly: Does the `.then(adapterExecutableCommand => adapterExecutableCommand)` actually do something? It seems like it is equivalent to just setting the command to registerAdapterExecutionCommand.

src/coreclr-debug/activate.ts
```diff
@@ -113,3 +110,46 @@ function showDotnetToolsWarning(message: string) : void
...
 }
+
+interface AdapterExecutableCommand {
+    command: string;
+}
+
+// The default extension manifest calls this command as the adapterExecutableCommand
+// If the debugger components have not finished downloading, the proxy displays an error message to the user
+// Else it will launch the debug adapter
+export async function registerAdapterExecutionCommand(channel: vscode.OutputChannel): Promise<AdapterExecutableCommand> {
```
gregg-miskelly: nit: It seems like a better name for this method is 'getAdapterExecutionCommand` since it doesn't seem like it registers anything.

src/coreclr-debug/activate.ts
```diff
@@ -113,3 +110,46 @@ function showDotnetToolsWarning(message: string) : void
...
+        // the order in which files are dealt with is this:
+        // 1. install.Begin is created
+        // 2. install.Lock is created
+        // 3. install.Begin is deleted
+        // 4. install.complete is created
+
+        // install.Lock does not exist, need to wait for packages to finish downloading.
+        let installLock: boolean = await common.installFileExists(common.InstallFileType.Lock);
+        if (!installLock) {
+            throw new Error('The omnisharp-csharp extension is still downloading packages. Please see progress in the output window below.');
```
gregg-miskelly: Show that output channel?

src/coreclr-debug/activate.ts
```diff
@@ -113,3 +110,46 @@ function showDotnetToolsWarning(message: string) : void
...
+        }
+        // install.complete does not exist, check dotnetCLI to see if we can complete.
+        else if (!CoreClrDebugUtil.existsSync(util.installCompleteFilePath())) {
+            let success: boolean = await completeDebuggerInstall(logger, channel);
+
+            if (!success)
+            {
+                throw new Error('Failed to complete the installation of the omnisharp-csharp extension. Please see the error in the output window below.');
```
gregg-miskelly: Show the output channel?

src/coreclr-debug/activate.ts
gregg-miskelly: Oh, and extension is called 'C#' in the extensions gallery and also that is the name of the output channel, so I would probably suggest using 'C#' instead of 'omnisharp-csharp' in the two error messages.

src/features/commands.ts
WardenGnaw: Nope, I'm just terrible at Typescript.

gregg-miskelly:
LGTM

PR #1938, opened 2017-12-22
Changing the "# References" to display before the set of attributes(if any)

akshita31:
Fixes: #429 

To display the # of references before the set of attributes, the provideCodeLens must return a range for the attributes of the node. So changed the omnisharpLens object to contain two ranges - one the `actualRange` of the node ( to be consumed by the resolveCodeLens method) and the other `attributeRange` that will be consumed to determine the position to display the references.
Also added `attributeSpantoRange` helper to convert span to a range in the document.


akshita31:
Omnisharp-roslyn side : https://github.com/OmniSharp/omnisharp-roslyn/pull/1075

akshita31:
codelens1

The "run test" and "debug test" now shows up as a separate codelens. Do we need to shift that as well alongside the reference codelens.

rchande:
@akshita31 Yes, we need to move the test codelens too.

src/features/codeLensProvider.ts
```diff
@@ -131,12 +133,24 @@ export default class OmniSharpCodeLensProvider extends AbstractProvider implemen
...
     }
+
+    private _attributeSpanToRange(document: vscode.TextDocument, node: protocol.Node): vscode.Range {
+        //If the node has some attributes then returns the range of the attributes, else returns the range from the actual node location
+        if (node.AttributeSpanStart) {
+            let line = document.positionAt(node.AttributeSpanStart).line;
+            let column = document.positionAt(node.AttributeSpanStart).character;
+            let endLine = document.positionAt(node.AttributeSpanEnd).line;
+            let endColumn = document.positionAt(node.AttributeSpanEnd).character;
+            return new Range(line, column, endLine, endColumn);
```
rchande: I think there's a range constructor that takes 2 `Positions`s (instead of 2 l,c coordinates)

src/features/codeLensProvider.ts
```diff
@@ -16,23 +16,25 @@ import { OmniSharpServer } from '../omnisharp/server';
...
 
     fileName: string;
+    actualRange: vscode.Range;
 
-    constructor(fileName: string, range: vscode.Range) {
-        super(range);
+    constructor(fileName: string, actualrange: vscode.Range, attributeRange: vscode.Range) {
+        super(attributeRange);
         this.fileName = fileName;
+        this.actualRange = actualrange;
```
rchande: casing doesn't match that of `attributeRange`

src/features/codeLensProvider.ts
```diff
@@ -53,35 +55,35 @@ export default class OmniSharpCodeLensProvider extends AbstractProvider implemen
...
-        if (!this._options.showReferencesCodeLens && !this._options.showTestsCodeLens)
-        {
+        if (!this._options.showReferencesCodeLens && !this._options.showTestsCodeLens) {
             return [];
         }
 
         return serverUtils.currentFileMembersAsTree(this._server, { FileName: document.fileName }, token).then(tree => {
             let ret: vscode.CodeLens[] = [];
-            tree.TopLevelTypeDefinitions.forEach(node => this._convertQuickFix(ret, document.fileName, node));
+            tree.TopLevelTypeDefinitions.forEach(node => this._convertQuickFix(ret, document, node));
```
rchande: Not your fault, but can we make this method async (and thus not having to do the `then` chaining)?

filipw:
I love this change, but AFAIR the idea was to not diverge the VS Code experience from full VS. Is that not the case anymore?

rchande:
@filipw From talking to other Roslyn folks at lunch, it sounds like people are generally in favor of the change. I'm not aware of any requirement that the VS Code experience exactly match the VS one. Thoughts @DustinCampbell ? I would argue that if this is a better experience (and I think it is) that we should do it.

DustinCampbell:
This requires a new OmniSharp build, correct?

DustinCampbell:
@rchande : The experience for C# in VS vs. VS Code doesn't need to be precisely the same, but that is our starting point for designing the experience. If we decide that the experience should be different, we need to be able to articulate why, and also understand we aren't changing VS to match.

I would not split references from the test code lenses here. That looks pretty bad to me. IMO, all of the code lenses should be above the attributes to match VS unless we've got a reason not to change VS.

rchande:
@DustinCampbell It looks like VS currently puts the codelens adornment underneath attributes. I'll sync with the IDE folks to see what they think and determine if they want to make the same change in VS.

image

DustinCampbell:
@rchande : Yes, that was a conscious decision, and I think we should keep it there. When we implemented CodeLens in VS, I was the one who made the case for keeping the adornment directly above the method declaration and not allowing it to drift above the attributes. Here's an example of why we made that decision from the Roslyn code base:

image

It's an extreme case, but it's a real concern.

DustinCampbell:
cc @JinuJoseph and @kuhlenh for consistency between VS and VS Code.

DustinCampbell:
I'm also curious to know if any thought has been given to XML doc comments, which are defined grammatically to appear above attributes. Does this still look good when the code lens is between XML doc comments and attributes?

DustinCampbell:
Also cc'ing @Pilchie in case he remembers any other reasons why we chose the position of code lens in VS that we did.

Pilchie:
@DustinCampbell's comments match my recollection.

danielsig:
This is marked as "Blocked on OmniSharp".
What's blocking this? I want this!

filipw:
it was "Blocked on OmniSharp" because it required a corresponding change in the Omnisharp language server.
however, it was decided that this change will not be implemented for reasons listed in the discussion above

danielsig:
That's very disappointing to hear ☹️. I've been wanting this to be fixed for ages, but haven't voiced my frustration until now. Can't we just have a configurable attribute line limit with a default value of 0? If the number of lines occupied by attributes is above that limit, # references is displayed as it is now. This would allow the same behavior as it is now by default yet still allow people like me to configure this the way we think is best for our use cases 😃

Issue #1909, opened 2017-12-05
Enable users to download alternative copies of omnisharp


rchande:
We should provide capabilities like the CLI does, where we enable downloading of different feeds of omnisharp.

rchande:
Our new downloading code should also address https://github.com/OmniSharp/omnisharp-vscode/issues/1896

DustinCampbell:
What is the end user scenario?

rchande:
@DustinCampbell This primarily enhances dev scenarios, but I could see this benefiting users, too:

* As a dev (or user), I can start C# extension for VSCode and have it download a version of OmniSharp other than the version it has hard coded.

This will let us easily set up CI builds for testing the extension with newer/older/etc bits of OmniSharp and also improve the local dev scenario for consuming new OmniSharp in the extension. @jonsequitur also thinks we can leverage this in try.net.

DustinCampbell:
It might be worth re-titling this issue to reflect that. :smile:

akshita31:
This has been resolved in https://github.com/OmniSharp/omnisharp-vscode/pull/2065

PR #1972, opened 2018-01-22
Enable usage of multiple versions of omnisharp instead of just the released version

akshita31:
This is the first step in a series of changes to solve the issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1909

Changes : 

1. Added option `alternateVersion` in which a string value can be specified, and if the particular version is present in .omnisharp/experimental folder, the server will be started using that.

2. Added option `useLatestExperimentalBuild` which if set will start the server from the latest version present in the .omnisharp/experiment folder.

Next steps:

1. For the latest option , download a `version.txt` kind of file and get the latest version from that. If the version is not present then download and install the appropriate packages and then start the server.

2. For the release option , if the version is not present download and install it and then start the server.


package.json
```diff
@@ -363,6 +363,11 @@
           "default": null,
           "description": "Specifies the full path to the OmniSharp server."
         },
+        "omnisharp.experimentOmnisharp":{
+          "type":"string",
+          "default": null,
+          "description": "Specifies if the user wants to use the experiment build. Possible values - latest or some version number"
```
rchande: I think there's a way of specifying possible values so they show up in completions. Don't block on this, but something to consider.

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,25 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import * as utils from '../common';
+import * as fs from 'fs';
+
+export class OmnisharpDownloader {
+
+    public GetLatestExperimentVersion(): string {
```
rchande: Since this tells you the most recently installed version, maybe `GetLatestInstalledExperimentalVersion`?

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,25 @@
...
+        let basePath = path.resolve(utils.getExtensionPath(), ".omnisharp/experiment");
+        let compareVersions = require('compare-versions');
+        let latestVersion: string;
+        let items = fs.readdirSync(basePath);
+        if (items) {
+            items.sort(compareVersions);
```
rchande: This could just be a for loop (or whatever JS/TS equivalent there is to Enumerable.Min).

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,25 @@
...
+            latestVersion = items[items.length - 1];
```
rchande: Also, what if there are folders whose names aren't parseable versions?

src/omnisharp/launcher.ts
```diff
@@ -253,8 +251,14 @@ function launch(cwd: string, args: string[]): Promise<LaunchResult> {
...
         }
 
-        // If the user has not provided a path, we'll use the locally-installed OmniSharp
-        const basePath = path.resolve(util.getExtensionPath(), '.omnisharp');
+        let basePath: string;
+        if (experimentVersion) {
+            basePath = path.resolve(util.getExtensionPath(), `.omnisharp/experiment/${experimentVersion}`);
+        }
+        else {
+            // If the user has not provided a path, we'll use the locally-installed OmniSharp
```
rchande: This comment needs updated.

src/omnisharp/options.ts
```diff
@@ -18,7 +18,8 @@ export class Options {
         public useFormatting?: boolean,
         public showReferencesCodeLens?: boolean,
         public showTestsCodeLens?: boolean,
-        public disableCodeActions?: boolean) { }
+        public disableCodeActions?: boolean,
+        public experimentOmnisharp?: string) { }
```
rchande: Can we replace all instances of `experiment` with `experimental`? :)

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,25 @@
...
+            //get the latest version after sorting
+        }
+
+        return latestVersion;
```
rchande: This looks like something we could unit test :).

src/omnisharp/OmnisharpDownloader.ts
colombod: probably better to do a loop and use package like semver to address also the non-semantic version string (this will return a null object) and also handling pre-release versions

src/omnisharp/launcher.ts
```diff
@@ -253,8 +251,14 @@ function launch(cwd: string, args: string[]): Promise<LaunchResult> {
...
 
-        // If the user has not provided a path, we'll use the locally-installed OmniSharp
-        const basePath = path.resolve(util.getExtensionPath(), '.omnisharp');
+        let basePath: string;
+        if (experimentalVersion) {
+            basePath = path.resolve(util.getExtensionPath(), `.omnisharp/experimental/${experimentalVersion}`);
+        }
+        else {
+            // If the user has not provided a path and some experimental version, we'll use the locally-installed OmniSharp
+            basePath = path.resolve(util.getExtensionPath(), '.omnisharp');
```
rchande: Why does this computation happen in server.ts and this file?

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,32 @@
...
+    public GetLatestInstalledExperimentalVersion(basePath: string) {
+        const semver = require('semver');
+
+        let latestVersion: string;
+        if (fs.existsSync(basePath)) {
+            let installedItems = fs.readdirSync(basePath);
+            if (installedItems && installedItems.length > 0) {
+                let validVersions = installedItems.filter(value => semver.valid(value));
+                if (validVersions && validVersions.length > 0) {
+                    latestVersion = validVersions.reduce((latestTillNow, element) => {
```
rchande: Nice!

package.json
```diff
@@ -363,6 +363,11 @@
           "default": null,
           "description": "Specifies the full path to the OmniSharp server."
         },
+        "omnisharp.experimentalOmnisharp": {
+          "type": "string",
+          "default": null,
+          "description": "Specifies if the user wants to use the experiment build. Possible values - latest or some version number"
+        },
```
DustinCampbell: Please provide "enum" values to help with JSON completion and schema validation. See "omnisharp.loggingLevel" below for an example.

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,32 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fs from 'fs';
+
+export class OmnisharpDownloader {
+
+    public GetLatestInstalledExperimentalVersion(basePath: string) {
+        const semver = require('semver');
```
DustinCampbell: Why not just `import * as semver from 'semver'` at the top of the module?

package.json
DustinCampbell: Also, please clean up the grammar in the description. This is a user-facing string.

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,32 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fs from 'fs';
+
+export class OmnisharpDownloader {
```
DustinCampbell: I'm confused by this name since this class doesn't actually download anything at all.

src/omnisharp/server.ts
```diff
@@ -259,6 +260,23 @@ export class OmniSharpServer {
             args.push('--debug');
         }
 
+        let experimentalVersion : string;
+        let experimentalOption = this._options.experimentalOmnisharp;
+        let extensionPath = utils.getExtensionPath();
+        
+        if(experimentalOption == "latest"){
+            let downloader = new OmnisharpDownloader();
```
DustinCampbell: Is this expected to actually download something from the Internet somehow? If so, should this be plugged into the package downloader that already exists in omnisharp-vscode?

package.json
DustinCampbell: Also, **omnisharp**.experimental**OmniSharp** is a bit redundant. It's also not clear to me why this is "experimental". Maybe `omnisharp.alternateVersion` or something like that? @rchande?

src/omnisharp/latestOptionHelper.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
```
colombod: would not be better to call this file VersionSelectors.ts ?

TheRealPiotrP:
@DustinCampbell I share your concern about duplication. We took a good hard look at the existing downloader and in fact started by modifying it. However, that code is very lightly tested and the current structure doesn't lend itself to testability. We intend to:
- merge this feature
- temporarily fork the downloader, refactoring it to be testable and tested
- once we're happy with the downloader, extend it to provide for the O# scenario as well

Let's chat about this next week?

DustinCampbell:
@TheRealPiotrP : Sure, although this PR doesn't actually do any downloading. :smile:

akshita31:
The approach to the feature changed a bit https://github.com/OmniSharp/omnisharp-vscode/pull/2028 and  https://github.com/OmniSharp/omnisharp-vscode/pull/2039, hence the functionality added here is no more needed. Closing this for now

PR #1995, opened 2018-02-01
Update to omnisharp 1.29

rchande:
cc @akshita31 @TheRealPiotrP


rchande:
@akshita31 It looks like we changed the hover behavior to only show 'Returns trueif object is tagged with tag" in the hover integration test. I'm surprised that we don't show the summary part created in the test 'Checks if object is tagged with the tag.". Thoughts?

akshita31:
The test was written before https://github.com/OmniSharp/omnisharp-roslyn/pull/1064 was merged, hence it is failing now.

PR #2028, opened 2018-02-09
Enable usage of omnisharp.path option for downloading multiple versions of omnisharp

akshita31:
This is a part of the change to add the feature : https://github.com/OmniSharp/omnisharp-vscode/issues/1909

Brief Outline:

1.  The server checks if the `omnisharp.path` is set and if not proceeds with the normal default launch
2. If some path has been specified, the `GetExperimentalOmnisharpPath` is called which returns the appropriate path that will be used by the launcher.

`GetExperimentOmnisharpPath` : First checks whether the path is an existing path on disk, if so the same path is returned. Else we try to treat this path as a version, then invoke `InstallVersionAndReturnLaunchPath` which installs the packages and returns the path to be used.

If at any point in the install and download process ,an error is encountered like 'Invalid Version' or 'Download failure', the error is returned upto the server and the launcher is not called.
omnisharpdownload

Also attached is a representation of the feature. Note that the "latest" download is not a part of this PR.


src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+import * as vscode from 'vscode';
+import { PackageManager, Package } from '../packages';
+import { PlatformInformation } from '../platform';
+import { Logger } from '../logger';
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { GetPackagesFromVersion } from './experimentalOmnisharp.PackageCreator';
+import { GetDependenciesAndDownloadPackages, GetStatus, GetAndLogPlatformInformation, ReportInstallationError, SendInstallationTelemetry } from '../experimentalOmnisharp.Download.Helper';
+
+export class ExperimentalOmnisharpDownloader {
+
```
rchande: Extra blank line

src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+    public constructor(
+        private channel: vscode.OutputChannel,
+        private logger: Logger,
+        private reporter: TelemetryReporter /* optional */,
```
rchande: use optional parameter syntax?

src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+        private packageJSON: any) {
```
rchande: We don't have a more specific type for this?

src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public async DownloadAndInstallExperimentalVersion(version: string, serverUrl: string, installPath: string) {
+        if (!version) {
+            throw new Error('Invalid version');
+        }
+
+        this.logger.append('Getting the version packages...');
```
rchande: I don't know what "getting the version packages" means.

src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.logger.appendLine();
+        this.channel.show();
+
+        let statusItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right);
```
rchande: The pattern here is: "call raw VSCode APIs to add status bar item, then call a helper to wrap the VSCode status item into something easy to manipulate". Can we push the VSCode API call into the "GetStatus" helper (and call it something like "SetStatus")?

src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+        let status = GetStatus(statusItem);
+
+        let telemetryProps: any = {};
+        let installationStage = '';
```
rchande: Should installationStage be an enum?

src/omnisharp/experimentalOmnisharp.PackageCreator.ts
```diff
@@ -0,0 +1,58 @@
...
+    for (let inputPackage of runTimeDependencies) {
+        if (inputPackage.experimentalPackageId) {
+            versionPackages.push(GetExperimentPackage(inputPackage, serverUrl, version, installPath));
+        }
+    }
+
+    return versionPackages;
+}
+
+export function GetExperimentPackage(inputPackage: Package, serverUrl: string, version: string, installPath: string): Package {
```
rchande: Experimental

src/omnisharp/experimentalOmnisharp.PackageCreator.ts
```diff
@@ -0,0 +1,58 @@
...
+    let installBinary: string;
+    if (inputPackage.experimentalPackageId == "win-x86" || inputPackage.experimentalPackageId == "win-x64") {
```
rchande: You're saying we should update the package.json schema for each package to list its architecture(?) as `experimentalPackageId`?

src/omnisharp/experimentalOmnisharp.PackageCreator.ts
```diff
@@ -0,0 +1,58 @@
...
+        installBinary = "Omnisharp.exe";
```
rchande: Casing?

src/experimentalOmnisharp.Download.Helper.ts
```diff
@@ -0,0 +1,82 @@
...
+import { PackageManager, Status, PackageError, Package } from './packages';
+import { PlatformInformation } from './platform';
+import { Logger } from './logger';
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+export async function GetDependenciesAndDownloadPackages(packages: Package[],status: Status, platformInfo: PlatformInformation, packageManager: PackageManager,logger: Logger) { 
+    const config = vscode.workspace.getConfiguration();
+    const proxy = config.get<string>('http.proxy');
+    const strictSSL = config.get('http.proxyStrictSSL', true);
+    await packageManager.DownloadPackages(logger, status, proxy, strictSSL);
```
rchande: How do other callers do this setup before calling download packages? Should proxy configuration just be a feature of DownloadPackages?

src/omnisharp/experimentalOmnisharp.Manager.ts
```diff
@@ -0,0 +1,73 @@
...
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { PlatformInformation } from '../platform';
+
+export class ExperimentalOmnisharpManager {
+    public constructor(
+        private channel: vscode.OutputChannel,
+        private logger: Logger,
+        private reporter: TelemetryReporter,
+        private packageJSON: any
+    ) {
```
rchande: Weird spacing

src/omnisharp/experimentalOmnisharp.Manager.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public async GetExperimentalOmnisharpPath(optionPath: string, useMono: boolean, platformInfo: PlatformInformation, serverUrl: string, installPath: string, extensionPath: string): Promise<string> {
```
rchande: omnisharpPath?

src/omnisharp/experimentalOmnisharp.Manager.ts
```diff
@@ -0,0 +1,73 @@
...
+        // Looks at the options path, installs the dependencies and returns the path to be loaded by the omnisharp server
+        // To Do : Add the functionality for the latest option
+
+        if (await util.fileExists(optionPath)) {
+            return optionPath;
+        }
+        //If the path is not a valid path on disk, treat it as a version
```
rchande: Newlines

rchande: I would make this a little more specific: if what they specified happens to be an absolute path that doesn't exist, print errors about a missing file instead of trying to download it. Otherwise, typos in the file path are going to lead to confusing errors about file downloads.

src/omnisharp/experimentalOmnisharp.Manager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    public async InstallVersionAndReturnLaunchPath(version: string, useMono: boolean, serverUrl: string, installPath: string, extensionPath: string, platformInfo: PlatformInformation) {
+        if (IsValidSemver(version)) {
+            let downloader = new ExperimentalOmnisharpDownloader(this.channel, this.logger, this.reporter, this.packageJSON);
+            await downloader.DownloadAndInstallExperimentalVersion(version, serverUrl, installPath);
+
+            return await GetLaunchPathForVersion(platformInfo, version, installPath, extensionPath, useMono);
+        }
+        else {
+            throw new Error('Bad Input to Omnisharp Path');
```
rchande: You know what the problem is, why don't you just print that the Omnisharp.Path isn't a "valid absolute path or semantic version" or something.

src/omnisharp/experimentalOmnisharp.Manager.ts
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+}
+
+function IsValidSemver(version: string): boolean {
```
rchande: Do we really need this method?

test/unitTests/experimentalOmnisharpDownloader.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        tmpDir = tmp.dirSync();
+        util.setExtensionPath(tmpDir.name);
+    });
+    
+    test('Packages are downloaded from the specified server url and installed at the specified path', async () => {
+        /* Download a test package that conatins a install_check_1.2.3.txt file and check whether the 
+           file appears at the expected path */
+        let version = "1.2.3";
+        let downloader = GetOmnisharpDownloader();
+        let serverUrl = "https://roslynomnisharp.blob.core.windows.net";
```
rchande: How do we feel about unit tests that depend on the internet @TheRealPiotrP ?

test/unitTests/experimentalOmnisharpDownloader.test.ts
```diff
@@ -0,0 +1,172 @@
...
+function GetOmnisharpDownloader() {
+    let channel = vscode.window.createOutputChannel('Experiment Channel');
+    let logger = new Logger(text => channel.append(text));
+    return new ExperimentalOmnisharpDownloader(channel, logger, null, GetTestPackageJSON());
+}
+
+//Since we need only the runtime dependencies of packageJSON for the downloader create a testPackageJSON
+//with just that
+export function GetTestPackageJSON() {
+let testpackageJSON = {
```
rchande: I bet you could make this read in the package.json that's in the repo...

package.json
```diff
@@ -361,7 +367,7 @@
             "null"
           ],
           "default": null,
-          "description": "Specifies the full path to the OmniSharp server."
+          "description": "Specifies pre-release OmniSharp to use. Can be one of \"latest\", a specific version number, or the absolute path to a local OmniSharp folder."
```
DustinCampbell: FWIW, I suspect that "pre-release" is not the right text since this can be set to any OmniSharp -- even those that are released. :smile:

src/experimentalOmnisharp.Download.Helper.ts
```diff
@@ -0,0 +1,82 @@
...
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import * as vscode from 'vscode';
+import { PackageManager, Status, PackageError, Package } from './packages';
+import { PlatformInformation } from './platform';
+import { Logger } from './logger';
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+export async function GetDependenciesAndDownloadPackages(packages: Package[],status: Status, platformInfo: PlatformInformation, packageManager: PackageManager,logger: Logger) {
```
DustinCampbell: nit: missing space between `,` and `status`

src/omnisharp/experimentalOmnisharp.Downloader.ts
```diff
@@ -0,0 +1,72 @@
...
+export class ExperimentalOmnisharpDownloader {
+
+    public constructor(
+        private channel: vscode.OutputChannel,
+        private logger: Logger,
+        private reporter: TelemetryReporter /* optional */,
+        private packageJSON: any) {
+    }
+
+    public async DownloadAndInstallExperimentalVersion(version: string, serverUrl: string, installPath: string) {
```
DustinCampbell: What is the plan to unify this with the code in https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/CSharpExtDownloader.ts where this is copied and pasted from?

src/omnisharp/launcher.ts
```diff
@@ -207,9 +207,9 @@ export interface LaunchResult {
     usingMono: boolean;
 }
 
-export function launchOmniSharp(cwd: string, args: string[]): Promise<LaunchResult> {
+export function launchOmniSharp(cwd: string, args: string[], experimentalLaunchPath: string): Promise<LaunchResult> {
```
DustinCampbell: Should this just be called "launchPath"?

src/omnisharp/experimentalOmnisharp.PackageCreator.ts
akshita31: Yes that way we dont have to parse the url or any other parameter to get the os and architecture to form the new download urls

src/experimentalOmnisharp.Download.Helper.ts
akshita31: There is only one caller - the CSharpExtDownloader other than the OmnisharpDownloader and I have configured it the same way. There might be other ways to set proxy and strictSSL so moving this code to the DownloadPackages may not be a good idea. Thoughts ?

test/unitTests/experimentalOmnisharpDownloader.test.ts
akshita31: We tried using the mock server so we dont have to depend on the internet, but seems that the code creates an HTTPS request and we could not get the mock server running on the 443 port. Hence we have uploaded some dummy folders on the real server (like https://roslynomnisharp.blob.core.windows.net/releases/1.2.3/omnisharp-win-x64.zip) just for the testing purpose

src/omnisharp/experimentalOmnisharp.Downloader.ts
akshita31: Once this feature has been added and tested satisfactorily, we plan to unify the two in a separate PR

package.json
```diff
@@ -367,7 +367,7 @@
             "null"
           ],
           "default": null,
-          "description": "Specifies pre-release OmniSharp to use. Can be one of \"latest\", a specific version number, or the absolute path to a local OmniSharp folder."
+          "description": "Specifies the OmniSharp to use. Can be one of \"latest\", a specific version number, or the absolute path to a local OmniSharp folder."
```
rchande: "Specifies how to acquire OmniSharp?" "the OmniSharp" is a weird phrasing

package.json
```diff
@@ -367,7 +367,7 @@
             "null"
           ],
           "default": null,
-          "description": "Specifies the OmniSharp to use. Can be one of \"latest\", a specific version number, or the absolute path to a local OmniSharp folder."
+          "description": "Specifies how to acquire the OmniSharp to use. Can be one of \"latest\", a specific version number, or the absolute path to a local OmniSharp folder."
```
rchande: Still seems awkward...I think the problem is the phrase "the Omnisharp". Maybe "The copy of Omnisharp"? @DustinCampbell ?

DustinCampbell: I think it might be easier to be plain and then add a bit more explaining text at the end. Also, I believe that "the absolute path to a local OmniSharp folder" is actually incorrect. IIRC, the path needs to point to the file that will be launched.

Here's a stab a some new text. Additional word-smithing is welcome!

"Specifies the path to OmniSharp. This can be the absolute path to an OmniSharp executable, a specific version number, or \"latest\". If a version number or \"latest\" is specified, the appropriate version of OmniSharp will be downloaded on your behalf."

akshita31: Do we need to specify that "latest" implies the latest CI build, because just the word "latest" might also imply the latest released version to the user ?

PR #2039, opened 2018-02-15
Enable download and usage of latest version of omnisharp

akshita31:
Issue : https://github.com/OmniSharp/omnisharp-vscode/issues/1909

This is in continuation to the PR : https://github.com/OmniSharp/omnisharp-vscode/pull/2028

Added the capability to donwload the "latest" version, by first downloading and reading a versionInfo file that contains the latest version information. On getting this version, we download the packages and the server is started from the returned launch path, else appropriate errors are thrown and the server is not started.
Tests have also been added for the same.


src/omnisharp/OmnisharpManager.ts
```diff
@@ -32,24 +31,31 @@ export class OmnisharpManager {
...
+            return await this.LatestInstallAndReturnLaunchPath(downloader, useMono, serverUrl, versionFilePathInServer, installPath, extensionPath, platformInfo);
+        }
+        //If the path is neither a valid path on disk not the string "latest", treat it as a version 
+        return await this.InstallVersionAndReturnLaunchPath(downloader, omnisharpPath, useMono, serverUrl, installPath, extensionPath, platformInfo);
     }
 
-    public async InstallVersionAndReturnLaunchPath(version: string, useMono: boolean, serverUrl: string, installPath: string, extensionPath: string, platformInfo: PlatformInformation) {
+    public async LatestInstallAndReturnLaunchPath(downloader: OmnisharpDownloader, useMono: boolean, serverUrl: string, versionFilePathInServer: string, installPath: string, extensionPath: string, platformInfo: PlatformInformation) {
+        let version = await downloader.GetLatestVersion(serverUrl, versionFilePathInServer);
+        await downloader.DownloadAndInstallOmnisharp(version, serverUrl, installPath);
```
rchande: I would expect this to call InstallVersionAndLaunchPath with the version you determined through downloading.

src/omnisharp/OmnisharpManager.ts
```diff
@@ -32,24 +31,31 @@ export class OmnisharpManager {
                 throw new Error('Invalid path specified');
             }
         }
-        //If the path is not a valid path on disk, treat it as a version 
-        return await this.InstallVersionAndReturnLaunchPath(omnisharpPath, useMono, serverUrl, installPath, extensionPath, platformInfo);
+        else if (omnisharpPath == "latest") {
+            return await this.LatestInstallAndReturnLaunchPath(downloader, useMono, serverUrl, versionFilePathInServer, installPath, extensionPath, platformInfo);
+        }
+        //If the path is neither a valid path on disk not the string "latest", treat it as a version
```
rchande: Newline after }

src/omnisharp/OmnisharpManager.ts
```diff
@@ -32,24 +31,31 @@ export class OmnisharpManager {
                 throw new Error('Invalid path specified');
```
rchande: Is this really a condition where we need to throw an exception? We know exactly what went wrong...

src/omnisharp/OmnisharpPackageCreator.ts
```diff
@@ -55,4 +55,13 @@ function GetPackageFromArchitecture(inputPackage: Package, serverUrl: string, ve
     };
 
     return versionPackage;
-}
\ No newline at end of file
+}
+
+export function GetVersionFilePackage(serverUrl: string, pathInServer: string): Package {
```
rchande: This is confusing. Passing around `packages` to represent the actual packages we download, sure. Creating a `package` to represent downloading a text file...strange. Can we at least add a comment explaining why?

src/omnisharp/OmnisharpPackageCreator.ts
```diff
@@ -55,4 +55,13 @@ function GetPackageFromArchitecture(inputPackage: Package, serverUrl: string, ve
...
+    let filePackage = <Package>{
```
rchande: you can just `return <Package> ...`

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -4,69 +4,96 @@
...
             throw error;// throw the error up to the server
         }
         finally {
-            SendInstallationTelemetry(this.logger, this.reporter, telemetryProps, installationStage, platformInfo, statusItem);
+            SendInstallationTelemetry(this.logger, this.reporter, this.telemetryProps, installationStage, this.platformInfo, this.statusItem);
+        }
+    }
+
+    public async GetLatestVersion(serverUrl: string, versionFilePathInServer): Promise<string> {
+        let installationStage = 'getLatestVersionInfoFile';
```
rchande: In our current pattern, anyone who wants to add an installation stage and print errors on failure has to set up an identical try/catch/finally block. I wonder if we could add a `RunInstllationStage` helper that takes an async func that does the work, the stage name, and the logger and does all that for you.

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -4,69 +4,96 @@
...
+
+        let statusObject = SetStatus();
+        this.status = statusObject.Status;
+        this.statusItem = statusObject.StatusItem;
+
+        let networkObject = GetNetworkDependencies();
+        this.proxy = networkObject.Proxy;
+        this.strictSSL = networkObject.StrictSSL;
+
+        this.telemetryProps = {};
```
rchande: What is this empty object for?

akshita31: https://github.com/OmniSharp/omnisharp-vscode/blob/experiment_download/src/OmnisharpDownload.Helper.ts#L47

This is used for the logging. This initialisation is the same as in the CSharpExtDownloader - https://github.com/OmniSharp/omnisharp-vscode/blob/experiment_download/src/CSharpExtDownloader.ts#L52

src/omnisharp/OmnisharpDownloader.ts
TheRealPiotrP: Do you see this as a regression in the source? There is a lot of refactoring to be done in this area of the code and I'd like to limit the changes in this PR lest it grows too big...

src/omnisharp/OmnisharpPackageCreator.ts
TheRealPiotrP: let's add the comment for now, then refactor more in a future PR? all of the great work that @DustinCampbell did around proxy negotiation, etc. is currently best accessed via the Package construct.

test/unitTests/OmnisharpPackageCreator.test.ts
```diff
@@ -194,4 +194,16 @@ suite('GetPackagesFromVersion : Gets the experimental omnisharp packages from a
...
+    test('Contains the expected description', () => {
+        let testPackage = GetVersionFilePackage("someUrl", "somePath");
+        expect(testPackage.description).to.equal('Latest version information file');
+    }); 
+
+    test('Contains the url based on serverUrl and the pathInServer', () => {
+        let testPackage = GetVersionFilePackage("someUrl", "somePath");
+        expect(testPackage.url).to.equal('someUrl/somePath');
+    }); 
 });
```
TheRealPiotrP: Are we able to provide an end-to-end integration test at this stage? something that will read the 'latest' property value, download a test version file, and based on that downloads a test O#? This can be a subsequent PR, but I'd like to get the full feature behind a test once we're ready, if it's reasonable...

src/omnisharp/OmnisharpPackageCreator.ts
rchande: Works for me.

src/omnisharp/OmnisharpDownloader.ts
rchande: Technically it's not a pattern this PR introduced, but this PR does create more instances of it. @akshita31 I'd be OK with getting this refactoring in a followup PR.

rchande:
@akshita31 I test this out and it seems pretty good, except there's one issue. It appears that if you use "omnisharp.path" to set a file path, that that value gets cached. 
If this isn't a new issue (ie, your PR didn't cause this), then I think we're good to go. Otherwise, we should see if it would be easy to fix.

1. Set omnisharp.path to an invalid path-like value, restart VS Code.
2. OMnisharp Log will have an error about invalid path
3. Set omnisharp.path to a valid path value, use "restart omnisharp" command

Expected:
omnisharp starts
Actual:
message printed about invalid path

This goes away if you restart VS Code.

akshita31:
@rchande Which platform did you find this for ? I am not getting a repro on my windows machine.

rchande:
@akshita31 I did this on my mac. I realize now that you don't have one, so I'll go head and try this out.

PR #1958, opened 2018-01-12
Using Structured Documentation for Signature Help and improving Parameter Documentation

akshita31:
Fixes : #1940 

Using the structured documentation to display parsed xml in signature help. Also displaying the parameter documentation properly.

 Omnisharp-roslyn side : https://github.com/OmniSharp/omnisharp-roslyn/pull/1085

PR #2050, opened 2018-02-17
Make hover integration test pass

akshita31:
The hover integration test is failing because of a possible race condition between the omnisharp process and the vscode process to access the file that has been added. Hence, if we restart the server before we ask vscode to open it, omnisharp will have read the file and no problem should occur.

PS: I am not sure whether this is the best approach to solve this, because we will have to restart the server whenever we add a file for testing.

### DIFF
--- a/test/integrationTests/hoverProvider.integration.test.ts
+++ b/test/integrationTests/hoverProvider.integration.test.ts
@@ -48,15 +48,11 @@ namespace Test
        }
    }
 }`;
-        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("test1.cs", program);
-
+        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("hover.cs", program);
         await omnisharp.restart();
         await omnisharp.waitForEmptyEventQueue();
-
-        await vscode.workspace.openTextDocument(fileUri);
-
+        await vscode.commands.executeCommand("vscode.open", fileUri);
         let c = await vscode.commands.executeCommand("vscode.executeHoverProvider", fileUri, new vscode.Position(10, 29));
-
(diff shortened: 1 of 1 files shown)


rchande:
@akshita31 The solution we discussed was to just always include the "hover" file in the test solution. However, if this works, I'm happy to take it :). I restarted your PR build to see if it's flaky.

test/integrationTests/hoverProvider.integration.test.ts
```diff
@@ -48,15 +48,11 @@ namespace Test
...
 }`;
-        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("test1.cs", program);
-
+        let fileUri = await testAssetWorkspace.projects[0].addFileWithContents("hover.cs", program);
         await omnisharp.restart();
         await omnisharp.waitForEmptyEventQueue();
-
-        await vscode.workspace.openTextDocument(fileUri);
-
+        await vscode.commands.executeCommand("vscode.open", fileUri);
```
rchande: Why did you move away from the openTextDocument helper that you switched to?

akshita31: No specific reason. I just thought it would maintain consistency with the codeActionRename test. Is openTextDocument a better way ?

akshita31:
@rchande I was trying to add a test for signature help for this PR : https://github.com/OmniSharp/omnisharp-vscode/pull/1958 , following the same approach as above, but it doesn't seem to work :(

akshita31:
This approach doesn't work. Closing this in favor of https://github.com/OmniSharp/omnisharp-vscode/pull/2058

PR #2065, opened 2018-02-23
Enable multiple download of omnisharp 

akshita31:
Final PR including all the changes.


src/OmnisharpDownload.Helper.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    if (reporter) {
+        reporter.sendTelemetryEvent('Acquisition', telemetryProps);
+    }
+
+    logger.appendLine();
+    installationStage = '';
+    logger.appendLine('Finished');
+
+    statusItem.dispose();
```
rchande: Seems weird that this method takes statusItem just to call dispose on it.

src/OmnisharpDownload.Helper.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function SendInstallationTelemetry(logger: Logger, reporter: TelemetryReporter, telemetryProps: any, installationStage: string, platformInfo: PlatformInformation, statusItem: vscode.StatusBarItem) {
+    telemetryProps['installStage'] = installationStage;
+    telemetryProps['platform.architecture'] = platformInfo.architecture;
+    telemetryProps['platform.platform'] = platformInfo.platform;
+    if (platformInfo.distribution) {
+        telemetryProps['platform.distribution'] = platformInfo.distribution.toTelemetryString();
+    }
+    if (reporter) {
```
rchande: Is reporter ever null?

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.proxy = networkObject.Proxy;
+        this.strictSSL = networkObject.StrictSSL;
+
+        this.telemetryProps = {};
+        this.packageManager = new PackageManager(this.platformInfo, this.packageJSON);
+    }
+
+    public async DownloadAndInstallOmnisharp(version: string, serverUrl: string, installPath: string) {
+        if (!version) {
+            throw new Error('Invalid version');
```
rchande: If this happens, this is basically a bug in our code, right? Since this method is not expecting a null version. "invalid version" makes it sound like a user mistake, which it isn't.

src/omnisharp/OmnisharpManager.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (path.isAbsolute(omnisharpPath)) {
+            if (await util.fileExists(omnisharpPath)) {
+                return omnisharpPath;
+            }
+            else {
+                throw new Error('The system could not find the specified path');
+            }
+        }
+        else if (omnisharpPath == "latest") {
+            return await this.LatestInstallAndReturnLaunchPath(downloader, useMono, serverUrl, versionFilePathInServer, installPath, extensionPath, platformInfo);
```
rchande: call this method `InstallLatestAndReturnLaunchPath`

src/omnisharp/OmnisharpManager.ts
```diff
@@ -0,0 +1,74 @@
...
+        return await this.InstallVersionAndReturnLaunchPath(downloader, version, useMono, serverUrl, installPath, extensionPath, platformInfo);
+    }
+
+    public async InstallVersionAndReturnLaunchPath(downloader: OmnisharpDownloader, version: string, useMono: boolean, serverUrl: string, installPath: string, extensionPath: string, platformInfo: PlatformInformation) {
+        if (semver.valid(version)) {
+            await downloader.DownloadAndInstallOmnisharp(version, serverUrl, installPath);
+            return GetLaunchPathForVersion(platformInfo, version, installPath, extensionPath, useMono);
+        }
+        else {
+            throw new Error('Invalid omnisharp version specified');
```
rchande: Can we include the specified version in this message?

src/omnisharp/OmnisharpManager.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function GetLaunchPathForVersion(platformInfo: PlatformInformation, version: string, installPath: string, extensionPath: string, useMono: boolean) {
+    if (!version) {
+        throw new Error('Invalid Version');
+    }
+
+    let basePath = path.resolve(extensionPath, installPath, version);
+
+    if (platformInfo.isWindows()) {
```
rchande: `platformInfo.isWindows() || useMono` ?

src/omnisharp/OmnisharpPackageCreator.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    return GetPackageFromArchitecture(inputPackage, serverUrl, version, inputPackage.experimentalPackageId, installPath, installBinary);
+}
+
+function GetPackageFromArchitecture(inputPackage: Package, serverUrl: string, version: string, architectureInfo: string, installPath: string, installBinary: string): Package {
+    if (!version) {
+        throw new Error('Invalid version');
+    }
+
+    let versionPackage = <Package>{
```
rchande: Can you use the `...` operator and just change the fields you care about?
`{...package, url = zzz, experimentalPackageId = xxx`
Means the same thing, but might be easier to tell what fields you're changing.

package.json
```diff
@@ -39,6 +39,7 @@
     "postinstall": "node ./node_modules/vscode/bin/install"
   },
   "dependencies": {
+    "chai-as-promised": "^7.1.1",
```
colombod: should this be a dev dependency only?

package.json
```diff
@@ -93,7 +94,8 @@
       "architectures": [
         "x86"
       ],
-      "installTestPath": "./.omnisharp/OmniSharp.exe"
+      "installTestPath": "./.omnisharp/OmniSharp.exe",
+      "experimentalPackageId": "win-x86"
```
TheRealPiotrP: let's rename `experimentalPackageId` to `platformId`

src/omnisharp/server.ts
```diff
@@ -259,6 +268,25 @@ export class OmniSharpServer {
             args.push('--debug');
         }
 
+        let launchPath: string;
+        if (this._options.path) {
+            try {
+                let serverUrl = "https://roslynomnisharp.blob.core.windows.net";
```
rchande: Maybe move these into some obvious constant definitions somewhere?

src/packages.ts
```diff
@@ -24,6 +24,7 @@ export interface Package {
     architectures: string[];
     binaries: string[];
     tmpFile: tmp.SynchrounousResult;
+    experimentalPackageId?: string;
```
rchande: Change to whatever name Peter came up with :)

src/omnisharp/OmnisharpManager.ts
akshita31: No, in case of useMono we append an extra "omnisharp" to the path

src/OmnisharpDownload.Helper.ts
akshita31: I dont think so. But I just followed by the pattern followed by CSharpExtDownloader.

Currently this pattern helps because in testing I am conveniently passing this field as null, but in future we might want to test the Telemetry as well.

For the current PR, I think we can retain this pattern. What do you think ?

src/omnisharp/server.ts
akshita31: Where do you suggest can these be moved? Should we define some object that contains these fields in the package.json and just read from it here ?

package.json
akshita31: This is automatically added when I npm install the package. What do you think should be changed here?

colombod: I would just move this dependency in the devdependecies list and fix it to 7.1.1 instead of compatible versions

src/omnisharp/server.ts
rchande: You could just move them up to be const fields in server.ts

src/omnisharp/OmnisharpManager.ts
rchande: Ok.

src/OmnisharpDownload.Helper.ts
rchande: Sure, we can refactor later.

PR #2072, opened 2018-02-26
Refactor CSharpExtDownloader to use the extracted helper functions

akshita31:
Helper functions were extracted from the CSharpDownloader to be used by the OmnisharpDownloader (https://github.com/OmniSharp/omnisharp-vscode/pull/2065).

Refactoring the code to use these methods.


src/CSharpExtDownloader.ts
```diff
@@ -9,34 +9,27 @@ import * as util from './common';
...
         private channel: vscode.OutputChannel,
         private logger: Logger,
         private reporter: TelemetryReporter /* optional */,
         private packageJSON: any) {
     }
 
-    public installRuntimeDependencies(): Promise<boolean> {
+    public async installRuntimeDependencies(): Promise<boolean> {
         this.logger.append('Installing C# dependencies...');
+        this.logger.appendLine();
```
TheRealPiotrP: why?

src/CSharpExtDownloader.ts
```diff
@@ -9,34 +9,27 @@ import * as util from './common';
...
-            setMessage: text => {
-                statusItem.text = text;
-                statusItem.show();
-            },
-            setDetail: text => {
-                statusItem.tooltip = text;
-                statusItem.show();
-            }
-        };
+        let statusObject = SetStatus();
```
TheRealPiotrP: Wow, the naming of this object is really tough with the current factoring!

How about we rename `SetStatus` to `GetRightAlignedStatusBarItem`? It's still not perfect, but we'll need to do deeper refactoring around this object to make it better, and that will be better done in a dedicated PR.

src/CSharpExtDownloader.ts
```diff
@@ -46,95 +39,52 @@ export class CSharpExtDownloader
...
-            })
-            .then(() => {
-                installationStage = 'completeSuccess';
-                success = true;
-            })
-            .catch(error => {
-                if (error instanceof PackageError) {
-                    // we can log the message in a PackageError to telemetry as we do not put PII in PackageError messages
-                    telemetryProps['error.message'] = error.message;
+            packageManager = new PackageManager(platformInfo, this.packageJSON);
```
TheRealPiotrP: The original code had a `this.logger.appendLine();` after this line.

src/CSharpExtDownloader.ts
```diff
@@ -46,95 +39,52 @@ export class CSharpExtDownloader
...
-                    if (error.pkg) {
-                        telemetryProps['error.packageUrl'] = error.pkg.url;
-                    }
+            installationStage = 'downloadPackages';
 
-                } else {
-                    // do not log raw errorMessage in telemetry as it is likely to contain PII.
-                    errorMessage = error.toString();
-                }
+            let networkObject = GetNetworkDependencies();
```
TheRealPiotrP: how about:
`let networkConfiguration = GetNetworkConfiguration();`? Rename the variable and the method? this seems clearer.

DustinCampbell:
LGTM. I added @gregg-miskelly to review, since he's the original author of this file.

src/CSharpExtDownloader.ts
```diff
@@ -9,34 +9,27 @@ import * as util from './common';
...
-                statusItem.show();
-            },
-            setDetail: text => {
-                statusItem.tooltip = text;
-                statusItem.show();
-            }
-        };
+        let statusObject = SetStatus();
+        let status = statusObject.Status;
+        let statusItem = statusObject.StatusItem;
```
gregg-miskelly: nit: Instead of having `SetStatus` return the `StatusItem`, I think the code would be a lot more understandable if if you just returned `Status`, and `Status` had a `Dispose` method to dispose the `StatusItem`.

PR #2078, opened 2018-02-28
Enable travis to use the latest omnisharp build

akshita31:
Added settings.json file to the integration tests folder to enable the latest CI build of omnisharp to be consumed by Travis during testing.

### DIFF
--- a/test/integrationTests/testAssets/singleCsproj/.vscode/settings.json
+++ b/test/integrationTests/testAssets/singleCsproj/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "omnisharp.path": "latest"
+}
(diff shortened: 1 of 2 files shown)


rchande:
@akshita31 Change looks good, but how are you verifying travis is actually downloading something different? Can you share the relevant log snippet from your build after it completes?

akshita31:
Here is the log : https://api.travis-ci.org/v3/job/347408371/log.txt

Pasting relevant screenshots that show the path from which the server is starting. The path contains the version that is being used.
travis
travis1

rchande:
@akshita31 Thanks. Another question: why are we downloading omnisharp twice?

akshita31:
@rchande: The first time the machine is downloading the C# dependencies which includes the debugger as well. The next download is particular to the "latest" download, where it first downloads the file that contains the information about the latest download and then downloads/skips the appropriate version packages.

rchande:
@akshita31 Sounds good, thanks. LGTM.

PR #2091, opened 2018-03-05
Add mocha+wallaby tests

TheRealPiotrP:
eventually the feature tests should be removed and most of our tests should become unit tests that are runnable from the command line or via wallaby.

npm run tdd will enable using mocha's command line tdd capability


colombod:
This is very good job

package-lock.json
```diff
@@ -621,6 +646,12 @@
         "ms": "2.0.0"
       }
     },
+    "decamelize": {
```
rchande: What are these new packages used for? Not obvious to me from code changes (but maybe it should be).

TheRealPiotrP: this one is a transient dependency. The new packages are used in mocha.opts to enable running TS tests in mocha

Issue #2086, opened 2018-03-04
"Debug adapter executable 'vsdbg-ui.exe' does not exist" on Windows x86


SridharPasham:
## Environment data
`dotnet --info` output:
.NET Command Line Tools (2.1.4)

Product Information:
 Version:            2.1.4
 Commit SHA-1 hash:  5e8add2190

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x86
 Base Path:   C:\Program Files\dotnet\sdk\2.1.4\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.5
  Build    : 17373eb129b3b05aa18ece963f8795d65ef8ea54

VS Code version:
1.20.1

C# Extension version:
ms-vscode.csharp, version: 1.14.0
## Steps to reproduce
Created a asp.net core mvc application
Commnads:
dotnet new mvc
## Expected  behavior
When I run the application with debugger or without debugger. It should open the browser and load the application.
## Actual behavior
Showing this error message
Debug adapter executable 'C:\\Users\\Sridhar\\.vscode\extensions\\ms-vscode.csharp-1.14.0\\.debugger\\vsdbg-ui.exe' does not exist.

Tried to uninstall and reinstall C# extension but no use.

gregg-miskelly:
@SridharPasham The first time you activate the C# extension, the C# extension will open up an output window pain and show you information about the packages that it is downloading. Are you seeing this? What does it say?

![download-dependencies](https://user-images.githubusercontent.com/11078489/36992125-d0bfeb70-205e-11e8-8a93-ac10f95f4255.png)

SridharPasham:
@gregg-miskelly Tried to uninstall and reinstall C# extension. Did not find any output in OUTPUT window.
If I remember correctly I think it was giving x86 is not supported.

If you need exact message let me know how I can get that initial message.

gregg-miskelly:
Ah. Got it. Yes, the root problem is that we don't support x86 Windows (https://github.com/OmniSharp/omnisharp-vscode/issues/844).

But it looks like we broke our error messages for this scenario. @WardenGnaw Do you have the cycles to try and fix the error messages here?

WardenGnaw:
@gregg-miskelly Yes. I'll take a look.

SridharPasham:
[...]

PR #2094, opened 2018-03-06
Add Architecture check in AdapterExectuableCommand

WardenGnaw:
x86 messages only show during the first installation. But it still
registeres the vsdbg-ui.exe command. This adds a check so it will
error instead of returning the command.

Addresses #2086

### DIFF
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 'use strict';
 
-import * as os from 'os';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import * as common from './../common';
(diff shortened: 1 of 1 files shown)


src/coreclr-debug/activate.ts
```diff
@@ -23,32 +23,39 @@ export async function activate(thisExtension : vscode.Extension<any>, context: v
...
 
     if (!CoreClrDebugUtil.existsSync(_debugUtil.debugAdapterDir())) {
-        let platformInformation: PlatformInformation;
-        
-        try {
-            platformInformation = await PlatformInformation.GetCurrent();
-        }
-        catch (err) {
-            // Somehow we couldn't figure out the platform we are on
+        let isInvalidArchitecture: boolean = await checkForInvalidArchitecture(logger, channel);
```
gregg-miskelly: I think this is going to be really annoying for anyone who is using the extension for editing C#. I think we want to only do this when we first try to install, and when the user tries to debug.

gregg-miskelly: Never mind. I guess this is what we always have done.

src/coreclr-debug/activate.ts
```diff
@@ -23,32 +23,39 @@ export async function activate(thisExtension : vscode.Extension<any>, context: v
     _logger = logger;
 
     if (!CoreClrDebugUtil.existsSync(_debugUtil.debugAdapterDir())) {
```
gregg-miskelly: I think you should probably just delete this check all together and move both checks into `completeDebuggerInstall`.

PR #2101, opened 2018-03-10
Add product-wide code coverage + codecov.io integration

TheRealPiotrP:
Title says most of it :D

Several new scripts were added:
- `npm run cov:instrument`: rebuilds your sources, then instruments them for coverage. Subsequent `npm run test` will generate coverage data into the `.nyc_output` directory
- `npm run cov:merge-html`: merges all reports from `.nyc_output` and puts a locally viewable coverage report into `coverage` directory


tslint.json
```diff
@@ -1,6 +1,6 @@
 {
     "rules": {
-        "tslint-no-unused-expression-chai": true,
```
akshita31: We edited this in a recent PR(https://github.com/OmniSharp/omnisharp-vscode/pull/2092). Why are we changing it here ?

TheRealPiotrP: the syntax here is peculiar and the previous PR did not fully enable this linting rule override.

PR #2133, opened 2018-03-23
Refactoring status using rxjs

akshita31:
Fixes: #2146 
Changes introduced:

1. Removed status.ts and divided it into 4 observers:
- InformationMessageObserver
- WarningMessageObserver
- ProjectStatusObserver
- OmnisharpStatusBarObserver
Each of these observers handles a separate responsibility

2.  The extension was popping up a single `statusBarItem` which showed both the `OmnisharpServerStatus` and the status of the currently open project in the workspace. We have now made two separate status bar items for these.

3. The adding of the listeners and the corresponding disposables for the events like - OnServerStart, OnServerStop, ProjectAdded, etc (done previously in status.ts) are now handled directly in server.ts.

4. For the events in which a project changes, specifically- `ProjectAdded`, `ProjectRemoved` and `ProjectChange` the updateTracker listener is being used to debounce the update project request using rx debounce( We need the debounce with `leading: true`, so the logic with updateTracker is implemented so that the first request will always be serviced and the corresponding requests are debounced)

5.  The logic where the status bar depended on the `ActiveTextEditor` has been removed because we intend the status bar to be a workspace element and not depending on the currently opened document.

6. OmnisharpStatusBarObserver deals with all the Omnisharp events like Server Start, Server Stop, Install, etc. The corresponding status bar element has been moved to the left side of the editor window. It is a clickable object and on click will show the `Omnisharp Log` channel.

running

7. ProjectStatusBarObserver deals with the project that has been selected and displays the name of the currently selected project or the "Select Project" option if there are multiple launch targets. Clicking this will provide the user the option to pick projects from the Command Palette. The "folder" icon is used to represent this element as shown below.

running

When a server stop event is created both these status bars are reset and hidden.

Corresponding tests have also been added for the observers.


src/features/status.ts
```diff
@@ -1,270 +0,0 @@
-/*---------------------------------------------------------------------------------------------
```
TheRealPiotrP: @DustinCampbell  :D :D

src/observers/OmnisharpServerStatusObserver.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface ExecuteCommand<T> {
+    (command: string, ...rest: any[]): Thenable<T | undefined>;
+}
+
+export class OmnisharpServerStatusObserver {
+    private _messageHandle: NodeJS.Timer;
+    private subject: Subject<ObservableEvent.BaseEvent>;
+
+    constructor(private showWarningMessage: ShowWarningMessage<MessageItemWithCommand>, private executeCommand: ExecuteCommand<string>, scheduler?: Scheduler) {
+        this.subject = new Subject<ObservableEvent.BaseEvent>();
```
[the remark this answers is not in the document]
TheRealPiotrP: The observer owns projecting the event into vscode, so it will be responsible for interpreting that configuration.

TheRealPiotrP: subject is not a very informative name here.

src/observers/OmnisharpServerStatusObserver.ts
```diff
@@ -0,0 +1,55 @@
...
+                console.log(value);
+                this.executeCommand(value.command);
+            }
+        });
+    }
+
+    public post = (event: ObservableEvent.BaseEvent) => {
+        switch (event.constructor.name) {
+            case ObservableEvent.OmnisharpServerOnError.name:
+                this.subject.onNext(event);
```
TheRealPiotrP: Maybe `this.projectErrorDebouncer.onNext(...)`?

src/observers/OmnisharpServerStatusObserver.ts
```diff
@@ -0,0 +1,55 @@
+/*---------------------------------------------------------------------------------------------
```
TheRealPiotrP: How about we call this the `ShowWarningMessageObserver`?

src/omnisharp/loggingEvents.ts
```diff
@@ -21,78 +25,90 @@ export class OmnisharpDelayTrackerEventMeasures extends TelemetryEventWithMeasur
...
 
-export class OmnisharpServerMsBuildProjectDiagnostics implements BaseEvent{
+export class OmnisharpServerMsBuildProjectDiagnostics implements BaseEvent {
     constructor(public diagnostics: protocol.MSBuildProjectDiagnostics) { }
 }
 
-export class OmnisharpServerUnresolvedDependencies implements BaseEvent{
-    constructor(public unresolvedDependencies: protocol.UnresolvedDependenciesMessage) { }
+export class OmnisharpServerUnresolvedDependencies implements BaseEvent {
+    constructor(public unresolvedDependencies: protocol.UnresolvedDependenciesMessage, public server: OmniSharpServer, public eventStream: EventStream) { }
```
TheRealPiotrP: This change is temporary, right? We'll get rid of the extra parameters soon?

src/omnisharp/server.ts
```diff
@@ -258,16 +304,16 @@ export class OmniSharpServer {
...
             }
             catch (error) {
-                this.eventStream.post(new OmnisharpFailure(`Error occured in loading omnisharp from omnisharp.path\nCould not start the server due to ${error.toString()}`, error));
+                this.eventStream.post(new ObservableEvents.OmnisharpFailure(`Error occured in loading omnisharp from omnisharp.path\nCould not start the server due to ${error.toString()}`, error));
                 return;
             }
         }
 
-        this.eventStream.post(new OmnisharpInitialisation(new Date(), solutionPath));
+        this.eventStream.post(new ObservableEvents.OmnisharpInitialisation(new Date(), solutionPath));
```
TheRealPiotrP: Why do we need to add the `ObservableEvents` specifier everywhere? It makes the code much more busy..

akshita31: The import from logging events was becoming too long so I changed that to `import * as ObservableEvents from loggingEvents` . What do you think is better ?

src/omnisharp/loggingEvents.ts
akshita31: Yes we will use the vscode api's to do the restore then we can get rid of this here.

src/observers/OmnisharpServerStatusObserver.ts
akshita31: Done!

src/observers/OmnisharpServerStatusObserver.ts
akshita31: I renamed it as `warningMessageDeboucer`. Does that seem alright ?

src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -0,0 +1,189 @@
...
+                this.handleOmnisharpServerOnStop();
+                break;
+            case ObservableEvent.OmnisharpServerOnStart.name:
+                this.handleOmnisharpServerOnStart(<ObservableEvent.OmnisharpServerOnStart>event);
+                break;
+        }
+    }
+
+    private updateProjectInfo = (server: OmniSharpServer) => {
+        serverUtils.requestWorkspaceInformation(server).then(info => {
```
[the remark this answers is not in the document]
akshita31: We have removed the status.ts and moved its functionality across three files and this is one of them.

src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -0,0 +1,189 @@
...
+            case ObservableEvent.OmnisharpOnBeforeServerInstall.name:
+                SetStatus(this.defaultStatus, '$(flame) Installing OmniSharp...', 'o.showOutput', '');
+                this.render();
+                break;
+            case ObservableEvent.OmnisharpOnBeforeServerStart.name:
+                SetStatus(this.defaultStatus, '$(flame) Starting...', 'o.showOutput', '');
+                this.render();
+                break;
+            case ObservableEvent.ActiveTextEditorChanged.name:
+                this.render();
```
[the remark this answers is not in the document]
akshita31: The `Status` class is a wrapper around the StatusBarItem. There are two possible values that we might display in the StatusBarItem object - defaultStatus or projectStatus. On calling render we check whether any of those values are set for the class and if they are then we display them according to some logic.

src/features/status.ts
DustinCampbell: :smile:

src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -0,0 +1,191 @@
...
+    private defaultStatus: Status;
+    private projectStatus: Status;
+    private localDisposables: CompositeDisposable;
+    private updateProjectDebouncer: Subject<OmniSharpServer>;
+    private firstUpdateProject: boolean;
+
+    constructor(private vscode: vscode, private statusBar: StatusBarItem, scheduler?: Scheduler) {
+        this.defaultStatus = new Status(defaultSelector);
+        this.updateProjectDebouncer = new Subject<OmniSharpServer>();
+        this.updateProjectDebouncer.debounce(1500, scheduler).subscribe((server: OmniSharpServer) => { this.updateProjectInfo(server); });
```
DustinCampbell: If we're using debouce from RX, can we remove the dependency on `lodash.debounce` in package.json?

src/main.ts
```diff
@@ -31,34 +36,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<{ init
...
     let omnisharpChannel = vscode.window.createOutputChannel('OmniSharp Log');
     let omnisharpLogObserver = new OmnisharpLoggerObserver(omnisharpChannel);
     let omnisharpChannelObserver = new OmnisharpChannelObserver(omnisharpChannel);
+    eventStream.subscribe(omnisharpLogObserver.post);
+    eventStream.subscribe(omnisharpChannelObserver.post);
 
-    const eventStream = new EventStream();
-    eventStream.subscribe(dotnetChannelObserver.post);
-    eventStream.subscribe(dotnetLoggerObserver.post);
+    let omnisharpServerStatusObserver = new WarningMessageObserver(vscode);
```
DustinCampbell: Should this be called `warningMessageObserver` to be consistent with `informationMessageObserver` below?

src/main.ts
```diff
@@ -31,34 +36,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<{ init
...
+    eventStream.subscribe(omnisharpServerStatusObserver.post);
 
-    eventStream.subscribe(csharpchannelObserver.post);
-    eventStream.subscribe(csharpLogObserver.post);
+    let informationMessageObserver = new InformationMessageObserver(vscode);
+    eventStream.subscribe(informationMessageObserver.post);
+
+    let omnisharpStatusBar = new StatusBarItemAdapter(vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, Number.MIN_VALUE));
```
DustinCampbell: Should the `omnisharpStatusBar` be owned by `OmniSharp` in https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/extension.ts?

src/main.ts
```diff
@@ -31,34 +36,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<{ init
...
+    eventStream.subscribe(dotnetChannelObserver.post);
+    eventStream.subscribe(dotnetLoggerObserver.post);
 
     let csharpChannel = vscode.window.createOutputChannel('C#');
     let csharpchannelObserver = new CsharpChannelObserver(csharpChannel);
     let csharpLogObserver = new CsharpLoggerObserver(csharpChannel);
+    eventStream.subscribe(csharpchannelObserver.post);
+    eventStream.subscribe(csharpLogObserver.post);
 
     let omnisharpChannel = vscode.window.createOutputChannel('OmniSharp Log');
```
DustinCampbell: Should `omnisharpChannel` (and its observers) be owned by `OmniSharp` in https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/extension.ts?

src/observers/InformationMessageObserver.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    public post = (event: ObservableEvent.BaseEvent) => {
+        switch (event.constructor.name) {
+            case ObservableEvent.OmnisharpServerUnresolvedDependencies.name:
+            this.handleOmnisharpServerUnresolvedDependencies(<ObservableEvent.OmnisharpServerUnresolvedDependencies>event);
+            break;
+        }
+    }
+
+    private handleOmnisharpServerUnresolvedDependencies(event: ObservableEvent.OmnisharpServerUnresolvedDependencies) {
```
DustinCampbell: FWIW, it'd probably be a good idea to specify return types in general, but especially when a random Promise is the result. It's really unclear from the signature what this is expected to return, which makes the code hard to understand from a cursory read.

src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -0,0 +1,191 @@
...
+            SetStatus(this.projectStatus, '$(flame) ' + label, 'o.pickProjectAndStart');
+            SetStatus(this.defaultStatus, '$(flame) Switch projects', 'o.pickProjectAndStart');
+            this.render();
+        });
+    }
+
+    private render = () => {
+        let activeTextEditor = this.vscode.window.activeTextEditor;
+        if (!activeTextEditor) {
+            this.statusBar.hide();
```
DustinCampbell: FWIW, I've wondered at times whether we should tweak this bit of code. Essentially, this means that the status bar item is not visible unless a file is open in the editor, which is a little weird IMO. What do you think?

src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -0,0 +1,191 @@
...
+];
+
+export class OmnisharpStatusBarObserver {
+    private defaultStatus: Status;
+    private projectStatus: Status;
+    private localDisposables: CompositeDisposable;
+    private updateProjectDebouncer: Subject<OmniSharpServer>;
+    private firstUpdateProject: boolean;
+
+    constructor(private vscode: vscode, private statusBar: StatusBarItem, scheduler?: Scheduler) {
```
DustinCampbell: Should `statusBar` be `statusBarItem`?

src/omnisharp/server.ts
```diff
@@ -516,11 +560,11 @@ export class OmniSharpServer {
         const request = this._requestQueue.dequeue(packet.Command, packet.Request_seq);
 
         if (!request) {
-            this.eventStream.post(new OmnisharpServerMessage(`Received response for ${packet.Command} but could not find request.`));
+            this.eventStream.post(new ObservableEvents.OmnisharpServerMessage(`Received response for ${packet.Command} but could not find request.`));
             return;
         }
 
-        this.eventStream.post(new OmnisharpServerVerboseMessage(`handleResponse: ${packet.Command} (${packet.Request_seq})`));
+        this.eventStream.post(new ObservableEvents.OmnisharpServerVerboseMessage(`handleResponse: ${packet.Command} (${packet.Request_seq})`));
```
DustinCampbell: FWIW, "OmniSharp" has a capital 'S'

src/observers/OmnisharpStatusBarObserver.ts
akshita31: Done!

DustinCampbell: thanks!

src/main.ts
akshita31: @DustinCampbell Right now we are trying to initialise all the observers in the main.ts so that it is that component that does the wiring up. But if you still feel that we need to move those observers into extension.ts we can do that.

colombod:
Looks good to me

src/observers/InformationMessageObserver.ts
```diff
@@ -0,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as ObservableEvent from "../omnisharp/loggingEvents";
+import { vscode } from '../vscodeAdapter';
+
+export class InformationMessageObserver {
```
rchande: Why does the info message observer not need to debounce when the warning one does? 
Also, why is there not an error message observer (maybe it was added in another PR)?

Generally, it seems like we might be painting ourselves into a corner here. If for some reason, we want to be able to configure the errorlevel of a message, we now have to go teach each error level observer how to print that message. Maybe it would be clearer to only have one WindowPostMessageObserverThingy and have it know how to translate event -> VS Code Warning level?

src/observers/WarningMessageObserver.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface MessageItemWithCommand extends MessageItem {
+    command: string;
+}
+
+export class WarningMessageObserver {
+    private warningMessageDebouncer: Subject<BaseEvent>;
+
+    constructor(private vscode: vscode, scheduler?: Scheduler) {
+        this.warningMessageDebouncer = new Subject<BaseEvent>();
+        this.warningMessageDebouncer.debounce(1500, scheduler).subscribe(async event => {
```
rchande: We've gotten feedback that having a ton of these messages appear is annoying. The command we have here doesn't appear to scroll the output to a specific location. Instead of "Debounce", should we change this observer to just raise one notification if it gets that event?

src/omnisharp/server.ts
```diff
@@ -68,9 +70,8 @@ const latestVersionFileServerPath = 'releases/versioninfo.txt';
 export class OmniSharpServer {
 
     private static _nextId = 1;
-
     private _readLine: ReadLine;
-    private _disposables: vscode.Disposable[] = [];
+    private _disposables: CompositeDisposable = new CompositeDisposable(Disposable.empty);
```
rchande: Why?

src/omnisharp/server.ts
```diff
@@ -286,11 +342,24 @@ export class OmniSharpServer {
         });
     }
 
-    public stop(): Promise<void> {
-
-        while (this._disposables.length) {
-            this._disposables.pop().dispose();
+    // we could move this to a project updated observer and pass the server there. But we will need the event stream also there :(
+    private updateTracker = () => {
+        if (this.firstUpdateProject) {
```
rchande: I don't really understand what this is trying to? The first time, it just sends the message, but subsequent project updates are debounced?

src/omnisharp/server.ts
```diff
@@ -84,12 +85,17 @@ export class OmniSharpServer {
 
     private _omnisharpManager: OmnisharpManager;
     private eventStream: EventStream;
+    private updateProjectDebouncer: Subject<ObservableEvents.ProjectModified>;
+    private firstUpdateProject: boolean;
```
rchande: Just initialize this here.

src/omnisharp/server.ts
```diff
@@ -286,11 +342,24 @@ export class OmniSharpServer {
         });
     }
 
-    public stop(): Promise<void> {
-
-        while (this._disposables.length) {
-            this._disposables.pop().dispose();
+    // we could move this to a project updated observer and pass the server there. But we will need the event stream also there :(
```
rchande: I don't understand this comment, but I agree with the `:(`

akshita31: Sorry I had added in while working but forgot to delete :(

src/omnisharp/server.ts
akshita31: We are trying to remove the dependency on vscode wherever possible. So here we are using the Rx disposables instead.

rchande: I meant, why is this initialized to a empty composite disposable when line 286 replaces it with a real disposable?

src/omnisharp/server.ts
akshita31: The first time when we get one of the three events - 'onProjectAdded` , `onProjectChanged` and `onProjectRemoved` , the updateProjectInfo function is invoked directly. After this if we get any of the events, then we invoke the `updateProjectInfo` function via the debouncer. This is the functional equivalent of lodash.debounce with leading: true - https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-1

rchande: What's the consequence if you always debounce? The first message is delayed by debounce interval?

akshita31: If we keep getting the request then the status bar will not be displayed until there is a silence of 1500 ms. This way with leading true, we service the first request that we get so status bar shows something, and then when for the next requests we see a silence of 1500ms then we service them. Basically it is just so that we have something to display for the first time.

src/omnisharp/server.ts
akshita31: Yes, you are right. Removed the empty initialization.

src/omnisharp/server.ts
rchande: Can you remind me what actual status/message this is displaying? If we can make this code  simpler and the consequence is that a message gets delayed by 1.5 seconds, I would be in favor of that. Up to you though...

src/observers/InformationMessageObserver.ts
akshita31: The error event and MSbuildDiagnostics event shows the warning message hence they are both present in the warning message observer

src/observers/WarningMessageObserver.ts
akshita31: But if the server restarts then ?

rchande: Ah, this was me misunderstanding debouce. Looks good to me.

Issue #2137, opened 2018-03-25
cant create vsix offline


ronmara:
## Environment data
`dotnet --info` output:
.NET Command Line Tools (2.1.101)

Product Information:
 Version:            2.1.101
 Commit SHA-1 hash:  6c22303bf0

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.101\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.6
  Build    : 74b1c703813c8910df5b96f304b0f2b78cdf194d

VS Code version: 1.21.1
C# Extension version: 1.15.0-beta3

## Steps to reproduce

> run 
> npm i
> npm run compile

Open up gulpfile.js which is located at the root level of the repo and on line 95 change:
`throw new Error('Do not build offline packages on windows. Runtime executables will not be marked executable in *nix packages.');`

to:
`console.log('Do not build offline packages on windows. Runtime executables will not be marked executable in *nix packages.');`

now run

> node node_modules/gulp/bin/gulp.js package:offline

(followed this [guide](https://www.towfeek.se/2017/07/05/enabling-csharp-and-dotnet-core-debugging-in-vs-code-from-an-offline-environment/))

## Expected  behavior
Ill get my offline VSIX package on windows

## Actual behavior
getting exception

> TypeError: eventStream.post is not a function
> at downloadPackage (E:\omnisharp-vscode\out\src\packages.js:141:17)
> at doesPackageTestPathExist.then (E:\omnisharp-vscode\out\src\packages.js:132:20)
> at <anonymous<anonymous>>

rchande:
@akshita31 It looks like this might have gotten regressed with your eventStream changes. Want to take a look?

akshita31:
Yes, it seems that the downloadPackage function expects an eventStream but we are not providing it any.  This part got ignored probably while we made those changes. Will make the necessary changes now.

rchande:
Also, we've had a couple of customers report bugs in this area (offline packaging). Should we add an "offline packaging" integration test that runs whatever scripts a user is expected to run a verifies that they work?

akshita31:
[...]

PR #2138, opened 2018-03-26
Fix offline packaging using event stream

akshita31:
Fixes: #2137


gulpfile.js
```diff
@@ -102,7 +102,7 @@ function doPackageSync(packageName) {
 
 function doOfflinePackage(platformInfo, packageName, packageJSON) {
     if (process.platform === 'win32') {
-        throw new Error('Do not build offline packages on windows. Runtime executables will not be marked executable in *nix packages.');
+        //throw new Error('Do not build offline packages on windows. Runtime executables will not be marked executable in *nix packages.');
```
rchande: Should this get enabled?

test/releaseTests/offlinePackage.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as chai from 'chai';
+import * as glob from 'glob-promise';
+import * as path from 'path';
+import { invokeCommand } from './testAssets/testAssets';
+import { PlatformInformation } from '../../src/platform';
+import * as fs from 'async-file';
+
+suite("Offline packaging of VSIX", function () {
+    let vsixFiles: string[];
+    this.timeout(2000000);
```
rchande: If this timeout is in milliseconds, that means this test takes nearly 30 minutes to run in travis? Probably not desirable to run that on every build?

test/releaseTests/testAssets/testAssets.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+import * as path from 'path';
+
+export function invokeCommand(args: string[]){
```
rchande: invokeNode? invokeCommand sounds like it could invoke any program

rchande:
@akshita31 Can we add a check that checks if ("TRAVIS_PULL_REQUEST_BRANCH" == """) and then does the release test? That will not run these expensive tests on people's PRs, but will fail releases.

rchande:
@TheRealPiotrP @colombod Thoughts on when we should run this slow and infrequently regressed test?

rchande:
Also @akshita31 in my earlier comment, I was referring to travis.yml :)

.travis.yml
```diff
@@ -34,7 +34,10 @@ script:
   - npm run cov:instrument
   - npm test --silent
   - npm run cov:report
-  - npm run test:release
+  - npm run test:artifacts
+  - if [ $TRAVIS_PULL_REQUEST_BRANCH == "" ]; then
```
rchande: One other thing: can you also check that `travis tag` != `master`? That way we'll only run this when someone starts a release (the release process on GitHub creates a new branch).

.travis.yml
```diff
@@ -34,7 +34,8 @@ script:
   - npm run cov:instrument
   - npm test --silent
   - npm run cov:report
-  - npm run test:release
+  - npm run test:artifacts
+  - test $TRAVIS_TAG != "master" && npm run test:release
```
TheRealPiotrP: Is this sufficient? We don't need to also check for PR build?

.travis.yml
```diff
@@ -35,7 +35,7 @@ script:
   - npm test --silent
   - npm run cov:report
   - npm run test:artifacts
-  - 'if [ "$TRAVIS_TAG" != "master" && "$TRAVIS_PULL_REQUEST" = "false" ]; then npm run test:release; fi'
+  - 'if [[ "$TRAVIS_TAG" != "master" && "$TRAVIS_PULL_REQUEST" = "false" ]]; then npm run test:release; fi'
```
rchande: Why do we need double brackets?

akshita31: "&&" operator requires double brackets - https://stackoverflow.com/questions/15993062/bash-scripting-missing/15993083. Without using that I was getting an error

rchande:
@akshita31 BTW, we should probably "Squash merge"t his one :)

PR #2142, opened 2018-03-28
Reduce the size of vsix

akshita31:


TheRealPiotrP:
Are we able to add a test here?

rchande:
@akshita31 Can you explain how this fix worked?

TheRealPiotrP:
This is a partial fix. We have some dependencies with rather bloated npm packages. This is typically not an issue because of bundling tools that trim unused code...

PR #2143, opened 2018-03-28
1.15.0-beta4

akshita31:
Checking if the size is reduced by changes


DustinCampbell:
You can build the VSIX locally too. :smile:

PR #2156, opened 2018-04-04
Test the vsix, not the build layout

TheRealPiotrP:
The VSCode C# Extension build process follows the VS Code docs and runs tests directly inside of its repo root. This unfortunately gives a false sense of security because bugs can be introduced during VSIX packaging [particularly due to missing content in node_modules or excluded via .vscodeignore]. 

This change addresses this problem by moving our CI tests to execute the VSIX instead of the build's intermediate artifacts. Specifically:
 - build the vsix
 - unpackage the vsix
 - instrument the unpackaged vsix
 - run tests with VS Code Host pointing to the unpackaged vsix

This makes our CI tests ~= to the user's runtime experience and will greatly help us with size reduction efforts.

To support this change, I also moved our build system from package.json to Gulp. This makes the build scripts significantly easier to understand, provides intellisense for build scripts, and build-time type checking for their contents.

I also strengthened the repo's use of .vscodeignore by creating a copy of the file for each scenario [online packages and offline packages]. The new gulp packaging scripts take advantage of these files to produce packages with predictable contents regardless of when packaging occurs. [small caveat, @akshita31 will be adding a test that validates that net-new content does not start sneaking into the vsix package].


tasks/commandLineArguments.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as minimist from 'minimist';
+import * as path from 'path';
+
+let argv = minimist(process.argv.slice(2), { 
+    boolean: ['retainVsix']
+});
+
+export const commandLineOptions ={
+    retainVsix: !!argv['retainVsix'],
+    vsixPackagingOutputFolder:  makePathAbsolute(argv['o']),
```
akshita31: Should we name this just as an output folder so that any other gulp tasks(that may or may not be related to vsixs) can also specify a `-o` option to list their output folder.

tasks/offlinePackagingTasks.ts
```diff
@@ -0,0 +1,133 @@
...
+import { commandLineOptions } from '../tasks/commandLineArguments';
+import { CsharpLoggerObserver } from '../src/observers/CsharpLoggerObserver';
+import { EventStream } from '../src/EventStream';
+import { getPackageJSON } from '../tasks/packageJson';
+import { Logger } from '../src/logger';
+import { PackageManager } from '../src/packages';
+import { PlatformInformation } from '../src/platform';
+
+gulp.task('vsix:offline:package', () => {
+    del.sync(vscodeignorePath);
```
akshita31: Do we need to delete the vscodeingorePath twice, when we start the task as well as when we exit ?

tasks/onlinePackagingTasks.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as debugUtil from '../src/coreclr-debug/util';
+import * as del from 'del';
+import * as fs from 'fs';
+import * as gulp from 'gulp';
+import * as path from 'path';
+import * as unzip from 'unzip2';
+import * as util from '../src/common';
+
+import { offlineVscodeignorePath, onlineVscodeignorePath, rootPath, unpackedVsixPath, vscePath, vscodeignorePath } from './projectPaths';
+
```
akshita31: blank line

tasks/offlinePackagingTasks.ts
```diff
@@ -0,0 +1,133 @@
...
+    let stdoutObserver = new CsharpLoggerObserver(logger);
+    eventStream.subscribe(stdoutObserver.post);
+    const debuggerUtil = new debugUtil.CoreClrDebugUtil(path.resolve('.'));
+
+    return packageManager.DownloadPackages(eventStream, undefined, undefined, undefined)
+        .then(() => {
+            return packageManager.InstallPackages(eventStream, undefined);
+        })
+        .then(() => {
+
```
akshita31: blank line

tasks/onlinePackagingTasks.ts
```diff
@@ -0,0 +1,54 @@
...
+    fs.createReadStream(packageName).pipe(unzip.Extract({ path: unpackedVsixPath }));
+});
+
+gulp.task('vsix:online:package', (onError) => {
+    del.sync(vscodeignorePath);
+
+    fs.copyFileSync(onlineVscodeignorePath, vscodeignorePath);
+
+    let onDone = (reason) => {
+
```
akshita31: blank line

tasks/projectPaths.ts
```diff
@@ -0,0 +1,52 @@
...
+export const codecovPath = path.join(nodeModulesPath, 'codecov', 'bin', 'codecov');
+export const vscodeTestHostPath = path.join(nodeModulesPath, 'vscode', 'bin', 'test');
+
+export const packageJsonPath = path.join(rootPath, "package.json");
+
+export const packedVsixOutputRoot = commandLineOptions.vsixPackagingOutputFolder || rootPath;
+export const unpackedVsixPath = path.join(rootPath, "vsix");
+export const unpackedExtensionPath = path.join(unpackedVsixPath, "extension");
+
+
```
akshita31: blank line

tasks/onlinePackagingTasks.ts
```diff
@@ -0,0 +1,54 @@
...
+import { CsharpLoggerObserver } from '../src/observers/CsharpLoggerObserver';
+import { EventStream } from '../src/EventStream';
+import { Logger } from '../src/logger';
+import { PackageManager } from '../src/packages';
+import { PlatformInformation } from '../src/platform';
+import { getPackageJSON } from './packageJson';
+import spawnNode from './spawnNode';
+
+gulp.task('vsix:online:unpackage', () => {
```
rchande: "Release" packaging?

tasks/projectPaths.ts
```diff
@@ -0,0 +1,52 @@
+/*---------------------------------------------------------------------------------------------
```
rchande: I like this file :)

tasks/offlinePackagingTasks.ts
TheRealPiotrP: yes. it's a transient build artifact. This is exactly the scenario for which we want the vsce --vscodeignorefile option

TheRealPiotrP:
Addressed PR feedback. Once this is green, let's merge.

PR #2159, opened 2018-04-06
Enable `noImplicitAny` and `alwaysStrict` in `tsconfig.json

TheRealPiotrP:
Along the way, fixed a few bugs that were in place due to implicit anys. Also removed dependency on deprecated vscode API.


gulpfile.ts
```diff
@@ -10,6 +10,8 @@ import * as optionsSchemaGenerator from './src/tools/GenerateOptionsSchema';
 import * as packageDependencyUpdater from './src/tools/UpdatePackageDependencies';
 
 import tslint from 'gulp-tslint';
+import { PlatformInformation } from './src/platform';
```
akshita31: We are not using these imports anywhere ?

src/features/diagnosticsProvider.ts
```diff
@@ -162,17 +162,17 @@ class DiagnosticsProvider extends AbstractSupport {
...
-        if (this._diagnostics[key]) {
+        if (this._diagnostics.get(key)) {
             didChange = true;
-            this._diagnostics[key].dispose();
-            delete this._diagnostics[key];
+            this._diagnostics.delete(key);
         }
-        if (this._documentValidations[key]) {
+        
+        if (this._documentValidations[key.toString()]) {
```
rchande: I think we did `key.toString()` enough times to justify introducing a local :)
BTW, why is `key` directly used as a ... key in `_diagnostics` but we toString it to use it as a key in `_documentValidations`?

tasks/offlinePackagingTasks.ts
```diff
@@ -58,7 +58,7 @@ function doPackageOffline() {
         new PlatformInformation('linux', 'x86_64')
     ];
 
-    let promise = Promise.resolve();
+    let promise: Promise<Result> = Promise.resolve<Result>(null);
```
DustinCampbell: Does this need type on the right side even after you've specified the generic argument?

test/unitTests/logging/OmnisharpChannelObserver.test.ts
```diff
@@ -21,16 +20,4 @@ suite("OmnisharpChannelObserver", () => {
         observer.post(event);
         expect(hasShown).to.be.true;
     });
-
```
akshita31: Why did we remove the test here?

src/observers/BaseChannelObserver.ts
```diff
@@ -14,13 +13,8 @@ export abstract class BaseChannelObserver {
...
     abstract post: (event: BaseEvent) => void;
 
-    public showChannel(preserveFocusOrColumn?: boolean | ViewColumn, preserveFocus?: boolean) {
-        if (typeof(preserveFocusOrColumn) === "boolean") {
-            this.channel.show(preserveFocusOrColumn as boolean);
-        }
-        else {
-            this.channel.show(preserveFocusOrColumn as ViewColumn, preserveFocus);
-        }
+    public showChannel(preserveFocusOrColumn?: boolean) {
```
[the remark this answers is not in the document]
akshita31: The ViewColumn thing is deprecated, so we removed that from the part of code that was using it.

test/featureTests/OmnisharpDownloader.test.ts
```diff
@@ -12,13 +13,12 @@ import { rimraf } from 'async-file';
...
 import { EventStream } from '../../src/EventStream';
 
-const tmp = require('tmp');
 const chai = require("chai");
 chai.use(require("chai-as-promised"));
 let expect = chai.expect;
 
 suite("DownloadAndInstallExperimentalVersion : Gets the version packages, downloads and installs them", () => {
-    let tmpDir = null;
+    let tmpDir: tmp.SynchrounousResult = null;
```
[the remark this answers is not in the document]
TheRealPiotrP: sorry, I don't own the @types/tmp package :(

test/unitTests/logging/OmnisharpChannelObserver.test.ts
TheRealPiotrP: This test was testing the deprecated column functionality. Since we removed that parameter, we don't need to test for it.

PR #2160, opened 2018-04-06
Remove the status handling from the package manager

akshita31:
Making use of events already present in the package manager and observing them in the `OmnisharpStatusBarObserver` to show the corresponding status.

Also modified the `flame Running` thing to just `flame` in green color and showing the message in the tooltip. The 'Error' and 'Starting' case have also be modified similarly.
falmegreen1
flamegreen

Added tests for the same


src/observers/BaseStatusBarItemObserver.ts
```diff
@@ -11,17 +11,19 @@ export abstract class BaseStatusBarItemObserver {
     constructor(private statusBarItem: StatusBarItem) {
     }
 
-    public SetAndShowStatusBar(text: string, command: string, color?: string) {
+    public SetAndShowStatusBar(text: string, command: string, color: string, tooltip: string) {
```
rchande: Consider making tooltip an optional parameter? That way the callsites that pass empty string for tooltip can just pass nothing instead? Marginally more readable, I guess.

src/observers/BaseStatusBarItemObserver.ts
```diff
@@ -11,17 +11,19 @@ export abstract class BaseStatusBarItemObserver {
...
         this.statusBarItem.text = text;
         this.statusBarItem.command = command;
         this.statusBarItem.color = color;
```
rchande: Consider using a "themecolor"? https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-namethemecoloraspan-classcodeitem-id247themecolorspan

src/observers/BaseStatusBarItemObserver.ts
rchande: I would probably do the same for "color", since you seem to pass green or nothing.

PR #2162, opened 2018-04-08
tsconfig.json: noUnusedLocals, noFallThroughCaseInSwitch, tslint.json: promise-function-async

TheRealPiotrP:
If this is too much context to review, I can break into 3 PRs.

Adds noUnusedLocals to tsconfig.json to keep our sources clean
Adds noFallThroughCaseInSwitch in tsconfig.json to prevent unintended switch behavior
Adds promise-function-async to tslint.json to force all async functions to be marked as async. This is a building block towards eliminating promises in favor of async/await.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -295,8 +295,8 @@ function hasAddOperations(operations: Operations) {
     return operations.addLaunchJson || operations.addLaunchJson;
 }
 
-function getOperations(generator: AssetGenerator) {
-    return getBuildOperations(generator.tasksJsonPath).then(operations =>
+async function getOperations(generator: AssetGenerator) {
+    return getBuildOperations(generator.tasksJsonPath).then(async operations =>
         getLaunchOperations(generator.launchJsonPath, operations));
 }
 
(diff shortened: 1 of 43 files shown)


TheRealPiotrP:
Coverage is a bit lower after this change. I don't believe that any scenarios are actually degraded, however. This is a combination of:
 - character-based coverage %. I added more characters like `async` which, for uncovered lines that return promises, gave more weight to those lines.
 - I made fallthroughs in some switches explicit. These scenarios were not being covered, and are now not covered with more source code characters.

src/assets.ts
```diff
@@ -295,8 +295,8 @@ function hasAddOperations(operations: Operations) {
     return operations.addLaunchJson || operations.addLaunchJson;
 }
 
-function getOperations(generator: AssetGenerator) {
-    return getBuildOperations(generator.tasksJsonPath).then(operations =>
+async function getOperations(generator: AssetGenerator) {
+    return getBuildOperations(generator.tasksJsonPath).then(async operations =>
```
colombod: should this become await pattern then

PR #2168, opened 2018-04-10
unify rx usage to rxjs

TheRealPiotrP:
The current codebase uses both rx [v4] and rxjs [v5] implementations. This PR consolidates our use of rx onto the v5 libraries.


test/unitTests/logging/WarningMessageObserver.test.ts
```diff
@@ -77,39 +84,91 @@ suite('WarningMessageObserver', () => {
...
             });
         });
     });
-});
\ No newline at end of file
+});
+
+function timeToMarble(timeinMilliseconds: number): string {
+    let marble = "";
+    for (let i = 0; i < (timeinMilliseconds / 10); i++) {
```
rchande: Wat

test/unitTests/logging/WarningMessageObserver.test.ts
```diff
@@ -77,39 +84,91 @@ suite('WarningMessageObserver', () => {
...
+                    [getMSBuildDiagnosticsMessage("errorFile", "", "", 5, 6, 7, 8)]);
+
+                let eventC = getOmnisharpMSBuildProjectDiagnosticsEvent("CFile",
+                    [getMSBuildDiagnosticsMessage("warningFile", "", "", 1, 2, 3, 4)],
+                    [getMSBuildDiagnosticsMessage("errorFile", "", "", 5, 6, 7, 8)]);
+
+                let marble_event_map = { a: event, b: eventB, c: eventC };
+                let eventList = scheduler.createHotObservable(marble, marble_event_map);
+                eventList.subscribe(e => observer.post(e));
+                scheduler.expectObservable(assertionObservable).toBe(`${timeToMarble(3520)}a`, { a: '[1] Some projects have trouble loading. Please review the output for more details.' });
```
akshita31: We can also check here that in an intermediate time say 2000, there is no output because of the debouncing

akshita31: And the event that will be emitted here will be event c, because of the debouncing event a and b will not be emitted and since c will have an emission silence it will be emitted

Issue #1990, opened 2018-02-01
Extension complains about project.json files in node_modules


bdukes:
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (2.1.4)

Product Information:
 Version:            2.1.4
 Commit SHA-1 hash:  5e8add2190

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.4\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.5
  Build    : 17373eb129b3b05aa18ece963f8795d65ef8ea54
```
VS Code version: 1.19.3
C# Extension version: 1.13.1

## Steps to reproduce
- Use Yarn or npm to install the `edge` or `edge-js` package
- Open the root folder (i.e. the folder containing `node_modules`) in VS Code

## Expected  behavior
- No warnings

## Actual behavior
- Warnings about `project.json` and requests to restore dependencies

Related to agracio/edge-js#20

DustinCampbell:
Thanks for reporting! OmniSharp should just leave node_modules out when searching for projects.

PR #2171, opened 2018-04-10
Pass a semicolon-delimited list of files excluded from VS Code.

rchande:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/1990


TheRealPiotrP:
Can we make the list an array instead of a string? This will be easier for users to edit, to diff..

TheRealPiotrP:
Also, a test would be useful :)

TheRealPiotrP:
And are the files consisting of relative paths? Relative to what? What happens if I switch launch targets?

rchande:
@TheRealPiotrP 
> Can we make the list an array instead of a string? This will be easier for users to edit, to diff..
Users define this data by editing artifacts in their repo or in vs code settings. The string generated here will be passed to O# as a command line arg. Arguably, it could be json, but that's not very user-friendly for anyone invoking o# outside of the C# extension.

> And are the files consisting of relative paths? Relative to what? What happens if I switch launch targets?
They're paths relative to the workspace root. Maybe there's some normalizing we need to do if the working directory for o# isn't the workspace root. 

> Also, a test would be useful :)
Ok fine :)

firelizzard18:
Is this still being worked?

akshita31:
cc @rchande

rchande:
@kvalev Thanks for the ping.
@filipw I've implemented your suggestion to use the normal omnisharp configuration mechanism. Any other feedback?

WeAthFoLD:
Is this still being worked on? It would be a very useful feature.

rchande:
Better late than never, this is ready to merge.

Issue #2176, opened 2018-04-11
C# output pane is always displayed with "omnisharp.path" is set to "latest"


DustinCampbell:
I set my "omnisharp.path" to "latest" and found that this has the effect of popping the C# output every time I open VS Code, even when it doesn't need to update anything:

```
Getting latest build information...

Downloading package 'Latest version information file' (1 KB) .................... Done!
Installing Omnisharp Version = 1.29.2-beta.133...
Platform: win32, x86_64

Skipping package 'OmniSharp for Windows (.NET 4.6 / x64), Version = 1.29.2-beta.133' (already downloaded).
Finished
```

It would be nice if the output pane was only shown if the extension needs to download a new OmniSharp.

akshita31:
Seems we regressed this behavior during some of the refactorings.

PR #2178, opened 2018-04-11
Show the C# channel when there is an actual download

akshita31:
Fixes: #2176 

Stop the channel observer from listening to `PackageInstallation` event and instead listen to `DownloadStart` event.


DustinCampbell:
That was fast! :smile:

akshita31:
The refactoring effort that we have put so far has really made logging easy to understand and debug. Thanks to @TheRealPiotrP @colombod 👯‍♂️

PR #2179, opened 2018-04-11
group completions by key and label

colombod:
completions are grouped by kind and label.


rchande:
@colombod Can add screenshots showing how the new behavior is different?

colombod:
Closing the PR, is not allowed by the language

PR #2216, opened 2018-04-25
Update debugger links for 1.15.0-beta6

chuckries:
This updates the debugger packages to the latest corelr-debug which is
now built on .NET Core 2.1 Preview 2. We are skipping beta5 due to an
internal build of the debugger that was marked beta5 but will not be
shipping.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-vscode",
-  "version": "1.15.0-beta4",
+  "version": "1.15.0-beta6",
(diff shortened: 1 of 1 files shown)


chuckries:
@gregg-miskelly @DustinCampbell @TheRealPiotrP

gregg-miskelly:
I already said this in person, but I think we should re-upload those zips to `15-1`. I would expect that this is the last insertion of the debugger for 15.0. So I don't think we want 'beta' in the URL name.

chuckries:
abondoning in favor of https://github.com/OmniSharp/omnisharp-vscode/pull/2226

PR #2219, opened 2018-04-25
Using a separate Package class for resolved packages

akshita31:
Renamed the existing 'Package' interface to `PackageJSONPackage` and created a new `Package` class which is formed by resolving the paths using the formed extension paths, which is then used for the download and installl process


src/CSharpExtDownloader.ts
```diff
@@ -54,9 +54,10 @@ export class CSharpExtDownloader {
     }
 }
 
-export function GetRunTimeDependenciesPackages(packageJSON: any): Package[] {
+export function GetRunTimeDependenciesPackages(packageJSON: any, extensionPath: string): PackageJSONPackage[] {
     if (packageJSON.runtimeDependencies) {
-        return JSON.parse(JSON.stringify(<Package[]>packageJSON.runtimeDependencies));
+        let dependencies = <PackageJSONPackage[]>packageJSON.runtimeDependencies;
+        return dependencies.map(pkg => new Package(pkg, extensionPath));
```
TheRealPiotrP: I don't get it. This method returns an array of `Package`s. So why is the return type an array of `PackageJSONPackage`?

src/omnisharp/OmnisharpDownloader.ts
```diff
@@ -42,6 +43,12 @@ export class OmnisharpDownloader {
         }
     }
 
+    private GetResolvedPackagesFromVersion(version: string, packageJSON: any, serverUrl: string, installPath: string, extensionPath: string) {
```
TheRealPiotrP: I think this should be `GetResolvedPackagesForVersion`

src/CSharpExtDownloader.ts
akshita31: Yes you are right. Will correct that.

src/omnisharp/OmnisharpPackageCreator.ts
```diff
@@ -3,24 +3,28 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Package } from "../packageManager/Package";
+import { PackageJSONPackage } from "../packageManager/PackageJSONPackage";
```
TheRealPiotrP: There should be exactly one component that knows/cares what a PackageJSONPackage is. Why do we need to look at the Package.json in two places?

akshita31:
I think we need to redo some things here for a better shape. Closing this for now.

Issue #2182, opened 2018-04-12
Why is it still "Preview"?


Timothy-Liu:
I'm new to VS Code and OmniSharp. I saw there is a red "Preview" tag in VS Code and the description "Welcome to the C# extension for Visual Studio Code! This preview provides the following features inside VS Code:" ...
May I ask a dummy question: if the version is already 1.14, why is it still preview?

Regards.

gregg-miskelly:
Sorry for the delay in responding. We talked about this internally and we decided that there is no reason this extension should still be considered a "preview". For our next release (1.15), we will remove the preview branding.

PR #2220, opened 2018-04-25
Mark the C# extension as non-preview

gregg-miskelly:
This changes the branding on the C# extension so that it is no longer labeled a 'preview'.

Two changes:
1. Change `preview` to `false` in package.json
2. Update the license that is used in official builds of the C# extension. This new EULA is no longer a pre-release EULA and it also has the latest text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 [![Wallaby.js](https://img.shields.io/badge/wallaby.js-configured-green.svg)](https://wallabyjs.com)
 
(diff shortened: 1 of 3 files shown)


gregg-miskelly:
This fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2182

DustinCampbell:
@TheRealPiotrP: I think there's some noise in our code coverage due to particular performance telemetry events not firing.

PR #2221, opened 2018-04-25
Ensure that OmniSharp is launched on global mono even when a version or 'latest' specifed

DustinCampbell:
While @rchande and I were investigating a different issue, we noticed that OmniSharp was always being launched with the run script when `"omnisharp.path"` is set to `"latest"`. Essentially, it would only try to launch OmniSharp on the globally-installed Mono if the user hasn't set `"omnisharp.path"` at all, which really isn't what we want.

Here are the desired behaviors for Linux/OSX:

1. If the user hasn't set `"omnisharp.path"` at all, launch the default OmniSharp. If a valid version of Mono is installed, launch with Mono; otherwise, use the run script.
2. If the user has set `"omnisharp.path"` to an absolute path, assume that path is OmniSharp and launch it. If the user also sets `"omnisharp.useMono"` to `true`, launch with Mono. If a valid version of Mono can't be found, throw an error.
3. If the user has set `"omnisharp.path"` to a version or `"latest"`, first, ensure the appropriate version of OmniSharp is downloaded. Then, if a valid version of Mono is installed, launch with Mono; otherwise, use the run script.

This change fixes `#3` above.

As part of this change, I've done a bit of refactoring to simplify things a bit. The OmniSharpManager has been changed to return two paths: the path to the run script, and the path to launch with Mono. In addition, the calculation of the path to the default OmniSharp has been pushed into the OmniSharpManager. The launcher code has also been simplified a bit. Finally, I changed the logging to only say "started with Mono" if OmniSharp was actually launched with Mono, and I added the version number of Mono.

### DIFF
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -325,7 +325,8 @@ export class OmniSharpServer {
         this.eventStream.post(new ObservableEvents.OmnisharpInitialisation(new Date(), solutionPath));
         this._fireEvent(Events.BeforeServerStart, solutionPath);
 
-        return launchOmniSharp(cwd, args, launchInfo).then(async launchResult => {
+        try {
+            let launchResult = await launchOmniSharp(cwd, args, launchInfo);
             this.eventStream.post(new ObservableEvents.OmnisharpLaunch(launchResult.monoVersion, launchResult.command, launchResult.process.pid));
 
             this._serverProcess = launchResult.process;
@@ -333,16 +334,15 @@ export class OmniSharpServer {
             this._setState(ServerState.Started);
             this._fireEvent(Events.ServerStart, solutionPath);
 
-            return this._doConnect();
-        }).then(() => {
-            // Start telemetry reporting
+            await this._doConnect();
+
             this._telemetryIntervalId = setInterval(() => this._reportTelemetry(), TelemetryReportingDelay);
-        }).then(() => {
             this._requestQueue.drain();
-        }).catch(async err => {
+        }
+        catch (err) {
             this._fireEvent(Events.ServerError, err);
             return this.stop();
-        });
+        }
     }
 
     private debounceUpdateProjectWithLeadingTrue = () => {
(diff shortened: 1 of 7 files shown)


src/omnisharp/server.ts
```diff
@@ -312,25 +312,23 @@ export class OmniSharpServer {
...
+            this.eventStream.post(new ObservableEvents.OmnisharpFailure(`Error occured in loading omnisharp from omnisharp.path\nCould not start the server due to ${error.toString()}`, error));
+            return;
         }
 
         this.eventStream.post(new ObservableEvents.OmnisharpInitialisation(new Date(), solutionPath));
         this._fireEvent(Events.BeforeServerStart, solutionPath);
 
-        return launchOmniSharp(cwd, args, launchPath).then(async value => {
-            this.eventStream.post(new ObservableEvents.OmnisharpLaunch(value.usingMono, value.command, value.process.pid));
+        return launchOmniSharp(cwd, args, launchInfo).then(async launchResult => {
```
akshita31: Can we use async await here as well ?

DustinCampbell: Sure

akshita31: Thanks!

DustinCampbell:
Hmmm... I"m not sure what's up with codecov. It seems to show lots of uncovered blocks in OmniSharpManager.ts, but those blocks are clearly covered by feature tests. @TheRealPiotrP, is this a known issue?

akshita31:
@DustinCampbell  We do not have code coverage for feature tests right now. I have created an issue and will look into it - https://github.com/OmniSharp/omnisharp-vscode/issues/2227

DustinCampbell:
Thanks @akshita31! Are you all ok if I go ahead and merge this in that case? The OmniSharpManager.ts file is really quite well-covered if feature tests are included.

akshita31:
@DustinCampbell Sure go ahead 😀

PR #2229, opened 2018-04-27
Remove platform information Get Current

akshita31:
Modified the code so that we get the platform information only once and use it wherever required. This makes the code more testable as we can pass a custom platform information to the function while testing.

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -44,7 +44,7 @@ export default function registerCommands(server: OmniSharpServer, eventStream: E
     let d7 = vscode.commands.registerCommand('dotnet.generateAssets', async () => generateAssets(server));
 
     // Register command for remote process picker for attach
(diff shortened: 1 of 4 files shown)


rchande:
Serious question: is this code more maintainable? We have to do a lot of plumbing up and down the stack in order to be able to use PlatformInfo now. Is there a way we could manipulate what GetPlatformInfo() returns in tests? 

Not saying this change is right or wrong, but interested in discussion.

TheRealPiotrP:
@rchande very fair point. The issue with the current implementation is that we're essentially using a dressed-up `global` to share state... along with all the fun that globals bring. 

I agree with you it's pretty anoying to have to plumb through this settings type. Globals are one way around that... but I wonder if there is some model that would, I don't know, [Inject the Dependency](https://github.com/young-steveo/bottlejs) only where we need it...?

PR #2237, opened 2018-04-30
Removing a possible race condition between the downloader and installer - To remove the `Immediate Zip File error` 

akshita31:
In the downloader we were piping the response to the output stream with `end: false` due to which we were not getting the writestream `finish` event which indicates when the response was completely written to the filestream. This led to the `Immediate Zip FIle Error` when using the Download and Install Pipeline as we didnt wait for the data to be completely put into the filestream and trying to unzip it created that error.

So in this PR, I have removed {end: false} condition and am resolving the promise of the dowloader when the filestream finish event is seen, so that there is no race condition.
Also instead of using open file descriptors, modified the code to use the file paths so there is no problem, even if the file doesnot have an open fd.

### DIFF
--- a/src/omnisharp/OmnisharpDownloader.ts
+++ b/src/omnisharp/OmnisharpDownloader.ts
@@ -49,7 +49,7 @@ export class OmnisharpDownloader {
         try {
             this.eventStream.post(new LatestBuildDownloadStart());
             tmpFile = await CreateTmpFile();
-            await DownloadFile(tmpFile.fd, description, this.eventStream, this.networkSettingsProvider, url);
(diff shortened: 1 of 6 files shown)


TheRealPiotrP:
Overall looks ok. I'm not convinced by the fd -> path change. Does this open us up to a bunch of issues?
- Perf issues due to multiple OS interactions?
- Race conditions? What prevents the file we used to have an fd for from being taken by other processes
- Special case of the above: downloaders/installers often fight with AntiVirus tools because the latter grabs locks on recently written files. Are we sure that this change won't open up that set of issues?

akshita31:
Closing this in favor of #2245

PR #2238, opened 2018-05-01
Installing the default omnisharp in a version folder

akshita31:
I have added a property defaults in the package json which will include the default omnisharp version being used. This will be used at runtime to launch the default omnisharp. Along with this change we can completely drop the "experimental" suffix for the "latest" omnisharp download


akshita31:
To Do: Also change test-plan.md

package.json
```diff
@@ -143,7 +146,7 @@
       "description": "OmniSharp for Windows (.NET 4.6 / x86)",
       "url": "https://download.visualstudio.microsoft.com/download/pr/11655913/afaf4c0f0bee0304c8b1c55dfc190a9f/omnisharp-win-x86-1.29.1.zip",
       "fallbackUrl": "https://omnisharpdownload.blob.core.windows.net/ext/omnisharp-win-x86-1.29.1.zip",
-      "installPath": ".omnisharp",
+      "installPath": ".omnisharp/1.29.1",
```
rchande: Can we now delete "installPath" and determine it at runtime by concatting defaults.omnisharp and ".omnisharp"?

src/omnisharp/server.ts
```diff
@@ -310,7 +310,7 @@ export class OmniSharpServer {
         let launchInfo: LaunchInfo;
         try {
             let extensionPath = utils.getExtensionPath();
-            launchInfo = await this._omnisharpManager.GetOmniSharpLaunchInfo(this._options.path, serverUrl, latestVersionFileServerPath, installPath, extensionPath);
+            launchInfo = await this._omnisharpManager.GetOmniSharpLaunchInfo(this.packageJSON.defaults.omniSharp,this._options.path, serverUrl, latestVersionFileServerPath, installPath, extensionPath);
```
rchande: spacing around comma

src/omnisharp/OmnisharpManager.ts
```diff
@@ -20,10 +20,10 @@ export class OmnisharpManager {
         private platformInfo: PlatformInformation) {
     }
 
-    public async GetOmniSharpLaunchInfo(omnisharpPath: string, serverUrl: string, latestVersionFileServerPath: string, installPath: string, extensionPath: string): Promise<LaunchInfo> {
+    public async GetOmniSharpLaunchInfo(defaultOmnisharpVersion: string, omnisharpPath: string, serverUrl: string, latestVersionFileServerPath: string, installPath: string, extensionPath: string): Promise<LaunchInfo> {
```
rchande: Consider pushing all of these parameters into some "omnisharpSettings" class to make their relationship clearer?

package.json
akshita31: We could do it for omnisharp but we use this for the debugger packages as well and at this point I think we should not try to diverge from this.

src/omnisharp/OmnisharpManager.ts
akshita31: Yes that will be a part of the next refactoring step that we will be taking up, where we will be introducing these settings classes.

PR #2239, opened 2018-05-01
A few small option description tweaks

DustinCampbell:


package.json
```diff
@@ -521,7 +521,7 @@
         "omnisharp.useMono": {
           "type": "boolean",
           "default": false,
-          "description": "Launch OmniSharp with Mono."
+          "description": "Launch OmniSharp with Mono. Note that this option is only valid when 'omnisharp.path' is set to an absolute path."
```
TheRealPiotrP: Is this not true if `path` is `latest` or a specific version number?

akshita31: From what I understand from the code we are trying to use the mono irrespective of whether the omnisharp.path is set and if set what is the value. The launcher as of now just cares that if the useMono is set then it will try to launch using mono: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/launcher.ts#L245

rchande: Maybe the docs should explicitly visualize the table we're talking about :)

DustinCampbell: @akshita31: Yes, there is essentially a bug that I think we should tweak here. Your PR to make the default OmniSharp use the same infrastructure as everything else actually makes it easier to fix that. I'll go ahead and do it.

DustinCampbell:
Closed in favor of #2244

PR #2244, opened 2018-05-01
Add/update 'omnisharp.useGlobalMono' option

DustinCampbell:
This change updates the name of the 'omnisharp.useMono' option to 'omnisharp.useGlobalMono', and makes it a tri-state value rather than a boolean. It now has three possible values:

* "auto": Launch OmniSharp on the globally-installed Mono if it's available.
* "always": Always try to launch OmniSharp on the globally-installed Mono and error if it's not available.
* "never": Never launch OmniSharp on the globall-installed Mono.

### DIFF
--- a/package.json
+++ b/package.json
@@ -521,10 +521,20 @@
           "default": null,
           "description": "Specifies the path to OmniSharp. This can be the absolute path to an OmniSharp executable, a specific version number, or \"latest\". If a version number or \"latest\" is specified, the appropriate version of OmniSharp will be downloaded on your behalf."
         },
-        "omnisharp.useMono": {
-          "type": "boolean",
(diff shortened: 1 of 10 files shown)


akshita31:
@DustinCampbell Since this is user-facing change should there be an update to the ChangeLog as well ?

DustinCampbell:
Yes there should. Thanks!

akshita31:
@DustinCampbell Should we also update the test-plan.md (https://github.com/OmniSharp/omnisharp-vscode/blob/master/test-plan.md#omnisharpusemono-for-linux) in accordance with these changes.

DustinCampbell:
@akshita31 : Yes, it needs updating.

PR #2245, opened 2018-05-01
Use a buffer instead of a temporary file for download and install

akshita31:
To avoid the conflicts that occur with using file streams to store the downloaded files, using a buffer instead.


src/packageManager/FileDownloader.ts
```diff
@@ -90,15 +89,12 @@ async function downloadFile(fd: number, description: string, urlString: string,
             });
 
             response.on('end', () => {
-                resolve();
+                resolve(Buffer.concat(buffers));
             });
 
             response.on('error', err => {
                 reject(new NestedError(`Reponse error: ${err.message || 'NONE'}`, err));
```
TheRealPiotrP: I hadn't noticed this before. This exception isn't super descriptive. Should it be somethigng like `Failed to download from ${uri}. Error Message: $err.message || 'NONE'}`?

src/packageManager/FileDownloader.ts
```diff
@@ -53,15 +54,13 @@ async function downloadFile(fd: number, description: string, urlString: string,
...
-            reject(new NestedError("Temporary package file unavailable"));
-        }
+    let buffers: any[] = [];
 
+    return new Promise<Buffer>((resolve, reject) => {
         let request = https.request(options, response => {
             if (response.statusCode === 301 || response.statusCode === 302) {
                 // Redirect - download from new location
-                return resolve(downloadFile(fd, description, response.headers.location, eventStream, networkSettingsProvider));
+                return resolve(downloadFile(description, response.headers.location, eventStream, networkSettingsProvider));
```
TheRealPiotrP: why on this invocation of `resolve` we return its return value, but on other invocations we ignore it?

akshita31: I am not sure about that. Do you think we need to change this?

PR #2255, opened 2018-05-03
Check for run script instead of mono run time

akshita31:
Since the path to the mono runtime binary has changed, we must check for the presence of the run script to determine whether the package exists or not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,7 +182,7 @@
         "./mono.osx",
         "./run"
       ],
-      "installTestPath": "./.omnisharp/mono.osx",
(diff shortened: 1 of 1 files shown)


akshita31:
@DustinCampbell Should we change the path for the mono binary - https://github.com/OmniSharp/omnisharp-vscode/pull/2255/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L182 to be `bin/mono.osx` ?

akshita31:
Closing in favor of : https://github.com/OmniSharp/omnisharp-vscode/pull/2256

PR #2257, opened 2018-05-03
Set the correct installBinary to test for

### LINKED ISSUE #2253: OmniSharp - "latest" is being downloaded on every restart
The install test path ends with mono.linux-x86 which is not present in the latest OmniSharp versions, hence the package is being downloaded on every restart of the server when the omnisharp path is set to "latest".

@TheRealPiotrP @DustinCampbell @rchande @colombod

DustinCampbell:
Fixes #2253


akshita31:
Super cool 💃

DustinCampbell:
@TheRealPiotrP: I'm not sure it makes sense to collapse the tests just because the expected file is the same. After all, they still test different platforms. Note that there are also two tests above these for Windows(x86) and Windows(x64) that also share the same assertion.

PR #2263, opened 2018-05-04
Display prompt to restart OmniSharp when the OmniSharp config changes

akshita31:
Listening to the workspace config changed event and responding by checking if the options have changed. If yes then  prompting the user to click the button to restart omnisharp. On clicking this button, omnisharp is restarted and the OmniSharp log channel is shown.

options1


akshita31:
@DustinCampbell 
1. For the multiple prompts I have added the debounce of 1.5 seconds.
2. What are the options for which we should display the prompt . Currently I have added 4 of them https://github.com/OmniSharp/omnisharp-vscode/pull/2263/files#diff-70fe6bcbbf03e1cfb59a12901ac73e18R50.
3. Modified the display message - https://github.com/OmniSharp/omnisharp-vscode/pull/2263/files#diff-70fe6bcbbf03e1cfb59a12901ac73e18R81

src/features/codeActionProvider.ts
```diff
@@ -13,26 +13,23 @@ import { Options } from '../omnisharp/options';
...
-        this._resetCachedOptions();
-
-        let d1 = vscode.workspace.onDidChangeConfiguration(this._resetCachedOptions, this);
-        let d2 = vscode.commands.registerCommand(this._commandId, this._runCodeAction, this);
-        this.addDisposables(new CompositeDisposable(d1, d2));
-    }
-
-    private _resetCachedOptions(): void {
         this._options = Options.Read(vscode);
+        let configChangedDisposable = new Disposable(optionStream.subscribe(options => this._options = options));
```
rchande: Do we actually need to make this a disposable now? Are we not allowed to dispose the underlying observable if there are still observers?

rchande:
@akshita31 I think the biggest change in this PR is the idea that we're putting our options themselves into an obvservable and that features that care about options will use this observable, not the VS Code options API. I actually think that's a really cool idea, because we have a ton of repeated code where features try to figure out the most appropriate way to handle options changing over the liftetime of the extension.

How would you feel about breaking this into 2 PRs:
1. that changes all of our existing options use to use this new system
2. that adds the "options changed, consider restarting omnisharp" observer?

I noticed in your existing changes that there are still features that call options.read(), but I bet we can get rid of all of those in favor of subscribing to the new options observable. Thoughts?

Overall I really like the concept here but I think there's still some cleanup in order to make everything as consistent as possibel

src/features/codeActionProvider.ts
```diff
@@ -13,26 +13,23 @@ import { Options } from '../omnisharp/options';
...
 
-        this._resetCachedOptions();
-
-        let d1 = vscode.workspace.onDidChangeConfiguration(this._resetCachedOptions, this);
-        let d2 = vscode.commands.registerCommand(this._commandId, this._runCodeAction, this);
-        this.addDisposables(new CompositeDisposable(d1, d2));
-    }
-
-    private _resetCachedOptions(): void {
         this._options = Options.Read(vscode);
```
rchande: I suspect that features that use the options observable shouldn't also call options.read().

src/features/commands.ts
```diff
@@ -54,7 +54,9 @@ export default function registerCommands(server: OmniSharpServer, eventStream: E
     return new CompositeDisposable(d1, d2, d3, d4, d5, d6, d7, d8, d9, d10);
 }
 
-function restartOmniSharp(server: OmniSharpServer) {
+function restartOmniSharp(server: OmniSharpServer, eventStream: EventStream) {
+    eventStream.post(new ShowOmniSharpChannel());
```
rchande: Arguably, we shouldn't show the message if o# isn't running.

src/features/commands.ts
```diff
@@ -54,7 +54,9 @@ export default function registerCommands(server: OmniSharpServer, eventStream: E
...
+    //show the channel when restarting omnisharp
```
rchande: Shouldn't deciding whether to show the channel be a feature of the server component?

akshita31:
@rchande I poked around a bit and realised that there are two types of classes where Options.Read() is being used:
1. That store the options and use it at some point in time like - codeActionProvider , codeLensProvider. These care about the options being changed and listen to the `onDidChangeConfiguration` to reset their cached options.
2.  Once that dont store the options anywhere and invoke the Options.Read whenever they need the options. Examples of this type are launcher.ts, server.ts. 

From what I understand as of now, we definitely need the first type to subscribe to and use the OptionObservable that we are trying to create here, but I am not so sure as to if we need the second type to do that as well. They just read the options when they need to without caring about any previous history. Thoughts ?

src/observers/WarningMessageObserver.ts
```diff
@@ -46,12 +35,41 @@ export class WarningMessageObserver {
...
     private handleOmnisharpServerMsBuildProjectDiagnostics(event: OmnisharpServerMsBuildProjectDiagnostics) {
-        if (!this.disableMsBuildDiagnosticWarning() && event.diagnostics.Errors.length > 0) {
+        let options = this.readOmniSharpOptions();
+        if (!options.disableMSBuildDiagnosticWarning  && event.diagnostics.Errors.length > 0) {
             this.warningMessageDebouncer.next(event);
         }
     }
+
+    private async handleWorkspaceConfigurationChanged() {
+        let newOptions = this.readOmniSharpOptions();
```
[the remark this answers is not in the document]
akshita31: Yes that is right. I am working on that now.

rchande:
@akshita31 Here's how I would interpret features and their relationship to options:
1. Features that want to use the most up-to-date value of an option when they are invoked. The fact that `codeActionProvider` caches the value and listens for the changed event is an implementation detail--we just don't want to read the option every time VS Code calls the `codeActionProvider`, which could be very often. The actual feature just wants to use the most recent value. `server.ts` also wants the most recent value but no one bothered to cache the value because the feature doesn't run as often. 
I think this first category is both of the things you said.

2. Features that want to reason about changes between options. I think the only feature we have in this category is your "restart O#" prompt.

I'm not an rx expert like you (😄) but I think that if our internal representation of options was an observable, we could do something like this:
* Category (1) receives the Options observable and stores it internally as a LastObservable of itself. These features always just read the single value of the LastObservable and use those options when the are invoked. If we do this, we can delete all the individual feature code that tries to subscribe to option changes.
* Category (2) is implemented as an Observer of the Options observable and it does the right rx thing to be able to handle diffing option changes.

What do you think? I'm sure I messed up some rx details, but that's the general shape I'm imagining. My two goals are:
* Avoid having 2 separate options subsystems
* Avoid having to duplicate "on VS Code options change" functionality in multiple features

rchande:
@akshita31 After reading more about rx, I guess category (1) would receive some object that implements observer, subscribes to the options observable with .last(), and exposes a method/property that allows clients to see that value. I think :)

src/main.ts
```diff
@@ -42,6 +47,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
     util.setExtensionPath(extension.extensionPath);
 
     const eventStream = new EventStream();
+    let optionStream = new BehaviorSubject<Options>(Options.Read(vscode));
+    let optionChangedStream = new Subject<void>();
```
rchande: Can we take this a step further and put the options onto the same message bus as the rest of our notifications?

src/main.ts
```diff
@@ -97,16 +105,19 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
...
-    let omniSharpPromise = OmniSharp.activate(context, eventStream, extension.packageJSON, platformInfo, networkSettingsProvider);
+    let omniSharpPromise = OmniSharp.activate(context, eventStream, extension.packageJSON, platformInfo, networkSettingsProvider, optionStream);
 
     // register JSON completion & hover providers for project.json
     context.subscriptions.push(addJSONProviders());
     context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(() => {
         eventStream.post(new ActiveTextEditorChanged());
     }));
 
+    context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(e => optionChangedStream.next()));
```
rchande: For clarity,  I would move this setup code to where you create the subject earlier in file.

src/main.ts
```diff
@@ -42,6 +47,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
     util.setExtensionPath(extension.extensionPath);
 
     const eventStream = new EventStream();
+    let optionStream = new BehaviorSubject<Options>(Options.Read(vscode));
+    let optionChangedStream = new Subject<void>();
+    optionChangedStream.debounceTime(1500).subscribe(() => { optionStream.next(Options.Read(vscode)); });
```
rchande: Why do we need optionStream and optionChangedStream? Can't we set the debounce on optionStream and directly push it when we hear the event?

src/observers/OmniSharpConfigChangedObserver.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { vscode } from '../vscodeAdapter';
+import 'rxjs/add/operator/debounceTime';
+import { Options } from '../omnisharp/options';
+import showWarningMessage from './ShowWarningMessage';
+
+let options: Options;
+
+export async function WorkspaceConfigurationChangedHandle(vscode: vscode, newOptions: Options) {
+    if (options != null) {
```
rchande: Instead of trying to store the last option you see, I would hook this up to a "pairiwise" so you can compare each value to the previous one https://www.learnrxjs.io/operators/combination/pairwise.html

src/observers/OmniSharpConfigChangedObserver.ts
```diff
@@ -0,0 +1,27 @@
...
+        if (options.path != newOptions.path
+            || options.useGlobalMono != newOptions.useGlobalMono
+            || options.waitForDebugger != newOptions.waitForDebugger
+            || options.projectLoadTimeout != newOptions.projectLoadTimeout) {
+            let message = "OmniSharp configuration has changed. Would you like to relaunch the OmniSharp server with your changes?";
+            await showWarningMessage(vscode, message, { title: "Restart Now", command: 'o.restart' });
```
rchande: Is "warning" the appropriate presentation here? Is there an "info" presentation or something?

src/omnisharp/server.ts
```diff
@@ -90,7 +92,7 @@ export class OmniSharpServer {
     private updateProjectDebouncer = new Subject<ObservableEvents.ProjectModified>();
     private firstUpdateProject: boolean;
 
-    constructor(private vscode: vscode, networkSettingsProvider: NetworkSettingsProvider, private eventStream: EventStream, private packageJSON: any, private platformInfo: PlatformInformation) {
+    constructor(private vscode: vscode, networkSettingsProvider: NetworkSettingsProvider, private eventStream: EventStream, private packageJSON: any, private platformInfo: PlatformInformation, private optionStream: Observable<Options>) {
```
rchande: Couldn't you pass in the BehaviorSubject you created earlier?

src/omnisharp/server.ts
```diff
@@ -415,15 +417,18 @@ export class OmniSharpServer {
...
-            const options = Options.Read(this.vscode);
-
+            let options = await this.readOptions();
             await this._start(launchTarget, options);
         }
     }
 
-    public autoStart(preferredPath: string): Thenable<void> {
-        return findLaunchTargets().then(async launchTargets => {
+    private readOptions(): Promise<Options>{
```
rchande: Let's try to avoid having to copy-paste this into every feature :)

akshita31:
@rchande As per your advice, I am breaking up this PR into separate parts for clarity, the first of which I am working on in - https://github.com/OmniSharp/omnisharp-vscode/pull/2273/files

PR #2272, opened 2018-05-08
Warning message observer clean up

akshita31:
Just a small clean up of the warning message observer and the tests


src/observers/WarningMessageObserver.ts
```diff
@@ -20,7 +21,7 @@ export class WarningMessageObserver {
         this.warningMessageDebouncer = new Subject<BaseEvent>();
         this.warningMessageDebouncer.debounceTime(1500, scheduler).subscribe(async event => {
             let message = "Some projects have trouble loading. Please review the output for more details.";
-            await showWarningMessage(this.vscode, message, { title: "Show Output", command: 'o.showOutput' });
+            await ShowWarningMessage(this.vscode, message, { title: "Show Output", command: 'o.showOutput' });
```
TheRealPiotrP: if we're doing this, is there any further need to take a vscode into this observer? Why not inject a ShowWarningMessage implementaiton that already has its vscode reference?

PR #2273, opened 2018-05-08
Using option stream to get the latest options

akshita31:
Create an OptionStream class that maintains the value of the latest options and has a subscription to the vscode- `onDidiChangeConfiguration` event.

For the classes like codeActionProvider that need to cache the value of the options we are subscribing to the option stream and for methods like restart where we just care about the latest value of the options we are exposing a `GetLatestOptions` function that returns the latest value of the options present in the option stream.

Note: After this PR, except the OptionStream.ts there is only one file that invokes the 'Options.Read' which is the main.ts for the `disableMSBuildDiagnosticWarning` option for the WarningMessageObserver. Changing that to use the optionStream will be done in a separate PR for clarity.


src/observables/OptionStream.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public post() {
+        this.optionChangeDebouncer.next();
+    }
+
+    public subscribe(observer: (options: Options) => void): Disposable {
+        return new Disposable(this.optionStream.subscribe(observer));
+    }
+
+    public async GetLatestOptions(): Promise<Options> {
```
rchande: I would just call this `Options()`

src/features/codeActionProvider.ts
```diff
@@ -13,30 +13,24 @@ import { Options } from '../omnisharp/options';
...
-    constructor(server: OmniSharpServer) {
+    constructor(server: OmniSharpServer, optionStream: OptionStream) {
         super(server);
 
         this._commandId = 'omnisharp.runCodeAction';
-
-        this._resetCachedOptions();
-
-        let d1 = vscode.workspace.onDidChangeConfiguration(this._resetCachedOptions, this);
+        let d1 = optionStream.subscribe(options => this._options = options);
```
rchande: Remove this and call optionStream.GetLatestOptions()

src/features/codeActionProvider.ts
```diff
@@ -13,30 +13,24 @@ import { Options } from '../omnisharp/options';
...
         this.addDisposables(new CompositeDisposable(d1, d2));
     }
 
-    private _resetCachedOptions(): void {
-        this._options = Options.Read(vscode);
-    }
-
     public async provideCodeActions(document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext, token: vscode.CancellationToken): Promise<vscode.Command[]> {
-        if (this._options.disableCodeActions) {
+        if (this._options && this._options.disableCodeActions) {
```
rchande: Remove field and call optionStream

src/features/codeLensProvider.ts
```diff
@@ -29,19 +30,13 @@ export default class OmniSharpCodeLensProvider extends AbstractProvider implemen
...
     private _options: Options;
 
-    constructor(server: OmniSharpServer, testManager: TestManager) {
+    constructor(server: OmniSharpServer, testManager: TestManager, optionStream: OptionStream) {
         super(server);
 
-        this._resetCachedOptions();
-
-        let configChangedDisposable = vscode.workspace.onDidChangeConfiguration(this._resetCachedOptions, this);
+        let configChangedDisposable = optionStream.subscribe(options => this._options = options);
```
rchande: Remove this, store field of optionStream, call optionStream.GetLatestOptions() where applicable

src/observables/OptionStream.ts
rchande: I don't think this needs to be promise or async. It looks like there's always synchronously a value in the BehaviorSubject?

src/observables/OptionStream.ts
```diff
@@ -0,0 +1,34 @@
...
+import Disposable from "../Disposable";
+
+export default class OptionStream {
+    private optionStream: BehaviorSubject<Options>;
+    private optionChangeDebouncer: Subject<void>;
+
+    constructor(vscode: vscode) {
+        this.optionStream = new BehaviorSubject<Options>(Options.Read(vscode));
+        this.optionChangeDebouncer = new Subject<void>();
+        this.optionChangeDebouncer.debounceTime(1000).subscribe(() => { this.optionStream.next(Options.Read(vscode)); });
```
rchande: Is the subscription to the options change event coming in a later PR? Might make sense to include that in this PR so we can see that whole flow.

src/omnisharp/extension.ts
```diff
@@ -34,21 +33,22 @@ import { EventStream } from '../EventStream';
...
-export async function activate(context: vscode.ExtensionContext, eventStream: EventStream, packageJSON: any, platformInfo: PlatformInformation, provider: NetworkSettingsProvider) {
+export async function activate(context: vscode.ExtensionContext, packageJSON: any, platformInfo: PlatformInformation, provider: NetworkSettingsProvider, eventStream: EventStream, optionStream: OptionStream) {
     const documentSelector: vscode.DocumentSelector = {
         language: 'csharp',
         scheme: 'file' // only files from disk
     };
 
-    const options = Options.Read(vscode);
-    const server = new OmniSharpServer(vscode, provider, eventStream, packageJSON, platformInfo);
+    const options = await optionStream.GetLatestOptions();
```
rchande: I don't think this is used?

src/observables/OptionStream.ts
rchande: Oh, I see you did it below.
If this class takes a vscode in its constructor, it should handle all the event subscribing itself.

src/omnisharp/extension.ts
akshita31: It is used below in the file

src/observables/OptionStream.ts
akshita31: take(1) returns an `Observable(Options)` and we need to convert it to a Promise in order to be able to get the options.

src/observables/OptionStream.ts
akshita31: We are adding the subscription disposable to the set of disposables of the extension so that the resources are freed up when the vscode window closes. Hence I think it will be better to keep the subscription to the `vscode.onDidChangeConfiguration` inside the main.ts only where other similar subscriptions are present. Thoughts ?

src/features/codeActionProvider.ts
```diff
@@ -9,34 +9,26 @@ import AbstractProvider from './abstractProvider';
...
-    constructor(server: OmniSharpServer) {
+    constructor(server: OmniSharpServer, private optionStream: OptionStream) {
         super(server);
 
         this._commandId = 'omnisharp.runCodeAction';
-
-        this._resetCachedOptions();
-
-        let d1 = vscode.workspace.onDidChangeConfiguration(this._resetCachedOptions, this);
         let d2 = vscode.commands.registerCommand(this._commandId, this._runCodeAction, this);
```
DustinCampbell: Maybe give `d2` a better name?

src/features/codeActionProvider.ts
```diff
@@ -9,34 +9,26 @@ import AbstractProvider from './abstractProvider';
...
-        this.addDisposables(new CompositeDisposable(d1, d2));
-    }
-
-    private _resetCachedOptions(): void {
-        this._options = Options.Read(vscode);
+        this.addDisposables(new CompositeDisposable(d2));
```
DustinCampbell: Since there's just a single `Disposable`, does it need to be wrapped in a `CompositeDisposable`?

src/main.ts
```diff
@@ -98,15 +100,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
...
     // activate language services
-    let omniSharpPromise = OmniSharp.activate(context, eventStream, extension.packageJSON, platformInfo, networkSettingsProvider);
+    let omniSharpPromise = OmniSharp.activate(context, extension.packageJSON, platformInfo, networkSettingsProvider, eventStream, optionStream);
 
     // register JSON completion & hover providers for project.json
     context.subscriptions.push(addJSONProviders());
     context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(() => {
         eventStream.post(new ActiveTextEditorChanged());
     }));
+    context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(e => optionStream.post()));
```
DustinCampbell: Maybe add a line break before this to make sure it doesn't look like the "// register JSON completion" comment applies to this as well?

src/features/codeActionProvider.ts
akshita31: The addDisposables takes a CompositeDisposable as an argument

src/omnisharp/server.ts
```diff
@@ -438,7 +438,7 @@ export class OmniSharpServer {
                         watcher.dispose();
                         resolve();
                     });
-                }).then(() => {
+                }).then(async () => {
```
rchande: Why async? We're not waiting in that lambda?

test/unitTests/optionStream.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    test("Gives the latest options if there are changes in omnisharp config", () => {
+        let changingConfig = "omnisharp";
+        updateConfig(vscode, changingConfig, 'path', "somePath");
+        listenerFunction.forEach(listener => listener(getConfigChangeEvent(changingConfig)));
+        let options = optionStream.Options();
+        expect(options.path).to.be.equal("somePath");
+    });
+
+    test("Gives the latest options if there are changes in omnisharp config", () => {
```
rchande: "csharp config"

test/unitTests/optionStream.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        let changingConfig = 'csharp';
+        updateConfig(vscode, changingConfig, 'omnisharpUsesMono', true);
```
rchande: is this an option in the C# config?

src/observables/OptionStream.ts
```diff
@@ -0,0 +1,42 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Options } from "../omnisharp/options";
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
+import { vscode } from "../vscodeAdapter";
+import 'rxjs/add/operator/take';
+import Disposable, { IDisposable } from "../Disposable";
+
+export default class OptionStream {
```
rchande: The API that this class isn't really stream-y. Do we really need to have "stream" in the name?

rchande: Should we break this into 2 concepts? One thing is the "optionstream" that subscribes to the VS Code option system and and acts like an Observable. There are then 2 observers: one provides the Options() functionality currently described in this file, and the options changed observer you're going to add later?

src/observables/OptionStream.ts
```diff
@@ -0,0 +1,34 @@
...
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
+import Disposable, { IDisposable } from "../Disposable";
+import { Subject } from "rxjs/Subject";
+
+export default class OptionStream {
+    private optionStream: Subject<Options>;
+    private disposable: IDisposable;
+
+    constructor(vscode: vscode) {
+        this.optionStream = new BehaviorSubject<Options>(Options.Read(vscode));
```
rchande: I thought you were making this into a regular Subject?

akshita31: An subject wont retain the most recent value for the observers that subscribe to it after the value is pushed into the subject. Hence we need to use the behavior subject here.

PR #2303, opened 2018-05-15
Codecov unit test coverage for all files

akshita31:


akshita31:
The drop in coverage from unit tests seems an indicator that the change works. Though I am not 100% sure.

TheRealPiotrP:
help me understand, @akshita31. Why would the drop in coverage indicate improvement? We're not seeing new files added whose coverage is 0. We are just seeing existing files lose their coverage...

akshita31:
@TheRealPiotrP Yes, I think you are right. I think I need a better understanding of what that "all" option implies. Closing this for now.

PR #2316, opened 2018-05-16
Display prompt [Information Message]  to restart the server when the omnisharp configuration changes

akshita31:
Added an option change observer that subscribes to an Observable<Options> and shows the information message when there is a change in the configuration the server cares about and then restarts the server when "Restart" is clicked. Added unit tests for each of the components involved.

### DIFF
--- a/src/observables/CreateOptionStream.ts
+++ b/src/observables/CreateOptionStream.ts
@@ -10,7 +10,7 @@ import 'rxjs/add/operator/publishBehavior';
 import { Observable } from "rxjs/Observable";
 import { Observer } from "rxjs/Observer";
 
-export function createOptionStream(vscode: vscode): Observable<Options> {
+export default function createOptionStream(vscode: vscode): Observable<Options> {
     return Observable.create((observer: Observer<Options>) => {
         let disposable = vscode.workspace.onDidChangeConfiguration(e => {
             //if the omnisharp or csharp configuration are affected only then read the options
--- a/src/observers/OptionChangeObserver.ts
+++ b/src/observers/OptionChangeObserver.ts
@@ -1,0 +1,32 @@
+/*--------------------------------------------------------------------------------------------- 
+ *  Copyright (c) Microsoft Corporation. All rights reserved. 
+ *  Licensed under the MIT License. See License.txt in the project root for license information. 
+ *--------------------------------------------------------------------------------------------*/
+
+import OptionStream from "../observables/OptionStream";
+import { vscode } from "../vscodeAdapter";
+import { Options } from "../omnisharp/options";
+import ShowInformationMessage from "./utils/ShowInformationMessage";
+
+export function ShowOmniSharpConfigHasChanged(optionStream: OptionStream, vscode: vscode) {
+    let options: Options;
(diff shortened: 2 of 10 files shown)


akshita31:
@DustinCampbell Are there any other options that we should consider here ?

akshita31:
@colombod This test is so cool. Thanks for the idea 🥇

DustinCampbell:
Where is the code where we decide which options should cause us to prompt an OmniSharp restart? Right now, it looks like we'll prompt on any change to the 'omnisharp' or 'csharp' configurations.

DustinCampbell:
@akshita31; Thanks! I was having trouble spotting that code for some reason. Looks good to me!

colombod:
@DustinCampell the magic of rx!

src/observables/CreateOptionStream.ts
```diff
@@ -0,0 +1,24 @@
...
+export default function createOptionStream(vscode: vscode): Observable<Options> {
+    return Observable.create((observer: Observer<Options>) => {
+        let disposable = vscode.workspace.onDidChangeConfiguration(e => {
+            //if the omnisharp or csharp configuration are affected only then read the options
+            if (e.affectsConfiguration('omnisharp') || e.affectsConfiguration('csharp')) {
+                observer.next(Options.Read(vscode));
+            }
+        });
+
+        return () => disposable.dispose();
```
rchande: This and the line below are beyond my rx-fu, can you explain them?

src/observers/OptionChangeObserver.ts
```diff
@@ -0,0 +1,42 @@
...
+        options = newOptions;
+        return changed;
+    });
+}
+
+export function ShowOmniSharpConfigChangePrompt(optionObservable: Observable<Options>, vscode: vscode): Disposable {
+    let subscription = ConfigChangeObservable(optionObservable)
+        .subscribe(_ => {
+            let message = "OmniSharp configuration has changed. Would you like to relaunch the OmniSharp server with your changes?";
+            ShowInformationMessage(vscode, message, { title: "Restart Now", command: 'o.restart' });
```
rchande: I would have the button say "Restart OmniSharp"

src/observers/OptionChangeObserver.ts
```diff
@@ -0,0 +1,42 @@
...
+        });
+
+    return new Disposable(subscription);
+}
+
+function hasChanged(oldOptions: Options, newOptions: Options): boolean {
+    if (oldOptions.path != newOptions.path ||
```
rchande: Nit, you could write this `return ... `

src/observables/CreateOptionStream.ts
akshita31: publishBehavior is used to create a behavior observable which has the initial value as Option.Read(). refCount() is used so that the dispose function on the observable is called only when all the observers who susbcribed to this observable invoke unsubscribe. This test verifies this behavior https://github.com/OmniSharp/omnisharp-vscode/pull/2316/files#diff-dd9d0618b9edb503c00be5f531debc27R78

PR #2317, opened 2018-05-16
Update diagnostics when focusing an editor (DiagnosticsProvider is lazy)

SirIntruder:
I noticed I often have to insert random whitespace for vscode to update error reports, so I checked what extension is doing. 

Currently, extension updates diagnostics on two events: _onDidChangeTextDocument_ and _onDidOpenTextDocument_. _onDidOpenTextDocument_ is called only when first openning an editor for a file [(docs)](https://code.visualstudio.com/docs/extensionAPI/vscode-api). If you modify a file "A.cs":

- diagnostics inside changed file are updated almost instantly, because a document validation request was sent for "A.cs" 
- If project is below HugeProjectLimit, project-wide validation will be requested as well, so changes made in "A.cs" will propagate to other open editors ("B.cs", "C.cs" etc) with a couple of seconds of latency. depending on the size of the project. 
- If your project is above huge project limit, tough luck, outside of the modified file you have to make a random change (to invoke _onDidChangeTextDocument_ ) to see updated diagnostics

This PR would add two vscode events: _onDidChangeActiveTextEditor_ and _onDidChangeWindowState_, and call the same pathway as _onDidOpenTextDocument_/_onDidChangeTextDocument_  when user focuses an editor. This both fixes issue with huge projects (without any project wide check, just updates info for the newly focused document) and makes diagnostics as snappy as possible. 

Additional requests this PR would introduce are already invoked on every keystroke anyway, so server load should not increase meaningfully.


akshita31:
@SirIntruder Thanks for your contribution. I pulled your branch and realized that because we have the `onDidChangeWindowState` every time we change the window - opening, closing , minimise, resize , etc we are doing the diagnostics request. Do you have a specific scenario in mind where the `onDidChangeActiveTextEditor` would not work and `onDidChangeWindowState` would be helpful?
Thoughts @DustinCampbell ?

SirIntruder:
I added onDidChangeWindowState as support for alt tabbing - that is useful for when you make changes from outside vscode and return, in my case usually git pull. I checked focus flag so it should not send request on minimizing or closing, but I was not aware it is invoked on resize as well. I could add a tracking flag so it would only be sent when focused is actually changed from false to true, would that be ok?

akshita31:
@SirIntuder I set a breakpoint on https://github.com/OmniSharp/omnisharp-vscode/pull/2317/files#diff-f7f698ea628a7fd476ec3cd2355e1927R163 and it is being hit on minimize and closing as well. Can you try that?

SirIntruder:
It looks like I always get two onDidChangeWindowState events when I debug from vscode - always one windowState.focused === false, one windowState.focused === true. It looks like it is just issue with focus jumping to debugger and back. When I turned breakpoint into logpoint, and logged windowState.focused, it worked as you would expect (true only when returning to focus.)

I didn't debug this before, I compiled it with tsc and was printing out some logs to see what is going on - and according to logs it works like it should (focused == true only when returning to focus).

akshita31:
@SirIntruder Thanks for the analysis. Looks good to me.

DustinCampbell:
Thanks very much for the contribution @SirIntruder!

PR #2318, opened 2018-05-16
Clear the OmniSharp channel before restarting the server

akshita31:
Clear the `Omnisharp log` channel when we restart the server to avoid having to navigate through logs from the previous start.

A relevant unit test also added to the OmnisharpChannelObserver

### DIFF
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -321,7 +321,6 @@ export class OmniSharpServer {
 
         try {
             let launchResult = await launchOmniSharp(cwd, args, launchInfo, this.platformInfo, options);
-            this.eventStream.post(new ObservableEvents.ShowOmniSharpChannel());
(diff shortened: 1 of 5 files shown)


src/omnisharp/server.ts
```diff
@@ -321,6 +321,7 @@ export class OmniSharpServer {
 
         try {
             let launchResult = await launchOmniSharp(cwd, args, launchInfo, this.platformInfo, options);
+            this.eventStream.post(new ObservableEvents.ShowOmniSharpChannel());
```
DustinCampbell: This would show the OmniSharp log everytime the user launches C# for VS Code? I think that would be very annoying. I would recommend against that behavior.

akshita31:
Understood your concern. Removed the show thing. Now this PR just clears the omnisharp channel whenever there is a restart.

filipw:
thank you - this is a very welcome change!

Issue #2276, opened 2018-05-09
Add option to restore all projects in solution


akshita31:
In reference to : https://github.com/OmniSharp/omnisharp-vscode/issues/240

I propose the following change:
.NET: Restore All Projects : Restore all the projects in the workspace.
.NET Restore Project: Shows a command palette menu to select one of the projects to be restored.

How does that sound ?

Also we have a 'There are unresolved dependencies..' warning that pops up for every project that has an unresolved dependency. How do we change that? It is irritating especially when there are a number of projects in the solution. Should we modify that to come up when there is atleast one project that has an unresolved dependency 
`There are unresolved dependencies in the solution. Please execute the restore command to continue.`
and then when the user clicks restore we perform the 'Restore all projects' as stated above ?

akshita31:
Please share your thoughts @rchande @TheRealPiotrP @DustinCampbell

rchande:
@akshita31 I'm on board for having two commands for restoring projects. I don't have strong opinions about the names of those commands. I completely agree with your solution for the warning. It's annoying because we show one notification per project. The clear solution (as you said) is to only show one notification that offers to restore all projects. I'm having trouble imagining scenarios where users would want to only restore a subset of projects in response to the warning.

DustinCampbell:
FYI that there's also this: https://github.com/OmniSharp/omnisharp-vscode/issues/240

Issue #240, opened 2016-04-22
Add separate command for 'dotnet restore' on a specific project


DustinCampbell:
Currently the `dotnet:Restore Packages` command forces the user to select a specific project for which to restore packages. Instead, the command should just restore packages for all projects. A separate command should be added to restore a specific project.

akshita31:
This has been resolved in #2323.

PR #2323, opened 2018-05-17
Display prompt to restore all projects

akshita31:
Fixes: #2276 #240 

The prompt that asks to restore the project dependencies appears just once and clicking on "Restore" , restores all the projects.
The command palette has 2 options:
1. .NET : Restore Project - Displays a menu to select one project to be restored.
2. .NET : Restore Packages - Restores all the projects.


akshita31:
I am not sure what should be the names of the commands in the command pallette.

package.json
```diff
@@ -607,7 +608,12 @@
...
       },
       {
-        "command": "dotnet.restore",
+        "command": "dotnet.restore.project",
+        "title": "Restore Project",
+        "category": ".NET"
+      },
+      {
+        "command": "dotnet.restore.all",
         "title": "Restore Packages",
```
DustinCampbell: Should we rename this to "Restore All Projects" for symmetry with "Restore Project"?

akshita31: Done.

src/observers/InformationMessageObserver.ts
```diff
@@ -20,10 +20,11 @@ export class InformationMessageObserver {
     }
 
     private async handleOmnisharpServerUnresolvedDependencies(event: ObservableEvent.OmnisharpServerUnresolvedDependencies) {
+        //to do: determine if we need the unresolved dependencies message
         let csharpConfig = this.vscode.workspace.getConfiguration('csharp');
         if (!csharpConfig.get<boolean>('suppressDotnetRestoreNotification')) {
-            let message = `There are unresolved dependencies from '${this.vscode.workspace.asRelativePath(event.unresolvedDependencies.FileName)}'. Please execute the restore command to continue.`;
-            return showInformationMessage(this.vscode, message, { title: 'Restore', command: 'dotnet.restore', args: event.unresolvedDependencies.FileName });
+            let message = `There are unresolved dependencies'. Please execute the restore command to continue.`;
```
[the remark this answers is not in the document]
DustinCampbell: Maybe. If we do, the text should probably be `in abc.sln` rather than `from abc.sln`.

However, I'm interested in the semantics of how this message appears. If several "unresolved dependencies" events trigger for different projects, will we just display a single message? If so, would it be better if we triggered `dotnet restore` for each project rather than running it over the whole workspace? Doing the latter might be unexpected.

src/features/commands.ts
```diff
@@ -24,35 +24,29 @@ export default function registerCommands(server: OmniSharpServer, platformInfo:
...
     let d3 = vscode.commands.registerCommand('o.showOutput', () => eventStream.post(new ShowOmniSharpChannel()));
-    let d4 = vscode.commands.registerCommand('dotnet.restore', fileName => {
-        if (fileName) {
-            dotnetRestoreForProject(server, fileName, eventStream);
-        }
-        else {
-            dotnetRestoreAllProjects(server, eventStream);
-        }
-    }); 
+    let d4 = vscode.commands.registerCommand('dotnet.restore.project', () => pickProjectAndDotnetRestore(server, eventStream));
```
rchande: This is getting a bit silly. Can we just start pushing these into an array and construct our composite disposable from that?

src/features/commands.ts
```diff
@@ -120,58 +114,40 @@ function projectsToCommands(projects: protocol.ProjectDescriptor[], eventStream:
...
-        });
-    });
 }
 
-export async function dotnetRestoreForProject(server: OmniSharpServer, filePath: string, eventStream: EventStream) {
+async function dotnetRestoreAllProjects(server: OmniSharpServer, eventStream: EventStream): Promise<void> {
+    let descriptors = await getProjectDescriptors(server);
+    eventStream.post(new CommandDotNetRestoreStart());
+    for (let descriptor of descriptors) {
+        await dotnetRestore(descriptor.Directory, eventStream);
```
rchande: What does the log output look like while we restore multiple projects?

akshita31: It shows the logs from the restore of each project
restore

src/observers/InformationMessageObserver.ts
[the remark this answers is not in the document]
rchande: I think one problem is that discovery of unrestored projects is asynchronous. If we implement the feature so that it only restores projects for which it received the "unrestored project" message, there's a race condition between this popup and the discovery. Eg, one unrestored project is discovered and the warning is shown, and the user clicks restore. O# continues discovering projects and finds more, triggering a new prompt that the user needs to click to restore the newly discovered projects. 

Restoring the entire sln file at once avoids that race condition. I suppose restoring the whole solution could be slow, but what else is "unexpected" about it? 

There's also the case of multiple csproj's floating around with no sln file. I'm not sure how you would coordinate restoring all of those projects simulatenously without an sln...

rchande:
@DustinCampbell @akshita31 Do we still have outstanding concerns about this PR or is it ready to merge? 😄

DustinCampbell:
None from me.

PR #2326, opened 2018-05-18
Add type info in integration tests

akshita31:
1. Add type information in integration test
2. initialisationFinished is a function so we should await that
3. remove logging from spawnGit.

### DIFF
--- a/test/integrationTests/testAssets/spawnGit.ts
+++ b/test/integrationTests/testAssets/spawnGit.ts
@@ -25,8 +25,8 @@ export default async function spawnGit(args?: string[], options?: SpawnOptions):
     
     let spawned = spawn('git', args, optionsWithFullEnvironment);
 
-    spawned.stdout.on('data', (data) => console.log(data.toString()));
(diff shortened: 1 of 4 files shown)


test/integrationTests/testAssets/spawnGit.ts
```diff
@@ -25,8 +25,8 @@ export default async function spawnGit(args?: string[], options?: SpawnOptions):
     
     let spawned = spawn('git', args, optionsWithFullEnvironment);
 
-    spawned.stdout.on('data', (data) => console.log(data.toString()));
-    spawned.stderr.on('data', (data) => console.log(data.toString()));
+    //spawned.stdout.on('data', (data) => console.log(data.toString()));
+    //spawned.stderr.on('data', (data) => console.log(data.toString()));
```
DustinCampbell: Did you mean to leave this commented or should it just be deleted?

akshita31: I have left if commented. It might be useful to enable it if sometimes we need to debug.

PR #2347, opened 2018-05-28
Using a class for TestZip , MockHttpServer

akshita31:
Refactoring of the unit tests


test/unitTests/testAssets/TestZip.ts
```diff
@@ -4,30 +4,28 @@
...
+        return this._buffer.length;
+    }
 
+    get files(): TestFile[]{
+        return this._files;
+    }
+}
 
-export async function createTestZipAsync(filesToAdd: Array<{ content: string, path: string }>): Promise<Buffer> {
+export async function createTestZipAsync(...filesToAdd: TestFile[]): Promise<TestZip> {
```
rchande: I think you could make this a static method on TestZip and avoid having to import TestZip and this method in every file

akshita31: Done. Thanks!

PR #2349, opened 2018-05-28
Show the red omnisharp flame when there is std error in omnisharp

akshita31:


src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -3,13 +3,14 @@
...
 
-import { OmnisharpServerOnServerError, BaseEvent, OmnisharpOnBeforeServerInstall, OmnisharpOnBeforeServerStart, OmnisharpServerOnStop, OmnisharpServerOnStart, DownloadStart, InstallationStart, DownloadProgress } from "../omnisharp/loggingEvents";
+import { OmnisharpServerOnServerError, BaseEvent, OmnisharpOnBeforeServerInstall, OmnisharpOnBeforeServerStart, OmnisharpServerOnStop, OmnisharpServerOnStart, DownloadStart, InstallationStart, DownloadProgress, OmnisharpServerOnStdErr } from "../omnisharp/loggingEvents";
 import { BaseStatusBarItemObserver } from './BaseStatusBarItemObserver';
 
 export class OmnisharpStatusBarObserver extends BaseStatusBarItemObserver {
     public post = (event: BaseEvent) => {
         switch (event.constructor.name) {
             case OmnisharpServerOnServerError.name:
+            case OmnisharpServerOnStdErr.name:
```
rchande: Do we only get the OnStdErr event when starting omnisharp? The message on line 14 makes it sound like that.

akshita31: It is produced when the standard error stream on the omnisharp process produces an error. Modified the case to show that message instead

src/observers/OmnisharpStatusBarObserver.ts
```diff
@@ -12,6 +12,10 @@ export class OmnisharpStatusBarObserver extends BaseStatusBarItemObserver {
             case OmnisharpServerOnServerError.name:
                 this.SetAndShowStatusBar('$(flame)', 'o.showOutput', 'rgb(218,0,0)', 'Error starting OmniSharp');
                 break;
+            case OmnisharpServerOnStdErr.name:
+                let msg = (<OmnisharpServerOnStdErr>event).message;
+                this.SetAndShowStatusBar('$(flame)', 'o.showOutput', 'rgb(218,0,0)', `OmniSharp process errored:${msg}`);
```
rchande: Can you extract out constants for the colors and call them names like "red" and "green"?

test/unitTests/logging/OmnisharpStatusBarObserver.test.ts
```diff
@@ -27,17 +27,31 @@ suite('OmnisharpStatusBarObserver', () => {
...
-    test('OnServerError: Status bar is shown with the error text', () => {
-        let event = new OmnisharpServerOnServerError("someError");
+    [
+        new OmnisharpServerOnServerError("someError"),
+    ].forEach((event: BaseEvent) => {
+        test(`${event.constructor.name}: Status bar is shown with the error text`, () => {
+            observer.post(event);
+            expect(showCalled).to.be.true;
+            expect(statusBarItem.text).to.equal(`$(flame)`);
+            expect(statusBarItem.command).to.equal('o.showOutput');
```
rchande: Verify color?

PR #2360, opened 2018-06-01
Implement VS Code syntax-based folding API

rchande:
Uses https://github.com/OmniSharp/omnisharp-roslyn/pull/1209


src/features/structureProvider.ts
```diff
@@ -0,0 +1,23 @@
+import { FoldingRangeProvider, TextDocument, FoldingContext, CancellationToken, FoldingRange } from "vscode";
+import AbstractSupport from './abstractProvider';
+import { blockStructure } from "../omnisharp/utils";
+import { Request } from "../omnisharp/protocol";
+
+export class structureProvider extends AbstractSupport implements FoldingRangeProvider
```
akshita31: Should this be in capital ?

DustinCampbell: I agree

src/omnisharp/protocol.ts
```diff
@@ -623,6 +624,18 @@ export namespace V2 {
...
+    {
+        Spans: BlockSpan[]
+    }
+
+    export interface BlockSpan {
+        TextSpan: Range;
+        HintSpan: Range;
+        BannerText: string;
+        Type: string;
+    }
```
DustinCampbell: This protocol should match other OmniSharp endpoints better. It'd probably better to have just a single `Range` property rather than all of the other properties that are ultimately unused.

DustinCampbell: `BlockStructureRequest` appears to be missing

src/features/structureProvider.ts
rchande: Will fix. Also--it looks like many other classes start with Capital letters but have filenames that start with lowercase letters. Should we try to make this consistent or is this some typescriptism?

src/omnisharp/protocol.ts
rchande: @DustinCampbell This is something the extension codebase is inconsistent about: for request types that derive Request but don't add any properties, we haven't been adding corresponding entries in protocol.ts (instead just preferring to use Request directly) (case in point: there's no FileMembersAsTree request type in the extension). I was following that pattern here. We could also switch to adding request-deriving types that don't add properties. What do you think?

src/omnisharp/launcher.ts
```diff
@@ -228,6 +228,7 @@ async function launch(cwd: string, args: string[], launchInfo: LaunchInfo, platf
         args.push(`formattingOptions:useTabs=${!getConfigurationValue(globalConfig, csharpConfig, 'editor.insertSpaces', true)}`);
         args.push(`formattingOptions:tabSize=${getConfigurationValue(globalConfig, csharpConfig, 'editor.tabSize', 4)}`);
         args.push(`formattingOptions:indentationSize=${getConfigurationValue(globalConfig, csharpConfig, 'editor.tabSize', 4)}`);
+        args.push('-z')
```
DustinCampbell: I don't think you want this. If you launch OmniSharp with `--zero-based-indeces`, you'll need to update all of the places in C# for VS Code that expect 1-based indeces today.

src/omnisharp/protocol.ts
```diff
@@ -623,6 +624,16 @@ export namespace V2 {
...
+
+    export interface BlockStructureResponse
+    {
+        Spans: BlockSpan[]
+    }
+
+    export interface BlockSpan {
+        Range: Range;
+        Type: string;
+    }
```
DustinCampbell: I think you're missing your `BlockStructureRequest` type.

src/omnisharp/utils.ts
```diff
@@ -19,6 +19,10 @@ export async function currentFileMembersAsTree(server: OmniSharpServer, request:
     return server.makeRequest<protocol.CurrentFileMembersAsTreeResponse>(protocol.Requests.CurrentFileMembersAsTree, request, token);
 }
 
+export async function blockStructure(server: OmniSharpServer, request: protocol.Request, token: vscode.CancellationToken) {
```
DustinCampbell: In the protocol you defined on the OmniSharp side, this takes a `BlockStructureRequest`, which inherits from `SimpleFileRequest` rather than `Request`.

src/features/structureProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+            return FoldingRangeKind.Comment
+        }
+        else if (type == "Imports")
+        {
+            return FoldingRangeKind.Imports
+        }
+        else if (type == "Region")
+        {
+            return FoldingRangeKind.Region
+        }
```
DustinCampbell: switch?

src/features/structureProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class StructureProvider extends AbstractSupport implements FoldingRangeProvider
+{
+    async provideFoldingRanges(document: TextDocument, context: FoldingContext, token: CancellationToken): Promise<FoldingRange[]> {
+        let request: Request = {
+            FileName: document.fileName,
+        }
+        
+        let response = await blockStructure(this._server, request, token)
+        let ranges : FoldingRange[] = [];
```
DustinCampbell: nit: format this file? This colon is in a different spot than it is above.

src/features/structureProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+        for (let member of response.Spans)
+        {
+            ranges.push(new FoldingRange(member.Range.Start.Line - 1, member.Range.End.Line - 1, this.GetType(member.Type)))
+        }
+
+        return ranges;
+    }
+
+    GetType(type: string) : FoldingRangeKind
```
DustinCampbell: `GetKind`?

DustinCampbell:
When will we update OmniSharp in VS Code so that this works without specifying `"omnisharp.path": "latest"`?

rchande:
@DustinCampbell let's try to do it this week--I've requested opinions in the o# slack.

PR #2364, opened 2018-06-05
Add snapshot flag to attach options.

nbilling:
@WardenGnaw @gregg-miskelly @chuckries

### DIFF
--- a/src/tools/OptionsSchema.json
+++ b/src/tools/OptionsSchema.json
@@ -463,6 +463,11 @@
           "default": {
             "*": { "enabled": true }
           }
+        },
+        "useSnapshotDebugger": {
+          "type": "boolean",
+          "description": "If true then Snapshot Debugger will be used rather than live debugger.",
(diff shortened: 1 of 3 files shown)


src/tools/OptionsSchema.json
```diff
@@ -463,6 +463,11 @@
           "default": {
             "*": { "enabled": true }
           }
+        },
+        "useSnapshotDebugger": {
+          "type": "boolean",
+          "description": "If true then Snapshot Debugger will be used rather than live debugger.",
```
gregg-miskelly: Do we want this to be a user-exposed option? If not, there are better places to put this. If so, we need a better description?

nbilling: User's shouldn't have to mess with this manually. Let's discuss where else to put it.

nbilling:
There's a better way to do this, Gregg and I spoke offline.

PR #2365, opened 2018-06-05
Use omnisharp download settings class

akshita31:


akshita31:
I think we need a better refactoring than this. Closing this for now.

PR #2371, opened 2018-06-07
Use the correct install test path for omnisharp in package.json

akshita31:
We recently introduced the version folder concept for omnisharp default download but forgot to modify the test path accordingly, hence fixing that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,7 @@
       "architectures": [
         "x86"
       ],
-      "installTestPath": "./.omnisharp/OmniSharp.exe",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -151,7 +151,7 @@
       "architectures": [
         "x86"
       ],
-      "installTestPath": "./.omnisharp/OmniSharp.exe",
+      "installTestPath": "./.omnisharp/1.30.1/OmniSharp.exe",
```
rchande: There are 21 instances of the string "1.30.1" in this file. We need a better approach. Thoughts?

PR #2385, opened 2018-06-20
WIP - Separate the "Package" into two interfaces [ Do not review]

akshita31:
The "Package" interface currently is very confusing and error prone as the same interface is used to read packages from packageJSON where there are relative paths and same is used in the download pipeline with resolved absolute paths. 

This PR defines three interfaces:
1. IPackage : A generic interface that includes description, url , and other properties not involving paths.
2. RuntimeDependency : The packages as they appear in packageJSON(with relative installPath, installTestPath, binaries)
3. InstallablePackage : The package where the relative paths in the runtimeDependency are converted to absolute paths.


akshita31:
Closing this in favor of #2403

PR #2388, opened 2018-06-21
Preselect feature in the completion item

akshita31:


src/features/completionItemProvider.ts
```diff
@@ -66,6 +66,8 @@ export default class OmniSharpCompletionItemProvider extends AbstractSupport imp
                 completion.commitCharacters = response.IsSuggestionMode
                     ? OmniSharpCompletionItemProvider.CommitCharactersWithoutSpace
                     : OmniSharpCompletionItemProvider.AllCommitCharacters;
+                
+                completion.preselect = response.Preselect;
```
rchande: It'd be cool if we could test this somehow...

package.json
```diff
@@ -269,7 +269,7 @@
     }
   ],
   "engines": {
-    "vscode": "^1.18.0"
+    "vscode": "*"
```
rchande: As discussed let's not merge this until the feature moves into the latest stable build.

TheRealPiotrP:
This PR drops 30% of our code coverage. What's going on?

akshita31:
@TheRealPiotrP The integration tests are failing because the desired vscode version has not been released. Once that is in place, we can get the real coverage analysis.

rchande:
@akshita31 @TheRealPiotrP in https://github.com/OmniSharp/omnisharp-vscode/pull/2360 I changed the version of vs code used to run the integration tests. You might need to do something similar to get tests passing.

akshita31:
Blocked on : https://github.com/Microsoft/vscode/issues/53749

SirIntruder:
Did latest vscode version fix the issue?

package.json
```diff
@@ -269,7 +269,7 @@
     }
   ],
   "engines": {
-    "vscode": "^1.23.0"
+    "vscode": "^1.26.1"
```
DustinCampbell: Note that this will mean that users will need to install at least 1.26.1 in order to install C# for VS Code. Is that intentional? Does this API only appear in 1.26.1? Or, is it in an earlier version (e.g. 1.26.0).

PR #2393, opened 2018-06-27
tiny typo in debugger-launchjson.md

qayshp:
Remove a stray asterisk.

Does not look intentional, and does not work as it was.


gregg-miskelly:
@qayshp Thanks for fixing this!

PR #2403, opened 2018-07-02
Using absolute path and installable package classes

akshita31:
At a number of places in the code, we require that the paths that are associated with the packages must be absolute. However the package interface imposed no such restrictions which made the download and install pipeline error prone. 
Hence added an `InstallablePackage` class that requires that the `installPath`, `installTestPath` and the `binaries` must have absolute and not relative paths.


src/packageManager/InstallablePackage.ts
```diff
@@ -0,0 +1,59 @@
...
+function getAbsoluteBinaries(pkg: Package, extensionPath: string): AbsolutePath[] { 
+    let basePath = getAbsoluteInstallPath(pkg, extensionPath).value;
+    if (pkg.binaries) { 
+        return pkg.binaries.map(value => AbsolutePath.getAbsolutePath(basePath, value)); 
+    } 
+ 
+    return null; 
+} 
+ 
+function getAbsoluteInstallPath(pkg: Package, extensionPath: string): AbsolutePath {
```
rchande: extensionPath is kind of a confusing name. I think this is the base directory of the extension (or something?) but it winds up being the `pathToPrepend` despite being called extension?

src/packageManager/Package.ts
```diff
@@ -3,15 +3,17 @@
...
     fallbackUrl?: string;
-    installPath?: string;
     platforms: string[];
     architectures: string[];
-    binaries: string[];
     platformId?: string;
-    // Path to use to test if the package has already been installed
+}
+
+export interface Package extends IPackage{
```
rchande: `RelativePathPackage`?

src/packageManager/InstallablePackage.ts
```diff
@@ -0,0 +1,59 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IPackage, Package } from "./Package";
+import { AbsolutePath } from "./AbsolutePath";
+
+export class InstallablePackage implements IPackage{
```
rchande: `AbsolutePathPackage`?

src/packageManager/AbsolutePath.ts
```diff
@@ -0,0 +1,18 @@
...
+import { isAbsolute, resolve } from "path";
+
+export class AbsolutePath{
+    constructor(public value: string) {
+        if (!isAbsolute(value)) {
+            throw new Error("The path must be absolute");
+        }
+    }
+
+    public static getAbsolutePath(pathToPrepend: string, path: string): AbsolutePath {
```
rchande: thoughts on making the signature here just look like the signature of path.resolve?

src/packageManager/Package.ts
akshita31: I am not sure we should do that. The paths here can be absolute or relative there is no specific restriction that we are imposing right(like isRelative(path) or something). Thoughts ?

src/packageManager/AbsolutePath.ts
akshita31: Done

src/packageManager/InstallablePackage.ts
akshita31: Done

PR #2377, opened 2018-06-11
Use new `/codestructure` endpoint from OmniSharp

DustinCampbell:
Fixes #2132
Fixes #2192
Fixes #2307
Fixes #2329

This change takes advantage of the new `/codestructure` endpoint in two providers: the `CodeLensProvider` and the `DocumentSymbolProvider`.

* `CodeLensProvider` has been overhauled. The logic to compute the references and test code lenses has been unified. In addition, there was an issue with using `await` in a loop within the `provideCodeLenses` method that could cause the method to return to VS Code before the code lenses were actually added to the result array.
* The `DocumentSymbolProvider` gets a lot of benefit from this change because the `/codestructure` endpoint includes all types and members, and not just a strange handful:

    image

PR #2410, opened 2018-07-04
Integration test for the implementation provider

akshita31:


test/integrationTests/implementationProvider.test.ts
```diff
@@ -0,0 +1,33 @@
...
+import CSharpExtensionExports from "../../src/CSharpExtensionExports";
+import * as path from "path";
+import testAssetWorkspace from "./testAssets/testAssetWorkspace";
+import { expect } from "chai";
+
+suite(CSharpImplementationProvider.name, () => {
+    let fileUri: vscode.Uri;
+    
+    suiteSetup(async () => {
+        let csharpExtension = vscode.extensions.getExtension<CSharpExtensionExports>("ms-vscode.csharp");
```
rchande: I think all our tests share this boilerplate. Can we find a way to share it?

test/integrationTests/testAssets/singleCsproj/implementation.cs
```diff
@@ -0,0 +1,7 @@
+using System;
+
+namespace minimal
```
rchande: You can leave this out if you really want...

test/integrationTests/implementationProvider.test.ts
akshita31: @DustinCampbell has already added the helper methods for the integration tests
https://github.com/OmniSharp/omnisharp-vscode/pull/2377. Once that change is in, I can modify the tests here to use that.

PR #2425, opened 2018-07-19
Add option to run with a custom mono installation

shana:
👋 

When `useGlobalMono` is set to `always` or `auto`, the launcher uses whatever mono it finds in the path, which may or may not be a usable mono version for omnisharp. The user may have a suitable mono installation for omnisharp somewhere else, but not set as the default system one, or they may want to have omnisharp launched with a specific mono for omnisharp for debugging purposes. In my case, I have:

```
drwxr-xr-x  4.2.2.30
drwxr-xr-x  4.6.2.16
drwxr-xr-x  4.8.1.0
drwxr-xr-x  5.12.0.226_1
drwxr-xr-x  5.14.0.169_1
drwxr-xr-x  5.4.1.6
drwxr-xr-x  5.8.1.0_1
```

and I switch between them often, depending on what I need to do. Having vscode tied to whatever happens to be currently set on my system is really brittle, and if I need to have vscode use a particular mono version because of some mono bug, it's hard.

This PR adds a `monoPath` configuration option, and changes the launcher so that the environment variables PATH and MONO_GAC_PREFIX are changed to include the `monoPath` value, if set, when launching mono. If the `monoPath` option isn't set, it behaves as normal (also if `useGlobalMono` is set to `never`). I figured maybe this is useful for other people out there too 😃

### DIFF
--- a/src/omnisharp/launcher.ts
+++ b/src/omnisharp/launcher.ts
@@ -199,6 +199,7 @@ export interface LaunchResult {
     process: ChildProcess;
     command: string;
     monoVersion?: string;
+    monoPath?: string;
 }
 
 export async function launchOmniSharp(cwd: string, args: string[], launchInfo: LaunchInfo, platformInfo: PlatformInformation, options: Options): Promise<LaunchResult> {
@@ -250,12 +251,12 @@ async function launch(cwd: string, args: string[], launchInfo: LaunchInfo, platf
 
         const launchPath = launchInfo.MonoLaunchPath || launchInfo.LaunchPath;
 
-        return launchNixMono(launchPath, monoVersion, cwd, args, childEnv);
+        return launchNixMono(launchPath, monoVersion, options.monoPath, cwd, args, childEnv);
     }
 
     // If we can launch on the global Mono, do so; otherwise, launch directly;
     if (options.useGlobalMono === "auto" && isValidMonoAvailable && launchInfo.MonoLaunchPath) {
-        return launchNixMono(launchInfo.MonoLaunchPath, monoVersion, cwd, args, childEnv);
+        return launchNixMono(launchInfo.MonoLaunchPath, monoVersion, options.monoPath, cwd, args, childEnv);
     }
     else {
         return launchNix(launchInfo.LaunchPath, cwd, args);
@@ -312,7 +313,7 @@ function launchNix(launchPath: string, cwd: string, args: string[]): LaunchResul
     };
 }
 
-function launchNixMono(launchPath: string, monoVersion: string, cwd: string, args: string[], environment: NodeJS.ProcessEnv): LaunchResult {
+function launchNixMono(launchPath: string, monoVersion: string, monoPath: string, cwd: string, args: string[], environment: NodeJS.ProcessEnv): LaunchResult {
     let argsCopy = args.slice(0); // create copy of details args
     argsCopy.unshift(launchPath);
     argsCopy.unshift("--assembly-loader=strict");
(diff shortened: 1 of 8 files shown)


akshita31:
@shana Thank you for taking time to contribute to the repo.
I notice that the integration tests are failing somehow. The timeout suggests that there is a problem with starting the "OmniSharp" process. Did you try running the integration tests on your machine?
You could run them by using `npm run test:integration`.

rchande:
@shana Thanks for your contribution! Looks good to me, once tests pass.

shana:
Thanks for the review! Looks like the return value from the settings file was null instead of undefined, which meant the block overriding the env vars was always running, and I imagine screwing up the environment for spawning omnisharp in the process. I ran the integration tests locally with an update default system mono with no monoPath override, and with an older system mono and the monoPath override, and everything passes locally.

Can't really add much coverage for the integration tests though, not without bundling mono just for that purpose :/

src/omnisharp/launcher.ts
```diff
@@ -306,14 +312,15 @@ function launchNix(launchPath: string, cwd: string, args: string[]): LaunchResul
     };
 }
 
-function launchNixMono(launchPath: string, monoVersion: string, cwd: string, args: string[]): LaunchResult {
+function launchNixMono(launchPath: string, monoVersion: string, cwd: string, args: string[], environment: NodeJS.ProcessEnv): LaunchResult {
     let argsCopy = args.slice(0); // create copy of details args
     argsCopy.unshift(launchPath);
     argsCopy.unshift("--assembly-loader=strict");
```
akshita31: Should we add some logging here, like printing the mono launch path, just like we print the version?

We create the log event here from the launch result: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/server.ts#L324. We might also return a monoLaunchPath in the launch result and display it here: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/observers/OmnisharpLoggerObserver.ts#L58. We will have to add the appropriate property to `OmnisharpLaunch` event here: https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/loggingEvents.ts#L30.
Basically to do logging in omnisharp-vscode, we post the events into an event stream to which some observers subscribe. Whenever we post an event to the event stream, all the observers are invoked and take action appropriately. Does that make sense ?

shana: Yeah, I hacked up some manual logging directly to a separate window just so I could see what was going on while I was putting this together, but adding some more information to the event stream probably makes sense, I can give that a try.

src/omnisharp/server.ts
```diff
@@ -321,7 +321,7 @@ export class OmniSharpServer {
 
         try {
             let launchResult = await launchOmniSharp(cwd, args, launchInfo, this.platformInfo, options);
-            this.eventStream.post(new ObservableEvents.OmnisharpLaunch(launchResult.monoVersion, launchResult.command, launchResult.process.pid));
+            this.eventStream.post(new ObservableEvents.OmnisharpLaunch(launchResult.monoVersion, launchResult.command, launchResult.process.pid, options.monoPath));
```
akshita31: I think it would be better if the LaunchResult returned the monoLaunchPath(if any) that it used.

shana: Oh yes, that makes way more sense, not sure how I missed that. ☕️ ☕️ ☕️

shana:
@akshita31 Logging fixed! 😄

akshita31:
@shana Thank you so much for patiently going through the feedback.
If you like, you can do a changelog update in a separate PR, otherwise I will do that.
Thanks again for taking time to make the repo better!

shana:
@akshita31 Sorry, I didn't see your comment. Yay merge! I was not aware of the need to update the changelog manually, how does that work? It's just adding a new entry for the change?

akshita31:
@shana Yes that is right.

PR #2448, opened 2018-07-31
Using the in box omnisharp for integration tests

akshita31:


DustinCampbell:
How is this the "in box" OmniSharp. The version listed in the package.json is `1.30.1`

akshita31:
I initially started with that, but then later we switched to the more recent versions. Apologies for the confusion.

DustinCampbell:
FWIW, PR titles are editable. :smile:

PR #2451, opened 2018-08-03
1.16.0-beta1

akshita31:


DustinCampbell:
Does this beta version require setting `"omnisharp.path": "latest"`? If so, that's going to be a hiccup for anyone trying it. Could we update the default OmniSharp first?

rchande:
@akshita31 Can you also do the thing Dustin said?

PR #2460, opened 2018-08-15
Update CDN links

rchande:
cc @DustinCampbell @akshita31 @gregg-miskelly


DustinCampbell:
I downloaded and checked each archive. They appear to have the correct folder structure.

PR #2465, opened 2018-08-17
Use the fsPath instead of path in the virtual document tracker

akshita31:
The virtual document tracker was sending a file change request with "path" instead of fsPath. The path was not a valid file system path and was causing omnisharp to crash when a deleted file was being opened in the git side-bar( https://github.com/OmniSharp/omnisharp-roslyn/pull/1252#issuecomment-413059744).

### DIFF
--- a/src/features/virtualDocumentTracker.ts
+++ b/src/features/virtualDocumentTracker.ts
@@ -63,7 +63,7 @@ function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): 
 }
 
 async function openVirtualDocument(document: TextDocument, server: OmniSharpServer, eventStream: EventStream) {
-    let req = { FileName: document.uri.path, changeType: FileChangeType.Create };
+    let req = { FileName: document.uri.fsPath, changeType: FileChangeType.Create };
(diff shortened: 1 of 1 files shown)


rchande:
cc @NTaylorMullen

src/features/virtualDocumentTracker.ts
```diff
@@ -63,7 +63,7 @@ function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer):
 }
 
 async function openVirtualDocument(document: TextDocument, server: OmniSharpServer, eventStream: EventStream) {
-    let req = { FileName: document.uri.path, changeType: FileChangeType.Create };
+    let req = { FileName: document.uri.fsPath, changeType: FileChangeType.Create };
```
NTaylorMullen: You need to fallback to `.path`. Aka,

```TypeScript
let path = document.uri.fsPath;

if (!path) {
    path = document.uri.path;
}
```

Same below.

akshita31: Done. Thanks!

Issue #2414, opened 2018-07-06
Adopt improved DocumentSymbolProvider API


jrieken:
With the 1.25 release we have finalised the API for the [outline tree](https://code.visualstudio.com/updates/v1_25#_outline-view). A `DocumentSymbolProvider` should now return [document symbols](https://github.com/Microsoft/vscode/blob/bd953248ad6a8e2f8896bfb78b3ae40e0fd47426/src/vs/vscode.d.ts#L2474) which can represent a hierarchy and that can differentiate between the full range of  a symbol and the selection range, e.g. the name of a function. 

*Note* that we synthesise trees using range-containment when a provider returns `SymbolInformation`-objects and that that isn't perfect. Adopting the new API makes the outline more robust and support features like follow-cursor and diagnostics association. 

Let me know if need more information is needed. Happy Coding!

jrieken:
It will show with the label but only when selected. Like so: 
<img width="197" alt="screen shot 2018-09-06 at 08 53 08" src="https://user-images.githubusercontent.com/1794099/45140188-4d102980-b1b2-11e8-84eb-a49c303be6e7.png">

akshita31:
@jrieken I adopted the new API in this PR: https://github.com/OmniSharp/omnisharp-vscode/pull/2490. However the integration test is failing. It is still returning the symbol information type and when i query on the "detail", I am getting undefined. Any idea what might be happening here ?

jrieken:
> t is still returning the symbol information type and when i query on the "detail",

Are you using the api command in that test?

akshita31:
@jrieken This is [the test](https://github.com/OmniSharp/omnisharp-vscode/pull/2490/files#diff-3914e9c63d0b3334f21cb14aba14f661R43). I am using : 
`<vscode.DocumentSymbol[]>await vscode.commands.executeCommand("vscode.executeDocumentSymbolProvider", fileUri);` to get the document symbols. The symbols are as expected h but they are of type SymbolInformation.

jrieken:
Yeah, there is currently a backwards compatibility challenge. It's tracked here and will be tackled this month: https://github.com/Microsoft/vscode/issues/57984

akshita31:
From @DustinCampbell 
[...]

PR #2490, opened 2018-09-05
Return document symbols from the documentSymbolProvider

akshita31:
Fixes: #2414


src/features/documentSymbolProvider.ts
```diff
@@ -83,3 +74,9 @@ function toSymbolKind(kind: string): vscode.SymbolKind {
 
     return kinds[kind];
 }
+
+function toVscodeRange(range: protocol.V2.Range):  vscode.Range{
```
rchande: I think we might have an implementation of this somewhere else already

src/features/documentSymbolProvider.ts
```diff
@@ -14,43 +14,34 @@ import SymbolRangeNames = protocol.V2.SymbolRangeNames;
...
-        element.DisplayName,
-        toSymbolKind(element.Kind),
-        parentElementName,
-        new vscode.Location(vscode.Uri.file(fileName), vscodeRange)
-    );
+function createSymbolForElement(element: Structure.CodeElement): vscode.DocumentSymbol {
+    const fullRange = element.Ranges[SymbolRangeNames.Full];
+    const nameRange = element.Ranges[SymbolRangeNames.Name];
+    
+    return new vscode.DocumentSymbol(element.DisplayName, JSON.stringify(element.Properties), toSymbolKind(element.Kind), toVscodeRange(fullRange), toVscodeRange(nameRange));
```
rchande: Can you share an example of where this gets presented visually so we can see how the detail is shown? I don't think you want a json string.

src/features/documentSymbolProvider.ts
akshita31: Flavors of this one are present, not exactly this one. However I added this method to that file itself. Now there are three methods that are essentially doing the same thing. Do you think we should refactor a bit here ?

src/features/documentSymbolProvider.ts
akshita31: detail

The detail is shown in the outline view. And yes you are right we should not put a json string. Should we use the filename instead ?

akshita31:
Ok, adding the filename doesn't make it look good.
filename

Any suggestions ?

src/features/documentSymbolProvider.ts
```diff
@@ -25,32 +26,23 @@ export default class OmnisharpDocumentSymbolProvider extends AbstractSupport imp
...
-        element.DisplayName,
-        toSymbolKind(element.Kind),
-        parentElementName,
-        new vscode.Location(vscode.Uri.file(fileName), vscodeRange)
-    );
+function createSymbolForElement(element: Structure.CodeElement, filename: string): vscode.DocumentSymbol {
+    const fullRange = element.Ranges[SymbolRangeNames.Full];
+    const nameRange = element.Ranges[SymbolRangeNames.Name];
+    
+    return new vscode.DocumentSymbol(element.DisplayName, filename, toSymbolKind(element.Kind), toRange3(fullRange), toRange3(nameRange));
```
DustinCampbell: Is this still hierarchical? Do we need to populate the [`children`](https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts#L2569-L2572) property on [`DocumentSymbol`](https://code.visualstudio.com/docs/extensionAPI/vscode-api#DocumentSymbol)? Or, does that happen automatically passed on the ranges that are passed in?

rchande:
@oscarg933 Thanks for taking the time to review this PR. It looks like you marked it as "approved" several times without leaving any comments. Was that intentional? Did you have feedback you wanted to leave?

src/features/documentSymbolProvider.ts
```diff
@@ -11,10 +11,11 @@ import * as vscode from 'vscode';
 import Structure = protocol.V2.Structure;
 import SymbolKinds = protocol.V2.SymbolKinds;
 import SymbolRangeNames = protocol.V2.SymbolRangeNames;
+import { toRange3 } from '../omnisharp/typeConvertion';
```
rchande: Can we rename "typeConvertion" -> typeConversion? I'm tired of seeing that typo

PR #2535, opened 2018-09-17
Add generic telemetry event

akshita31:
cc @NTaylorMullen


test/unitTests/logging/TelemetryObserver.test.ts
```diff
@@ -60,6 +60,14 @@ suite('TelemetryReporterObserver', () => {
         });
     });
 
+    test(`${TelemetryEvent.name}: SendTelemetry event is called with the name, properties and measures`, () => {
```
TheRealPiotrP: since this will be used in some cases with raw json events, can you add one test that validates that scenario? @NTaylorMullen  & co can then use that test as a template for their work.

akshita31: @TheRealPiotrP The properties and events that I have provided in the above test are json objects only. Can you give an example of the data that you are talking about here ?

Issue #2521, opened 2018-09-13
Red OmniSharp flame icon in status bar with "OmniSharp process errored:" tooltip (1.16.1)


zvinless:
## Environment data
`dotnet --info` output:
```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.402
 Commit:    3599f217f4

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.13-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.1.402/

Host (useful for support):
  Version: 2.1.4
  Commit:  85255dde3e

.NET Core SDKs installed:
  2.1.402 [/usr/local/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.4 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.4 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.4 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download
```

VS Code version: 1.27.1
C# Extension version: 1.16.1

## Steps to reproduce
1. Create a new project: `dotnet new console`
1. Open containing folder with Visual Studio Code
1. Observe that the OmniSharp Log appears in the OUTPUT pane and focuses itself automatically (happens every time I open a C# project since 1.16.0)
1. Observe that the OmniSharp flame icon in the taskbar is red instead of green
1. Mouse over the icon and observe "OmniSharp process errored:" tooltip
1. Observe that there are no errors in the output from step 3

## Expected  behavior
The extension should load normally without opening the OUTPUT pane in VS Code. The OmniSharp flame icon should be green as it was prior to version 1.16.0.

## Actual behavior
<img width="254" alt="statusbar" src="https://user-images.githubusercontent.com/24303763/45467069-98569a80-b6d2-11e8-9be4-76219dca35c8.png">

Upon loading a C# project, the OUTPUT pane opens up with the message (this is with log level set to "debug"; usually I have it set to "critical" and it still focuses itself upon opening a project):
```
Starting OmniSharp server at 9/12/2018, 9:19:49 PM
    Target: /Users/zach/Desktop/OmniTest

[...]

PR #2540, opened 2018-09-18
Remove the bom from the data

akshita31:
Fixes: #2521 

We were getting the red flame because of a mono behavior, where if we set the encoding to UTF8 then a BOM(Byte Object Mark) is pushed into the stream (output as well as error) as the first character. In the extension we were listening to this input on the error stream and making the flame red.

So the workaround is to remove the "BOM" from the buffer and firing the "STDERR" event only if the length of the BOM is greater than 0.

This will also remove the stray character in the "OmniSharp" log that appears when using mono on Linux/Mac.
screenshot from 2018-09-18 15-20-36

### DIFF
--- a/src/omnisharp/server.ts
+++ b/src/omnisharp/server.ts
@@ -29,8 +29,8 @@ import CompositeDisposable from '../CompositeDisposable';
 import Disposable from '../Disposable';
 import OptionProvider from '../observers/OptionProvider';
 import { IMonoResolver } from '../constants/IMonoResolver';
-const removeBomBuffer = require("remove-bom-buffer");
-const removeBomString = require("strip-bom");
+import { removeBOMFromBuffer, removeBOMFromString } from '../utils/removeBOM';
+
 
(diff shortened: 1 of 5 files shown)


rchande:
It's worth noting that this is a mono behavior. A console app that does nothing but set `Console.OutputEncoding = Encoding.UTF8` will print a BOM after the encoding is set.

src/omnisharp/server.ts
```diff
@@ -515,8 +515,11 @@ export class OmniSharpServer {
 
     private async _doConnect(options: Options): Promise<void> {
 
-        this._serverProcess.stderr.on('data', (data: any) => {
-            this._fireEvent('stderr', String(data));
+        this._serverProcess.stderr.on('data', (data: Buffer) => {
```
rchande: Shouldn't we add the same handling to stdio?

akshita31: Done. Thanks!

Issue #2538, opened 2018-09-18
Intelllisense errors in a vscode dotnet core 2.1 / C# project (with git diff view)


llint:
[Intellisense.zip](https://github.com/OmniSharp/omnisharp-vscode/files/2392732/Intellisense.zip)
#### `dotnet --info` output:

.NET Core SDK (reflecting any global.json):
 Version:   2.1.402
 Commit:    3599f217f4

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.402\

Host (useful for support):
  Version: 2.1.4
  Commit:  85255dde3e

.NET Core SDKs installed:
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.400 [C:\Program Files\dotnet\sdk]
  2.1.401 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.3-servicing-26724-03 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download

#### VS Code version:

Version: 1.27.2 (user setup)
Commit: f46c4c469d6e6d8c46f268d1553c5dc4b475840f
Date: 2018-09-12T16:17:45.060Z
Electron: 2.0.7
[...]

PR #2552, opened 2018-09-21
Add vitual documents with a definite scheme

akshita31:
Fixes: #2538 

When git diff is opened we add a virtual document with the same path as the original document and when the diff is closed we are sending omnisharp a delete event from the virtualDocumentTracker that deletes the file from the workspace.

Hence to avoid virtual documents from various extensions interfering with omnisharp, the virtual document tracker will only track virtual documents that obey a specific scheme.

cc @NTaylorMullen

### DIFF
--- a/src/features/virtualDocumentTracker.ts
+++ b/src/features/virtualDocumentTracker.ts
@@ -50,7 +50,7 @@ function trackFutureVirtualDocuments(server: OmniSharpServer, eventStream: Event
 }
 
 function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): boolean {
-    if (document.uri.scheme === 'file' || document.languageId !== 'csharp' || document.uri.scheme !== "virtualCSharp") {
+    if (document.uri.scheme !== "virtualCSharp" || document.languageId !== 'csharp') {
         // We're only interested in non-physical CSharp documents.
         return true;
     }
--- a/src/features/virtualDocumentTracker.ts
+++ b/src/features/virtualDocumentTracker.ts
@@ -50,7 +50,7 @@ function trackFutureVirtualDocuments(server: OmniSharpServer, eventStream: Event
 }
 
 function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): boolean {
-    if (document.uri.scheme === 'file' || document.languageId !== 'csharp') {
+    if (document.uri.scheme === 'file' || document.languageId !== 'csharp' || document.uri.scheme !== "virtualCSharp") {
         // We're only interested in non-physical CSharp documents.
         return true;
     }
--- a/src/features/virtualDocumentTracker.ts
+++ b/src/features/virtualDocumentTracker.ts
@@ -50,7 +50,7 @@ function trackFutureVirtualDocuments(server: OmniSharpServer, eventStream: Event
 }
 
 function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): boolean {
-    if (document.uri.scheme !== "virtualCSharp" || document.languageId !== 'csharp') {
+    if (document.uri.scheme.startsWith("virtualCSharp-") || document.languageId !== 'csharp') {
         // We're only interested in non-physical CSharp documents.
         return true;
     }
--- a/src/features/virtualDocumentTracker.ts
+++ b/src/features/virtualDocumentTracker.ts
@@ -50,7 +50,7 @@ function trackFutureVirtualDocuments(server: OmniSharpServer, eventStream: Event
 }
 
 function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): boolean {
-    if (document.uri.scheme.startsWith("virtualCSharp-") || document.languageId !== 'csharp') {
+    if (!document.uri.scheme.startsWith("virtualCSharp-") || document.languageId !== 'csharp') {
         // We're only interested in non-physical CSharp documents.
         return true;
     }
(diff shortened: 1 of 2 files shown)


src/features/virtualDocumentTracker.ts
```diff
@@ -50,7 +50,7 @@ function trackFutureVirtualDocuments(server: OmniSharpServer, eventStream: Event
 }
 
 function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): boolean {
-    if (document.uri.scheme === 'file' || document.languageId !== 'csharp') {
+    if (document.uri.scheme === 'file' || document.languageId !== 'csharp' || document.uri.scheme !== "virtualCSharp") {
```
rchande: Could this be simplified to `document.uri.scheme !== "virtualCSharp" || document.languageId !== 'csharp'

rchande:
@NTaylorMullen Note that this would be a breaking change for you--you'd need to start using the new scheme.

NTaylorMullen:
> @NTaylorMullen Note that this would be a breaking change for you--you'd need to start using the new scheme.

Yup, totally get that. Thinking about the solution end-to-end a little more I think we need to be a little more relaxed in the scheme in order for this to properly work with virtual documents. One way virtual documents are created today are with a `TextDocumentContentProvider`. A providers tell `vscode` that they can provide the file contents of files under a certain `scheme` but vscode restricts there to be a 1-1 mapping of schemes => content providers; therefore, if in the `virtualDocumentTracker` we restrict it to just `virtualCSharp` that means only one extension can provide virtual files. I think the requirement should be a contains check or something similar on the scheme (could even imagine it being a `startsWith('virtualCSharp-')` or endsWIth equivalent.

NTaylorMullen:
It should also be a strict `endsWith` or `startsWith`. Shouldn't allow consumers to bind to the `virtualCSharp` scheme directly; that'll just open a can of worms.

rchande:
@NTaylorMullen Ah, makes sense. I'm not terribly familiar with this layer of VS Code extensibility but that sounds reasonable to me.

src/features/virtualDocumentTracker.ts
```diff
@@ -50,7 +50,7 @@ function trackFutureVirtualDocuments(server: OmniSharpServer, eventStream: Event
 }
 
 function shouldIgnoreDocument(document: TextDocument, server: OmniSharpServer): boolean {
-    if (document.uri.scheme === 'file' || document.languageId !== 'csharp') {
+    if (document.uri.scheme.startsWith("virtualCSharp-") || document.languageId !== 'csharp') {
```
NTaylorMullen: Need to also ensure that `uri.scheme !== "virtualCSharp"`. Don't want to encourage bad behavior 😄

akshita31: Wouldn't the dash at the end "virtualCSharp-" take care of that?

NTaylorMullen: Mistyped, need to make sure people don't bind to `virtualCSharp-`

test/integrationTests/virtualDocumentTracker.integration.test.ts
```diff
@@ -15,7 +15,7 @@ chai.use(require('chai-arrays'));
 chai.use(require('chai-fs'));
 
 suite(`Virtual Document Tracking ${testAssetWorkspace.description}`, function () {
-    let virtualScheme: string = "virtual";
+    let virtualScheme: string = "virtualCSharp-test";
```
rchande: We should probably add a negative test, ie documents with a scheme that doesn't start with virtualCsharp- aren't tracked.

akshita31:
@NTaylorMullen I tried to add the negative test as @rchande  mentioned above. However even if I change the scheme to "test", the virtualDocumentTracker ignores it, however I am still getting the completions. Any idea what might be happening here ?

NTaylorMullen:
@akshita31 Poked a bit more around the code and there's two more things you'll have to tweek:
- Change the diagnostic provider to only listen to files or `virtualCsharp-*` https://github.com/OmniSharp/omnisharp-vscode/commit/24663d92a4fe14193c6f0024b273a1c75a81841f#diff-f7f698ea628a7fd476ec3cd2355e1927
- Don't fire change events for virtual documents that don't meet the `virtualCSharp-* || file` scheme

SirIntruder:
thanks for this 😄 I think I also had problems with opening metadata documents, this will fix them too?

NTaylorMullen:
Any update on this one?

NTaylorMullen:
Found another edge case with the virtual document tracker. Basically there are times when documents are automatically added to the Misc workspace where instead they should be added to the appropriate project.

I'll send a PR with the fix.

NTaylorMullen:
> @NTaylorMullen I tried to add the negative test as @rchande mentioned above. However even if I change the scheme to "test", the virtualDocumentTracker ignores it, however I am still getting the completions. Any idea what might be happening here ?

Ya looks like bits are added to the miscellaneous workspace and then still operated on :(

test/integrationTests/virtualDocumentTracker.integration.test.ts
NTaylorMullen: Gonna have to find a better way to test the virtual documents. The server still understands requests that are specifically invoked on virtual documents but it treats them as misc docs therefore they don't impact anything surrounding them.

akshita31:
Closing this in favor of #2562

Issue #168, opened 2016-04-11
No intellisense available for razor


egamma:
_From @Felsig on April 8, 2016 23:51_
- VSCode Version: 0.10.11 (0.10.11)
- OS Version: OSX El Capitan 10.11.2 

Specifically when writing code like @Model.Id or the similar I get no sort of intellisense. I assumed that would be a feature as the website states support for Razor. Is this just not a feature?

_Copied from original issue: Microsoft/vscode#5113_

DustinCampbell:
Unfortunately, there is no way to get IntelliSense for Razor in VS Code through OmniSharp today. The Razor parser does not run as it does in Visual Studio. We will look at this in the future though.

Felsig:
Thanks!

eamodio:
Is there anyway to help out with this? Personally this is my number 1 issue with using VS Code for aspnet core development -- since refactoring/usage/etc is no longer safe because usages within cshtml aren't seen -- and cshtml views aren't even compiled on build (not sure if there is a way to make that happen with the new core stuff like there used to be).

DustinCampbell:
@david-driscoll was looking at it awhile back over in https://github.com/OmniSharp/omnisharp-roslyn. Maybe drop by [Slack](https://goo.gl/Ovnqr1) and chat about it?

eamodio:
@DustinCampbell Thanks -- I'll hop over there an find out.

rasmus91:
Have the same Issues (but from Ubuntu) really missing that Razor Intellisense, it'd really make all the difference in the world.

Will i be able to spot progress on this feature here?

eimik:
Agree. Now that we have tabs, razor-intellisense is the "one big thing" missing from VS Code for asp.net core devs.

david-driscoll:
We'll be adding support to OmniSharp eventually to support Razor, sadly I don't have a specific timeline.  This is mainly because it's not a very easy problem to solve, but it is something we want to get working.

mchomer:
Some updates?
I also really miss it.

winuxue:
I need it also

DustinCampbell:
To be clear, this is a fairly big work item. It is currently in our "future" milestone because we don't yet have a good sense of when it will happen..

eamodio:
[...]

PR #2554, opened 2018-09-24
Add preview Razor support

SteveSandersonMS:
As discussed previously, this adds Razor support behind a new `omnisharp.preview` options flag.

It works by downloading a platform-specific Razor Language Server executable (equivalent to the OmniSharp-Roslyn language server) and then running some VSCode-side logic that knows how to launch and interact with it.

### DIFF
--- a/test/integrationTests/documentSymbolProvider.integration.test.ts
+++ b/test/integrationTests/documentSymbolProvider.integration.test.ts
@@ -36,7 +36,14 @@ suite(`DocumentSymbolProvider: ${testAssetWorkspace.description}`, function () {
 
     test("Returns all elements", async function () {
         let symbols = await GetDocumentSymbols(fileUri);
-        expect(symbols.length).to.equal(25);
+
+        // The count can vary:
+        // Some builds of OmniSharp return a tree data structure with one root element
+        // Some have fixes for a duplicate symbols issue and return fewer than we
+        // used to assert
+        // For now, just assert any symbols came back so that this passes locally and in CI
+        // (where we always use the latest build)
+        expect(symbols.length).to.be.greaterThan(0);
     });
 });
 


package.json
```diff
@@ -73,6 +73,7 @@
     "http-proxy-agent": "2.1.0",
     "https-proxy-agent": "2.2.1",
     "jsonc-parser": "1.0.0",
+    "microsoft.aspnetcore.razor.vscode": "https://razorvscodetest.blob.core.windows.net/npm/microsoft.aspnetcore.razor.vscode-1.0.0-alpha1-20180924.3.tgz",
```
SteveSandersonMS: We're consuming this via URL, rather than from npmjs.org, because there's no reason why any other code should want to take a dependency on it. There's just no reason for this to package to be on npmjs.org.

package.json
```diff
@@ -84,7 +85,7 @@
     "tmp": "0.0.33",
     "vscode-debugprotocol": "1.6.1",
     "vscode-extension-telemetry": "0.0.15",
-    "yauzl": "2.9.1"
+    "yauzl": "2.10.0"
```
SteveSandersonMS: Needed to get the newer version of `yauzl` for compatibility with the Razor Language Server zipfiles.

package.json
```diff
@@ -268,6 +269,56 @@
         "./vsdbg"
       ],
       "installTestPath": "./.debugger/vsdbg-ui"
+    },
+    {
+      "description": "Razor Language Server (Windows / x64)",
+      "url": "https://razorvscodetest.blob.core.windows.net/languageserver/RazorLanguageServer-win-x64-1.0.0-alpha1-20180924.3.zip",
```
SteveSandersonMS: These URLs will need to change before this ships. We want the downloads to come from a particular CDN which we're still getting access to.

package.json
```diff
@@ -2712,6 +2796,37 @@
           }
         }
       }
-    ]
+    ],
+    "languages": [
+      {
+        "id": "aspnetcorerazor",
+        "extensions": [
+          ".cshtml"
```
SteveSandersonMS: We're overriding VS Code's built-in `.cshtml` handling by mapping it to a different language ID. This separate `aspnetcorerazor` language ID lets us control exactly what behaviors are applied to it, so we can give a better experience (e.g., around HTML completion) that is more deeply aware of Razor language semantics.

src/razor/razor.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
```
SteveSandersonMS: The logic in this file uses `vscode` APIs directly, not going through `vscodeAdapter`, for reasons we've discussed previously. The `microsoft.aspnetcore.razor.vscode` NPM package uses `vscode` directly extensively, so nothing extra would be gained by putting this tiny bit of wrapper logic through additional abstraction.

Later on we do want to move `microsoft.aspnetcore.razor.vscode` entirely to share OmniSharp-VSCode's `vscodeAdapter` pattern for testability, but that's a bit further out.

src/razor/syntaxes/aspnetcorerazor.tmLanguage.json
```diff
@@ -0,0 +1,9 @@
+{
+	"name": "ASP.NET Core Razor",
+	"scopeName": "text.aspnetcorerazor",
+	"patterns": [
+		{
+			"include": "text.html.cshtml"
```
SteveSandersonMS: This causes the `aspnetcorerazor` language to retain VSCode's built-in colorization, but we override everything else.

package.json
```diff
@@ -2712,6 +2796,37 @@
...
+        "language": "aspnetcorerazor",
+        "scopeName": "text.aspnetcorerazor",
+        "path": "./src/razor/syntaxes/aspnetcorerazor.tmLanguage.json"
+      }
+    ],
+    "menus": {
+      "editor/title": [
+        {
+          "command": "extension.showRazorCSharpWindow",
+          "when": "resourceLangId == razor"
```
SteveSandersonMS: Note: Change language here and below

package.json
NTaylorMullen: Shoot, just thought of one side effect of this. If the `omnisharp.preview` flag is false but we're still overriding the normal Razor language id I think this will nuke some of the default .cshtml editing experiences (no HTML completion).

Does VSCode allow us to toggle the language on/off?

package.json
```diff
@@ -602,6 +653,29 @@
           "type": "boolean",
           "default": false,
           "description": "Specifies whether notifications should be shown if OmniSharp encounters warnings or errors loading a project. Note that these warnings/errors are always emitted to the OmniSharp log"
+        },
+        "omnisharp.preview": {
```
NTaylorMullen: This doesn't already exist? @rchande / @akshita31 / @TheRealPiotrP I thought you guys already had an `omnisharp.preview` flag that we were sitting behind?

package.json
```diff
@@ -602,6 +653,29 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "Specifies whether to enable OmniSharp preview features."
+        },
+        "razor.languageServer.path": {
```
NTaylorMullen: Since this is really pointing to the directory we should consider changing this to `razor.languageServer.directory`

package.json
rchande: @NTaylorMullen We didn't have a flag like that before.

rchande:
@SteveSandersonMS I pulled this down and played around with it. Seems to work--very cool!

package.json
NTaylorMullen: Ah, alright. Well 👍

filipw:
what is RazorLanguageServer? is it available somewhere on Github?

NTaylorMullen:
> what is RazorLanguageServer? is it available somewhere on Github?

Soon, working on getting the repo open source 😉

package.json
```diff
@@ -602,6 +653,29 @@
...
+        },
+        "razor.languageServer.debug": {
+          "type": "boolean",
+          "default": false,
+          "description": "Specifies whether to wait for debug attach when launching the language server."
+        },
+        "razor.languageServer.trace": {
+          "type": "string",
+          "default": "Messages",
+          "description": "Specifies whether to output all messages [Verbose], some messages [Messages] or not at all [Off]."
```
akshita31: Are these the only possible inputs the user must specify. If so, we can use an enum type here, similar to https://github.com/OmniSharp/omnisharp-vscode/blob/master/package.json#L528

akshita31:
Also, I am curious to understand the logging scenario here. Should we have some log that says "Razor activated"? If we get some issues after the release, what is the best way to understand that the razor server is activated(other than asking the users if the preview flag is true)?
cc @rchande

package.json
```diff
@@ -602,6 +653,29 @@
...
+          "description": "Overrides the path to the Razor LanguageServer. If specified, this should be the absolute path to the language server directory."
+        },
+        "razor.languageServer.debug": {
+          "type": "boolean",
+          "default": false,
+          "description": "Specifies whether to wait for debug attach when launching the language server."
+        },
+        "razor.languageServer.trace": {
+          "type": "string",
+          "default": "Messages",
```
NTaylorMullen: Lets actually default to `Off`

NTaylorMullen:
> Also, I am curious to understand the logging scenario here. Should we have some log that says "Razor activated"? If we get some issues after the release, what is the best way to understand that the razor server is activated(other than asking the users if the preview flag is true)?

We'll have a Razor log output window that will indicate that. If you guys feel that there should also be a log in the Omnisharp output window we can definitely add one there as well.

rchande:
@NTaylorMullen Let's keep the logs separate for now, unless we find that there are a lot of issues that require closely correlating timestamps between both logs.

src/main.ts
```diff
@@ -134,6 +135,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
         coreClrDebugPromise = coreclrdebug.activate(extension, context, platformInfo, eventStream);
     }
 
+    if (optionProvider.GetLatestOptions().preview) {
+        activateRazorExtension(context);
```
NTaylorMullen: Should you be awaiting this?

SteveSandersonMS: Good point, yes, that's better.

package.json
SteveSandersonMS: OK, done

package.json
SteveSandersonMS: Good tip - thanks. Done.

package.json
SteveSandersonMS: That's a great point. Unfortunately I don't have any easy fix in mind for it.

I checked, and as expected, having `omnisharp.preview` set to `false` means that `.cshtml` files:

* Still get correct colorization, since that's defined declaratively
* No longer get completions for HTML/JS
* No longer get automatic closing-tag insertion for HTML

As such it's still a working experience, but it is degraded in functionality compared with the default.

Possible ways forwards:

* Decide that Razor support is looking good already and turn it on by default in OmniSharp-VSCode
* Or, move Razor support out into a completely separate extension (not the installation experience we wanted)
* Or, ask the VS Code team to provide some feature by which an extension can respond to the `onLanguage:(languageId)` event and programmatically decline the event, causing VS Code to drop that `language` entry from its config and proceed to activate the next extension that matches the selector. Or any other design the VS Code team prefers.
    * Then we wait for it to be implemented and ship
* Or, try to implement our `.cshtml` support *without* defining a new language ID.
    * Unfortunately that is what we tried originally and it didn't work well because VS Code consulted both our extension and the built-in Razor feature for completions (etc), leading to duplication in the UI and incorrect awareness of what should show up in HTML blocks vs C# blocks etc. For example, if you try to type `List<string>` in a C# context, the built-in HTML tag completion kicks in and inserts `</string>` right after, which is insane.
    * At the very least, going this route is likely to involve some work with and by the VS Code folks to find some new way of suppressing the default behaviors programmatically.

gregg-miskelly: Assuming you don't think things are already good enough that you just want to turn it on, have you thought about maybe having a 'razor' branch in this repo where you can release beta versions of this extension with the razor support enabled, and folks who want to try it can install this new extension rather than toggling an option?

NTaylorMullen: @SteveSandersonMS lets talk more offline in our sync up today.

@gregg-miskelly we have. Our goals are to get the extension into as many peoples hands as possible. Forcing users to go and download a pre-release version would be an experience blocker and would prohibit us from getting enough feedback to make educated decisions about future releases.

NTaylorMullen:
You'll actually also need to update this section in packages.json:

```
      "type": "coreclr",
        "label": ".NET Core",
        "enableBreakpointsFor": {
          "languageIds": [
            "csharp",
            "razor"
          ]
        },
```

rchande:
@SteveSandersonMS @NTaylorMullen How hard will it be for you to react to https://github.com/OmniSharp/omnisharp-vscode/pull/2552 when it merged? Would we be better off waiting on that PR?

NTaylorMullen:
> @SteveSandersonMS @NTaylorMullen How hard will it be for you to react to #2552 when it merged? Would we be better off waiting on that PR?

Already reacted. We'll just need to update the npm package / Language Server versions that are pulled down. I don't believe this PR currently is targeting the npm/language server bits with the reaction but either way it's super easy to add; let us know 😉

welkie:
Came here from https://github.com/OmniSharp/omnisharp-vscode/issues/168. I notice one of the comments above states that making the Razor support available as a separate extension would solve some of these early issues but wouldn't be the installation experienced that was originally wanted. I just wanted to chime in and mention that I think it is a good idea to make it a different extension. Not all of the ASP.NET Core projects I develop in Visual Studio Code use Razor. In fact, some people may create C# projects that have nothing to do with the web.

NTaylorMullen:
@welkie I totally get your concern. Even though the Razor experience is planned to be included as part of the C# extension it will only operate on Razor files and should not have any perf-impact to that of C# files.

NTaylorMullen:
@SteveSandersonMS is this good to merge now?

SteveSandersonMS:
@NTaylorMullen As of the commit I just added (63c4ed9), yes. This should include the extra stuff you had in your PR.

welkie:
@NTaylorMullen Ah, in that case, this looks super exciting. Can't wait to see it in action. It'll save me from needing to buy a Rider subscription when I eventually get into Razor stuff. Great work!

src/razor/razor.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as Razor from 'microsoft.aspnetcore.razor.vscode';
+import { EventStream } from '../EventStream';
+
+export async function activateRazorExtension(context: vscode.ExtensionContext, eventStream: EventStream) {
+    const razorConfig = vscode.workspace.getConfiguration('razor');
+    const configuredLanguageServerDir = razorConfig.get<string>('languageServer.directory', null);
+    const languageServerDir = configuredLanguageServerDir || path.join(context.extensionPath, '.razor');
```
akshita31: should we pass the extensionPath here instead of using context.extensionPath ?

akshita31:
Apologies, I have been off-track for a while here. Just wanted to confirm, if we have razor.disabled - true, it would still lead to degraded experience for the html completions right, how or why do we feel that is better than having razor disabled by default ? My guess is that user will know they have opted out of the razor experience because they put off the flag, hence this is a better scenario or is there anything I am missing out.

src/omnisharp/server.ts
```diff
@@ -304,6 +304,15 @@ export class OmniSharpServer {
             '--loglevel', options.loggingLevel
         ];
 
+        if (!options.razorDisabled) {
+            const razorPluginPath = path.join(
+                utils.getExtensionPath(),
```
akshita31: server takes an extension path in the constructor. We should use that here

NTaylorMullen:
> would still lead to degraded experience for the html completions right

Yup that's correct. It's not ideal but the primary purpose of the razor.disabled flag exists as a backup in-case we explode something unintentionally and break another piece of VSCode unintentionally.

> why do we feel that is better than having razor disabled by default

Even if Razor is disabled by default you still wouldn't get the vanilla Razor experience that exists today. Sadly this is all a restriction in how VSCode extensions can extend/replace languages 😢

src/razor/razor.ts
SteveSandersonMS: If you think that's better, it's fine with me. Done.

src/omnisharp/server.ts
SteveSandersonMS: Good plan. Done.

SteveSandersonMS:
The remaining Travis CI failure appears unrelated to this PR. It happens if you run the integration tests on `master` currently too.

rchande:
@SteveSandersonMS I'm looking into your CI failure now.

rchande:
@NTaylorMullen @SteveSandersonMS is this ready to merge?

SteveSandersonMS:
Yes, and thanks for resolving the CI issue!

PR #2558, opened 2018-09-27
If the config looks functional return it, otherwise force VSCode to open config file

isidorn:
Hi VSCode dev here,

VSCode changed the functionality when to automatically open a launch.json file.
Previosuly if the return configuraiton would not have a `.type` field we would automatically open it, however we decided that the extensions must explictly request this by returning `null`. More details about the change and the motivation behind it can be found here
https://github.com/Microsoft/vscode/issues/54213

It would be great if we could merge this as soon as possible since it creates a regression, more details about it here https://github.com/Microsoft/vscode/issues/59487

fyi @gregg-miskelly @DustinCampbell @akshita31 
sorry for pinging everyone

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -140,9 +140,9 @@ export class CSharpConfigurationProvider implements vscode.DebugConfigurationPro
             config = this.parseEnvFile(config.envFile.replace(/\${workspaceFolder}/g, folder.uri.path), config);
         }
 
-        // vsdbg will error check the debug configuration fields      
-        return config;
-    }   
+        // If the config looks functional return it, otherwise force VSCode to open a configuration file https://github.com/Microsoft/vscode/issues/54213
+        return config && config.type ? config : null;
+    }
 
     private static async showFileWarningAsync(message: string, fileName: string) {
         const openItem: MessageItem = { title: 'Open envFile' };


isidorn:
Also this PR will not break the existing vscode stable. 
The next vscode stable will be released in a week and ideally we could merge this before than. If that is not an option please let me know.

src/configurationProvider.ts
```diff
@@ -140,9 +140,9 @@ export class CSharpConfigurationProvider implements vscode.DebugConfigurationPro
             config = this.parseEnvFile(config.envFile.replace(/\${workspaceFolder}/g, folder.uri.path), config);
```
gregg-miskelly: @isidorn Will VS Code sometimes call this function with a null or undefined `config` argument, and so this if statement should be `if (config && config.envFile)`?

WardenGnaw:
@isidorn We do the same the C++ extension.

https://github.com/Microsoft/vscode-cpptools/blob/96e338628836fa4d579088c838e2ac4c6a0d80a3/Extension/src/Debugger/configurationProvider.ts#L63

isidorn:
Hi guys. Just a friendly reminder to merge this if possible. We are soon to release a new vscode stable build and it would be cool if this change is part of your extension before that.

gregg-miskelly:
@isidorn I had one question that I was hoping you would answer before I merged this.

isidorn:
@gregg-miskelly sorry! Somehow I missed the question. No vscode should never call that with a `null` or `undefined` config. Config should be an empty object at worst. [Code pointer](https://github.com/Microsoft/vscode/blob/a1fa2a89d6d664b3527af1abe027401a6307a951/src/vs/workbench/parts/debug/electron-browser/debugService.ts#L363)
If you are seeing `null` or `undefined` than that is a bug which shuold be fixed on the vscode side.

gregg-miskelly:
Cool. Thanks for confirming. I will commit this as soon as the build rules are satisfied.

PR #2562, opened 2018-09-29
Make Virtual document tracking apply to specific scheme.

NTaylorMullen:
- Also fixed issue where we'd attempt to initialize virtual documents before the project world was ready. This is an issue because virtual documents don't exist as part of the MSBuild piece of the project and therefore would more-often than not get put into the miscellaneous project on the server; resulting in bad IntelliSense.
- Fixed the diagnostic provider to also obey the new `virtualCSharp-*` scheme requirement
- Changed the default change forwarder to ignore virtual documents to ensure that it doesn't try and trigger any changes for them until the server is ready.

Fixes OmniSharp/omnisharp-vscode#2538
Fixes aspnet/Razor.VSCode#105

@rchande @akshita31 @SteveSandersonMS

@akshita31 I ported your changes to this PR and made the few PR requested changes. Also, I as well couldn't add a negative test for the virtual document tracking for reasons [here](https://github.com/OmniSharp/omnisharp-vscode/pull/2552#discussion_r221405541).

### DIFF
--- a/src/features/virtualDocumentTracker.ts
+++ b/src/features/virtualDocumentTracker.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import {workspace, TextDocument, Uri} from 'vscode';
-import {OmniSharpServer} from '../omnisharp/server';
+import { workspace, TextDocument, Uri } from 'vscode';
+import { OmniSharpServer } from '../omnisharp/server';
 import * as serverUtils from '../omnisharp/utils';
 import { FileChangeType } from '../omnisharp/protocol';
 import { IDisposable } from '../Disposable';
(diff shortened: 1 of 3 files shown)


src/features/virtualDocumentTracker.ts
```diff
@@ -102,8 +143,21 @@ function logSynchronizationFailure(uri: Uri, error: any, server: OmniSharpServer
...
-    let disposable = trackFutureVirtualDocuments(server, eventStream);
-    
-    return disposable;
+    let disposable: vscode.Disposable;
+    const registration = server.onProjectAdded(async () => {
+        registration.dispose();
+
+        // We wait until project state on the server has been initialized to add any virtual documents.
+
+        server.waitForEmptyEventQueue().then(async () => {
```
NTaylorMullen: @rchande @akshita31 this is ugly but essentially we need to wait until all of the projects have been initialized so we wait for the project initialization process to start and then in the background (we're not awaiting this) queue up a virtual document tracker initialization step. All of this is to ensure that we don't try and initialize a virtual document before its corresponding project is ready.

I thought about trying to assume the project in this virtual document tracker but figured that'd break functionality for anyone who truly just wanted a miscellaneous project only virtual document.

If there's a better way to do this i'm all ears.

rchande: @NTaylorMullen Why are we trying to wait for all projects to be initialized?

NTaylorMullen: > I believe it should get "promoted" to being a non-misc file if a project loads it after it's added as a misc file. If that's not the case, we should make it so. Have you actually seen it fail to promote the file?

Seems that way, or the promotion takes a super long time. From what I looked at it seemed that only MSBuild included file are promoted once a project has been loaded. Probably missing something though. Want to sit down and person and debug?

src/features/virtualDocumentTracker.ts
```diff
@@ -3,27 +3,39 @@
...
+import { OmniSharpServer } from '../omnisharp/server';
 import * as serverUtils from '../omnisharp/utils';
 import { FileChangeType } from '../omnisharp/protocol';
 import { IDisposable } from '../Disposable';
 import CompositeDisposable from '../CompositeDisposable';
 import { EventStream } from '../EventStream';
 import { DocumentSynchronizationFailure } from '../omnisharp/loggingEvents';
 
-function trackCurrentVirtualDocuments(server: OmniSharpServer, eventStream: EventStream) {
-    let registration = server.onProjectAdded(async () => {
```
NTaylorMullen: This isn't needed any more but because I removed it that also means we fire and forget the initial synchronization requests (aka we don't `await trackCurrentVirutalDocuments`). @rchande what's your opinions here?

NTaylorMullen:
@rchande @akshita31 the failure that's happening on travis in this PR is happening in master on my Mac as well. Looks like an unrelated issue, that known?

rchande:
@NTaylorMullen Is this ready to merge?

NTaylorMullen:
Yup go for it.

PR #2574, opened 2018-10-03
Add Razor verification test plan.

NTaylorMullen:
/cc @danroth27 @SteveSandersonMS @rchande @akshita31

### DIFF
--- a/test-plan.md
+++ b/test-plan.md
@@ -93,6 +93,34 @@ The easist way to verify that a project was successfully loaded is to open a .cs
 * In a project that uses globbing (.NET Core), use the VS Code file explorer to add a new file next to the csproj. Intellisense/sighelp/etc should be available in the new file
 * Add a new file and reference a type in it from a different file. Deleting from disk the file containing the referenced type  should produce error messages
 
+#### Razor
+The Razor experience is available when you open a .cshtml file in a valid OmniSharp project. To setup a test project to verify on you can do:  
+1. `dotnet new razor`
+2. Open `Pages/Index.cshtml`
+
+##### C# Completion
+* Typing `@DateTime.Now` and `@(DateTime.Now)` provides completions throughout typing.
+* Completion is available for types that exist in the project (i.e. `Program`)
+* Typing `@model DateTime` prompts for completion for the `model` symbol and the `DateTime` symbol.
+
+#### C# Signature Help
+* Typing `@Html.Raw()` prompts for signature help inside of the `()`.
+
+#### C# Diagnostics
+* Typing `@ThisDoesNotExist` results in an error being created and squiggled in the .cshtml file. NOTE: This error squiggly will be misaligned due to known issues.
+
+#### Razor Options
+
+##### razor.disabled
+This option can be set to `true` to disable the above described C# experience.
+
+##### razor.languageServer.trace
+This option should always be displayed in the View --> Output --> Razor Log window
+This option can be set to any of the following values:
+  * "Off" - Will launch Razor Language server with its log output set to 'Off'. The header in the Razor Log output window will be shown but no other content will be shown.
+  * "Messages" - Will launch Razor Language server with its log output set to 'Messages'. Limited messages will be shown such as "Opening document xyz in project abc".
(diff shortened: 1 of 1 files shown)


test-plan.md
```diff
@@ -93,6 +93,38 @@ The easist way to verify that a project was successfully loaded is to open a .cs
 * In a project that uses globbing (.NET Core), use the VS Code file explorer to add a new file next to the csproj. Intellisense/sighelp/etc should be available in the new file
 * Add a new file and reference a type in it from a different file. Deleting from disk the file containing the referenced type  should produce error messages
 
+#### Razor
```
NTaylorMullen: Added this as a sub-item of `Validating C# Extension for VS Code`, is that the right place?

akshita31:
@NTaylorMullen It would be great if we can provide a link to a sample repository or something with which the vendors can test.( We have been thinking of adding the test samples in the repo root itself, but havent been able to do that so far), but we do provide a set of commands that the vendors might use to create some sample C# projects. I am not sure if such a thing exists for razor. If not maybe we can create a sample repo somewhere on github and just link to that. That will make it easier for the vendors specially when they will do it for the first time.

test-plan.md
```diff
@@ -93,6 +93,38 @@ The easist way to verify that a project was successfully loaded is to open a .cs
...
+#### C# Diagnostics
+* Typing `@ThisDoesNotExist` results in an error being created and squiggled in the .cshtml file. NOTE: This error squiggly will be misaligned due to known issues.
+
+#### Razor Options
+
+##### razor.disabled
+This option can be set to `true` to disable the above described C# experience.
+
+##### razor.languageServer.directory
+Setting this path to any of the values as listed below, should start the Razor Language server.
```
akshita31: What do you mean by "start the razor language server"? Probably we just want to say here - There should or should not be a razor channel with some expected contents(assuming that is the visual hint to show that the razor server started).

test-plan.md
```diff
@@ -93,6 +93,38 @@ The easist way to verify that a project was successfully loaded is to open a .cs
...
+  * undefined - Razor Language server must start using the copy of Razor Language server shipped with the extension.
+  * Some absolute path to a directory - Razor Language server must start using the local copy of Razor Language server that exists in the path pointed to by the config.
```
akshita31: How does a vendor know what path the razor language server used? Do we print that in the "Razor channel", like "Razor server started...with path ..."

test-plan.md
```diff
@@ -93,6 +93,38 @@ The easist way to verify that a project was successfully loaded is to open a .cs
...
+  Example:C:/Razor.VSCode/artifacts/build/win-x64/ 
+
+##### razor.languageServer.trace
+This option should always be displayed in the View --> Output --> Razor Log window
+This option can be set to any of the following values:
+  * "Off" - Will launch Razor Language server with its log output set to 'Off'.
```
akshita31: Can we say something like "If the option is set to "Off", nothing will show up or something ?

test-plan.md
NTaylorMullen: Sadly we don't print where the server is being run from. We just print what version is being used (the assembly version of the server). I've filed an issue to add that information: https://github.com/aspnet/Razor.VSCode/issues/169

test-plan.md
NTaylorMullen: Content still shows up, you just don't get as verbose of logs. I'll clarify

NTaylorMullen:
🆙 📅

test-plan.md
SteveSandersonMS: I'd guess the only aspect of `razor.languageServer.directory` that a vendor could verify is that it causes an error if you point to some directory that doesn't exist or doesn't contain `rzls[.exe]`. They won't be able to verify it actually works if you point to a real alternate language server, because they won't have a real alternate language server.

TBH I don't think this is something the vendors really need to verify. The option only exists for people working on Razor Language Server.

NTaylorMullen: Alright fair enough, i'll nuke this requirement in the test plan.

NTaylorMullen:
Alright, this is good to be merged.

test-plan.md
rchande: Works for me.

Issue #2519, opened 2018-09-12
Show proper messages to the users when the downloads fail


akshita31:
For issues like these : https://github.com/OmniSharp/omnisharp-vscode/issues/2516, we should have a proper story to bring to the users notice that the omnisharp download did not succeed and should show them url they could themselves try to use for the download.

 Also, the message "Starting Omnisharp..." before we actually know whether omnisharp has started or not sounds like a bug.

Related issue: https://github.com/OmniSharp/omnisharp-vscode/issues/2348

cc @rchande

PR #2577, opened 2018-10-04
Add url in the download failure

akshita31:
Related issue: #2519 

When the download fails from a url, we should display the url for them so that they can go ahead and download the packages manually.


rchande:
@akshita31 Looks like some test baselines need to be updated.

PR #2584, opened 2018-10-04
Add to offline ignore vscode

akshita31:


NTaylorMullen:
@rchande @akshita31 Will need to have the release tag for the release include this change so [these](https://github.com/OmniSharp/omnisharp-vscode/wiki/Installing-the-C%23-extension-to-a-computer-without-internet-connectivity) instructions work correctly in offline restore.

PR #2592, opened 2018-10-08
Update readme for the issue reporter capability

akshita31:


NTaylorMullen:
Cool idea, we should probably do the same thing for Razor.VSCode 😄

PR #2614, opened 2018-10-17
Update Razor configurations.

NTaylorMullen:
- Added a `razor.devmode` configuration that suppresses the default Razor attachment to `aspnetcorerazor` documents. This is required to properly run the https://github.com/aspnet/Razor.VSCode extension in a dev environment.
- Added a `razor.plugin.path` configuration to enable easy debugging and deving of the Razor plugin.
- Changed `razor.languageServer.trace` to `razor.trace` since the configuration controls both the client and server log levels of the Razor component.
- Added a `RazorLoggerObserver` that understands users setting `razor.devmode` and the `razor.plugin.path`.
- Added `RazorLoggerObserver.test` to verify the 3 added logging events.

aspnet/Razor.VSCode#190
aspnet/Razor.VSCode#189

@rchande @akshita31 @SteveSandersonMS

### DIFF
--- a/src/omnisharp/extension.ts
+++ b/src/omnisharp/extension.ts
@@ -28,7 +28,7 @@ import WorkspaceSymbolProvider from '../features/workspaceSymbolProvider';
 import forwardChanges from '../features/changeForwarding';
 import registerCommands from '../features/commands';
 import { PlatformInformation } from '../platform';
-import { ProjectJsonDeprecatedWarning, OmnisharpStart } from './loggingEvents';
+import { ProjectJsonDeprecatedWarning, OmnisharpStart, RazorDevModeActive } from './loggingEvents';
 import { EventStream } from '../EventStream';
 import { NetworkSettingsProvider } from '../NetworkSettings';
 import CompositeDisposable from '../CompositeDisposable';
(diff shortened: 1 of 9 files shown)


src/omnisharp/extension.ts
```diff
@@ -152,8 +152,14 @@ export async function activate(context: vscode.ExtensionContext, packageJSON: an
             });
     }));
 
-    // read and store last solution or folder path
-    disposables.add(server.onBeforeServerStart(path => context.workspaceState.update('lastSolutionPathOrFolder', path)));
+    disposables.add(server.onBeforeServerStart(path => {
+        if (options.razorDevMode) {
```
NTaylorMullen: I couldn't find a better place to put this. If I try and dispatch a event before the server's start processes has begun the log message gets completely lost.

src/omnisharp/server.ts
```diff
@@ -337,6 +337,14 @@ export class OmniSharpServer {
             let launchResult = await launchOmniSharp(cwd, args, launchInfo, this.platformInfo, options, this.monoResolver);
             this.eventStream.post(new ObservableEvents.OmnisharpLaunch(launchResult.monoVersion, launchResult.monoPath, launchResult.command, launchResult.process.pid));
 
+            if (razorPluginPath && options.razorPluginPath) {
```
NTaylorMullen: Doing this here so the log messages show up after the important OmniSharp logs.

NTaylorMullen:
This is what the omnisharp output channel looks like with `razor.devmode: true` and `razor.plugin.path: ....dll`

image

NTaylorMullen:
Good to merge whenever you are then 😉

PR #2625, opened 2018-10-25
Do not send a symbol search request to OmniSharp if search term is shorter than "omnisharp.minFindSymbolsFilterLength" 

filipw:
Instead of doing the validation on the server, we do it on the client.

### DIFF
--- a/src/features/workspaceSymbolProvider.ts
+++ b/src/features/workspaceSymbolProvider.ts
@@ -23,10 +23,15 @@ export default class OmnisharpWorkspaceSymbolProvider extends AbstractSupport im
         let options = this.optionProvider.GetLatestOptions();
         let minFilterLength = options.minFindSymbolsFilterLength > 0 ? options.minFindSymbolsFilterLength : undefined;
         let maxItemsToReturn = options.maxFindSymbolsItems > 0 ? options.maxFindSymbolsItems : undefined;
-        return serverUtils.findSymbols(this._server, { Filter: search, MinFilterLength: minFilterLength, MaxItemsToReturn: maxItemsToReturn, FileName: '' }, token).then(res => {
+
+        if (minFilterLength != undefined && search.length < minFilterLength) {
+            return [];
+        }
+
+        return serverUtils.findSymbols(this._server, { Filter: search, MaxItemsToReturn: maxItemsToReturn, FileName: '' }, token).then(res => {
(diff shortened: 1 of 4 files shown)


src/features/workspaceSymbolProvider.ts
```diff
@@ -23,10 +23,15 @@ export default class OmnisharpWorkspaceSymbolProvider extends AbstractSupport im
         let options = this.optionProvider.GetLatestOptions();
         let minFilterLength = options.minFindSymbolsFilterLength > 0 ? options.minFindSymbolsFilterLength : undefined;
```
dmgonch: >minFindSymbolsFilterLength [](start = 38, length = 26)

Sounds like you could remove the setting as well.

filipw: the setting makes sense, it's just we evaluate it on the client rather than server

dmgonch: you are right - sorry for the confusion.

src/features/workspaceSymbolProvider.ts
```diff
@@ -23,10 +23,15 @@ export default class OmnisharpWorkspaceSymbolProvider extends AbstractSupport im
         let options = this.optionProvider.GetLatestOptions();
         let minFilterLength = options.minFindSymbolsFilterLength > 0 ? options.minFindSymbolsFilterLength : undefined;
         let maxItemsToReturn = options.maxFindSymbolsItems > 0 ? options.maxFindSymbolsItems : undefined;
-        return serverUtils.findSymbols(this._server, { Filter: search, MinFilterLength: minFilterLength, MaxItemsToReturn: maxItemsToReturn, FileName: '' }, token).then(res => {
```
dmgonch: But FindSymbolsRequest.MinFilterLength can be removed right, along with removing it on the server side, right?

filipw: yes, I removed it from the protocol.ts, thanks for the pointer 👍

dmgonch:
@filipw : let me know if you working on removing MinFilterLength from OmniSharp/omnisharp-roslyn as well - I will pick up the cleanup otherwise. thanks.

filipw:
we will not remove it yet to not cause a breaking change on the API, since it already shipped. I will remove it in the future, but not yet

Issue #2546, opened 2018-09-20
Accessing a resource scoped configuration without providing a resource is not expected


StephenWeatherford:
Issue Type: <b>Bug</b>

[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '[csharp]', provide the URI of a resource or 'null' for any resource.

[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'editor.insertSpaces', provide the URI of a resource or 'null' for any resource.
[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'editor.tabSize', provide the URI of a resource or 'null' for any resource.

[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'editor.tabSize', provide the URI of a resource or 'null' for any resource.

Extension version: 1.16.1
VS Code version: Code 1.27.2 (f46c4c469d6e6d8c46f268d1553c5dc4b475840f, 2018-09-12T16:17:45.060Z)
OS version: Windows_NT x64 10.0.17134

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Xeon(R) CPU E31230 @ 3.20GHz (4 x 3193)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|15.96GB (5.09GB free)|
|Process Argv|C:\Program Files\Microsoft VS Code\Code.exe|
|Screen Reader|no|
|VM|0%|

</details>

akshita31:
@StephenWeatherford I often get these in the "Debug Console" when I debug the extension but I am not sure what are the implications of these. The extension seems to work fine for me, so I assumed these are some sort of warning messages. Can you elaborate on what is the problem that these messages cause?

akshita31:
Also, @rchande do you have any ideas here ?

StephenWeatherford:
https://github.com/Microsoft/vscode/issues/37041

[...]

PR #2631, opened 2018-10-29
fixed the integration tests/debugging warnings raised from configuration

filipw:
Fixes these warnings:

```
[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '[csharp]', provide the URI of a resource or 'null' for any resource.
[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'editor.insertSpaces', provide the URI of a resource or 'null' for any resource.
[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'editor.tabSize', provide the URI of a resource or 'null' for any resource.
[ms-vscode.csharp] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'editor.tabSize', provide the URI of a resource or 'null' for any resource.
```


akshita31:
Fixes: #2546

PR #2639, opened 2018-11-01
Delete flaky CodelensProvider test

akshita31:
This flaky test is causing a lot of confusion, and we have to restart the builds so that it passes.


akshita31:
@rchande I've filed this issue here to keep a track of this work: https://github.com/OmniSharp/omnisharp-vscode/issues/2643. I would rather leave it open in case someone in the community is interested to take this up. Does that sound ok?

Issue #2650, opened 2018-11-03
Inconsistent changelog


KillyMXI:
## Environment data
`dotnet --info` output: irrelevant
VS Code version: 1.28.2
C# Extension version: 1.17.0

## Steps to reproduce
1. (optional) have outdated version of extension installed;
2. look in the Changelog tab in VS Code what changes were made in the extension update.

## Expected  behavior
Changelog tab has the final info about published version.

## Actual behavior
Changelog wasn't updated for the release, showing outdated info.
Readme tab duplicates some update info as an excerpt.

This look amateurish tbh. Didn't expect that from an extension with 9.4M installs...
Please improve your release publishing procedure to ensure consistency of information.

rchande:
@KillyMXI Thanks for the feedback. We'll discuss this and improve our process to avoid this in the future.

PR #2652, opened 2018-11-04
Update Changelog for the 1.17.0 release

akshita31:


akshita31:
@rchande Once we update this, will it be possible to change the tag to point to the new changelog

filipw:
once the tag is changed it should fix #2650

PR #2679, opened 2018-11-13
Update to alpha 2 Razor.VSCode.

NTaylorMullen:
/cc @rchande @akshita31

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "csharp",
-  "version": "1.17.0-beta6",
+  "version": "1.17.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "csharp",
-  "version": "1.17.0-beta6",
```
NTaylorMullen: I can undo this, but saw that it was a little messed up?

rchande: Arguably  it should be 1.17.1?

rchande: Though I don't think this file gets used in any releasea activities.

PR #2699, opened 2018-11-21
Move invalid CHANGELOG entry into more accurate position

jakubfijalkowski:
Hey,

After merging my PR #1877 (thanks!) my then-added `CHANGELOG.md` entry has been placed under version `1.17.0` which was then-to-be-released, but now is invalid. This PR moves that entry into more appropriate 1.17.2 (if that will be the version released).


akshita31:
For now let us do "1.18.0"

akshita31:
Thanks!

Issue #2478, opened 2018-08-29
Eliminate install.Lock from the download and install pipeline


akshita31:
The extension needs a better mechanism to be able to identify whether the required dependencies are installed or not.

cc @rchande

gregg-miskelly:
@akshita31 what don't you like about install.Lock?

akshita31:
@gregg-miskelly install.Lock is a file, the presence of which tells us currently that the default omnisharp and the debugger packages are installed. However if the user somehow deletes these folders, until the install.Lock is present, a new download wont be triggered. Wouldn't it be better if we could somehow check the "run" file associated with the packages to ensure that those folders are present ?

DustinCampbell:
What if we had two install.Lock files -- one for OmniSharp and one for the debugger? The files could then be written to the `.debugger` and `.omnisharp` folders respectively after extracting the archives. Then, users would just need to delete the appropriate folder to re-download.

gregg-miskelly:
Sounds good to me

Issue #2694, opened 2018-11-21
Validate OmniSharp payload


rchande:
https://github.com/OmniSharp/omnisharp-vscode/issues/2690 is an example of an issue where the OmniSharp download was either incomplete or did not extract properly. Let's consider adding a mechanism to validate that the download resulted in the correct directory structure.

rchande:
@akshita31 Any big brainthoughts here?

akshita31:
Steps:

- [ ] Add the check in the extension for the integrity
- [ ] Update omnisharp to produce the integrity check sum

PR #2707, opened 2018-11-27
Use a separate install.Lock for each dependency

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2478

With this change we will have an install.Lock for each package that is downloaded rather than having a single install.Lock for the runtime dependencies.
This will be a step forward towards the validation of the downloads as listed here: https://github.com/OmniSharp/omnisharp-vscode/issues/2694.

Also, in this PR, I have added an "Id" attribute to the packages so that whatever part of the code needs access to a particular package to verify if the required dependency is installed or not, they could filter the packages from package.json easily, eg, for the case of the debugger packages as I have done in this PR.

cc @NTaylorMullen

### DIFF
--- a/src/CSharpExtDownloader.ts
+++ b/src/CSharpExtDownloader.ts
@@ -3,9 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as util from './common';
 import { PlatformInformation } from './platform';
-import { PackageInstallation, LogPlatformInfo, InstallationSuccess, InstallationFailure } from './omnisharp/loggingEvents';
+import { PackageInstallation, LogPlatformInfo, InstallationSuccess } from './omnisharp/loggingEvents';
 import { EventStream } from './EventStream';
 import { DownloadAndInstallPackages } from './packageManager/PackageManager';
 import { Package } from './packageManager/Package';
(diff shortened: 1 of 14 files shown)


rchande:
How do this work if I upgrade an existing installation? We'll just stop looking at my existing install.lock and be good to go?

src/packageManager/AbsolutePathPackage.ts
```diff
@@ -8,7 +8,8 @@ import { IPackage } from "./IPackage";
 import { AbsolutePath } from "./AbsolutePath";
 
 export class AbsolutePathPackage implements IPackage{
-    constructor(public description: string,
+    constructor(public Id: string,
```
gregg-miskelly: nit: should this be cased with a lower case 'id'?

src/packageManager/PackageManager.ts
```diff
@@ -13,24 +13,41 @@ import { EventStream } from '../EventStream';
...
+                    eventStream.post(new InstallationFailure(installationStage, packageError));
+                    throw packageError;
                 }
                 else {
+                    eventStream.post(new InstallationFailure(installationStage, error));
                     throw error;
                 }
             }
+            finally {
+                await deleteInstallFile(pkg.installPath, InstallFileType.Begin);
```
gregg-miskelly: Should we only do this if we created the Begin file? Or do we need to eat exceptions if it doesn't exist?

src/coreclr-debug/activate.ts
```diff
@@ -125,7 +126,14 @@ export async function getAdapterExecutionCommand(platformInfo: PlatformInformati
...
         // install.Lock does not exist, need to wait for packages to finish downloading.
-        let installLock: boolean = await common.installFileExists(common.InstallFileType.Lock);
+        let installLock = false;
+        let debuggerPackage = getRuntimeDependencyPackageWithId("Debugger", packageJSON, platformInfo, extensionPath);
+        if (debuggerPackage && debuggerPackage.installPath)
+        {
+            installLock = await common.installFileExists(debuggerPackage.installPath, common.InstallFileType.Lock);
+        }
+        
+        //todo: check if the debugger install.lock is present
```
gregg-miskelly: Should this comment still be here?

src/CSharpExtDownloader.ts
```diff
@@ -26,30 +25,18 @@ export class CSharpExtDownloader {
 
     public async installRuntimeDependencies(): Promise<boolean> {
```
gregg-miskelly: We are now always calling this method when the C# extension starts right? If so, shouldn't we avoid posting some of these events to the event stream until after we have determined if there are any missing packages?

akshita31: @gregg-miskelly You are right, but with the current code structure it is not possible and we will need some refactoring in order to achieve that. I will do that in a subsequent PR.

akshita31:
@NTaylorMullen Do you have any concerns with this change ?

NTaylorMullen:
> @NTaylorMullen Do you have any concerns with this change ?

Based on what you wrote in the description of the PR this sounds great! Means we wont have busted partial installs of dependencies it sounds like. Any negative implications to this?

akshita31:
@NTaylorMullen I am still unsure of what part fails on the user's systems when they have an improper download/install. With this PR, I think it will help us identify better which of the three dependencies got corrupted as each will have its own install.Lock file that *should* be put only when both the download and install completed without throwing an exception. 

And if for a dependency say OmniSharp, the download failed this time and the install.Lock was not put into the expected location, the next time the user opens vscode, it would detect that it needs to install Omnisharp as the install.Lock is not present.

NTaylorMullen:
Awesome, ya this would be a great addition, no concerns!

Issue #2348, opened 2018-05-28
Omnisharp flame is green even though the omnisharp server errored


akshita31:
Steps to reproduce:

Delete the Omnisharp executable(ensure that the install.Lock is present so that a new download is not triggered) and then restart the server.

Expected:
OmniSharp flame in the status bar observer should turn red.

Actual:
Omnisharp flame is green, which indicates that the omnisharp server is running.

![omnisharp_error](https://user-images.githubusercontent.com/12554141/40626617-681393a4-62d7-11e8-91ea-fe68e79694fd.PNG)

Diagnosis:

A simple tweak here is to add another event to the OmnisharpStatusBarObserver, and the  PR for that is here : https://github.com/OmniSharp/omnisharp-vscode/pull/2349. However this makes me think that should we have a "success" message returned from omnisharp once it has successfully spawned ( https://github.com/nodejs/help/issues/1191)., and wait for that message before we say that 'OmniSharp server has started'. Doing this might also help remove the flakiness that we have in the integration tests.

When we fire the event here : https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/server.ts#L329, we are not sure that the server has started because upto this point we haven't seen any such event that suggests so.

Thoughts @rchande @DustinCampbell @TheRealPiotrP @colombod

akshita31:
@rchande Yes that is a thing I will take up, but the issue I am trying to convey here is more than that.

Assuming that there is an std error here : https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/server.ts#L511, we are not catching this error as it will be produced asynchronously and we fire the "server started" event here : https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/omnisharp/server.ts#L329, even before we actually know that omnisharp has started successfully. 
We await the omnisharp running status in our integration test and my guess is that the flakiness there is also because of this reason.

If we can return an event from OmniSharp saying - "Hey, I have started" and then fire the server started event, I think that will solve a number of problems. Did that make sense ?

rchande:
@akshita31 Sure, makes sense. Do we really have to add a dedicated event to o# though? I bet we could just listen for any command line output from O# and consider that to mean that it started.

akshita31:
@rchande Yes, listening to any data on the standard output of the process should be enough. I'll look into how to do this further.

Issue #2709, opened 2018-11-27
Error shown in Dev Tools Console even when everything works fine


akshita31:
## Issue Description ##
The extension always shows errors in the developer console. This misleads users to think there is some error even when there isnt
## Steps to Reproduce ##
Open a simple console project and hit F12
## Expected Behavior ##
No errors are shown if intellisense is working fine.
## Actual Behavior ##
Intellisense works fine but errors regarding a pending code check request is shown in the developer console.
## Logs ##

![devconsole](https://user-images.githubusercontent.com/12554141/49109630-0cb7c900-f240-11e8-96d2-1e4df5d9386b.PNG)

## Environment information ##

**VSCode version**: 1.29.1
**C# Extension**: 1.17.1

<details><summary>Dotnet Information</summary>
.NET Core SDK (reflecting any global.json):
 Version:   2.1.500-preview-009398
 Commit:    14c7845f4b

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.500-preview-009398\

Host (useful for support):
  Version: 2.1.5
  Commit:  290303f510

.NET Core SDKs installed:
  2.0.2 [C:\Program Files\dotnet\sdk]
  2.1.2 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.100-preview-007326 [C:\Program Files\dotnet\sdk]
  2.1.100 [C:\Program Files\dotnet\sdk]
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.301 [C:\Program Files\dotnet\sdk]
  2.1.400-preview-009063 [C:\Program Files\dotnet\sdk]
  2.1.400-preview-009197 [C:\Program Files\dotnet\sdk]
  2.1.401 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.500-preview-009335 [C:\Program Files\dotnet\sdk]
  2.1.500-preview-009398 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #2710, opened 2018-11-27
Show the green flame when the server has actually started and do codecheck only when server is running

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2348, https://github.com/OmniSharp/omnisharp-vscode/issues/2709

We should fire the omnisharp started event when the "doConnect" promise has resolved which means that the omnisharp process has sent some output on the standard output. This would prevent the false appearance when omnisharp flame turns green and omnisharp process has not started(or waiting for debugger)

Also, the code check request would now only be sent if omnisharp is running. This would prevent error messages in the developer console of pending codecheck requests being cancelled.


src/features/diagnosticsProvider.ts
```diff
@@ -234,24 +234,26 @@ class DiagnosticsProvider extends AbstractSupport {
 
         let source = new vscode.CancellationTokenSource();
         let handle = setTimeout(() => {
-            serverUtils.codeCheck(this._server, { FileName: document.fileName }, source.token).then(value => {
+            if (this._server.isRunning()) {
```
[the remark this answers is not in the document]
akshita31: I realised that the makeRequest function to the server already checks if the server is running, so that would cover all the scenarios here. The main problem was that we were assuming that the server has been started even when it was not.

src/features/diagnosticsProvider.ts
```diff
@@ -249,9 +248,9 @@ class DiagnosticsProvider extends AbstractSupport {
 
                 // (re)set new diagnostics for this document
                 let diagnostics = quickFixes.map(DiagnosticsProvider._asDiagnostic);
-
                 this._diagnostics.set(document.uri, diagnostics);
-            });
+            }
+            catch (error) { }
```
rchande: Can we be consistent about catch block formatting?

src/features/documentSymbolProvider.ts
```diff
@@ -16,11 +16,14 @@ import { toRange3 } from '../omnisharp/typeConversion';
 export default class OmnisharpDocumentSymbolProvider extends AbstractSupport implements vscode.DocumentSymbolProvider {
 
     async provideDocumentSymbols(document: vscode.TextDocument, token: vscode.CancellationToken): Promise<vscode.DocumentSymbol[]> {
-        const response = await serverUtils.codeStructure(this._server, { FileName: document.fileName }, token);
-        
-        if (response && response.Elements) {
-            return createSymbols(response.Elements);
+        try {
+            const response = await serverUtils.codeStructure(this._server, { FileName: document.fileName }, token);
```
rchande: Why are these sometimes const and sometimes let?

src/features/documentSymbolProvider.ts
```diff
@@ -16,11 +16,14 @@ import { toRange3 } from '../omnisharp/typeConversion';
...
+
+            if (response && response.Elements) {
+                return createSymbols(response.Elements);
+            }
         }
+        catch (error) { }
```
rchande: Typically people put the return inside the catch block

src/features/implementationProvider.ts
```diff
@@ -19,6 +19,6 @@ export default class CSharpImplementationProvider extends AbstractSupport implem
             }
 
             return response.QuickFixes.map(fix => toLocation(fix));
-        });
+        }).catch();
```
rchande: Can we write these all the same way?

src/features/diagnosticsProvider.ts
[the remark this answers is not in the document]
akshita31: Added error handling when the rejected requests are received.

src/features/workspaceSymbolProvider.ts
```diff
@@ -28,11 +28,15 @@ export default class OmnisharpWorkspaceSymbolProvider extends AbstractSupport im
             return [];
         }
 
-        return serverUtils.findSymbols(this._server, { Filter: search, MaxItemsToReturn: maxItemsToReturn, FileName: '' }, token).then(res => {
+        try {
+            let res= await  serverUtils.findSymbols(this._server, { Filter: search, MaxItemsToReturn: maxItemsToReturn, FileName: '' }, token)
```
rchande: Spacing

src/features/codeLensProvider.ts
```diff
@@ -118,22 +120,26 @@ export default class OmniSharpCodeLensProvider extends AbstractProvider implemen
...
 
-        const result = await serverUtils.findUsages(this._server, request, token);
-
-        if (!result || !result.QuickFixes) {
-            return;
-        }
+        try {
+            let result = await serverUtils.findUsages(this._server, request, token);
+            if (!result || !result.QuickFixes) {
+                return;
```
rchande: This isn't a void method, don't you have to return something?

src/features/codeLensProvider.ts
```diff
@@ -147,9 +153,15 @@ export default class OmniSharpCodeLensProvider extends AbstractProvider implemen
...
         }
-        
-        const projectInfo = await serverUtils.requestProjectInformation(this._server, { FileName: codeLens.fileName });
 
+        let projectInfo: protocol.ProjectInformationResponse;
+        try {
+            projectInfo = await serverUtils.requestProjectInformation(this._server, { FileName: codeLens.fileName });
+        }
+        catch (error) {
+            return;
```
rchande: Not void, don't you have to return something?

src/features/implementationProvider.ts
akshita31: The function here is returning a ProviderResult and not a promise hence I am unable to put an await here, becuase that would require the return type to be a promise

src/features/completionItemProvider.ts
```diff
@@ -101,7 +102,10 @@ export default class OmniSharpCompletionItemProvider extends AbstractSupport imp
             // for short completions (up to 1 character), treat the list as incomplete
             // because the server has likely witheld some matches due to performance constraints
             return new CompletionList(result, wordToComplete.length > 1 ? false : true);
-        });
+        }
+        catch (error) {
+            return;
```
rchande: Same...

src/features/definitionProvider.ts
```diff
@@ -60,6 +60,9 @@ export default class CSharpDefinitionProvider extends AbstractSupport implements
                     return new Location(uri, new Position(gotoDefinitionResponse.Line - 1, gotoDefinitionResponse.Column - 1));
                 });
             }
-        });
+        }
+        catch (error) {
+            return;
```
rchande: I'm not going to make the same comment again but can you go through and make sure we return something (even if null in non-void methods)?

akshita31:
@rchande I have tried to resolve your concerns here. Does it look ok now ?

PR #2711, opened 2018-11-28
Allow to control what is posted to the event stream based on if there are any downloads

akshita31:
As @gregg-miskelly stated [here](https://github.com/OmniSharp/omnisharp-vscode/pull/2707#discussion_r236859807), we should not show "Installing C# dependencies" if no actual dependencies are being downloaded.

Hence the "installRuntimeDependencies" function first queries if there are any valid packages that it needs to download. If yes, it posts the message to the event stream accordingly.


akshita31:
@rchande I added one.

Issue #2720, opened 2018-11-30
A workspace with a lot of CS files prevents projects from being detected


ckaczor:
I have a workspace with 10 Visual Studio solutions and the "select project" list only shows the first project in the workspace - I can't switch to any of the others.

I did some digging and the problem is that findLaunchTargets in launcher.ts does a find for project files and CS files but limits it to maxProjectResults (250 by default) files.  This means if the first directory/project has more than 250 files it will be the only one detected.  If the projects are smaller you might get the first few projects but once the limit is reached you won't get any more projects.

It doesn't seem efficient to just raise the limit since that will require even more directory scanning.

akshita31:
This is fixed by https://github.com/OmniSharp/omnisharp-vscode/pull/2721

PR #2721, opened 2018-11-30
Fix project detection to handle many CS files

ckaczor:
Fix for #2720 

It seems that project files should take priority over CS files - this solution will look for project files first (up to the limit) and then look for CS files (up to the limit) so the project files come first.

It might make sense to skip the CS detection altogether if project files are located but I wanted to keep everything as close to the original implementation as possible to not break any existing use cases.


rchande:
cc @akshita31

akshita31:
Looks good to me. Thanks!!

PR #2730, opened 2018-12-05
Remove the affected vscode package

akshita31:
As per : https://code.visualstudio.com/blogs/2018/11/26/event-stream, we need to update the vscode package

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1790,7 +1790,7 @@
     },
     "deep-assign": {
       "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


akshita31:
cc @NTaylorMullen

rchande:
@akshita31 I think you need to update package.lock.json? Your build has this error: 
```
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

PR #2742, opened 2018-12-11
Validate integrity of the downloaded packages

akshita31:
Issue: https://github.com/OmniSharp/omnisharp-vscode/issues/2694

1. Added an integrity parameter in the package.json for the packages to be able to specify the integrity check sum(using SHA256 algorithm). 
2. If there is no integrity check sum specified the validate function will return true.
3. Only if the check passes shall we unzip the files and put the install.Lock, hence if the user restarts vscode the required dependency will download again as the install.Lock will not be present


src/observers/CsharpLoggerObserver.ts
```diff
@@ -58,9 +58,23 @@ export class CsharpLoggerObserver extends BaseLoggerObserver {
...
+                this.handleDownloadRetry(<Event.DownloadRetry>event);
+                break;
+            case Event.CorruptedDownloadError.name:
+                this.handleCorruptedDownloadError(<Event.CorruptedDownloadError>event);
+                break;
         }
     }
 
+    private handleCorruptedDownloadError(event: Event.CorruptedDownloadError){
+        this.logger.appendLine(`There was a problem downloading ${event.packageDescription}. Some functionalities may not work as expected. Please restart vscode to retrigger the download or download the package manually from ${event.url}`);
```
rchande: functionalities -> features
vscode -> Visual Studio Code
What am I supposed to do if I download the package (as a user)?

src/observers/CsharpLoggerObserver.ts
```diff
@@ -58,9 +58,23 @@ export class CsharpLoggerObserver extends BaseLoggerObserver {
...
+                break;
         }
     }
 
+    private handleCorruptedDownloadError(event: Event.CorruptedDownloadError){
+        this.logger.appendLine(`There was a problem downloading ${event.packageDescription}. Some functionalities may not work as expected. Please restart vscode to retrigger the download or download the package manually from ${event.url}`); 
+    }
+
+    private handleDownloadRetry(event: Event.DownloadRetry) {
+        this.logger.appendLine(`Corrupt download obtained for package ${event.packageDescription}. Retrying..`);
```
rchande: Package `package` failed integrity check...?

src/observers/ErrorMessageObserver.ts
```diff
@@ -23,9 +22,15 @@ export class ErrorMessageObserver {
             case DotNetTestDebugStartFailure.name:
                 this.handleDotNetTestDebugStartFailure(<DotNetTestDebugStartFailure>event);
                 break;
+            case CorruptedDownloadError.name:
+                this.handleCorruptedDownloadError(<CorruptedDownloadError> event);
         }
     }
 
+    handleCorruptedDownloadError(event: CorruptedDownloadError): any {
```
rchande: Constants seem to be duplicated?

src/packageManager/isValidInstallation.ts
```diff
@@ -0,0 +1,42 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as stream from "stream";
+import * as crypto from "crypto";
+
+const hash = crypto.createHash('sha256');
+
+export async function isValidDownload(buffer: Buffer, integrity: string): Promise<boolean> {
```
rchande: Tests for this thing?

src/packageManager/downloadAndInstallPackages.ts
```diff
@@ -23,11 +24,26 @@ export async function downloadAndInstallPackages(packages: AbsolutePathPackage[]
...
-                installationStage = 'downloadAndInstallPackages';
-                let buffer = await DownloadFile(pkg.description, eventStream, provider, pkg.url, pkg.fallbackUrl);
-                await InstallZip(buffer, pkg.description, pkg.installPath, pkg.binaries, eventStream);
-                installationStage = 'touchLockFile';
-                await touchInstallFile(pkg.installPath, InstallFileType.Lock);
+                let count = 0;
+                while (count < 2) {
+                    count++;
+                    let buffer = await DownloadFile(pkg.description, eventStream, provider, pkg.url, pkg.fallbackUrl);
+                    if (await isValidDownload(buffer, pkg.integrity)) {
```
rchande: Do you think it would be useful to print a "validating..." message?

src/observers/CsharpLoggerObserver.ts
```diff
@@ -58,9 +58,23 @@ export class CsharpLoggerObserver extends BaseLoggerObserver {
             case Event.LatestBuildDownloadStart.name:
                 this.logger.appendLine("Getting latest OmniSharp version information");
                 break;
+            case Event.DownloadRetry.name:
```
rchande: Observer tests?

src/observers/ErrorMessageObserver.ts
akshita31: What are the constants you are referring to here ?

src/observers/CsharpLoggerObserver.ts
akshita31: I was thinking we can have a wiki link with instructions on how to setup the packages manually and direct the users there. However in that case also their download might be corrupted. Or we can skip the last line and just say restart vscode. Thoughts ?

akshita31:
logs

This is the way the download and install looks with these changes

src/packageManager/downloadAndInstallPackages.ts
```diff
@@ -23,11 +24,28 @@ export async function downloadAndInstallPackages(packages: AbsolutePathPackage[]
...
+                    if (isValidDownload(buffer, pkg.integrity)) {
+                        eventStream.post(new IntegrityCheckSuccess());
+                        await InstallZip(buffer, pkg.description, pkg.installPath, pkg.binaries, eventStream);
+                        installationStage = 'touchLockFile';
+                        await touchInstallFile(pkg.installPath, InstallFileType.Lock);
+                        break;
+                    }
+                    else {
+                        if (count == 1) {
+                            eventStream.post(new IntegrityCheckFailure(pkg.description, pkg.url, true));
```
rchande: Maybe:
```
bool shouldRetry = count == 1;
eventStream.post(new IntegrityCheckFailure(pkg.description, pkg.url, shouldRetry)
```

test/unitTests/OmnisharpDownloader.test.ts
```diff
@@ -59,14 +59,16 @@ suite('OmnisharpDownloader', () => {
     });
 
     test('Events are created', async () => {
```
rchande: Test for the "bad download -> good download" and "bad download -> bad download>" cases?

rchande:
@akshita31 Should we really be printing "Integrity check succeeded" for downloads that don't have a checksum?

test/unitTests/OmnisharpDownloader.test.ts
akshita31: Added the test in the download validation

akshita31:
Note: I was facing difficulty in writing the test that uses the actual "isValidDownload" function in the downloadAndInstallpackages because I was not seeing a consistent buffer hash. Hence I have filed an issue here : https://github.com/archiverjs/node-archiver/issues/359

src/observers/CsharpLoggerObserver.ts
akshita31: I think for now I will just say "restart vscode" and if we people complain that they are getting these errors repeatedly we can write a wiki article that:

1. Helps them to identify their target OS
2. Identify which dependency are they missing
3. What is the link to download that dependency
4. How to set the appropriate permissions(on Linux/Mac)

akshita31:
cc @NTaylorMullen

NTaylorMullen:
To verify I'm understanding it all correctly. The integrity is the checksum of the zip file itself that's downloaded?

akshita31:
@NTaylorMullen Yes that is right

NTaylorMullen:
That sounds great! We'll have to start generating one for Razor 😄

Issue #2743, opened 2018-12-13
Deprication of DebugAdapterExecutableCommand


WardenGnaw:
> Please note: With the finalization of this API, we are deprecating the old (proposed) API DebugConfigurationProvider.debugAdapterExecutable and the very old command-based approach (contribution point Debuggers, attribute: adapterExecutableCommand). Please move off those APIs as we plan to remove support for them in the next release.
See https://code.visualstudio.com/updates/v1_30#_debugging-api

This results in the following code being deprecated https://github.com/OmniSharp/omnisharp-vscode/blob/ad8fea90b445f439adf0e9bf179253b9aeceea63/src/features/commands.ts#L50

We will need to update and use [DebugAdapterDescriptorFactory ](https://github.com/Microsoft/vscode/blob/a152a5c7818a13666a5879b4450eb26a4d45fbde/src/vs/vscode.d.ts#L8471)

PR #2746, opened 2018-12-13
Updating debug adapter executable to new API

WardenGnaw:
https://code.visualstudio.com/updates/v1_30#_debugging-api has
depricated adapterExecutableCommand. This PR updates the extension to
use DebugAdapterDescriptorFactory.

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/2743

### DIFF
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -48,7 +48,7 @@ async function checkForInvalidArchitecture(platformInformation: PlatformInformat
     return false;
 }
 
-async function completeDebuggerInstall(platformInformation: PlatformInformation, eventStream: EventStream): Promise<boolean> {
+export async function completeDebuggerInstall(platformInformation: PlatformInformation, eventStream: EventStream): Promise<boolean> {
     return _debugUtil.checkDotNetCli()
         .then(async (dotnetInfo: DotnetInfo) => {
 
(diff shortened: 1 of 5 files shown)


src/coreclr-debug/activate.ts
```diff
@@ -106,50 +106,63 @@ function showDotnetToolsWarning(message: string): void {
...
-                eventStream.post(new DebuggerNotInstalledFailure());
-                throw new Error('Failed to complete the installation of the C# extension. Please see the error in the output window below.');
+    constructor(platformInfo: PlatformInformation, eventStream: EventStream, packageJSON: any, extensionPath: string) {
+        this.platformInfo = platformInfo;
+        this.eventStream = eventStream;
+        this.packageJSON = packageJSON;
+        this.extensionPath = extensionPath;
+    }
+
+    createDebugAdapterDescriptor(session: vscode.DebugSession, executable: vscode.DebugAdapterExecutable | undefined): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
```
chuckries: >| undefined [](start = 105, length = 11)

why this?

WardenGnaw: Took defn from https://github.com/Microsoft/vscode/blob/3e560c1b1ed9c58d7d3871f048b035474f4f1e85/src/vs/vscode.d.ts#L8489

chuckries: I think you can remove it. The example here doesn't have it: [https://code.visualstudio.com/updates/v1_30](https://code.visualstudio.com/updates/v1_30)

---
In reply to: [241589154](https://github.com/OmniSharp/omnisharp-vscode/pull/2746#discussion_r241589154) [](ancestors = 241589154)

src/coreclr-debug/activate.ts
```diff
@@ -106,50 +106,63 @@ function showDotnetToolsWarning(message: string): void {
...
+        if (!CoreClrDebugUtil.existsSync(util.debugAdapterDir())) {
+            // our install.complete file does not exist yet, meaning we have not completed the installation. Try to figure out what if anything the package manager is doing
+            // the order in which files are dealt with is this:
+            // 1. install.Begin is created
+            // 2. install.Lock is created
+            // 3. install.Begin is deleted
+            // 4. install.complete is created
+
+            // install.Lock does not exist, need to wait for packages to finish downloading.
+            let installLock = new Promise<boolean>((resolve, reject) => { resolve(false); });
```
chuckries: >new Promise<boolean>((resolve, reject) => { resolve(false); }); [](start = 30, length = 63)

I believe `Promise.resolve(false)` is shorthand for this.

src/coreclr-debug/activate.ts
WardenGnaw: Technically, since we don't have any executable defined in the package.json. It will be undefined every time.

chuckries: oic.

---
In reply to: [241590911](https://github.com/OmniSharp/omnisharp-vscode/pull/2746#discussion_r241590911) [](ancestors = 241590911)

src/coreclr-debug/activate.ts
```diff
@@ -48,7 +48,7 @@ async function checkForInvalidArchitecture(platformInformation: PlatformInformat
     return false;
 }
 
-async function completeDebuggerInstall(platformInformation: PlatformInformation, eventStream: EventStream): Promise<boolean> {
+export async function completeDebuggerInstall(platformInformation: PlatformInformation, eventStream: EventStream): Promise<boolean> {
```
chuckries: >export [](start = 0, length = 6)

is this necessary? am I missing something?

WardenGnaw: oops, leftover from when I created the class in another file.

src/coreclr-debug/activate.ts
```diff
@@ -106,50 +106,72 @@ function showDotnetToolsWarning(message: string): void {
...
-        }
+            else
+            {
+                resolve();
+            }
+        }).then(() => {
+            // debugger has finished installation, kick off our debugger process
+            return new vscode.DebugAdapterExecutable(path.join(common.getExtensionPath(), ".debugger", "vsdbg-ui" + CoreClrDebugUtil.getPlatformExeExtension()));
+        }, error => {
+            throw error;
```
chuckries: >throw error [](start = 12, length = 11)

if you're using this style, I think you probably want to `reject(error)` here, since you have the rejection handler from your new Promise.

src/coreclr-debug/activate.ts
```diff
@@ -106,50 +106,72 @@ function showDotnetToolsWarning(message: string): void {
     }
 }
 
-interface AdapterExecutableCommand {
-    command: string;
-}
+export class DebugAdapterDecriptor implements vscode.DebugAdapterDescriptorFactory {
```
gregg-miskelly: 1. `DebugAdapterDecriptor` feels like a bad name to me. How about 'ClrDebugDescriptorFactory' or something like that?
2. Is there a reason we left this in activate.ts? Since it in a new class it feels better to put this in its own .ts file to me.

src/coreclr-debug/activate.ts
```diff
@@ -106,50 +106,72 @@ function showDotnetToolsWarning(message: string): void {
...
+    private extensionPath: string;
+
+    constructor(platformInfo: PlatformInformation, eventStream: EventStream, packageJSON: any, extensionPath: string) {
+        this.platformInfo = platformInfo;
+        this.eventStream = eventStream;
+        this.packageJSON = packageJSON;
+        this.extensionPath = extensionPath;
+    }
+
+    createDebugAdapterDescriptor(session: vscode.DebugSession, executable: vscode.DebugAdapterExecutable | undefined): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
```
gregg-miskelly: This function feels complicated enough to me that we should really write it using async await instead.

src/coreclr-debug/activate.ts
WardenGnaw: And I need it again because I moved the class to a different .ts file.

JoeRobich:
Closing. This PR supersceded by https://github.com/OmniSharp/omnisharp-vscode/pull/3912

PR #2753, opened 2018-12-19
Update the issue template to use the issue reporter

akshita31:


ISSUE_TEMPLATE
```diff
@@ -1,3 +1,5 @@
+Please prefer filing issues through the in-built extension command `CSharp: Start authoring a new issue on GitHub`. Know more - https://github.com/OmniSharp/omnisharp-vscode/wiki/Reporting-Issues.
```
gregg-miskelly: How about instead:
```
<!-- To prefill this information:
1. Open Visual Studio Code
2. Bring up the command palette (press <kbd>F1</kbd>)
3. Type `csharp.ReportIssue`
-->
```

gregg-miskelly: BTW: I choose that language because it is similar to what VS Code uses in its template -- https://github.com/Microsoft/vscode/issues/new?template=bug_report.md

gregg-miskelly: Actually, when I try this steps, the report issue command doesn't show up for me in C# extension version 1.17.1. Any idea why?

gregg-miskelly: Never mind. Figured it out.

rchande: +1

PR #2766, opened 2018-12-28
Fix 'npm run gulp updatePackageDependencies'

gregg-miskelly:
This commit improves the updatePackageDependencies gulp task so that hopefully it will work for all three teams needing to update package.json to point at new runtime dependancies in hopefully a safe way.

Fixes/improvements:
* Make the task work again since gulp no longer likes the command line argument hack that we used to have
* Add support for computting checksums
* Add verification that the primary and fallback URLs have the same hash and both work

### DIFF
--- a/src/tools/UpdatePackageDependencies.ts
+++ b/src/tools/UpdatePackageDependencies.ts
@@ -6,33 +6,47 @@
 import * as fs from 'fs';
 import * as os from 'os';
 import { Package } from '../packageManager/Package';
-
+import { DownloadFile } from '../packageManager/FileDownloader';
+import { EventStream } from '../EventStream';
+import * as Event from "../omnisharp/loggingEvents";
+import NetworkSettings, { NetworkSettingsProvider } from '../NetworkSettings';
+import { getBufferIntegrityHash } from '../packageManager/isValidDownload';
 
 interface PackageJSONFile
 {
     runtimeDependencies : Package[];
 }
 
-export function updatePackageDependencies() {
+export async function updatePackageDependencies() : Promise<void> {
 
-    const urlsIndex = process.argv.indexOf("--urls");
-    const newPrimaryUrls = urlsIndex >= 0 ? process.argv[urlsIndex+1] : undefined;
+    const newPrimaryUrls = process.env["NEW_DEPS_URLS"];
+    const newVersion = process.env["NEW_DEPS_VERSION"];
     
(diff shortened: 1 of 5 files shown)


src/tools/UpdatePackageDependencies.ts
```diff
@@ -6,33 +6,47 @@
...
-    if (newPrimaryUrls.length === 0) {
-        throw new Error("Invalid first argument to updatePackageDependencies. URL string argument expected.");
+    const newPrimaryUrlArray = newPrimaryUrls.split(',');
+    for (let urlToUpdate of newPrimaryUrlArray) {
+        if (!urlToUpdate.startsWith("https://")) {
+            throw new Error("Unexpected 'NEW_DEPS_URLS' value. All URLs should start with 'https://'.");
+        }
     }
 
+    if (! /^[0-9]+\.[0-9]+\.[0-9]+$/.test(newVersion)) {
```
akshita31: We could also do a semver check here

src/tools/UpdatePackageDependencies.ts
```diff
@@ -54,23 +68,79 @@ export function updatePackageDependencies() {
...
+            }
+        }
+    }
+
+    // Next take another pass to try and update to the URL  
+    const eventStream = new EventStream();
+    eventStream.subscribe((event: Event.BaseEvent) => {
+        switch (event.constructor.name) {
+            case Event.DownloadFailure.name:
+                console.log("Failed to download: " + (<Event.DownloadFailure>event).message);
```
akshita31: Like in the other places in the code, we could use a single eventStream that is passed to all these functions and in this file we could post the appropriate events to this stream and separately create a "ConsoleObserver" or "UpdatePackageDependencies" observer that subscribes to this event stream and does the logging accordingly.

src/tools/UpdatePackageDependencies.ts
gregg-miskelly: I don't think there is anything in 'semver' that supports searching for a semver in a string, and since we always want the input version to be subsequently searchable, I don't think it makes sense to use the semver library for this.

src/tools/UpdatePackageDependencies.ts
gregg-miskelly: While I think the eventStream pattern that we created in other places make a lot of sense for enabling unit tests, I don't really see a benefit in this case since unit tests don't seem to make sense for a gulp task that produces output that would be code reviewed.

PR #2779, opened 2019-01-05
UpdateDependencies to be able to update omnisharp and razor packages

akshita31:
The UpdateDependencies script used to match the last part of the url to match the urls against the runtimeDependencies. However the urls for omnisharp and razor include the version as well, hence the names would never directly match, so added some logic that trims the versions before performing the matching.

Also update the version in the installPath, installTestPath and the defaults property which is present in the package.json

### DIFF
--- a/src/tools/UpdatePackageDependencies.ts
+++ b/src/tools/UpdatePackageDependencies.ts
@@ -95,7 +95,7 @@ export async function updatePackageDependencies(): Promise<void> {
     const networkSettingsProvider: NetworkSettingsProvider = () => new NetworkSettings(/*proxy:*/ null, /*stringSSL:*/ true);
 
     const downloadAndGetHash = async (url: string): Promise<string> => {
-        console.log(`Downlodaing from '${url}'`);
+        console.log(`Downloading from '${url}'`);
         const buffer: Buffer = await DownloadFile(url, eventStream, networkSettingsProvider, url, null);
         return getBufferIntegrityHash(buffer);
     };
@@ -143,7 +143,7 @@ function replaceVersion(value: string, newVersion: string): string {
     dottedVersionRegExp.lastIndex = 0;
 
     if (!regex.test(value)) {
-        return value; //If the string doesnt contain any version return the same string
+        return value; //If the string doesn't contain any version return the same string
     }
 
     return value.replace(regex, newValue);
@@ -169,10 +169,10 @@ function verifyMatchCount(value: string, shouldContainVersion = false): void {
     const matchCount: number = dottedMatches + dashedMatches;
 
     if (shouldContainVersion && matchCount == 0) {
-        throw new Error(`Version number not found in fallback URL '${value}'.`);
+        throw new Error(`Version number not found in '${value}'.`);
     }
     if (matchCount > 1) {
-        throw new Error(`Ambiguous version pattern found in fallback URL '${value}'. Multiple version strings found.`);
+        throw new Error(`Ambiguous version pattern found in '${value}'. Multiple version strings found.`);
     }
 }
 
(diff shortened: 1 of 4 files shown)


akshita31:
@NTaylorMullen I tried to run this script using 
a razor url from one of the previous PR's - "https://download.visualstudio.microsoft.com/download/pr/3b0ae709-c067-48b4-a658-549ca972e437/e4808fd370ba60d528944a5a36543cf0/razorlanguageserver-osx-x64-1.0.0-alpha2-20181112.3.zip1.0.0-alpha1-20181019.5". The script is currently dependent on regex matching and can run only if we have specified a release version in the `NEW_DEPS_VERSION` argument in launch.json

NTaylorMullen:
Could you share some of the context around this script? Not quite sure I understand all this goodness

NTaylorMullen:
Ah, I see. Ya we don't currently have a fallback url available for the last release (blame Azure Pipelines retention policy 😢). Am I reading this correctly that this also calculates the checksum for dependencies? If so that's doppppeee

gregg-miskelly:
@NTaylorMullen Correct. The script will -
* Make sure all the new URLs actually work
* Update the integrity hashes to match the new URLs
* Make sure the fall back URLs and the CDN URLs have the same content

Next time you update, I would recommend adding a fall back URL on your blob storage. It has saved us a few times in the past when the CDN started misbehaving...

src/tools/UpdatePackageDependencies.ts
```diff
@@ -155,7 +137,55 @@ export async function updatePackageDependencies() : Promise<void> {
...
+    dottedVersionRegExp.lastIndex = 0;
+
+    if (!regex.test(value)) {
+        return value; //If the string doesn't contain any version return the same string
+    }
+
+    return value.replace(regex, newValue);
+}
+
+function verifyMatchCount(value: string, shouldContainVersion = false): void {
```
gregg-miskelly: nit: If I didn't write the code, I am not sure I would understand what "verifyMatchCount" does without reading the code. How about "verifyVersionSubstringCount"?

NTaylorMullen:
> Next time you update, I would recommend adding a fall back URL on your blob storage. It has saved us a few times in the past when the CDN started misbehaving...

Definitely on our plate for the next release 😄 https://github.com/aspnet/Razor.VSCode/issues/257

src/tools/UpdatePackageDependencies.ts
```diff
@@ -155,7 +137,55 @@ export async function updatePackageDependencies() : Promise<void> {
...
 }
 
-function getLowercaseFileNameFromUrl(url : string) : string {
+
+function replaceVersion(value: string, newVersion: string): string {
+    if (!value) {
+        return value; //if the value is null or undefined return the same one
+    }
+
+    let regex: RegExp = dottedVersionRegExp;
```
rchande: This code is working really hard to prevent a version number from changing from dashed to dotted. How important is it to enforce that?

src/tools/UpdatePackageDependencies.ts
```diff
@@ -155,7 +137,55 @@ export async function updatePackageDependencies() : Promise<void> {
     fs.writeFileSync('package.json', content);
 }
 
-function getLowercaseFileNameFromUrl(url : string) : string {
+
+function replaceVersion(value: string, newVersion: string): string {
```
rchande: value -> filename

src/tools/UpdatePackageDependencies.ts
gregg-miskelly: This part of the code isn't new. And it is extremely important - if it didn't do this, the resulting URL would be wrong.

rchande: Ok, cool.

Issue #310, opened 2016-05-11
Provide project selector to pick startup project


DustinCampbell:
Picking a new startup project, should modify the launch.json so that F5 launches the project output under the debugger.

gregg-miskelly:
Here is an idea -- instead of asking the user which project they want as the startup project (or maybe in addition to asking), maybe we should generate a different entry in launch.json and tasks.json for each launchable project. That way the user can easily switch around the startup project using the drop down built into VS Code.

gregg-miskelly:
Here is how I decided to make this work --
1. When we generate launch.json, if there are multiple startup projects, the user will now get asked which one they want to launch.
2. If a user decides to make a different decision later, they can always rerun the '.NET: Generate Assets for Build and Debug' command.

Issue #2703, opened 2018-11-25
Missing Launch.json Auto-Creation for dotnetcore classlib template


parsalotfy:
## Environment data
`dotnet --info` output:
**.NETCore** SDK version 2.1.400
**VSCode** version 1.29.1 (latest in time of writing this)
**CsharpExtension** version 1.17.1
**OS** Windows x64

## Steps to reproduce
All the problem is creating, building and debugging the dot net core class library.

## Expected  behavior
when I create a **Console** project using this command : `dotnet new console -o A1` , the C# extension automatically create **bin** folder and ask me : 

> required assets to build and debug are missing from Console, Add them ?
![image](https://user-images.githubusercontent.com/30390810/48985065-27dbe900-f119-11e8-8bbb-85c53a7e9aac.png)

.

and after clicking yes, **the extension creates and configured launch.json and task.json for me** :

![image](https://user-images.githubusercontent.com/30390810/48985089-6671a380-f119-11e8-8776-93f6c2b0e66c.png)
 and after pressing F5 , it compiles and builds successfully.
**But ...**

## Actual behavior
But when I create class library using this command : `dotnet new classlib -o A1`
the extension does NOT show me any notification that ask me if I would like to add assets or not
**I mean the extension does NOT create Launch.json and Task.json for me** and clearly does not configure them.
![image](https://user-images.githubusercontent.com/30390810/48985258-e9472e00-f11a-11e8-98f6-dd77cf854cb3.png)
.
When I press F5 the command palette asks me to select my environment and I click on .Net Core
Then it generates Launch.json for me **BUT Wrongly** and when I press F5 to build it , I get this Error :

> Could not find the task 'build'

![image](https://user-images.githubusercontent.com/30390810/48985319-65417600-f11b-11e8-9864-b71374121ab6.png)

this is too annoying to have this behavior in this rich text editor and extension.
Thank you for your good Extension 🥇

gregg-miskelly:
@parsalotfy Thanks for writing up such good steps. However, this is actually how things are designed to work.

The issue is this: if you have a class lib project, then that project isn't launchable by itself -- you would need to add another project, such as a console app, or ASP.NET app which referenced the class library if you wanted to launch the project.

[...]

Issue #2673, opened 2018-11-13
Simplify generated launch.json 


isidorn:
Hey,

VSCode dev here. This milestone I am looking into simplifing generated `launch.json` for various extensions Microsoft/vscode#62851

The `launch.json` that C# generates is attached at the end. This is far too complex for the avarage user. I suggest to simplify it the following way:

- [ ] `DebugConfigurationProvider` should use the `quickPick` to ask the user if he would like to debug the console or a web app or to attach. After that it should ask what is the target framework and what is the project name if he did not choose attach. Optimally it should be able to detect those things on its own, but if not possible use `quickPick` to ask the user.
- [ ] Remove default fields which are not necessery like `internalConsoleOptions`, `args`, `cwd`
- [ ] Do users ever change anything in the `launchBrowser`, `env` and `sourceFileMap` properties. If no than these properties should be append in the resolveConfiguraiton calls. That way the user will not see any of this and there will be far less clutter
- [ ] You are already contributing `configuraitonSnippets` which is great. However these snippets also need to be simplified. I suggest to remove all attributes that have the default value. Example: `stopAtEntry`, `cwd`, `args`

If you agree with the suggestions I am making here I am also here to help with any potential questions you might have. The changes should not require a lot of work but will simplify the flow a lot imho. It should be much less complex and not too much like a wizard experience

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": ".NET Core Launch (console)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "build",
            "program": "${workspaceFolder}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll",
            "args": [],
            "cwd": "${workspaceFolder}",
            "console": "internalConsole",
            "stopAtEntry": false,
            "internalConsoleOptions": "openOnSessionStart"
        },
        {
            "name": ".NET Core Launch (web)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "build",
            "program": "${workspaceFolder}/bin/Debug/<insert-target-framework-here>/<insert-project-name-here>.dll",
            "args": [],
            "cwd": "${workspaceFolder}",
            "stopAtEntry": false,
            "internalConsoleOptions": "openOnSessionStart",
[...]

PR #2780, opened 2019-01-05
Improvements to launch.json/tasks.json

gregg-miskelly:
This commit contains many improvements to our launch.json and task.json generation. Highlights:

- Schema related changes:
  - Make most of the 'launchBrowser' attributes optional, remove the extra properties from the templates, and improve the schema
  - In the configuration resolver, add a default value for 'cwd' if we are local debugging

- Generator related fixes:
  - Use a quick pick to allow the user to select which project to launch. Before we would always launch the first one OmniSharp returned.
  - Remove 'internalConsoleOptions' from the generated templates, and instead default it in the configuration resolver.
  - Fix several problems with regenerating launch/tasks.json in the case that the file already existed. We had code to delete the files and recreate it, but the logic was wrong leading us to sometimes duplicate content or not create the file at all.
  - Switch to using the '$tsc' problem matcher. We were using the '$msCompile' matcher, but that assumes that file names will be absolute paths, which isn't what 'dotnet build' provides.
  - Stop supporting project.json based projects for purposes of generating launch/tasks.json. Continuing to support these now that we let the user pick the startup project was going to be more work than it made sense to support.
  - In the case of generating launch.json through the configuration provider, we will no longer generate a set of generic configurations in error cases. We will now either use a fall back configuration which is designed to feel like an error, or we will put up an error prompt and return nothing.
  - Allow generating a tasks.json even if there is no launchable project. Note that we will not automatically generate a tasks.json in this case as I wasn't sure people would really like this, but the 'Generate Assets' command will force it.

Testing: Verified open folders/workspaces with the following conditions:
- A single class library project
- A single console app
- A single ASP.NET app
- A folder containing many launchable projects
- A workspace with multiple folders, each of which has a launchable project

This fixes: #310, #2673 and #2703

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -20,12 +20,9 @@ export class AssetGenerator {
     public tasksJsonPath: string;
     public launchJsonPath: string;
 
-    private hasProject: boolean;
-    private projectPath: string;
-    private projectFilePath: string;
-    private targetFramework: string;
-    private executableName: string;
-    private configurationName: string;
+    private executeableProjects: protocol.MSBuildProject[];
+    private startupProject: protocol.MSBuildProject | undefined;
+    private fallbackBuildProject: protocol.MSBuildProject;
 
     public constructor(workspaceInfo: protocol.WorkspaceInformationResponse, workspaceFolder: vscode.WorkspaceFolder = undefined) {
         if (workspaceFolder) {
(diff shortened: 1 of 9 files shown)


src/tools/OptionsSchema.json
```diff
@@ -176,35 +176,26 @@
     },
     "LaunchBrowserPlatformOptions": {
       "type": "object",
+      "required": [ "command" ],
       "properties": {
         "command": {
           "type": "string",
-          "description": "The command to execute for launching the web browser",
+          "description": "The executable to launching which will start the web browser",
```
WardenGnaw: This sounds weird. Did you want `The executable which will start the web server`?

src/assets.ts
```diff
@@ -20,12 +20,9 @@ export class AssetGenerator {
...
     public launchJsonPath: string;
 
-    private hasProject: boolean;
-    private projectPath: string;
-    private projectFilePath: string;
-    private targetFramework: string;
-    private executableName: string;
-    private configurationName: string;
+    private executeableProjects: protocol.MSBuildProject[];
+    private startupProject: protocol.MSBuildProject|undefined;
```
WardenGnaw: nit: space around `|`

src/tools/OptionsSchema.json
gregg-miskelly: Sounds reasonable.

Issue #2756, opened 2018-12-20
Stop forcing focus on "Output" tab when OmniSharp is logging


yhvicey:
VS Code version: 1.30.1
C# Extension version: 1.17.1

Every time when I open a folder with lots of *proj file (like aspnet/Extensions), it keeps focusing on "Output" and the drop down list shows "OmniSharp Log". It is really annoying because it will interrupt other action, like opening a menu or opening a terminal. Please at least add a switch to disable it.

akshita31:
@yhvicey Thanks for filing the issue. I agree that we should not focus the logs by default.

stewartadam:
I presume this is also the cause of https://github.com/Microsoft/vscode/issues/22187, which I ran into again today. Midway through typing a Markdown document, OmniSharp started updating dependencies and grabbed focus away which is irritating.

BrennanConroy:
This is a super annoying issue. Is there a known previous version that people can use instead until this is fixed?

rchande:
@BrennanConroy Actually, we just created a beta that includes the fix for this. Please give it a try and let us know if you're still having issues: https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.18.0-beta5

BrennanConroy:
Excellent! Worked for me, thanks for the good work.

justinmchase:
It still does this... it keeps stealing focus from the terminal while I'm typing.

hoyosjs:
@akshita31 Sorry to ping on a closed issue, but it feels like this regressed. I have to kill omnisharp to be able to use the terminal in large projects where it takes a long time to create the full dependency graph and cache setup.

stewartadam:
@hoyosjs probably better to report a new issue if it's happening again

JoeRobich:
Please set the "csharp.showOmnisharpLogOnError" setting to `false` if you do not want this behavior.

https://github.com/OmniSharp/omnisharp-vscode/blob/5270351278c86cb2514487ff6f6c0adf27c15a7a/package.json#L693-L697

justinmchase:
I want omnisharp errors to be logged, but I don't want the frickin' errors to change my focus.

filipw:
that's what the setting does 🙂

justinmchase:
Oh 🥰 

It sounds like its saying the errors won't get logged at all but I see what you're sayin'

TheOnlyWei:
[...]

PR #2828, opened 2019-01-28
Do not steal the focus when showing the channels

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2756

### DIFF
--- a/test/unitTests/logging/OmnisharpChannelObserver.test.ts
+++ b/test/unitTests/logging/OmnisharpChannelObserver.test.ts
@@ -11,14 +11,19 @@ suite("OmnisharpChannelObserver", () => {
 
     let hasShown: boolean;
     let hasCleared: boolean;
+    let preserveFocus: boolean;
     let observer: OmnisharpChannelObserver;
 
     setup(() => {
         hasShown = false;
         hasCleared = false;
+        preserveFocus = false;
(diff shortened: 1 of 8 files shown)


test/unitTests/logging/OmnisharpChannelObserver.test.ts
```diff
@@ -11,14 +11,19 @@ suite("OmnisharpChannelObserver", () => {
...
     let observer: OmnisharpChannelObserver;
 
     setup(() => {
         hasShown = false;
         hasCleared = false;
+        preserveFocus = false;
         observer = new OmnisharpChannelObserver({
             ...getNullChannel(),
-            show: () => { hasShown = true; },
+            show: (preserve) => {
```
rchande: Do we ever call this with false?

akshita31: Is there a case where we want to steal the focus

akshita31: As of now I couldnt think of any particular case hence right now we are using true everywhere

rchande: Why not make it even easier and remove the paramter/always pass true in this method?

filipw:
this is great - thanks!

Issue #2817, opened 2019-01-26
Extension causes high cpu load


Obscuriosity:
[ms-vscode.csharp-unresponsive.cpuprofile.txt](https://github.com/OmniSharp/omnisharp-vscode/files/2799017/ms-vscode.csharp-unresponsive.cpuprofile.txt)

- Issue Type: `Performance`
- Extension Name: `csharp`
- Extension Version: `1.17.1`
- OS Version: `Linux x64 4.15.0-43-generic`
- VSCode version: `1.30.2`

:warning: Make sure to **attach** this file from your *home*-directory: `/home/tony/ms-vscode.csharp-unresponsive.cpuprofile.txt` :warning:

Find more details here: https://github.com/Microsoft/vscode/wiki/Explain:-extension-causes-high-cpu-load

akshita31:
@Obscuriosity Thanks for filing an issue. We are working on it.

akshita31:
@Obscuriosity  The [beta version](https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.18.0-beta7) containing the fix for this issue is out. You can install it by following the instructions [here](https://github.com/OmniSharp/omnisharp-vscode/wiki/Installing-Beta-Releases).

We believe this would fix the issue. Please try it out.

akshita31:
@Obscuriosity  We haven't heard back from you in a while so I will close this issue for now. Please feel free to reopen if the problem still persists.

Issue #2757, opened 2018-12-21
Extension causes high cpu load


SinueeHernandez:
[ms-vscode.csharp-unresponsive.cpuprofile.txt](https://github.com/OmniSharp/omnisharp-vscode/files/2701493/ms-vscode.csharp-unresponsive.cpuprofile.txt)
- Issue Type: `Performance`
- Extension Name: `csharp`
- Extension Version: `1.17.1`
- OS Version: `Linux x64 4.15.0-43-generic`
- VSCode version: `1.30.1`

akshita31:
@rchande Can this be related to the similar thing that we were observing on MacOS ?

jrieken:
@akshita31 From looking at the profile this seems to be code loading. Using webpack or rollup would certainly help here

akshita31:
@jrieken We are working on using webpack for the C# extension. However, I had a question. how would you test that the bundle produced doesnt have any missing dependencies. I have run into cases where my integration tests and unit tests run fine but when i bundle the code and install it in vscode, I get some errors due to which the extension doesnt run

jrieken:
> I have run into cases where my integration tests and unit tests run fine

Those test, do they use the bundled output? We have learnt that webpack is a little too relaxed with error messages and internally we treat every warning as error so that they trigger a build failure.

akshita31:
@SinueeHernandez  Can you try upgrading to the latest mono version as suggested [here](https://github.com/OmniSharp/omnisharp-vscode/issues/2794#issuecomment-457050038) and checking if the problem still persists.
Also, we have a [beta release](https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.18.0-beta7) containing major improvements on the code load times. You can install it by following the instructions [here](https://github.com/OmniSharp/omnisharp-vscode/wiki/Installing-Beta-Releases). Please let us know if that helps.

akshita31:
@SinueeHernandez We havent heard from you in a while. I will close this issue now. Please feel free to reopen if the problem persists.

PR #2835, opened 2019-02-01
Use webpack to reduce the code loading times

akshita31:
This PR enables the usage of webpack to bundle the extension, so that the corresponding js produced and the code load times are lesser, to fix issues like : https://github.com/OmniSharp/omnisharp-vscode/issues/2817, https://github.com/OmniSharp/omnisharp-vscode/issues/2757.

1. While working, we were facing problems that when webpack is used in production, the names of the classes were removed from the js files and hence the type checking we had in place for the logging pipeline(which was based on names of the classes) was falling apart, hence we have now modified the Observers to match based on enum values.
2. This will reduce the extension size to around 300Kb from current 3Mb 💃 
3. Now in travis we dont need to unpack the extension we can run our tests against the dist folders directly, hence I have removed the `gulp vsix:release:unpackage` step and also the --codeExtensionPath flag being passed to the gulp tasks.

Also, this PR updates the vscode tasks to version 2.0.


akshita31:
Through this work we were able to reduce the code load time from ~430 ms to 40ms. 
This is a screenshot for the activation time of the extension for a simple workspace for extension version 1.17
image

 And here is the same for the 1.18-beta7
beta7_cpu_profile

PR #2873, opened 2019-02-24
Fadeout based on roslyn tag 'unnecessary'.

savpek:
Based on original PR #2646 from @akshita31 

Requires https://github.com/OmniSharp/omnisharp-roslyn/pull/1410 and https://github.com/OmniSharp/omnisharp-roslyn/pull/1076 to be fully functional.

~~Without backend support (previous PR:s), this has few hard coding to fade most common situations without full blown analysis:~~ (decided not to hard code anything at this point, requires analyzers.)
image

With backend support (theres Roslynator package that fadeouts some additional items from image):
image

Theres some mapping when fadeout occurs because VSCode doesn't support 'hidden' with fadeout, so those items must be converted to hints or fadeout doesn't work on them.


savpek:
This is ready for review @rchande @akshita31 

Requires latest version of omnisharp-roslyn master branch and analyzers enabled.

filipw:
I would recommend to wait with this until there is official omnisharp-roslyn with the analyzers feature + that version is consumed by the C# extension.
Otherwise there is a setting that only works for `"omnisharp.path": "latest"` which is a bit confusing.

src/features/diagnosticsProvider.ts
```diff
@@ -319,36 +324,50 @@ class DiagnosticsProvider extends AbstractSupport {
         });
     }
 
-    private static _shouldInclude(quickFix: protocol.QuickFix): boolean {
-        const config = vscode.workspace.getConfiguration('csharp');
```
akshita31: Shouldnt we retain this check here ?

src/features/diagnosticsProvider.ts
```diff
@@ -275,25 +283,22 @@ class DiagnosticsProvider extends AbstractSupport {
...
                     .sort((a, b) => a.FileName.localeCompare(b.FileName));
 
                 let entries: [vscode.Uri, vscode.Diagnostic[]][] = [];
                 let lastEntry: [vscode.Uri, vscode.Diagnostic[]];
 
-                for (let quickFix of quickFixes) {
-
-                    let diag = DiagnosticsProvider._asDiagnostic(quickFix);
-                    let uri = vscode.Uri.file(quickFix.FileName);
+                for (let quickFix of this._mapQuickFixesAsDiagnostics(quickFixes)) {
```
akshita31: Sorry for nitpicking but this is confusing me a bit. We are calling mapQuickFixesAsDiagnostics then again calling it quickFix.

akshita31:
@savpek Is it possible to add a test for this ?

akshita31:
@savpek I have a question. Will this require "enableRolsynAnalyzer" to be set to true?

src/features/diagnosticsProvider.ts
savpek: > Sorry for nitpicking but this is confusing me a bit. We are calling mapQuickFixesAsDiagnostics then again calling it quickFix.

I don't think that was nitpicking, naming is important and that is definetly mistake by me 😄

savpek:
> @savpek I have a question. Will this require "enableRolsynAnalyzer" to be set to true?

Yes only certain analysis hides things, for example 'unused using' has CSxxxx diagnostic and another hidden analyzer result with 'unnecessary' tag that actually hides it. This is likely some kind of usability optimization by roslyn but i don't know it's background. I think none of non roslyn analyzer diagnostic has 'unnecessary' tag. There was few hard codings for most common cases on original version, removed at commit https://github.com/OmniSharp/omnisharp-vscode/pull/2873/commits/e01ce48f3a8483574651aeb7c76c6fda85d6821d, should they be retained?

> @savpek Is it possible to add a test for this ?

Sure thing, i think i can manage to write tests next week depending if there is existing diagnostic tests around that can be used as base implementation.

src/features/diagnosticsProvider.ts
savpek: I think i removed it because it conflicts with new feature, many of 'unnecessary' diagnostics are 'hidden' by default (they should not show any other thing in UI except faded text, for example should not be visible in problems list etc). If they were supressed -> faded didn't work. 

VScode doesn't seem to have 'hidden' level at all, however if i set level as 'hint' and add fading tag on vscode diagnostic it isn't visible in any way except it's faded. 

Do you know origin of suppressHiddenDiagnostics? Originally it seemed be so that hidden was mapped as information diagnostics (which is bit strange, it is meant to be hidden but then it was elevated as visible) and this flag suppressed all hidden diagnostics.

savpek: Based on https://github.com/OmniSharp/omnisharp-vscode/issues/1429 `suppressHiddenDiagnostics` is solely to provide some kind of support for unused diagnostics before VSCode implemented faded out code.

I propose that lets remove `suppressHiddenDiagnostics` at this point and add hard coding for unused using. This way end user get better experience (fade out without analyzer support instead of warning) and it doesn't conlflict with version analyzers enabled. Does this shoud good idea @akshita31 ? With ruleset support user can configure any error level for unused usings if they want so (error even), i think this is on same line with visual studio behavior.

savpek:
@akshita31 is there easy way to get CI print omnisharp-roslyn server logs during test run? It looks like theres something broken since diagnostic return errors like `Predefined type \'System.Object\' is not defined or imported (CS0518)`? This doesn't occur when i ran tests on my computer. Usually that kind of stuff pops up when restore isn't made or it's broken 🤔

src/features/diagnosticsProvider.ts
savpek: image

Ended up following: Instead of promoting hidden as information it now promotes it 'hint'. This can be suppressed `suppressHiddenDiagnostics `. 

This gives meaningfull feature, now its easy to see for example where code actions are available. In same time it's suppressed as default because it causes kind of lot of clutter around -> is optional for those who like it. Suppression was default on original version too.

This way it doesn't conflict with analytics since they really don't have 'hint' level at all, it's just shows them now as small gray dots on screen.

akshita31:
@savpek I cloned your branch locally and I am getting failure for a different reason:

 DiagnosticProvider: single csproj at root of workspace
       Return fadeout diagnostics like unused usings:
     AssertionError: expected 'The variable \'notUsed\' is assigned but its value is never used (CS0219) [singleCsproj]|Unnecessary using directive. (CS8019) [singleCsproj]' to contain 'IDE0005'

akshita31:
The behavior is flaky :/ The tests sometime pass for me sometimes not

savpek:
Damn :/ I will try fix problems during weekend.

Can help debug it littlebit since i think that doesn't happen on my machine. For error it looks like that omnisharp-roslyn starts without analyzers enabled for some reason, like it doesn't use enableRoslynAnalyzers at all from time to time.

Can you check on failed test run that is there `Solution initialized -> queue all documents for code analysis. Initial document count` like line in omnisharp-roslyn logs? If it's missing its marker that it indeed starts without analyzers enabled for some reason.

If thats the case i could try set enabling flag directly on settings.json at test assets, if it still fails in .omnisharp.json, if tests passes after that theres something strange going on when enabling flag is passed to server at command line ❓ 

CI server problem is different, i will try it on linux machine if it can be reproed there.

akshita31:
@savpek You were right that the problem that is coming in CI comes when the workspace is not restored.
However we are explicitly calling testAssetWorkspace.restore() so the workspace should be restored.

src/features/diagnosticsProvider.ts
```diff
@@ -319,36 +324,59 @@ class DiagnosticsProvider extends AbstractSupport {
...
-            return true;
+    private _asDiagnostic(quickFix: protocol.QuickFix): vscode.Diagnostic | undefined {
+        // CS0162 & CS8019 => Unnused using and unreachable code.
+        // These hard coded values bring some goodnes of fading even when analyzers are disabled.
+        let isFadeout = (quickFix.Tags && !!quickFix.Tags.find(x => x.toLowerCase() == 'unnecessary')) || quickFix.Id == "CS0162" || quickFix.Id == "CS8019";
+
+
+        let severity = this._asDiagnosticSeverity(quickFix, isFadeout);
+
+        if(severity === undefined)
```
rchande: weird formatting

src/features/diagnosticsProvider.ts
```diff
@@ -319,36 +324,59 @@ class DiagnosticsProvider extends AbstractSupport {
...
+            diagnostic.tags = [vscode.DiagnosticTag.Unnecessary];
+        }
+
+        return diagnostic;
     }
 
-    private static _asDiagnosticSeverity(logLevel: string): vscode.DiagnosticSeverity {
-        switch (logLevel.toLowerCase()) {
+    private _asDiagnosticSeverity(quickFix: protocol.QuickFix, fadeOut: boolean): vscode.DiagnosticSeverity | undefined {
+        if(fadeOut && quickFix.LogLevel.toLowerCase() === 'hidden' || quickFix.LogLevel.toLowerCase() === 'none')
```
rchande: Would is be clearer to inline this additional fadeOut checking after line 330?

src/features/diagnosticsProvider.ts
```diff
@@ -319,36 +324,59 @@ class DiagnosticsProvider extends AbstractSupport {
...
+            case 'info':
                 return vscode.DiagnosticSeverity.Information;
+            case 'hidden':
+                if(vscode.workspace.getConfiguration('csharp').get('suppressHiddenDiagnostics', true))
+                {
+                    return undefined;
+                }
+                return vscode.DiagnosticSeverity.Hint;
+            default:
+                return undefined;
```
rchande: What does it even mean to hit this?

src/features/diagnosticsProvider.ts
```diff
@@ -319,36 +324,59 @@ class DiagnosticsProvider extends AbstractSupport {
...
             case 'error':
                 return vscode.DiagnosticSeverity.Error;
             case 'warning':
                 return vscode.DiagnosticSeverity.Warning;
-            // info and hidden
-            default:
+            case 'info':
                 return vscode.DiagnosticSeverity.Information;
+            case 'hidden':
+                if(vscode.workspace.getConfiguration('csharp').get('suppressHiddenDiagnostics', true))
```
rchande: This could potentially read the settings file every time

src/features/diagnosticsProvider.ts
```diff
@@ -319,36 +324,59 @@ class DiagnosticsProvider extends AbstractSupport {
...
-    // --- data converter
+        let message = `${quickFix.Text} [${quickFix.Projects.map(n => this._asProjectLabel(n)).join(', ')}]`;
 
-    private static _asDiagnostic(quickFix: protocol.QuickFix): vscode.Diagnostic {
-        let severity = DiagnosticsProvider._asDiagnosticSeverity(quickFix.LogLevel);
-        let message = `${quickFix.Text} [${quickFix.Projects.map(n => DiagnosticsProvider._asProjectLabel(n)).join(', ')}]`;
-        return new vscode.Diagnostic(toRange(quickFix), message, severity);
+        let diagnostic = new vscode.Diagnostic(toRange(quickFix), message, severity);
+
+        if(isFadeout)
```
rchande: Can we refactor this a little to handle all the fadeout special stuff in one place?

src/features/diagnosticsProvider.ts
```diff
@@ -259,6 +260,13 @@ class DiagnosticsProvider extends AbstractSupport {
         this._documentValidations[key] = source;
     }
 
+    private _mapQuickFixesAsDiagnosticsInFile(quickFixes: protocol.QuickFix[]): { diag: vscode.Diagnostic, fileName: string }[]
+    {
+        return quickFixes
+            .map(qf => ({ diag: this._asDiagnostic(qf), fileName: qf.FileName }))
+            .filter(item => item.diag !== undefined);
```
rchande: Might be nicer if we could write the filter/map more explicitly, maybe like:?
```qf
.filter(item => hasCorrectDiagnosticType)
.map()
```

savpek:
Got a bit of battle with `Code actions can rename and open files` test. It was made differently than other tests and dynamically created file with content `class C {}`. However it duplicated content as `class C {}class C {}` sometimes. It was pretty hard to repro locally but it can happen: just requires 10+ tries with cleans and switch between test projects. Likely it is some kind of ordering + timing issue with other tests. Based on code everything should be fine, it creates or replaces file with content `class C {}` but ends up with `class C {}class C {}` that breaks the test.

Resolved by refactoring to same template as other tests use which is predefined files in test projects. https://github.com/OmniSharp/omnisharp-vscode/pull/2873/commits/736071899195968ce549d5fd1b01e8f1feb7bff8

savpek:
For restore CI issue very likely cause is https://github.com/OmniSharp/omnisharp-roslyn/issues/1458, it can be fixed workarounds like manually triggering analysis during tests but i strongly prefer fixing underlaying problem at omnisharp-roslyn side. I made test placeholder which is skipped currenlty until fix is available. It doesn't directly refer this feature however or block it, but it blocks further testing which require analyzers enabled.

savpek:
Ready for re-review @akshita31 @rchande

savpek:
Could @rchande @akshita31 check this out? It should be good to go and i use this at daily basis (custom build). 

I'd like to use this PR test infra with #3089 since i ended up refactor diagnostic provider there a bit and it's good point to add more tests for diagnostics 🙂

akshita31:
Looks good to me. Thanks @savpek

akshita31:
@savpek Would you like to do another PR to add the changelog.

savpek:
@akshita31 changelog https://github.com/OmniSharp/omnisharp-vscode/pull/3105

PR #2880, opened 2019-03-01
Allow integration tests to specify what project to generate assets for

rchande:


akshita31:
This is great. It will reduce our build times definitely. Thanks :1st_place_medal:

PR #2925, opened 2019-03-19
Set dirname to false in the webpack config

akshita31:
One of the reasons the razor extension was not loading was because to read the package.json the razor extension was using [__dirname for printing the version number of razor](https://github.com/aspnet/Razor.VSCode/blob/master/src/Microsoft.AspNetCore.Razor.VSCode/src/RazorLogger.ts#L99). However when we use webpack,[ __dirname behaves differently](https://codeburst.io/use-webpack-with-dirname-correctly-4cad3b265a92) and returns a "/", due to which the logger could not find the package.json and the entire razor part was falling apart as an exception was being thrown. Setting __dirname would make it behave like it should in normal node.js context


rchande:
Is there some way we can add a test here to prevent regressions?

akshita31:
@rchande Good question. I think it is possible to add an integration test in the omnisharp-vscode repo to validate that the razor parts are working properly(maybe just one test to ensure that cross repo changes dont break stuff). @NTaylorMullen thoughts ?

NTaylorMullen:
Definitely possible. Should just steal one of our integration tests: https://github.com/aspnet/Razor.VSCode/blob/master/client/test/Completions.test.ts

They test these apps: https://github.com/aspnet/Razor.VSCode/tree/master/test/testapps

test/razorTests/razorTelemetry.integration.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        let telemetryEventResolver: (value?: any) => void;
+        // tslint:disable-next-line:promise-must-complete
+        const extensionActivated = new Promise(resolve => {
+            telemetryEventResolver = resolve;
+        });
+        
+        subscription = eventStream.subscribe(event =>{
+            if (event.type == EventType.TelemetryEvent) {
+                let possibleRazorEvent = <TelemetryEvent>event;
+                if (possibleRazorEvent.eventName == "VSCode.Razor.DocumentOpened") {
```
NTaylorMullen: 👏  Would probably be good to also ensure that you don't receive: `VSCode.Razor.ErrorOnActivation`

akshita31: I am not sure if we can make it work. By the time we obtain the event stream and subscribe to it,  the "ErrorOnActivation" would have already been raised and I dont think there is a way currently to get the past events from the event stream.

akshita31: Thoughts ?

NTaylorMullen: Ah, if it's out of reach then don't worry about it.

PR #2969, opened 2019-04-05
Added processCommand to attach config

andrew-w-ross:
Attaching to a process can get quite annoying if you're using `dotnet watch` as `processName` isn't specific enough as there are multiple processes called dotnet. 
This pr add a config option called `processCommand` to the attach config that will find the process by it's command.
This pr is just a proof of concept, I'm just looking for feedback on the idea. 
If you like the feature I'll do a proper one with all the unit testing and correct error messages, etc.


src/configurationProvider.ts
```diff
@@ -153,14 +151,26 @@ export class CSharpConfigurationProvider implements vscode.DebugConfigurationPro
...
             // If the config doesn't look functional force VSCode to open a configuration file https://github.com/Microsoft/vscode/issues/54213
             return null;
         }
 
-        if (config.request === "launch")
-        {
+        if (config.request === "attach" && config.processCommand) {
+            const processCommand = config.processCommand.replace(/\${workspaceFolder}/g, folder.uri.fsPath);
+            delete config.processCommand;
+            return this.attachItemsProvider.getAttachItems()
```
gregg-miskelly: >return [](start = 12, length = 6)

nit: make this an async method and use `await`.

src/configurationProvider.ts
```diff
@@ -153,14 +151,26 @@ export class CSharpConfigurationProvider implements vscode.DebugConfigurationPro
...
+                .then(attachItems => {
+                    const attachItem = attachItems.find(ai => ai.detail === processCommand);
+                    if (attachItem == null) {
+                        throw new Error("Can't find process");
+                    }
+                    config.processId = attachItem.id;
```
gregg-miskelly: >config.processId = attachItem.id; [](start = 20, length = 33)

We should probably fail if there are multiple processes with the command rather than picking the first.

src/configurationProvider.ts
```diff
@@ -153,14 +151,26 @@ export class CSharpConfigurationProvider implements vscode.DebugConfigurationPro
...
-        if (config.request === "launch")
-        {
+        if (config.request === "attach" && config.processCommand) {
+            const processCommand = config.processCommand.replace(/\${workspaceFolder}/g, folder.uri.fsPath);
+            delete config.processCommand;
+            return this.attachItemsProvider.getAttachItems()
+                .then(attachItems => {
+                    const attachItem = attachItems.find(ai => ai.detail === processCommand);
+                    if (attachItem == null) {
+                        throw new Error("Can't find process");
```
gregg-miskelly: >"Can't find process" [](start = 40, length = 20)

We might want to mention what we are looking for.

package.json
```diff
@@ -1409,6 +1409,11 @@
             "type": "object",
             "required": [],
             "properties": {
+              "processCommand": {
```
gregg-miskelly: >processCommand [](start = 15, length = 14)

I believe you also need a tweak [here](https://github.com/OmniSharp/omnisharp-vscode/blob/4a26dce82e038a76bb78ca2870a9fb1ec3af97e5/src/tools/GenerateOptionsSchema.ts#L119) to delete `processCommand` from the unit test options.

andrew-w-ross: Done.

src/configurationProvider.ts
andrew-w-ross: Done

src/configurationProvider.ts
andrew-w-ross: Now throws `Couldn't find process with command ${processCommand}`

src/configurationProvider.ts
andrew-w-ross: Done

andrew-w-ross:
@gregg-miskelly Let me know if that's good. I'm not too familiar with Mocha so I'm not entirely sure how to mock `DotNetAttachItemsProviderFactory` so I can test.

package.json
```diff
@@ -1414,6 +1414,11 @@
...
+                "type": "string",
+                "description": "Attaches to a process with this command",
+                "default": "If this is use 'processId' and 'processName' should not be used."
```
gregg-miskelly: There is a typo here. You want 'used' instead of 'use'.

But thinking about it more, you might want to instead use something like `dotnet exec /example/path/to/program.dll`  to provide an example of the kind of thing we might want.

src/configurationProvider.ts
```diff
@@ -151,16 +149,31 @@ export class CSharpConfigurationProvider implements vscode.DebugConfigurationPro
...
+            delete config.processCommand;
+            const attachItems = await this.attachItemsProvider.getAttachItems();
+            const foundAttatchItems = attachItems.filter(ai => ai.detail === processCommand);
+
+            if (foundAttatchItems.length === 0) {
+                throw new Error(`Couldn't find a process with the command "${processCommand}"`);
+            }
+
+            if (foundAttatchItems.length > 1) {
+                throw new Error(`Find ${foundAttatchItems.length} processes with the command "${processCommand}"`);
```
gregg-miskelly: How about instead `"Unable to select process. Multiple processes found with command line "${processCommand}".`

gregg-miskelly:
> @gregg-miskelly Let me know if that's good. I'm not too familiar with Mocha so I'm not entirely sure how to mock `DotNetAttachItemsProviderFactory` so I can test.

I am not an expert on it either, so there may be other options. But what I have always done is extract out the thing I want to unit test into its own class which takes appropriate inputs (ex: probably a process listing + input debug target in your case) and then I can unit test just the transformation without having to get into real object mocking.

JoeRobich:
@gregg-miskelly Is this a PR we would want to fix up and consider taking? or are things handled differently now with DevKit?

gregg-miskelly:
If @andrew-w-ross (the original author) was still interested in taking this work, I wouldn't be opposed to it. That said, I think it is probably less important now since (1) apphost mostly solves the problem, and (2) no one else has requested this feature, so it would be hard for my team to justify working on this right now. So, closing this for now makes sense to me. If someone wants to work on this though, I am happy to reconsider.

andrew-w-ross:
@gregg-miskelly Completely forgot about this pr, sorry I should have cleaned it up.

gregg-miskelly:
@andrew-w-ross no problem! Do you want to just open a new PR if you wind up having the cycles to get back to it?

Issue #2962, opened 2019-04-04
Please consider removing Ctrl+L L keybinding


alexandrudima:
Starting with VS Code 1.32, we have a proeminent keybinding at <kbd>Ctrl</kbd>+<kbd>L</kbd>. 

https://github.com/Microsoft/vscode/issues/59003 offers a great explanation about why we have made this change. **TL;DR** We have a command `expandLineSelection` that we shipped by mistake, bound to <kbd>Ctrl</kbd>+<kbd>I</kbd> for a few years. The mistake was due to ... I-`I` looking like l-`l` in GitHub's issues font.

I would like to kindly ask that you remove this contributed keybinding from your `package.json` and free up <kbd>Ctrl</kbd>+<kbd>L</kbd> -- https://github.com/OmniSharp/omnisharp-vscode/blob/1279c53b2fcfdd1a9eb1cb985f0871d1ffb289e8/package.json#L828-L832

Thanks!

PR #2979, opened 2019-04-11
Update omnisharp 1.32.18

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/2962, https://github.com/OmniSharp/omnisharp-vscode/issues/2980

### DIFF
--- a/src/omnisharp/protocol.ts
+++ b/src/omnisharp/protocol.ts
@@ -431,7 +431,7 @@ export interface UnresolvedDependenciesMessage {
 }
 
 export interface ProjectConfigurationMessage {
-    ProjectGuid: string;
+    ProjectId: string;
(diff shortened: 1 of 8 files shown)


src/omnisharp/protocol.ts
```diff
@@ -431,7 +431,7 @@ export interface UnresolvedDependenciesMessage {
 }
 
 export interface ProjectConfigurationMessage {
-    ProjectGuid: string;
+    ProjectId: string;
```
NTaylorMullen: Was this the bit causing the telemetry issues or is this just cleanup?

akshita31: This is the clean up step.

PR #3038, opened 2019-05-02
Consume omnisharp 1.32.19

rchande:


filipw:
👍👍👍

filipw:
There is an Azure outage at the moment, which I suspect is the reason for the connectivity failures, see https://azure.microsoft.com/en-us/status/ and https://twitter.com/AzureSupport/status/1124046510411460610

rchande:
Cool, I'll stop flailing for now :)

filipw:
> Cool, I'll stop flailing for now :)

all my work/production services are completely down too 😞

filipw:
I was really hoping that it was the azure outage that broke it but my goodness 😢

PR #3089, opened 2019-05-31
Re-Analyze project or all projects

savpek:
VScode implementation for https://github.com/OmniSharp/omnisharp-roslyn/pull/1507

Detailed information in https://github.com/OmniSharp/omnisharp-roslyn/pull/1507

image

loading-icon-analyzers


savpek:
Ready for review @rchande @akshita31 @david-driscoll @filipw 

Theres skipped tests until https://github.com/OmniSharp/omnisharp-roslyn/pull/1507 is merged, tested locally that they should work once omnisharp-roslyn side is available.

Main bullets:
- Added new options to re-analyse either current project or all projects from vscode.
- Refactored diagnostic provider to support new project analyzed events from omnisharp-roslyn
  - New events reduce pressure of diagnostic api greatly when analyzers are enabled because api doesn't need to be polled continously any more. This is good perf boost on large workspaces.
- Refactored previous hand made debounce (kind of) with RX based version on diagnostic provider. It's now throttling instead debounce because it works better with analyzer enabled feedback. In same time it works well with non analyzer enabled version. Now analyzer enabled version has more real-time feeling since results are continously returned once they are available.
- Added status bar for background work. Currently it only shows when analysis is going on, but later on it can be extended to support thigs like 'restoring project xyz.csproj' etc.
- Added tests for new features, they are partially disabled since 3 of them needs omnisharp-roslyn to be updated.

src/features/diagnosticsProvider.ts
```diff
@@ -124,47 +127,56 @@ class DiagnosticsProvider extends AbstractSupport {
...
+            .pipe(throttleTime(3000))
+            .subscribe(async () => {
+                try {
+                    if (this._validationAdvisor.shouldValidateAll()) {
+                        await this._validateSmallWorkspace();
+                    }
+                    else if (this._validationAdvisor.shouldValidateFiles()) {
+                        await this._validateLargeWorkspace();
+                    }
+                } catch { }
```
colombod: Why swallowing all exceptions here?

src/features/diagnosticsProvider.ts
```diff
@@ -256,10 +257,22 @@ class DiagnosticsProvider extends AbstractSupport {
...
-        }, 750);
+        }, 2000);
+    }
 
-        source.token.onCancellationRequested(() => clearTimeout(handle));
-        this._documentValidations[key] = source;
+    // On large workspaces (if maxProjectFileCountForDiagnosticAnalysis) is less than workspace size,
+    // diagnostic fallback to mode where only open documents are analyzed.
+    private async _validateLargeWorkspace(): Promise<void> {
+        await setTimeout(async () => {
```
colombod: could this just be returned  instead of awaited?

src/features/diagnosticsProvider.ts
```diff
@@ -268,61 +281,41 @@ class DiagnosticsProvider extends AbstractSupport {
             .filter(diagnosticInFile => diagnosticInFile !== undefined);
     }
 
-    private _validateProject(): void {
-        // If we've already started computing for this project, cancel that work.
-        if (this._projectValidation) {
-            this._projectValidation.cancel();
-        }
+    private async _validateSmallWorkspace(): Promise<void> {
+        await setTimeout(async () => {
```
colombod: could this just be returned  instead of awaited?

src/observers/BackgroundWorkStatusBarObserver.ts
```diff
@@ -0,0 +1,29 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { BaseStatusBarItemObserver } from './BaseStatusBarItemObserver';
+import { BaseEvent, OmnisharpProjectDiagnosticStatus } from '../omnisharp/loggingEvents';
+import { EventType } from '../omnisharp/EventType';
+import { DiagnosticStatus } from '../omnisharp/protocol';
+
+export class BackgroundWorkStatusBarObserver extends BaseStatusBarItemObserver {
+    public post = (event: BaseEvent) => {
+        if(event.type == EventType.ProjectDiagnosticStatus)
```
colombod: should be ===

src/omnisharp/protocol.ts
```diff
@@ -286,6 +287,18 @@ export interface ProjectInformationResponse {
     DotNetProject: DotNetProject;
 }
 
+export enum DiagnosticStatus
+{
+    Processing = 0,
```
colombod: could this be done as union type with literals?

test/integrationTests/reAnalyze.integration.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        await vscode.commands.executeCommand("vscode.open", interfaceImplUri);
+        await vscode.commands.executeCommand("vscode.open", interfaceUri);
+    });
+
+    suiteTeardown(async () => {
+        await testAssetWorkspace.cleanupWorkspace();
+    });
+
+    test("When interface is manually renamed, then return correct analysis after re-analysis of project", async function () {
```
rchande: The code makes some differentiation between "large" and "small" workspaces, do these tests cover those differences?

src/omnisharp/protocol.ts
savpek: It presents enum from omnisharp-roslyn api message. I personally prefer string serialization over number ones but existing apis return enums in number form -> i don't think this is good place to brake that consistency. For example `FileModificationType` on many of file operations.

src/features/diagnosticsProvider.ts
savpek: I think this moved from original version, i don't see any reason for it in current form -> remove.

savpek:
Made some tweaks for tests:
- Restore was made outside of server context: problem is that it didn't emit correct events to eventstream since it was just dummy version -> replaced to call real restore functionality (same as user executes restore all). This caused pretty unpredictable behavior of restore. Other thing was that this often caused restore to fail with locked files etc.
- git clean mechanism cleaned project folder, however that didn't remove .vscode/settings.json, .vscode/launch... etc files when ran on .sln since they are on upper folder at that case -> replaced git clean to be executed on workspace root. This issue caused lots of side effects between tests.
- Since vscode isn't restarted at any point of time all editors are left open, this can cause issue when file is removed on backround because vscode leave it open as deleted 'virtual' document.
- Added retry for clean, this wasn't that much of problem at CI but locally git clean pretty often failed things like locked file etc.

Some tests are still bit flaky however, need to dig bit deeper still.

test/integrationTests/reAnalyze.integration.test.ts
savpek: Added tests for large workspace scenario

savpek:
Review fixes ready and tests seems to be stable now @rchande @colombod

kurobon-jp:
Why isn't a [diagnostic event](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn/DiagnosticEventForwarder.cs#L23) listener implemented?

savpek:
> Why isn't a [diagnostic event](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Roslyn/DiagnosticEventForwarder.cs#L23) listener implemented?

It's actually part of LSP implementation i think. It's definetly possible to use those events to sync state of diagnostics between client and server, however i don't think it's in scope of this PR.

CC @david-driscoll @mholo65 can you give better insights of that part @kurobon-jp mentioned? Could it be good idea to consume those events in vscode or should it move LSP directly instead when that is fully ready?

kurobon-jp:
Aside from including the implementation in this PR, I think it makes sense.

Analyzing thousands of lines of document in Omnisharp can take several seconds.
So I'm thinking of an implementation like this.
名称未設定

This allows users to quickly get important diagnostic results.

src/features/diagnosticsProvider.ts
```diff
@@ -256,10 +258,22 @@ class DiagnosticsProvider extends AbstractSupport {
             catch (error) {
                 return;
             }
-        }, 750);
+        }, 2000);
+    }
+
+    // On large workspaces (if maxProjectFileCountForDiagnosticAnalysis) is less than workspace size,
+    // diagnostic fallback to mode where only open documents are analyzed.
+    private _validateLargeWorkspace(): NodeJS.Timeout {
```
JoeRobich: I feel like the name `_validateOpenDocuments` is more meaningful here.

src/features/diagnosticsProvider.ts
```diff
@@ -268,61 +282,41 @@ class DiagnosticsProvider extends AbstractSupport {
             .filter(diagnosticInFile => diagnosticInFile !== undefined);
     }
 
-    private _validateProject(): void {
-        // If we've already started computing for this project, cancel that work.
-        if (this._projectValidation) {
-            this._projectValidation.cancel();
-        }
+    private _validateSmallWorkspace(): NodeJS.Timeout {
```
JoeRobich: I feel like the name `_validateEntireWorkspace` is more meaningful here.

test/integrationTests/codeActionRename.integration.test.ts
```diff
@@ -19,8 +20,8 @@ suite(`Code Action Rename ${testAssetWorkspace.description}`, function () {
 
     suiteSetup(async function () {
         should();
-        await testAssetWorkspace.restore();
         await activateCSharpExtension();
+        await testAssetWorkspace.restore();
```
JoeRobich: I noticed this reordering of activation and workspace restore in the tests and was curious. What problem is this solving?

JoeRobich:
@savpek I extracted the test improvements into https://github.com/OmniSharp/omnisharp-vscode/pull/3432.

savpek:
> @savpek I extracted the test improvements into #3432.

Thanks, sorry for delay i think i will have time to check these out tomorrow finally 😄

test/integrationTests/codeActionRename.integration.test.ts
savpek: Don't remember for sure. Its possible that this change is unnecessary since i think latter one where i changed tests to use real restore eventing instead of dummy one was root cause for issues with workspace restore instability during tests.

savpek:
Broken atm, probably i missed something at merge. I will try fix this during upcoming week.

savpek:
@JoeRobich i think all concerns fixed now and merge done.

JoeRobich:
@savpek Thanks!

Issue #3138, opened 2019-06-25
Why is the "OmniSharp server is running" status bar icon green?


Tyriar:
![image](https://user-images.githubusercontent.com/2193314/60122847-f244c000-973a-11e9-95dd-d579e7147fba.png)

Extensions typically shouldn't do this, red makes sense for a failure state but if everything is working as intended it should just be white so as to not draw the eye to it.

rchande:
cc @akshita31

Tyriar:
Thanks @akshita31 😃

ihodev:
Hello, @akshita31! Can I somehow change that color?

![image](https://user-images.githubusercontent.com/9801624/61533710-530fa180-aa36-11e9-948c-d58fecd91ba1.png)

mario-d-s:
This actually confused me - I was used to the icon being green and now I though something was wrong...

Anyway: now the icon is white regardless of theme, and as @ihodev illustrates this doesn't make much sense on light themes. Can it please be changed to have the dark grey of other status bar indicators on light themes?

v-zubritsky:
I could relate to this, icon doesn't look very attractive http://tinyimg.io/i/zgW8qOr.png

Tyriar:
@v-zubritsky created https://github.com/OmniSharp/omnisharp-vscode/issues/4384

PR #3144, opened 2019-06-27
Do not draw attention when evrything is fine

akshita31:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/3138


filipw:
noooooooo it was so lovely before 😭

Issue #1198, opened 2017-02-07
Errors are duplicated in output window


nguerrera:
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-rc4-004769)

Product Information:
 Version:            1.0.0-rc4-004769
 Commit SHA-1 hash:  9cf4e9d1d0

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\1.0.0-rc4-004769
```

VS Code version: 1.9.0
C# Extension version: 1.6.2

## Steps to reproduce

* dotnet new console
* dotnet restore
* Edit program.cs and make a deliberate error
* code .
* Ctrl+Shift+B
* Configure build task -> .NET Core
* Ctrl+Shift+B again
* Look at output window

## Expected  behavior

Each error appears only once in output window

## Actual behavior

Errors are duplicated.

## Workaround

Edit tasks.json to disable the summary:

```
 "tasks": [
        {
            "taskName": "build",
            "args": [
                "${workspaceRoot}\\qqq.csproj",
                "/clp:NoSummary"
            ],
            "isBuildCommand": true,
            "problemMatcher": "$msCompile"
        }
]
```

## Notes
Visual Studio arranges not to duplicate errors in its output windows, but VS Code does not.

Related: dotnet/sdk#823, dotnet/cli#5607

DustinCampbell:
Hmmm... I'm not sure of switching over to "dotnet msbuild" as it might have other ramifications.

@piotrpMSFT: will there be other fallout if we do this? E.g. will we break "dotnet build" telemetry?

nguerrera:
I think we want to fix dotnet/cli#5607 and then pass /clp:NoSummary.

DustinCampbell:
Makes sense to me!

(Very glad to have you active here this morning @nguerrera :smile:)

nguerrera:
Turns out that `dotnet build [project] /clp:NoSummary` works fine.

The issue is with `dotnet build /clp:NoSummary` with implicit project. So, I think you can fix this without blocking on dotnet/cli#5607

achandlerwhite:
Just want add that I am still seeing this problem. I am using the 2.1-preview1 SDK, VS Code 1.21.0, omnisharp-vscode 1.15-beta1 on MacOS High Sierra.

[...]

Issue #1197, opened 2017-02-07
Errors in output window do not get source links


nguerrera:
## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-rc4-004769)

Product Information:
 Version:            1.0.0-rc4-004769
 Commit SHA-1 hash:  9cf4e9d1d0

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\1.0.0-rc4-004769
```

VS Code version: 1.9.0
C# Extension version: 1.6.2

## Steps to reproduce

* dotnet new console
* dotnet restore
* Edit program.cs and make a deliberate error
* code .
* Ctrl+Shift+B
* Configure build task -> .NET Core
* Ctrl+Shift+B again
* Look at output window

## Expected  behavior

I can click on each of the errors in the output window and go to the relevant source.

## Actual behavior

I can't click on the errors because they have project-relative paths.

## Workaround
Set `<GenerateFullPaths>true</GenerateFullPaths>` in csproj. 

## Notes
Visual Studio arranges to have full paths in its output window, but VS code does not.

Related: dotnet/sdk#823

DustinCampbell:
Is there a step missing here? There needs to be a tasks.json file present for Ctrl+Shift+B to work.

nguerrera:
Sorry, that was from memory. It prompted me to add necessary files and I said yes.

DustinCampbell:
Cool. Just wanted to check and make sure it didn't happen some other way. This is pretty easily solved.

nguerrera:
I've updated the steps.

rchande:
Thanks for finding the duplicate @akshita31 , I'll look for a fix (unless you want to :) )

CoderJason123:
so...on a mac, is this expected behavior, and the line for full paths only on a mac fixes the problem?  The issue i logged 2045 has been closed, i'm not real clear on the resolution or just that it's an on going issue.  please advise and thank you.

DustinCampbell:
@rchande: The right fix is to update the tasks.json generation to include `/p:GenerateFullPaths=true`.

Xarkam:
I my case, my build is done by a .bat file
Because we can not chain tasks in the launch.json file.
Using a build file, the error persists.

And yes I have the argument on the build line in my .bat file
[...]

PR #3145, opened 2019-06-28
Modify tasks.json tasks

akshita31:
Modified the tasks that are provided in the default tasks.json to:

1. Use msCompile as the problem matcher
2. GenerateFullPaths as msCompile expects full paths(https://github.com/OmniSharp/omnisharp-vscode/issues/1197)
3. Do not produce summary as that is leading to duplicate errors in the problem window(https://github.com/OmniSharp/omnisharp-vscode/issues/1198)


akshita31:
cc @nguerrera

PR #3164, opened 2019-07-10
Fix listening to cancellation tokens in completionItemProvider

### LINKED ISSUE #2472: Autocomplete often incorrectly replaces typed text
Original issue reported here: https://github.com/Microsoft/vscode/issues/57037
I was asked to refer this issue to the C# extension project instead.

Often, while typing in the editor, when I press space after a word, the editor will replace the perfectly valid code with something invalid.  I am currently working in C#, and have this happen frequently with several keywords.  Today, I tried the Insiders build as well, but still run into the issue.  I have also attached a video of the problem happening.

Some examples:
- `new` becomes `nameof`
- `private` becomes `partial`
- `return` becomes `ref`

[bad-autocomplete.zip](https://github.com/Microsoft/vscode/files/2311514/bad-autocomplete.zip)

## Environment data
<details>
<summary>`dotnet --info` output:</summary>
.NET Core SDK (reflecting any global.json):
 Version:   2.1.301
 Commit:    59524873d6

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.13-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.1.301/

Host (useful for support):
  Version: 2.1.1
  Commit:  6985b9f684

.NET Core SDKs installed:
  2.1.301 [/usr/local/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.1 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.1 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.1 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
</details>

VS Code version: 1.26.1 and 1.27.0-insiders
C# Extension version: 1.15.2

## Steps to reproduce
Type an entire name/keyword, verbatim, in the editor and press the spacebar.  It does not always happen, but happens enough to be extremely frustrating and to severely impact my productivity.

## Expected  behavior
Typed name or keyword matches the top entry in the suggestion list and should not be changed.

## Actual behavior
Typed name or keyword changes to an unrelated, and often semantically invalid, name or keyword.

christianscheuer:
Should fix #2472 

I haven't tested this, since I don't know how to build Omnisharp for VSCode, but looking into the [VSCode API](https://code.visualstudio.com/api/references/vscode-api#CancellationToken) it appears we should actually take the passed `CancellationToken` into account and not return items if a cancellation was requested.

See this for a discussion on what goes wrong when the requests take longer than expected and the cancellation isn't observed:
https://github.com/OmniSharp/omnisharp-vscode/issues/2472#issuecomment-509036624

Also here in the VSCode repo it's noted that we need to listen to cancellations:
https://github.com/microsoft/vscode/issues/60753#issuecomment-430190566

### DIFF
--- a/src/features/completionItemProvider.ts
+++ b/src/features/completionItemProvider.ts
@@ -41,9 +41,9 @@ export default class OmniSharpCompletionItemProvider extends AbstractSupport imp
         }
 
         try {
-            let responses = await serverUtils.autoComplete(this._server, req);
+            let responses = await serverUtils.autoComplete(this._server, req, token);
 
-            if (!responses || token && token.isCancellationRequested) {
+            if (!responses) {
                 return;
             }
 
--- a/src/features/completionItemProvider.ts
+++ b/src/features/completionItemProvider.ts
@@ -43,7 +43,7 @@ export default class OmniSharpCompletionItemProvider extends AbstractSupport imp
         try {
             let responses = await serverUtils.autoComplete(this._server, req);
 
-            if (!responses) {
+            if (!responses || token && token.isCancellationRequested) {
                 return;
             }
 
--- a/src/omnisharp/utils.ts
+++ b/src/omnisharp/utils.ts
@@ -7,8 +7,8 @@ import { OmniSharpServer } from './server';
 import * as protocol from './protocol';
 import * as vscode from 'vscode';
 
-export async function autoComplete(server: OmniSharpServer, request: protocol.AutoCompleteRequest) {
-    return server.makeRequest<protocol.AutoCompleteResponse[]>(protocol.Requests.AutoComplete, request);
+export async function autoComplete(server: OmniSharpServer, request: protocol.AutoCompleteRequest, token: vscode.CancellationToken) {
+    return server.makeRequest<protocol.AutoCompleteResponse[]>(protocol.Requests.AutoComplete, request, token);
 }
 
 export async function codeCheck(server: OmniSharpServer, request: protocol.Request, token: vscode.CancellationToken) {


src/features/completionItemProvider.ts
```diff
@@ -43,7 +43,7 @@ export default class OmniSharpCompletionItemProvider extends AbstractSupport imp
         try {
             let responses = await serverUtils.autoComplete(this._server, req);
```
rchande: If you look at the defintion of serverUtils.autoComplete, it calls server.MakeRequest, which supports passing though a CancellationToken. Interested in threading that through so you can pass it here?

rchande:
@christianscheuer Awesome sleuthing, thanks for the PR! I've left you one suggestion--let me know if you're interested in taking that up. I'm also happy to PR your branch if that's more convenient for you. Thanks again!

src/features/completionItemProvider.ts
christianscheuer: Ah yea that makes total sense. Will take a look.

christianscheuer:
@rchande thank you for the review. I've pushed a new change that should address this now.

rchande:
Tests are passing, I'll merge this now. Thanks!

silicakes:
Started happening to me following the recent update:

OS: macOS Mojave (10.14.5)
VSCode version: 1.36.1
language within editor: Javascript

flaky_autocomplete

PR #3189, opened 2019-07-24
 add a new configuration 'csharp.diagnosticLevel' to control what should be displayed in problems panel

fatfatson:
sometimes I don't want show all warnings ,
but vscode has no way to  control this itself,
so I add the option to omnisharp.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ out
 .razor
 dist/
 *.razor.json
+yarn.lock
 
 install.*
 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,19 +1,28 @@
 {
     "version": "0.2.0",
-    "configurations": [    
+    "configurations": [
+        {
+            "name": "Debug Extension",
+            "type": "extensionHost",
+            "request": "launch",
+            "runtimeExecutable": "${execPath}",
+            "args": [
+                "--extensions-dir=${env:HOME}/.vscode/debugext",
+                "--extensionDevelopmentPath=${workspaceRoot}"
+            ],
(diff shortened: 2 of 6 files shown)


WhiteBookmark:
This is a very useful feature. Can someone please merge it ?

WhiteBookmark:
After a while I actually found that this can be controlled using a .csproj file. I wanted to suppress CS0251 warnings so this is what I add in my .csproj file:

```XML
<PropertyGroup>
    <TargetFramework>netcoreapp2.2</TargetFramework>
    <LangVersion>7.3</LangVersion>
    <NoWarn>0251</NoWarn>
</PropertyGroup>
```

Notice NoWarn containing the warning code ? After I saved the file all CS0251 warnings were gone from dotnet cli as well as omnisharp problem panel.

savpek:
Theres ".ruleset" file support in vscode which is standard across .net platform. Can't it be used to solve this issue by controlling severity of code issues (warning -> hidden)?

filipw:
thanks for the PR but yep, I agree with @savpek. it should definitely be implemented with rulesets (instead of filtering on the client side) or with the NoWarn settings

JoeRobich:
@fatfatson Thanks for the PR. I understand the frustration that is behind it, but we don't think this is the right way to go.

In addition to the .ruleset mentioned above. With Roslyn's native .editorconfig support, diagnostic severity can be configured via .editorconfig files (https://github.com/dotnet/roslyn/pull/38886). There is still work required in OmniSharp to switch off the VisualStudio.CodingConvention library and I'll make myself a note to investigate effort.

-Joey

PR #3232, opened 2019-08-15
Make changes for 1.21.1

akshita31:


filipw:
@akshita31 @rchande as you recall there is a massive regression in omnisharp 1.34.1 - you really want to take 1.34.2 before pushing the extension out 
see https://github.com/OmniSharp/omnisharp-roslyn/blob/master/CHANGELOG.md#1342---not-yet-released and https://github.com/OmniSharp/omnisharp-roslyn/pull/1579

PR #3241, opened 2019-08-20
Update to 3.0.0-preview9 compatible version of Razor.

NTaylorMullen:
- This shouldn't be merged until a preview8 release has been completed.

aspnet/AspNetCore#12701

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "csharp",
-  "version": "1.21.1-beta1",
+  "version": "1.21.1-beta2",
   "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


filipw:
does this need Roslyn update in omnisharp-roslyn too?

NTaylorMullen:
> does this need Roslyn update in omnisharp-roslyn too?

It depends on 3.3.0 beta builds of Roslyn. So as long as `omnisharp-roslyn` is on 3.30 it should "just work" (tested everything out locally). That being said if Roslyn did some breaking changes between beta builds it's possible it falls over; at which point we can spend more time to align versions.

PR #3370, opened 2019-10-23
Update ctor snippet in csharp.json 

doggy8088:
This PR make `ctor` snippet a bit better. It because `ctor` can guess it's method name from it's file name for most of the case.


JoeRobich:
Thanks for this PR! A similar change has been merged and will be in an upcoming release.

PR #3385, opened 2019-10-31
MOD Use filename instead of ClassName

BaiShouDan:
ctor snippet use the filename without extension for ClassName, 
the user can continue to edit or modify it if necessary.


BaiShouDan:
Hello I don't want every time write again the class name so i founded this solution.

JoeRobich:
@BaiShouDan Thanks!

PR #3417, opened 2019-11-20
Enabling setting breakpoints for Q# files

bettinaheim:
It would be awesome if you would consider supporting also debugging of Q# files. Since we are actually code generating C# files during compilation, all that would be needed to do so is to list qsharp in the package.json. It would be awesome if we could add this extension to our recommendations and gain debugging support for our VS Code users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -870,6 +870,7 @@
           "languageIds": [
             "csharp",
             "razor",
+            "qsharp",
             "aspnetcorerazor"
           ]
         },
@@ -1938,6 +1939,7 @@
           "languageIds": [
             "csharp",
             "razor",
+            "qsharp",
             "aspnetcorerazor"
           ]
         },


gregg-miskelly:
Hi @bettinaheim I don't see any problem accepting your PR, provided that you aren't expecting us to test Q# support -- we certainly always try not to break things. But our test scenarios are just going to be C#. As long as that is okay, I will merge this.

bettinaheim:
> Hi @bettinaheim I don't see any problem accepting your PR, provided that you aren't expecting us to test Q# support -- we certainly always try not to break things. But our test scenarios are just going to be C#. As long as we have that understanding, I will merge this.

Awesome! Yes, fully understood and agreed. We are not expecting anything else but C# to work. Thank you so much!

PR #3468, opened 2019-12-17
Fix DefinitionProvider being overly defensive

ajaybhargavb:
Fixes https://github.com/aspnet/AspNetCore/issues/17867

This was an oversight. We don't need the if condition here because we can safely trust the results of `remap` as we try-catch around the whole block inside `remap`.


ajaybhargavb:
@JoeRobich

ajaybhargavb:
@JoeRobich can you merge this for me? @NTaylorMullen is out and I don't have access.

PR #3509, opened 2020-01-14
Change Linux distro for CI

gregg-miskelly:
Try changing the Linux distro for the CI build to see if this makes it pass.


gregg-miskelly:
Looks like these are down on all distros currently: https://github.com/dotnet/core/issues/4122

Giving up.

PR #3515, opened 2020-01-18
Update debugger version 1.22.0

WardenGnaw:


src/tools/VSSymbolSettings.json
```diff
@@ -70,6 +84,35 @@
...
+          "type": "string",
+          "description": "The HTTP Authorization header to use for this symbol server."
+        },
+        "errorInfo": {
+          "type": "string",
+          "description": "A user visible error message explaining why authorization could not be obtained for this server."
+        }
+      },
+      "required": [ "url" ]
+    },
```
gregg-miskelly: Sorry, when Chuck added this to VSCodeDebugAdapterHost, I failed to remember that that is the source for the public version. We want to grab the version of this file before Chuck's change, and we should redo Chuck's change as an extension.

PR #3524, opened 2020-01-24
Update Razor to include more C# actions

ryanbrandenburg:
CC @NTaylorMullen and @ajaybhargavb.


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.21.10 (Not yet released)
+* Updated Razor support (PR:[#3524](https://github.com/OmniSharp/omnisharp-vscode/pull/3524))
+    * Added quick info (hover) support for TagHelper and Blazor components. You can now hover over TagHelpers, Components and their attributes to understand what associated C# type you’re hovering over in addition to an attributes expected value type.
+    * Migrated Razor’s project understanding from the VSCode extension into the Language Server. This enables the language server to reboot without extra assistance (reliability) from an LSP client and also enables future Razor LSP clients to have richer functionality with less “work”.
+    * Added C# light bulbs to enable users to “Fully Qualify” members that are causing errors.
+    * Expanded Razor’s TextMate grammar colorization support to understand email addresses and `await foreach`.
+
```
NTaylorMullen: I'd add a generic line here about bug fixes:

`Several bug fixes`

package.json
```diff
@@ -498,17 +498,17 @@
                   "items": {
                     "type": "string"
                   },
-                  "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                  "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
```
NTaylorMullen: I believe this happens on npm install. Should undo along with other similar unintended changes to ensure we don't change functionality in VSCode.

NTaylorMullen:
@ryanbrandenburg the build failure looks legit. Since we removed a good amount of telemetry this test is now failing: https://github.com/OmniSharp/omnisharp-vscode/blob/2edb8414981d49185ce9fe783187008ab343678b/test/razorTests/razorTelemetry.integration.test.ts#L46

Sadly there isn't a Razor telemetry event that O# can capture during their tests because the suite teardown/restore is when all of our existing telemetry events fire and during test time you can't control exactly when that happens.

ryanbrandenburg:
@JoeRobich FYI the telemetry test is being removed because we no longer publish that telemetry event. That seemed to be the only test in it's suite so I removed the suite and its infrastructure too.

PR #3539, opened 2020-01-30
Update minimum Mono version to 6.6.0

JoeRobich:
Update to match the OmniSharp-Roslyn required Mono version. (https://github.com/OmniSharp/omnisharp-roslyn/blob/1e93b0748fd7d2337011a9a674ae682571280486/build.json#L5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: node_js
 node_js:
 - "10"
 
+mono:
+  - latest
+
 env:
 - CODE_VERSION=1.33.0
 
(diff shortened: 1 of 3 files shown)


JoeRobich:
@filipw Would there be any reason to allow the use of a earlier version?

filipw:
no makes sense to me 👌

filipw:
the server allows 6.4.0 by the way, when that change was made 6.6.0 was still in pre-release
https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/MSBuild/Discovery/Providers/MonoInstanceProvider.cs#L72

but it should be aligned at 6.6.0 since embedded mono is 6.6.0 based

Issue #3425, opened 2019-11-26
Improve error message when the C# extension is run on unsupported processors


gregg-miskelly:
## Steps to reproduce
Configure VS Code to remotely connect to an ARM device using VS Code extension remoting.

## Expected  behavior
Either the extension should work, or it should fail telling users to use the C# extension locally and remote debug.

## Actual behavior
Extension doesn't work, and doesn't provide a good error message.

cartermp:
Thanks @WardenGnaw for the PR, we'll see if we can get this in

PR #3581, opened 2020-02-15
Add Error Message for Unsupported Platforms

WardenGnaw:
This PR adds in a check to see what platform the extension is running on. 
If it is running via Remote SSH, we reccomend users to code on their host machine and debug remotely. 
If they are running VS Code somehow on an unsupported platform, we will mention that the C# extension is incompatiable with the current platform and arch.

Bumping up vscode engine to 1.36 for `extension.ExtensionKind` checks. [More Info](https://code.visualstudio.com/api/advanced-topics/remote-extensions#varying-behaviors-when-running-remotely-or-vs-onlines-browser-editor) Had to modify vscodeAdapter.ts to match vscode engine 1.36

Addresses #3425

Example error message:
image

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -121,6 +121,32 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
         eventStream.post(new ActivationFailure());
     }
 
+    if (!isSupportedPlatform(platformInfo)) {
+        const platform: string = platformInfo.platform ? platformInfo.platform : "this platform";
+        const architecture: string = platformInfo.architecture ? platformInfo.architecture : " and architecture";
+        let errorMessage: string = `The C# extension for Visual Studio Code (powered by OmniSharp) is incompatiable on ${platform} ${architecture}`;
+        const messageOptions: vscode.MessageOptions = {
+        };
+
+        // Check to see if VS Code is running remotely
+        if (extension.extensionKind === vscode.ExtensionKind.Workspace) {
+            const setupButton: string = "How to setup Remote Debugging";
+            errorMessage += ` with the VS Code Remote Extensions. To see avaliable workarounds, click on '${setupButton}'.`;
+
+            await vscode.window.showErrorMessage(errorMessage, messageOptions, setupButton).then((selectedItem: string) => {
+                if (selectedItem === setupButton) {
+                    let remoteDebugInfoURL = 'https://github.com/OmniSharp/omnisharp-vscode/wiki/Remote-Debugging-On-Linux-Arm';
+                    vscode.env.openExternal(vscode.Uri.parse(remoteDebugInfoURL));
+                }
+            });
+        } else {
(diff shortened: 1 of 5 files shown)


src/main.ts
```diff
@@ -121,6 +121,32 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
         eventStream.post(new ActivationFailure());
     }
 
+    if (!isSupportedPlatform(platformInfo)) {
+        const platform: string = platformInfo.platform ? platformInfo.platform : "this platform";
+        const architecture: string = platformInfo.architecture ? platformInfo.architecture : " and architecture";
```
gregg-miskelly: Might want to `"<Unknown processor architecture>"` rather then `"and architecture"` since, if the platform is understood, 'and architecture would be a bit weird -- "... is incompatible on Windows and architecture"

src/main.ts
```diff
@@ -171,6 +197,24 @@ export async function activate(context: vscode.ExtensionContext): Promise<CSharp
     };
 }
 
+function isSupportedPlatform(platform: PlatformInformation): boolean {
+    if (platform.isWindows()) {
+        return platform.architecture === "x86" || platform.architecture === "x86_64";
```
gregg-miskelly: Do you know if we get accurate information on **Windows** ARM64? Technically the language service side of the extension should work there. But if `platform.architecture` lies and says we are on x86 Windows then it doesn't matter.

WardenGnaw: Just added a check for ARM64 on windows. It would always default to x86_64 if it is not just x86.

gregg-miskelly: Do things work with the language service on ARM64 Windows? If so, maybe we shouldn't block it here, but rather just fail debugger launch?

WardenGnaw: I don't see a win-arm64 package in omnisharp-roslyn.

@JoeRobich Does the language service run on Windows ARM64?

JoeRobich: There presently is no WinArm64 package. I've been looking into LinuxArm64 but windows has not been on my radar. =)

gregg-miskelly: Yes, but on Windows arm64, the x86 OmniSharp will work.

JoeRobich:
@WardenGnaw is this ready to go in?

WardenGnaw:
@JoeRobich Yes. However, I was having issues with the CI passing with these changes.

JoeRobich:
@WardenGnaw Ah got it. I'll take a look

JoeRobich:
@WardenGnaw I verified that if you update the CODE_VERSION specified in the travis.yml to 1.36.0 it fixes the CI errors (https://github.com/OmniSharp/omnisharp-vscode/pull/3716)

WardenGnaw:
@JoeRobich Thanks! I forgot that's how we controlled the Code version in Travis.

PR #3592, opened 2020-02-20
Ignore virtual files

ryanbrandenburg:
Fixes https://github.com/dotnet/aspnetcore/issues/18927.

This seemed to require editing a lot of test files to exclude the Blazor project from running other tests, if anyone on the O# side has a better way to handle that I'm open.

CC @NTaylorMullen, @ajaybhargavb

### DIFF
--- a/src/features/diagnosticsProvider.ts
+++ b/src/features/diagnosticsProvider.ts
@@ -251,6 +251,12 @@ class DiagnosticsProvider extends AbstractSupport {
                     return;
                 }
 
+                // No problems published for virtual files
+                if(isVirtualCSharpDocument(document))
+                {
+                    return;
+                }
+
                 // (re)set new diagnostics for this document
                 let diagnosticsInFile = this._mapQuickFixesAsDiagnosticsInFile(quickFixes);
 
--- a/test/integrationTests/launchConfiguration.integration.test.ts
+++ b/test/integrationTests/launchConfiguration.integration.test.ts
@@ -5,10 +5,9 @@
 
 import * as fs from 'async-file';
 import * as vscode from 'vscode';
-import * as path from 'path';
 
 import { should, expect } from 'chai';
-import { activateCSharpExtension } from './integrationHelpers';
+import { activateCSharpExtension, isRazorWorkspace } from './integrationHelpers';
 import testAssetWorkspace from './testAssets/testAssetWorkspace';
 import { poll } from './poll';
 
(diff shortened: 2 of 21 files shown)


ryanbrandenburg:
@JoeRobich I'm not 100% sure this is the correct fix, maybe you could correct my thinking (or convince me I'm already right).

Repro steps:
1. `dotnet new blazorserver`
1. In Index.razor add the line `@DoesNotExist`.
1. Watch the `Problems` pane, you should see an entry for `Index.razor_virtual.cs` flash briefly (occasionally it stays permanently). If you didn't see it try again, it should repro each time.

When I track this down I can find requests to _validateDocument for the `Index.cshtml_virtual.cs` file, which has returned a result for Index.cshtml, even though the request was for _virtual.cs.

Hopefully that all makes sense, I can go into more detail offline if you need more context or a demo.

JoeRobich:
@ryanbrandenburg I tried the step you listed and this is what I see in the latest release 1.21.12. I do see a reported problem, but it is for Index.razor.

image

JoeRobich:
@ryanbrandenburg I spoke too soon. I see what you are talking about.

src/features/diagnosticsProvider.ts
```diff
@@ -251,6 +251,12 @@ class DiagnosticsProvider extends AbstractSupport {
                     return;
                 }
 
+                // No problems published for virtual files
+                if(isVirtualCSharpDocument(document))
```
JoeRobich: Maybe call `shouldIgnoreDocument` instead. Is this a place where we should be mapping diagnostic locations?

ryanbrandenburg: `shouldIgnoreDocument` has the following code:

```ts
 private shouldIgnoreDocument(document: TextDocument) {
        if (document.languageId !== 'csharp') {
            return true;
        }
        if (document.uri.scheme !== 'file' &&
            !isVirtualCSharpDocument(document)) {
            return true;
        }
        return false;
    }
```

So it would tell us that we should NOT ignore the virtual document.  

I'm not sure that mapping diagnostic locations is the correct response here because we already have a correctly mapped diagnostic for Index.cshtml in this case, so remapping the __virtual file would just lead to two identical diagnostics.

JoeRobich: whoops, I misread that one. sounds good

test/integrationTests/launchConfiguration.integration.test.ts
```diff
@@ -18,6 +19,12 @@ chai.use(require('chai-fs'));
 suite(`Tasks generation: ${testAssetWorkspace.description}`, function () {
     suiteSetup(async function () {
         should();
+
+        // These tests don't run on the BasicRazorApp2_1 solution
+        if (vscode.workspace.workspaceFolders[0].uri.fsPath.split(path.sep).pop() === 'BasicRazorApp2_1') {
```
JoeRobich: Maybe extract this out into a helper function `isRazorWorkspace(vscode.workspace)`

ryanbrandenburg:
@JoeRobich, I've updated it with your feedback but it seems I don't have permissions to merge.

JoeRobich:
Thanks @ryanbrandenburg !

PR #3593, opened 2020-02-20
Detect Blazor WASM projects and add/update appropriate config to allow launching and debugging

NTaylorMullen:
- Didn't want to pollute the omnisharp-roslyn language server with an understanding of Blazor WASM projects so built a heuristic on the client to decide if a project looks like a Blazor WASM project. This approach is similar to how web projects are detected today (look at the csproj and see if there's the line `sdk=microsoft.net.sdk.web` case insensitively). The heuristic for Blazor web assembly project requires the following from the project:
  1. User has the JavaScript Debugger (Nightly) extension installed
  1. User has the `debug.node.useV3` setting set to `true`
  1. User has the `debug.chrome.useV3` setting set to `true`
  1. Contains a `launchSettings.json` file containing the text `"inspectUri"`.
  1. (Blazor hosted only) Is executable
  1. (Blazor hosted only) References the web SDK
  1. (Blazor hosted only) Netcoreapp targetting

- Moved some core utility code such as `isWebProject` into the `util.ts` file.
- Created a new `ProgramLaunchType` enum to influence how launch.json's are created. There's a special case in this new launch enum that knows about Blazor web assembly project types.
- Added the understanding of the new `ProgramLaunchType.BlazorWebAssemblyStandalone` and  `ProgramLaunchType.BlazorWebAssemblyHosted` launch types to produce two `launch.json` configurations:
1. Launch the Blazor dev server
1. Launch the Blazor web assembly project in chrome while attached.

- Added asset tests to verify that we properly generate the new Blazor web assembly configurations.

Example launch.json output for a hosted application:
```
{
   // Use IntelliSense to find out which attributes exist for C# debugging
   // Use hover for the description of the existing attributes
   // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
   "version": "0.2.0",
   "configurations": [
        {
            "name": ".NET Core Launch (Blazor Hosted)",
            "type": "coreclr",
            "request": "launch",
            // If you have changed target frameworks, make sure to update the program path.
            "program": "${workspaceFolder}/Server/bin/Debug/netcoreapp3.1/BlazorWasmHosted1.Server.dll",
            "args": [],
            "cwd": "${workspaceFolder}/Server",
            "stopAtEntry": false,
            "env": {
                "ASPNETCORE_ENVIRONMENT": "Development"
            },
            "preLaunchTask": "build"
        },
        {
            "name": ".NET Core Debug Blazor Web Assembly in Chrome",
            "type": "pwa-chrome",
            "request": "launch",
            "timeout": 30000,
            // If you have changed the default port / launch URL make sure to update the expectation below
            "url": "http://localhost:5000",
            "webRoot": "${workspaceFolder}/Server",
            "inspectUri": "{wsProtocol}://{url.hostname}:{url.port}/_framework/debug/ws-proxy?browser={browserInspectUri}"
        }
    ]
}
```

https://github.com/dotnet/aspnetcore/issues/17549

/cc @pranavkm @SteveSandersonMS @mkArtakMSFT @rynowak @javiercn @danroth27

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -138,8 +138,12 @@ export class AssetGenerator {
             throw new Error("Startup project not set");
         }
 
-        if (this.startupProject.IsBlazorWebAssemblyProject) {
-            return ProgramLaunchType.BlazorWebAssembly;
+        if (this.startupProject.IsBlazorWebAssemblyStandalone) {
+            return ProgramLaunchType.BlazorWebAssemblyStandalone;
+        }
+
+        if (this.startupProject.IsBlazorWebAssemblyHosted) {
+            return ProgramLaunchType.BlazorWebAssemblyHosted;
         }
 
         if (this.startupProject.IsWebProject) {
@@ -172,7 +176,7 @@ export class AssetGenerator {
         return path.join('${workspaceFolder}', path.relative(this.workspaceFolder.uri.fsPath, startupProjectDir));
     }
 
-    public createLaunchJson(programLaunchType: ProgramLaunchType): string {
+    public createLaunchJsonConfigurations(programLaunchType: ProgramLaunchType): string {
         switch (programLaunchType) {
             case ProgramLaunchType.Console: {
                 const launchConfigurationsMassaged: string = indentJsonString(createLaunchConfiguration(this.computeProgramPath(), this.computeWorkingDirectory()));
@@ -192,13 +196,45 @@ export class AssetGenerator {
     ${attachConfigurationsMassaged}
 ]`;
             }
-            case ProgramLaunchType.BlazorWebAssembly: {
-                const webLaunchConfigurationsMassaged: string = indentJsonString(createBlazorWebAssemblyLaunchConfiguration(this.computeWorkingDirectory()));
+            case ProgramLaunchType.BlazorWebAssemblyHosted: {
+                const chromeLaunchConfigurationsMassaged: string = indentJsonString(createBlazorWebAssemblyLaunchConfiguration(this.computeWorkingDirectory()));
+                const hostedLaunchConfigurationsMassaged: string = indentJsonString(createWebAssemblyHostedLaunchConfiguration(this.computeProgramPath(), this.computeWorkingDirectory()));
+                return `
+[
+    ${hostedLaunchConfigurationsMassaged},
+    ${chromeLaunchConfigurationsMassaged}
+]`;
+            }
+            case ProgramLaunchType.BlazorWebAssemblyStandalone: {
+                const chromeLaunchConfigurationsMassaged: string = indentJsonString(createBlazorWebAssemblyLaunchConfiguration(this.computeWorkingDirectory()));
                 const devServerLaunchConfigurationMassaged: string = indentJsonString(createBlazorWebAssemblyDevServerLaunchConfiguration(this.computeWorkingDirectory()));
                 return `
 [
     ${devServerLaunchConfigurationMassaged},
-    ${webLaunchConfigurationsMassaged}
+    ${chromeLaunchConfigurationsMassaged}
+]`;
+            }
+        }
+    }
+
+    public createLaunchJsonCompounds(programLaunchType: ProgramLaunchType): string | undefined {
+        switch (programLaunchType) {
+            case ProgramLaunchType.Console:
+            case ProgramLaunchType.Web: {
+                return;
+            }
+            case ProgramLaunchType.BlazorWebAssemblyHosted: {
+                const launchAndDebugCompoundMassaged: string = indentJsonString(createBlazorWebAssemblyHostedLaunchCompound());
+                return `
+[
+${launchAndDebugCompoundMassaged}
+]`;
+            }
+            case ProgramLaunchType.BlazorWebAssemblyStandalone: {
+                const launchAndDebugCompoundMassaged: string = indentJsonString(createBlazorWebAssemblyStandaloneLaunchCompound());
+                return `
+[
+${launchAndDebugCompoundMassaged}
 ]`;
             }
         }
@@ -272,7 +308,8 @@ export class AssetGenerator {
 export enum ProgramLaunchType {
     Console,
     Web,
-    BlazorWebAssembly,
+    BlazorWebAssemblyHosted,
+    BlazorWebAssemblyStandalone,
 }
 
 export function createWebLaunchConfiguration(programPath: string, workingDirectory: string): string {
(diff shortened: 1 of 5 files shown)


src/assets.ts
```diff
@@ -272,6 +301,34 @@ export function createWebLaunchConfiguration(programPath: string, workingDirecto
...
+    return `
+{
+    "name": "Start Blazor web assembly dev server",
+    "type": "coreclr",
+    "request": "launch",
+    "program": "dotnet",
+    "args": ["run"],
+    "cwd": "${util.convertNativePathToPosix(workingDirectory)}",
+    "env": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
```
NTaylorMullen: Is this needed?

src/assets.ts
```diff
@@ -272,6 +301,34 @@ export function createWebLaunchConfiguration(programPath: string, workingDirecto
...
 }
 
+export function createBlazorWebAssemblyLaunchConfiguration(workingDirectory: string): string {
+    return `
+{
+    "name": "Launch Blazor web assembly project in Chrome",
+    "type": "pwa-chrome",
+    "request": "launch",
+    // If you have changed the default port / launch URL make sure to update the expectation below
+    "url": "https://localhost:5001",
```
NTaylorMullen: Leaving this as the https variant given I assume we want that as the default.

NTaylorMullen:
Note debugger folks: I wasn't able to actually hit breakpoints (but could run and attach) with the private debugger builds due to bugs so please verify my assumptions are correct in the launch.json pieces.

filipw:
we already sniff out the Unity projects on the server side in a somewhat similar situation https://github.com/OmniSharp/omnisharp-roslyn/blob/51cf29ca5785ad4973ad0579ece445e7d1992484/src/OmniSharp.MSBuild/Models/MSBuildProjectInfo.cs#L23

I think Blazor WASM could be handled in a similar way, it would be more consistent?

filipw:
actually, it probably doesn't matter since it can be done entirely on the client side, there is probably no point 👍
we can move it over once it becomes more complex e.g. needs to rely on some msbuild task

src/omnisharp/utils.ts
```diff
@@ -121,4 +133,46 @@ export async function debugTestStop(server: OmniSharpServer, request: protocol.V
...
+
+    return true;
+}
+
+function isWebProject(project: MSBuildProject): boolean {
+    let projectFileText = fs.readFileSync(project.Path, 'utf8');
+
+    // Assume that this is an MSBuild project. In that case, look for the 'Sdk="Microsoft.NET.Sdk.Web"' attribute.
+    // TODO: Have OmniSharp provide the list of SDKs used by a project and check that list instead.
+    return projectFileText.toLowerCase().indexOf('sdk="microsoft.net.sdk.web"') >= 0;
```
filipw: we can expose this TODO stuff via the existing workspace information endpoint

NTaylorMullen: This was already there, just moved the method :smi

NTaylorMullen:
@JoeRobich should I be concerned about the `codecov/patch` PR check?

JoeRobich:
@NTaylorMullen I don't think so. Their detail page even says everything is the same

src/assets.ts
```diff
@@ -272,6 +301,34 @@ export function createWebLaunchConfiguration(programPath: string, workingDirecto
 }`;
 }
 
+export function createBlazorWebAssemblyLaunchConfiguration(workingDirectory: string): string {
+    return `
+{
+    "name": "Launch Blazor web assembly project in Chrome",
+    "type": "pwa-chrome",
```
WardenGnaw: Is `pwa-chrome` a type built into VSCode or does it depend on another extension being installed?

NTaylorMullen: It's built into the node debugger for VSCode (not yet released though)

src/assets.ts
SteveSandersonMS: Yes

SteveSandersonMS:
> a heuristic on the client to decide if a project looks like a Blazor WASM project

This is great for the "standalone" Blazor WebAssembly case, but is there anything here to account for the "Blazor WebAssembly hosted in ASP.NET Core" case? This is equally important.

If you want a detailed heuristic for this, it could be something like "an ASP.NET Core project that references another project that matches the heuristic for Blazor WebAssembly". A simpler alternative would just be: "for all ASP.NET Core projects, offer the option to add the '"Launch Blazor web assembly project in Chrome"' configuration".

Note that in the "hosted" case, it's not necessary to add the configuration for launching the Blazor WebAssembly dev server. Instead, we only need the "Launch Blazor web assembly project in Chrome" part.

Does VS Code have a way to do simultaneous mixed-mode debugging (.NET *and* the JS debugger which we use for Mono WebAssembly)? If so that's optimal for this case, as the developer might want to have breakpoints in both their server-side and client-side .NET code. This is how it works in VS. If there isn't a way to do simultaneous mixed-mode debugging then I guess the scenario doesn't apply.

NTaylorMullen:
> Does VS Code have a way to do simultaneous mixed-mode debugging (.NET _and_ the JS debugger which we use for Mono WebAssembly)?

Yup, they just launch multiple configurations simultaneously.

> is there anything here to account for the "Blazor WebAssembly hosted in ASP.NET Core" case? This is equally important.

Nope, nothing here to account for that. In that case the project will be understood as a classic ASP.NET web project. I need to see what it takes to add a configuration option to launch and debug a blazor web assembly project in chrome.

-------------------

With help from others I found a way to utilize compound launch configuration to launch and debug a Blazor WASM project project in one go. Going to work on updating this PR to use that.

NTaylorMullen:
🆙 📅 

Add understanding of Blazor hosted applications.

- Changed how we detect a Blazor web assembly project. The new criteria consists of reading a projects launchSettings.json file and if it has the text `"inspectUri"` to treat it as a Blazor WASM project. This is similar to how VS detects Blazor WASM.
- Added a new `ProgramLaunchType` of `BlazorWebAssemblyHosted` and factored the existing `BlazorWebAssembly` launch type to be `BlazorWebAssemblyStandalone`.
- `BlazorWebAssemblyHosted` is based on the following criteria:
  1. A Blazor web assembly project (`launchSettings.json` containing `"inspectUri"`).
  2. Executable
  3. References the web SDK
  4. Netcoreapp targetting
- Added new tests for the Blazor web assembly hosted project types.

src/omnisharp/utils.ts
```diff
@@ -121,4 +134,58 @@ export async function debugTestStop(server: OmniSharpServer, request: protocol.V
...
 export async function isNetCoreProject(project: protocol.MSBuildProject) {
     return project.TargetFrameworks.find(tf => tf.ShortName.startsWith('netcoreapp') || tf.ShortName.startsWith('netstandard')) !== undefined;
+}
+
+function isBlazorWebAssemblyHosted(project: protocol.MSBuildProject): boolean {
+    if (!isBlazorWebAssemblyProject(project)) {
+        return false;
+    }
+
+    if (!project.IsExe) {
```
pranavkm: Out of curiosity, why is this check important? I'd think the latter two checks would suffice.

NTaylorMullen: Just following suite with how O# typically checks "is runnable web" project.

JoeRobich:
@NTaylorMullen Is this ready to merge?

NTaylorMullen:
> @NTaylorMullen Is this ready to merge?

No not quite. We're still trying to figure out timelines and approaches.

NTaylorMullen:
🆙 📅 

Restrict Blazor WASM asset additions
- Added a restriction to our Blazor detection/asset prefilling tech. The new restriction is:
  1. User must have the `JavaScript Debugger (Nightly)` extension.
  2. User must have the `debug.node.useV3` flag set to `true`
  3. User must have the `debug.chrome.useV3` flag set to `true`

- Removed the launch.json compounds due to timing issues in core VSCode. The new experience now relies on a user hitting "launch Standalone/hosted" and then doing "debug Blazor xyz". As part of this I removed the `priority`presentation order since we no longer need to force ordering with a compound.

/cc @SteveSandersonMS @danroth27 @mkArtakMSFT

src/omnisharp/utils.ts
```diff
@@ -77,6 +77,18 @@ export async function requestWorkspaceInformation(server: OmniSharpServer) {
             project.IsBlazorWebAssemblyHosted = blazorDetectionEnabled && isBlazorWebAssemblyHosted(project);
             project.IsBlazorWebAssemblyStandalone = blazorDetectionEnabled && !project.IsBlazorWebAssemblyHosted && isBlazorWebAssemblyProject(project);
         }
+
+        if (!blazorDetectionEnabled && response.MsBuild.Projects.some(project => isBlazorWebAssemblyProject(project))) {
+            // There's a Blazor Web Assembly project but VSCode isn't configured to debug the WASM code, show a notification
+            // to help the user configure their VSCode appropriately.
+            vscode.window.showInformationMessage('Your VSCode is not configured to debug Blazor web assembly applications.', 'Learn more', 'Close')
```
danroth27:
```suggestion
            vscode.window.showInformationMessage('Additional setup is required to debug Blazor WebAssembly apps.', 'Learn more', 'Close')
```

src/omnisharp/utils.ts
```diff
@@ -81,7 +81,7 @@ export async function requestWorkspaceInformation(server: OmniSharpServer) {
         if (!blazorDetectionEnabled && response.MsBuild.Projects.some(project => isBlazorWebAssemblyProject(project))) {
             // There's a Blazor Web Assembly project but VSCode isn't configured to debug the WASM code, show a notification
             // to help the user configure their VSCode appropriately.
-            vscode.window.showInformationMessage('Your VSCode is not configured to debug Blazor web assembly applications.', 'Learn more', 'Close')
+            vscode.window.showInformationMessage('Additional setup is required to debug Blazor WebAssembly apps.', 'Learn more', 'Close')
```
SteveSandersonMS: Not a big deal at all, but if you're open to adding further tweaks, I'd suggest:

```suggestion
            vscode.window.showInformationMessage('Additional setup is required to debug Blazor WebAssembly applications.', 'Learn more', 'Close')
```

src/assets.ts
SteveSandersonMS: Yes. Technically it would be possible to infer the HTTPS port from `launchSettings.json` but 5001 is indeed the default.

Inferring it doesn't make a huge difference because (1) the default is already correct, and (2) for a developer who's in the habit of changing it, once they've generated `launch.json` there would be nothing keeping it in sync with subsequent changes in `launchSettings.json`.

So basically I think this is fine!

src/assets.ts
```diff
@@ -382,43 +354,10 @@ export function createBlazorWebAssemblyDevServerLaunchConfiguration(workingDirec
     "type": "coreclr",
     "request": "launch",
     "program": "dotnet",
-    "args": ["run", "--no-build"],
+    "args": ["run"],
```
SteveSandersonMS: That's a good change. I was a bit concerned about people getting tripped up by `--no-build` before.

src/omnisharp/utils.ts
SteveSandersonMS: Oh I now see Dan specifically recommended the `apps` phrasing. Seems a bit unusual to me.

I would find `applications` more normal in this context but don't want to block this PR on such a minor detail. @danroth27 was this intentional?

SteveSandersonMS:
> Added a restriction to our Blazor detection/asset prefilling tech. The new restriction is:
> User must have the JavaScript Debugger (Nightly) extension.
> User must have the debug.node.useV3 flag set to true
> User must have the debug.chrome.useV3 flag set to true

What happens longer term? I'm guessing the V3 debugger becomes the default at that point, and then people will no longer have the nightly extension nor these config flags.

What's the best way for us to (a) know when that change is going to roll into the mainline product, so we can remove this logic in the same release, or (b) better still, set up the logic so that it automatically works once the V3 debugger is the default?

SteveSandersonMS:
@NTaylorMullen This looks super to me. I don't get the option to leave a proper sign-off but FWIW I would do.

I would still like to have your thoughts on how [we should best react when the V3 debugger becomes default](https://github.com/OmniSharp/omnisharp-vscode/pull/3593#issuecomment-591372815), and if there's any reasonable way to deal with it in advance. But besides that I think we're all good here!

src/omnisharp/utils.ts
NTaylorMullen: I like applications better as well. WIll commit the suggestion

NTaylorMullen:
> I would still like to have your thoughts on how [we should best react when the V3 debugger becomes default](https://github.com/OmniSharp/omnisharp-vscode/pull/3593#issuecomment-591372815), and if there's any reasonable way to deal with it in advance. But besides that I think we're all good here!

So we'll have to pre-emptively make a change and release to O# to account for this where we'll add an additional condition where if the release extension (`javascript-debugger` I assume) is of a certain version or higher to then skip our nightly extension checks.

As for how we'll know when they're about to RTM; hopefully they'll be kind enough to let us know a date once one is confirmed at which point we can schedule the work on our end.

PR #3648, opened 2020-03-10
Update OmniSharp to 1.34.14

JoeRobich:


JoeRobich:
@brettfo Please take a look

Issue #3565, opened 2020-02-09
[Request] Provide tokens for VS Code semantic syntax highlighting


dannymcgee:
[Semantic highlighting](https://github.com/microsoft/vscode/wiki/Semantic-Highlighting-Overview) is finally in preview as of VS Code 1.42.0. C# is an ideal candidate for taking advantage of this feature, as the available TextMate grammar is pretty lacking, and the syntax of C# is such that it's impossible to accurately classify many tokens with a simple line-by-line regex parser.

- [SemanticTokensProvider API (proposed)](https://github.com/microsoft/vscode/wiki/Semantic-Highlighting-Overview)
- [Basic extension sample](https://github.com/microsoft/vscode-extension-samples/tree/master/semantic-tokens-sample)
- [VS Code's implementation of semantic tokens for JavaScript](https://github.com/microsoft/vscode/blob/master/extensions/html-language-features/server/src/modes/javascriptSemanticTokens.ts)

I'd love to help out with this, but I'm not at all familiar with the codebase here. If this is a feature you would like to support, I'd be more than happy to open a PR if someone would be willing to provide some guidance or a push in the right direction.

JoeRobich:
@dannymcgee Please try out the [1.21.19-beta1 prerelease](https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.21.19-beta1)

JoeRobich:
Closing as SemanticHighlighting is available in the 1.22.0 release.

PR #3667, opened 2020-03-15
Add experimental Semantic Highlighter

JoeRobich:
Implementation of the Semantic Tokens provider (https://github.com/microsoft/vscode/issues/86415) requested in https://github.com/OmniSharp/omnisharp-vscode/issues/3565. Built on the Semantic Endpoint added to OmniSharp in https://github.com/OmniSharp/omnisharp-roslyn/pull/1734

Future enhancement:
- [ ] Add RegEx classifications

With Textmate Classification:
image

With Semantic Classification:
image

With Semantic Classification and Visual Studio 2019 theme tweaks:
image

<details>
  <summary>Applied theme tweaks: </summary>

```json
    "editor.tokenColorCustomizations": {
        "[Default Dark+]": {
            "textMateRules": [
                {
                    "scope": "entity.name.namespace",
                    "settings": {
                        "foreground": "#D4D4D4"
                    }
                },
                {
                    "scope": "entity.name.variable.field",
                    "settings": {
                        "foreground": "#D4D4D4"
                    }
                },
                {
                    "scope": "variable.other.property",
                    "settings": {
                        "foreground": "#D4D4D4"
                    }
                },
                {
                    "scope": "variable.other.constant",
                    "settings": {
                        "foreground": "#D4D4D4",
                    }
                },
                {
                    "scope": "variable.other.enummember",
                    "settings": {
                        "foreground": "#D4D4D4"
                    }
                },
                {
                    "scope": "entity.name.type.interface",
                    "settings": {
                        "foreground": "#b8d7a3"
                    }
                },
                {
                    "scope": "entity.name.type.enum",
                    "settings": {
                        "foreground": "#b8d7a3"
                    }
                },
                {
                    "scope": "entity.name.type.parameter",
                    "settings": {
                        "foreground": "#b8d7a3"
                    }
                },
                {
                    "scope": "storage.type.struct",
                    "settings": {
                        "foreground": "#86C691",
                    }
                },
                {
                    "scope": "entity.name.function.extension",
                    "settings": {
                        "foreground": "#DCDCAA",
                    }
                },
                {
                    "scope": "comment.documentation",
                    "settings": {
                        "foreground": "#608B4E",
                    }
                },
                {
                    "scope": "comment.documentation.attribute",
                    "settings": {
                        "foreground": "#C8C8C8",
                    }
                },
                {
                    "scope": "comment.documentation.cdata",
                    "settings": {
                        "foreground": "#E9D585",
                    }
                },
                {
                    "scope": "comment.documentation.delimiter",
                    "settings": {
                        "foreground": "#808080",
                    }
                },
                {
                    "scope": "comment.documentation.name",
                    "settings": {
                        "foreground": "#569CD6",
                    }
                }
            ]
        },
        "[Default Light+]": {
            "textMateRules": [
                {
                    "scope": "entity.name.namespace",
                    "settings": {
                        "foreground": "#222222"
                    }
                },
                {
                    "scope": "entity.name.variable.field",
                    "settings": {
                        "foreground": "#222222"
                    }
                },
                {
                    "scope": "variable.other.property",
                    "settings": {
                        "foreground": "#222222"
                    }
                },
                {
                    "scope": "variable.other.constant",
                    "settings": {
                        "foreground": "#222222",
                    }
                },
                {
                    "scope": "variable.other.enummember",
                    "settings": {
                        "foreground": "#222222"
                    }
                },
                {
                    "scope": "entity.name.type.interface",
                    "settings": {
                        "foreground": "#267f99"
                    }
                },
                {
                    "scope": "entity.name.type.enum",
                    "settings": {
                        "foreground": "#267f99"
                    }
                },
                {
                    "scope": "entity.name.type.parameter",
                    "settings": {
                        "foreground": "#267f99"
                    }
                },
                {
                    "scope": "storage.type.struct",
                    "settings": {
                        "foreground": "#267f99",
                    }
                },
                {
                    "scope": "entity.name.function.extension",
                    "settings": {
                        "foreground": "#795E26",
                    }
                },
                {
                    "scope": "comment.documentation",
                    "settings": {
                        "foreground": "#008000",
                    }
                },
                {
                    "scope": "comment.documentation.attribute",
                    "settings": {
                        "foreground": "#282828",
                    }
                },
                {
                    "scope": "comment.documentation.cdata",
                    "settings": {
                        "foreground": "#808080",
                    }
                },
                {
                    "scope": "comment.documentation.delimiter",
                    "settings": {
                        "foreground": "#808080",
                    }
                },
                {
                    "scope": "comment.documentation.name",
                    "settings": {
                        "foreground": "#808080",
                    }
                }
            ]
        }
    },
```
</details>


NTaylorMullen:
@ryanbrandenburg since you own our semantic classification you should take a look at this as well. It will also enable us to call through to Roslyn for semantic colorization as well 😄

src/omnisharp/extension.ts
```diff
@@ -92,6 +93,13 @@ export async function activate(context: vscode.ExtensionContext, packageJSON: an
         localDisposables.add(forwardChanges(server));
         localDisposables.add(trackVirtualDocuments(server, eventStream));
         localDisposables.add(vscode.languages.registerFoldingRangeProvider(documentSelector, new StructureProvider(server, languageMiddlewareFeature)));
+
+        // Do not set during tests
+        if (process.env.OSVC_SUITE === undefined && vscode.languages.registerDocumentSemanticTokensProvider !== undefined) {
+            const semanticTokensProvider = new SemanticTokensProvider(server, languageMiddlewareFeature);
+            localDisposables.add(vscode.languages.registerDocumentSemanticTokensProvider(documentSelector, semanticTokensProvider, semanticTokensProvider.getLegend()));
```
NTaylorMullen: I believe these will explode in a published extension. Aka, I don't believe you're allowed to publish an official extension with proposed APIs unless approved by VSCode folks; we've added TypeScripty equivalent ifdefs to ensure we don't explode in those circumstances.

src/features/semanticTokensProvider.ts
```diff
@@ -0,0 +1,315 @@
...
+    async _provideSemanticTokens(document: vscode.TextDocument, range: protocol.V2.Range, token: vscode.CancellationToken): Promise<vscode.SemanticTokens | null> {
+        let req = createRequest<protocol.V2.SemanticHighlightRequest>(document, new vscode.Position(0, 0));
+        req.Range = range;
+
+        let response = await serverUtils.getSemanticHighlights(this._server, req);
+
+        const semanticSpans = response.Spans;
+
+        const builder = new vscode.SemanticTokensBuilder();
+        for (let span of semanticSpans) {
```
NTaylorMullen: Maybe I'm misunderstanding the code but why not have the server just return the LSP compatible type vs. re-building it here?

src/features/semanticTokensProvider.ts
```diff
@@ -0,0 +1,315 @@
...
+            // supported as a readonly variable. Until we start getting more complete
+            // modifiers from the highlight service we can add the readonly modifier here.
+            if (span.Type === SemanticHighlightClassification.ConstantName) {
+                tokenModifiers += 2 ** DefaultTokenModifier.readonly;
+            }
+
+            var spanRange = toRange2(span);
+            for (let line = spanRange.start.line; line <= spanRange.end.line; line++) {
+                const startCharacter = (line === spanRange.start.line ? spanRange.start.character : 0);
+                const endCharacter = (line === spanRange.end.line ? spanRange.end.character : document.lineAt(line).text.length);
```
NTaylorMullen: Hmmm, you're making the assumption that document.lineAt(line - 1).text is equivalent to what the language server saw right? This is after several await's, at this point it's super possible the document representation has changed by now. Typically VSCode will map-forward whatever results you provide to the latest document they know about because they're aware of what doc version they originally requested

src/features/semanticTokensProvider.ts
```diff
@@ -0,0 +1,315 @@
...
+}
+
+enum SemanticHighlightModifier {
+    Static
+}
+
+export default class SemanticTokensProvider extends AbstractProvider implements vscode.DocumentSemanticTokensProvider, vscode.DocumentRangeSemanticTokensProvider {
+
+    getLegend(): vscode.SemanticTokensLegend {
+        return new vscode.SemanticTokensLegend(tokenTypes, tokenModifiers);
```
NTaylorMullen: The server's typically the one that defines the legend. Why re-create it on the client (vs. querying it)?

src/features/semanticTokensProvider.ts
NTaylorMullen: This is all based off the assumption that vscode.TextDocument is mutable. Which i'm pretttty sure it is 😄 . I don't believe they run on a snaphsot system

src/omnisharp/extension.ts
JoeRobich: Thanks, I'll have to check on that.

src/features/semanticTokensProvider.ts
JoeRobich: I was just following the [typescript extension](https://github.com/microsoft/vscode/blob/master/extensions/typescript-language-features/src/features/semanticTokens.ts#L65) as an example where they request information from their server then perform some post processing. I'll have to think more on whether to keep it like this.

src/features/semanticTokensProvider.ts
JoeRobich: Thanks for pointing this out. I added back a version check that was present in the TypeScript Extensions implementation.

src/features/semanticTokensProvider.ts
JoeRobich: Yeah goes back to whether the OmniSharp endpoint should just implement the LSP return types. I'll think on it.

src/features/semanticTokensProvider.ts
NTaylorMullen: Ah, TypeScript hasn't moved fully over to LSP yet which is why they follow that model. They plan to move fully over to LSP

JoeRobich:
@NTaylorMullen Would like to include this in the next release. We'll consider it experimental (defaulted off) and collect feedback about the performance. Ultimately we may implement an API that is closer to the LSP, but that may wait until Roslyn implements a classification service that returns tokens and modifiers.

NTaylorMullen:
> Ultimately we may implement an API that is closer to the LSP, but that may wait until Roslyn implements a classification service that returns tokens and modifiers.

Ah, when that comes to fruition that'll enable us to add semantic C# colorization to Razor's C# pieces. Is that something that's tracked somewhere? FYI @ryanbrandenburg 

> We'll consider it experimental (defaulted off) and collect feedback about the performance.

What type of switch do you plan to put it behind? Asking because our TagHelper semantic highlighting is cheap and is something we'd have "on" all the time however if we were to semantically highlight our C# pieces of the document we'd want it behind a similar switch to ensure it's only active when yours is active.

JoeRobich:
> Is that something that's tracked somewhere?

We didn't have an exact issue tracking this, so I made a proposal. https://github.com/dotnet/roslyn/issues/44155

> What type of switch do you plan to put it behind?

There is a `csharp.semanticHighlighting.enabled` setting which is available off of the `Options` class.

JoeRobich:
Although GitHub checks didn't update, this is the passing Travis CI for this PR - https://travis-ci.org/github/OmniSharp/omnisharp-vscode/builds/685827799

Merging

JoeRobich:
for good measure this is the passing master build following this merge https://travis-ci.org/github/OmniSharp/omnisharp-vscode/builds/685844034

PR #3678, opened 2020-03-18
Check for undefined

ryanbrandenburg:
`showWarningMessage` returns undefined in the case that the selection from the message is to press the `x` instead of one of the 'option' buttons, which causes an Error without some sort of check.


ajaybhargavb:
@JoeRobich

Issue #3664, opened 2020-03-14
C# plugin hangs and cause high CPU usage while opening a big project 


heshuimu:
## Issue Description ##
Since C# plugin 1.21.13, the C# plugin seems to hang when I open a project with more than 700 files. There is also a side effect that causes `extensionHosts` to max out a single CPU core. It can be reproduced 100% of the time and I have not experience the hang to finish at any point. 

**EDIT: ** My project is a Unity project

1.21.12 was the last version that was working properly for me. 

## Steps to Reproduce ##
1. Open my project with C# plugin 1.21.13 or 1.21.14 installed
2. Observe OmniSharp log to stop at a point
3. Observe `extensionHosts` using nearly 100 unit of CPU
![image](https://user-images.githubusercontent.com/7824592/76670500-d5694a80-655e-11ea-92b9-3fec00089d53.png)

## Expected Behavior ##
Project loads properly (Omnisharp prints "solution initialized") in about a minute, Code Lens and jump to declaration works properly. 

## Actual Behavior ##
Project does not finish loading, Reference Code Lens does not appear and jump to declaration does not work.

### OmniSharp log ###
The log contains file names to an already-live project, so I am hesitant to share. 
In summary, the log stops at the same line of text every time when the issue occurs.

### C# log ###
None

## Environment information ##

**VSCode version**: 1.43.0
**C# Extension**: 1.21.13

<details><summary>Mono Information</summary>
        OmniSharp using global mono :6.8.0</details>
<details><summary>Dotnet Information</summary>
.NET Core SDK (reflecting any global.json):
 Version:   3.0.101
 Commit:    bc5f8df0f5

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.15
 OS Platform: Darwin
 RID:         osx.10.15-x64
 Base Path:   /usr/local/share/dotnet/sdk/3.0.101/

Host (useful for support):
  Version: 3.0.1
  Commit:  32085cbc72

.NET Core SDKs installed:
  2.1.103 [/usr/local/share/dotnet/sdk]
  2.1.505 [/usr/local/share/dotnet/sdk]
  3.0.101 [/usr/local/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.9 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
[...]

PR #3679, opened 2020-03-18
Make launchSettings.json search performant.

NTaylorMullen:
- Stopped globbing for the `launchSettings.json` and instead we search for it at the root and in the Properties folder.

Fixes omnisharp/omnisharp-vscode#3664

@danroth27 is this awful? I don't think people move this file around all that much so wanted to take the approach that impacted O# the least.

### DIFF
--- a/src/omnisharp/utils.ts
+++ b/src/omnisharp/utils.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as fs from 'fs-extra';
-import * as glob from 'glob';
 import { OmniSharpServer } from './server';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


NTaylorMullen:
> looks reasonable to me. this should be shipped asap I think given the amount of issue we got 😀

Agree, I pinged @danroth27 offline to get the thumbs up from him to be sure I'm not making a big mistake here 😆

danroth27:
Sounds fine to me, but I think this is really @KathleenDollard's area.

NTaylorMullen:
Closing this in favor of https://github.com/OmniSharp/omnisharp-vscode/pull/3681

Issue #3673, opened 2020-03-16
C# Extension is entirely unresponsive


MartinHaeusler:
## Issue Description ##
The C# extension suddenly stopped working for me. Syntax highlighting is okay, but everything else loads indefinitely:
 - Code Completion
 - Formatting
 - Outline
 - ...

All the while, it is consuming 100% CPU resources on one of my cores. I even tried a full fresh install of everything, that didn't help.

## Logs ##
[exthost-c2991c.cpuprofile.txt](https://github.com/OmniSharp/omnisharp-vscode/files/4339926/exthost-c2991c.cpuprofile.txt)

[ms-dotnettools.csharp-unresponsive.cpuprofile.txt](https://github.com/OmniSharp/omnisharp-vscode/files/4339933/ms-dotnettools.csharp-unresponsive.cpuprofile.txt)

### OmniSharp log ###

<details><summary>OmniSharp Log</summary>

Starting OmniSharp server at 16.3.2020, 20:07:45
    Target: d:\git\HoBC\Unity\HoBC\HoBC.sln

OmniSharp server started.
    Path: C:\Users\Martin\.vscode\extensions\ms-dotnettools.csharp-1.21.14\.omnisharp\1.34.14\OmniSharp.exe
    PID: 4360

[info]: OmniSharp.Stdio.Host
        Starting OmniSharp on Windows 6.2.9200.0 (x64)
[info]: OmniSharp.Services.DotNetCliService
        DotNetPath set to dotnet
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        Located 2 MSBuild instance(s)
            1: Visual Studio Build Tools 2019 16.5.29911.84 - "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin"
            2: StandAlone 16.4 - "c:\Users\Martin\.vscode\extensions\ms-dotnettools.csharp-1.21.14\.omnisharp\1.34.14\.msbuild\Current\Bin"
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        Registered MSBuild instance: Visual Studio Build Tools 2019 16.5.29911.84 - "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin"
[info]: OmniSharp.Cake.CakeProjectSystem
        Detecting Cake files in 'd:\git\HoBC\Unity\HoBC'.
[...]

PR #3681, opened 2020-03-18
Use workspace.findFiles instead of glob

JoeRobich:
`isBlazorWebAssemblyProject` used `glob` to search for launchSettings.json files. It was being invoked 3 different times for each project within the workspace by `requestWorkspaceInformation`. 

Based on this advice:
```
	 * The workspace offers support for [listening](#workspace.createFileSystemWatcher) to fs
	 * events and for [finding](#workspace.findFiles) files. Both perform well and run _outside_
	 * the editor-process so that they should be always used instead of nodejs-equivalents.
```

Updated `isBlazorWebAssemblyProject` to use `vscode.workspace.findFiles` and also use the excluded paths defined in `files.exclude`

Potential fix for #3673

### DIFF
--- a/src/omnisharp/options.ts
+++ b/src/omnisharp/options.ts
@@ -119,26 +119,30 @@ export class Options {
         );
     }
 
-    public static getExcludedPaths(vscode: vscode): string[] {
-        let excludePaths: string[] = [];
-
+    public static getExcludedPaths(vscode: vscode, includeSearchExcludes: boolean = false): string[] {
         let workspaceConfig = vscode.workspace.getConfiguration();
         if (!workspaceConfig) {
-            return excludePaths;
+            return [];
         }
 
-        let excludeFilesOption = workspaceConfig.get<{ [i: string]: boolean }>('files.exclude');
-        if (!excludeFilesOption) {
-            return excludePaths;
-        }
+        let excludePaths = getExcludes(workspaceConfig, 'files.exclude');
 
-        for (let field in excludeFilesOption) {
-            if (excludeFilesOption[field]) {
-                excludePaths.push(field);
-            }
+        if (includeSearchExcludes) {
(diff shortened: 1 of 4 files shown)


src/omnisharp/options.ts
```diff
@@ -134,6 +119,28 @@ export class Options {
...
 
+    public static getExcludedPaths(vscode: vscode): string[] {
+        let excludePaths: string[] = [];
+
+        let workspaceConfig = vscode.workspace.getConfiguration();
+        if (!workspaceConfig) {
+            return excludePaths;
+        }
+
+        let excludeFilesOption = workspaceConfig.get<{ [i: string]: boolean }>('files.exclude');
```
NTaylorMullen: Out of curiosity does VSCode pre-fill node_modules in this list?

JoeRobich: unfortunately it wasn't set for the `dotnet new razor` project I created to test with.

NTaylorMullen: It might make sense for us to auto-include `node_modules` and whatever Unity's package folder is if it doesn't already exist.

JoeRobich: Maybe this should also include `search.exclude`.. hmm..

JoeRobich: Good news is that Unity seems to create a .vscode/settings.json which has an extensive `files.exclude` defined

JoeRobich: Updated to include `search.exclude` paths which by [default include `node_modules`](https://github.com/microsoft/vscode/blob/f68b1bd2a3045f4ca2153e0957a1fda168d5e467/src/vs/workbench/contrib/search/browser/search.contribution.ts#L687). Unity's settings.json includes the Library/PackageCache in its `files.exclude` so I think we are covered from these scenarios.

PR #3682, opened 2020-03-19
Update changelog for release

JoeRobich:


JoeRobich:
Merging so we can validate performance fix

PR #3772, opened 2020-05-09
Add vscode implemention of the 'RunTestsInContext' and `DebugTestsInContext' commands

333fred:
This is the omnisharp-vsocde side of https://github.com/OmniSharp/omnisharp-roslyn/pull/1782. I've implemented support for both of those commands, and bound them to the same default shortcuts as in VS proper. These commands can also be invoked from the command palette.

I've also done a small bit of refactoring to remove references to legacy project.json support. There is no support for loading this project type in Omnisharp anymore, so this was cruft that was unnecessary at this point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -911,7 +911,7 @@
       {
         "command": "dotnet.test.debugTestsInContext",
         "key": "ctrl+r ctrl+t",
-        "mac": "cmd+r ctrl+t",
+        "mac": "cmd+r cmd+t",
         "when": "editorLangId == csharp && editorTextFocus"
       }
     ],
--- a/package.json
+++ b/package.json
@@ -3272,6 +3272,18 @@
           "when": "resourceLangId == aspnetcorerazor"
         }
       ],
+      "editor/context": [
+        {
+          "command": "dotnet.test.runTestsInContext",
+          "when": "editorLangId == csharp",
+          "group": "2_dotnet@1"
+        },
+        {
+          "command": "dotnet.test.debugTestsInContext",
+          "when": "editorLangId == csharp",
+          "group": "2_dotnet@2"
+        }
+      ],
       "commandPalette": [
         {
           "command": "dotnet.test.runTestsInContext",
--- a/package.json
+++ b/package.json
@@ -884,6 +884,16 @@
         "command": "razor.reportIssue",
         "title": "Report a Razor issue",
         "category": "Razor"
+      },
+      {
+        "command": "dotnet.test.runTestsInContext",
+        "title": "Run Tests in Context",
+        "category": ".NET"
+      },
+      {
+        "command": "dotnet.test.debugTestsInContext",
+        "title": "Debug Tests in Context",
+        "category": ".NET"
       }
     ],
     "keybindings": [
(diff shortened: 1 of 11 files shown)


333fred:
Some screenshots:

image

image

image

image

image

image

Note: that last one is hard to actually get, you have to bind the command to your own keyboard shortcut without a `when` clause or edit the existing `when` clause.

package.json
```diff
@@ -884,13 +884,35 @@
...
     "keybindings": [
       {
         "command": "o.showOutput",
         "key": "Ctrl+Shift+F9",
         "mac": "Cmd+Shift+F9"
+      },
+      {
+        "command": "dotnet.test.runTestsInContext",
+        "key": "ctrl+r t",
+        "mac": "cmd+r t",
```
333fred, line 908: I don't actually know what the keyboard shortcuts for run/debug tests in VS for Mac are, so I've just done the equivalent. If there's a better one to use, I'm all ears.

src/omnisharp/protocol.ts
```diff
@@ -284,7 +284,6 @@ export interface AutoCompleteResponse {
 
 export interface ProjectInformationResponse {
     MsBuildProject: MSBuildProject;
-    DotNetProject: DotNetProject;
```
333fred, line 287: Omnisharp no longer has a project information provider for `project.json` projects, so anything that used this was not used.

src/omnisharp/protocol.ts
```diff
@@ -477,8 +476,10 @@ export namespace V2 {
         export const GetTestStartInfo = '/v2/getteststartinfo';
```
333fred, line 476: This command can probably be removed as well, as it was used by the `project.json` support. Let me know if you'd like me to go through and remove it as well.

src/features/dotnetTest.ts
```diff
@@ -435,6 +458,46 @@ export default class TestManager extends AbstractProvider {
...
+        targetFrameworkVersion: string,
+        debugEventListener: DebugEventListener): Promise<LaunchConfiguration | null> {
+
+        const listener = this._server.onTestMessage(e => {
+            this._eventStream.post(new DotNetTestMessage(e.Message));
+        });
+
+        const request: protocol.V2.DebugTestsInContextGetStartInfoRequest = {
+            FileName: fileName,
+            Line: line + 1,
```
RikkiGibson, line 476: is this due to vscode using zero-indexed positions while roslyn uses one-indexed positions?

333fred: No. It's because the deserializer expects the incoming position to be 1-indexed. Both vscode and roslyn actually expect 0-indexed.

package.json
JoeRobich: Those work for me

src/omnisharp/protocol.ts
JoeRobich: Thanks for removing!

src/omnisharp/protocol.ts
JoeRobich: If you are interested, you could remove in a follow up PR or open a tracking issue with the suggestion. Thanks!

333fred:
I've additionally added the commands to the right-click context menu in C# editors. They looks like this:
image

package.json
```diff
@@ -3249,7 +3271,29 @@
           "command": "razor.reportIssue",
           "when": "resourceLangId == aspnetcorerazor"
         }
+      ],
+      "editor/context": [
+        {
+          "command": "dotnet.test.runTestsInContext",
+          "when": "editorLangId == csharp",
+          "group": "2_dotnet@1"
```
333fred, line 3279: Vscode groups are documented here: https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups

`2_` puts them after find all refs/rename, but before copy/paste. The `@` symbol on the end denotes relative ordering of commands in a group.

JoeRobich: I dig it

PR #3790, opened 2020-05-16
Add Visual Studio 2019 themes with semantic colors

JoeRobich:
Adding color themes that show off the semantic highlighting and closely match Visual Studio 2019 editor colors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -389,6 +389,18 @@
     "workspaceContains:**/*.cake"
   ],
   "contributes": {
+    "themes": [
+      {
+        "label": "Visual Studio 2019 Dark",
+        "uiTheme": "vs-dark",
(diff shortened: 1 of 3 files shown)


NTaylorMullen:
@ryanbrandenburg would it make sense to include Razor's semantic goodness here?

ryanbrandenburg:
My current understand is that we will not need to make edits to themes to support Razor Semantic Colorization, we should be able to just use the default scopes.

JoeRobich:
@ryanbrandenburg @NTaylorMullen What we found was that colors were not defined for the scopes we were using which caused VS Code to fall back to the textmate grammar scopes. So to better show off the semantic colors we wanted to include themes that really made it apparent.

PR #3801, opened 2020-05-21
Updated Razor TextMate Grammar

TanayParikh:
Updates to the Razor TextMate Grammar (Implicit / Explicit Razor Expressions within attributes)


captainsafia:
cc: @JoeRobich This is the follow-on change we needed to make.

TanayParikh:
Here's the updated file @JoeRobich! 

FYI: @NTaylorMullen, @ryanbrandenburg, @captainsafia

PR #3802, opened 2020-05-21
Fix the color of comment punctuation

JoeRobich:


JoeRobich:
Merging as non-functional change.

PR #3807, opened 2020-05-26
Set IsWebProject property before it is used

captainsafia:
Closes https://github.com/dotnet/aspnetcore/issues/22151.

The `IsWebProject` property is used inside the `isBlazorWebAssemblyHosted` function before it is set. As a result, this causes the `isBlazorWebAssemblyHosted` to incorrectly not identify hosted Blazor apps.

To fix this, we set the `IsWebProject` property before it is used.

### DIFF
--- a/src/omnisharp/utils.ts
+++ b/src/omnisharp/utils.ts
@@ -77,10 +77,11 @@ export async function requestWorkspaceInformation(server: OmniSharpServer) {
         let blazorWebAssemblyProjectFound = false;
 
         for (const project of response.MsBuild.Projects) {
+            project.IsWebProject = isWebProject(project);
+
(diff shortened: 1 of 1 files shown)


sedat-eyuboglu:
I have the same issue as #22151 When i created a new project with dotnet new blazorwasm --hosted
and let VS Code to create launch.json, i get :

{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch and Debug Standalone Blazor WebAssembly App",
            "type": "blazorwasm",
            "request": "launch",
            "cwd": "${workspaceFolder}/Server"
        }
    ]
}


I have the c# extension 1.23.12.
SDK is 5.0.203 [C:\Program Files\dotnet\sdk]
What can do to overcome this?

PR #3800, opened 2020-05-21
Update Razor to add Blazor debugger and semantic colorization

captainsafia:
* Add support for `blazorwasm` debug adapter
* Adds support for semantic colorization

cc: @ryanbrandenburg @NTaylorMullen

PR #3809, opened 2020-05-27
Insert new version of Razor extension

captainsafia:
A new version of the Razor extension with support for attaching via the Blazor debugger and a bug fix for the rename issue.


TanayParikh:
Is the `package-lock.json` alright? I don't recall it ever being so large, and we didn't change anything fundamental since the last insert (https://github.com/OmniSharp/omnisharp-vscode/pull/3800/files#diff-32607347f8126e6534ebc7ebaec4853d).

JoeRobich:
> Is the `package-lock.json` alright?

Not all our packages are version locked, meaning if you run `npm i` or the asset updater (apparently). Package references are re-evaluated and some packages are allowed to roll-forward to a newer version, which can cause large changes. 

You may ask, "if packages are allowed to roll-forward, can your builds be deterministic?". On the CI side we install packages from information baked into the lock file, so there is no worry of accidentally taking a higher version than expected.

PR #3817, opened 2020-05-31
Adds support for different omnisharp "engines"

david-driscoll:
Adds support for different OmniSharp "engines" and moves to using 0-based indices.

@JoeRobich I'm mainly looking to test the waters on your end if this is something you would be okay with.  At the end of the day I think we only need one engine, but the idea here is that we would have a nice way to fallback to the old one in the event there were problems.

The goal here is to move the extension forward a little bit:
* Today the entire buffer is sent each request and updated on the server side
  * This is pretty inefficient and with big documents causes some problems.
* Today diagnostics / codecheck is done as a client pull instead of server push

This replaces the custom omnisharp protocol with the vscode language client.  However all of the features of the language client are disabled, this means that requests are processed the same as they always have been (with full buffers).  

Consider this phase 1, the goal of phase 2 could be to move away from sending the buffer every time and instead using the text document events built into lsp.  Phase 3 might involve removing some handlers for operations that don't have special behaviors (metadata, tests, etc).

Things that need to be done:

- [ ] Test with mono / linux / mac
- [x] Get the latest lsp shipped into omnisharp-roslyn

This depends on https://github.com/OmniSharp/omnisharp-roslyn/pull/1815 which adds the support for calling exising omnisharp endpoints via LSP.


JoeRobich:
I think this is a step in the right direction. Would definitely want to make sure changing the indexing doesn't break the Razor or Debugger extensions. 

@NTaylorMullen @gregg-miskelly

david-driscoll:
So the reason for the change of indexing is to follow the lsp spec (0 based vs 1 based).  The only reason omnisharp had been ones based originally was due to the legacy how emacs/vim did line numbers internally.

NOTE: Omnisharp does 0-based indexes today, and the configuration setting just turns on a jsonconverter that transforms the 0-based indexes to 1-based.  So today we're translating twice `(0 -> 1, 1 -> 0)` 😄 

I'll take a look and see if there are things that need to be done in razor plugin, but

src/features/diagnosticsProvider.ts
```diff
@@ -127,18 +127,20 @@ class DiagnosticsProvider extends AbstractSupport {
...
         this._analyzersEnabled = vscode.workspace.getConfiguration('omnisharp').get('enableRoslynAnalyzers', false);
         this._validationAdvisor = validationAdvisor;
         this._diagnostics = vscode.languages.createDiagnosticCollection('csharp');
         this._suppressHiddenDiagnostics = vscode.workspace.getConfiguration('csharp').get('suppressHiddenDiagnostics', true);
 
-        this._subscriptions.push(this._validateCurrentDocumentPipe
-            .asObservable()
-            .pipe(throttleTime(750))
-            .subscribe(async x => await this._validateDocument(x)));
+        if (!options.GetLatestOptions().enableLspDriver) {
```
JoeRobich, line 138: Is this because diagnostics are push based with LSP?

test/integrationTests/semanticTokensProvider.test.ts
```diff
@@ -60,19 +60,25 @@ suite(`SemanticTokensProvider: ${testAssetWorkspace.description}`, function () {
...
         let isWorkspaceLoaded = false;
 
-        const subscription = activation.eventStream.subscribe(event => {
-            if (event.type === EventType.WorkspaceInformationUpdated) {
-                isWorkspaceLoaded = true;
-                subscription.unsubscribe();
-            }
-        });
+        if (!activation.wasActive) {
+            const subscription = activation.eventStream.subscribe(event => {
```
JoeRobich, line 68: This won't stop execution. If you had a promise that resolved when `WorkspaceInformationUpdated` then you could await it to stop execution.

JoeRobich (commented):
Very excited for this work to get in.

src/features/diagnosticsProvider.ts
david-driscoll: yeah, diagnostics will be pushed on demand by the server now.

test/integrationTests/semanticTokensProvider.test.ts
david-driscoll: that's a possibility as well, I was just trying to think about the minimal change to the test structure, but having a built in promise would work as well.

JoeRobich: Could still define the promise in the test.

```js
const workspaceLoadedPromise = new Promise(resolve =>
{
    const subscription = activation.eventStream.subscribe(event => {
        console.log(event);
        if (event.type === EventType.WorkspaceInformationUpdated) {
            subscription.unsubscribe();
            resolve(true);
        }
});
isWorkspaceLoaded = await workspaceLoadedPromise;
```

JoeRobich: Also, I now see the polling below that waits for isWorkspaceLoaded. please disregard my suggestion.

src/omnisharp/engines/LspEngine.ts
```diff
@@ -0,0 +1,330 @@
...
+            run: {
+                command: launchInfo.MonoLaunchPath ?? launchInfo.LaunchPath,
+                args: ['-lsp'].concat(args),
+                options: {
+                    cwd,
+                },
+            },
+            debug: {
+                command: launchInfo.MonoLaunchPath ?? launchInfo.LaunchPath,
+                args: [/*"-d", */ '-lsp'].concat(args),
```
filipw, line 89: why is debug flag disabled in debug mode?

david-driscoll: I'm sure I was just testing the extension and didn't want the server to wait for a debugger to attach, I'll fix that.

david-driscoll:
updated to actually run on Mac and hopefully linux @333fred let me know

JoeRobich:
I have brought the changes from this PR up to date and pushed it as branch [feature/lsp-engine](https://github.com/OmniSharp/omnisharp-vscode/tree/feature/lsp-engine). We can continue to evolve these changes from the feature branch.

PR #3829, opened 2020-06-05
Fix for directory remove/move/rename

savpek:
VSCode file watcher has limitation where it doesn't notify any removed files if folder is removed/moved/renamed (https://github.com/microsoft/vscode/issues/90746). 

What happens is VSCode don't notify about removed file but send new one in another location in cases like directory rename. This causes duplicate documents on workspace -> cannot build because of duplicate types etc. Only way to currently fix this is restart which occurs fairly often if you refactor folders around.

Because file watcher doesn't support notifying removed files but it supports notifying about removed files i added new endpoint parameter at omnisharp-roslyn side to notify about removed folder -> omnisharp-roslyn can remove matching documents from folder at it's workspace.

PR at omnisharp-roslyn side https://github.com/OmniSharp/omnisharp-roslyn/pull/1821

Before after rename/move/delete directory operation at vscode:
image

After these fixes:
image


savpek:
Ready for review @filipw @JoeRobich @david-driscoll

PR #3837, opened 2020-06-09
Update Razor version.

NTaylorMullen:
- There isn't really any functional improvements in this insertion other than making our error telemetry Codespaces compliant

/cc @TanayParikh @ajaybhargavb @ryanbrandenburg


JoeRobich:
@NTaylorMullen Thanks!

PR #3849, opened 2020-06-13
Update Razor version to 5.0.0-preview.7.20312.9.

NTaylorMullen:
- This has a fix for the Razor `rename` issue .


NTaylorMullen:
It's our Azure blob feed which we use for fallback cases. Given the npm package is only used at restore time having a CDN isn't 100% necessary

Issue #3200, opened 2019-07-29
Minimum .NET Core SDK should be a supported version


richlander:
At present, Omnisharp [checks for a minimum .NET Core SDK version of 1.0.0](https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/coreclr-debug/util.ts#L12). .NET Core 1.0 is no longer supported and that particular servicing version hasn't been supported for multiple years.

The min-supported .NET Core version should align with the highest security patch for the oldest supported .NET Core SDK version. Right now, that would be `2.1.801`. I'm not suggesting a hard-line to updating this version on a schedule, but updating it to whatever is the correct version every time we publish a new omnisharp version.

We would probably need to [update the message](https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/coreclr-debug/util.ts#L109-L110) to make it more descriptive for the UX.

rchande:
@richlander I think what you've written is actually a feature request to notify the user that the detected .NET Core SDK is no longer supported. I don't believe we'll be able to provide a good implementation of this by adding/updating constants to our code. Rather, this should be data-driven by data we get from the CLI somehow. Thoughts?

As a footnote, I don't know why the min version is set to 1.0.0, but I'm also not sure why we would choose to fail project load/throw errors if someone was on an out of date SDK.

richlander:
You are right. Changing constants for this purpose is a bad idea. In fact, the constant can likely be removed entirely. It's not providing any obvious value.

Visual Studio does a good job of helping you keep up to date on SDKs. Part of that is keeping you secure but an equal part is that our test matrix is quite shallow between VS and .NET Core versions. That's presumably the case here as well.  I assume we no longer test omnisharp with .NET Core 1.0.0.

Let's assume we had a CLI API to give us any information we wanted. I think we'd want to do the following:

- Log that you were not using the latest SDK for your target framework, tell you which version you are are using, and what the latest is, and how to download it.
- Provide a visual way of showing you that there is a newer .NET Core SDK version available (maybe a little circle on top of the omnisharp icon, much like how VS Code does it for updates) that would open the browser when clicked.
- Prompt you to download the .NET Core SDK when you don't have one installed, either at all, or that is compatible with your current project.

gregg-miskelly:
[...]

PR #3856, opened 2020-06-16
Update minimum .NET CLI version requirement

cartermp:
Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/3200

### DIFF
--- a/src/coreclr-debug/util.ts
+++ b/src/coreclr-debug/util.ts
@@ -9,7 +9,7 @@ import * as semver from 'semver';
 import * as os from 'os';
 import { getDotnetInfo, DotnetInfo, DOTNET_MISSING_MESSAGE } from '../utils/getDotnetInfo';
 
-const MINIMUM_SUPPORTED_DOTNET_CLI: string = '1.0.0';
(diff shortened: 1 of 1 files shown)


gregg-miskelly:
@cartermp What does `semver.lt` do with the `x` in your constant? Did you test on a box with only .NET Core 2.1 installed?

gregg-miskelly:
Also, I am not sure if this change makes sense -- this check is here because way back in the day there were .NET Core versions that wouldn't work. So if it fails I believe we would entirely fail to try debugging. At this point, I think we should probably just delete line 83-88 and remove the constant from this file. 

If we wanted to implement what @richlander was really asking for, we would instead want to ship this somewhere else with some sort of warning about "hey we noticed your .NET Core is really old, you should update it" that maybe should be part of OmniSharp all up instead of this extension.

JoeRobich:
After reading the issues and seeing where this code is called, I agree with Greg. If we are going to check for a valid, up to date .NET Core SDK being installed, we wouldn't do it as part of debugger installation. Closing.

PR #3867, opened 2020-06-22
improved documentation entry for omnisharp.path

filipw:
At the moment it is not clear that `"omnisharp.path":"latest"` opts the user into the beta channel builds.


JoeRobich:
Thanks!

PR #3874, opened 2020-06-24
Add keywords in package.json

RudraSwat:
I've added some keywords in `package.json`


JoeRobich:
Thanks for the contribution but we do not want to make these changes at this time.

PR #3915, opened 2020-07-17
Make "Run/debug tests in context" position a link

RikkiGibson:
This modifies the format for the context position so that it formats as a link in VS Code. This is handy when you are reviewing the test output logs and you want to get back to the context position for some tests that you ran previously.

This imitates the formatting for positions used when the compiler reports diagnostics and in stack traces at runtime.

Before: `src/Compilers/CSharp/Test/Symbol/Symbols/Source/NullablePublicAPITests.cs 3015:15`
After: `src/Compilers/CSharp/Test/Symbol/Symbols/Source/NullablePublicAPITests.cs(3015,15)`

You can actually try just including these strings in a terminal command and you will see that the former does not parse as a link but the latter does.

/cc @333fred @JoeRobich


JoeRobich:
Thanks @RikkiGibson

PR #3931, opened 2020-07-26
Make updateBuffer use incremental updates

333fred:
Today, VSCode serializes the entire file whenever it sends a file update request. This can be very costly, particularly for files like Roslyn's NullableReferenceTypeTests.cs, which is 5+ MB. This causes large allocations in both vscode itself and in omnisharp, slowing down typing responsiveness. This commit moves the vscode extension to use the per-line changes that the omnisharp-roslyn host already supports. Editing experience in the afformentioned file is noticeably smoother. Completions are still somewhat slow and work needs to be done on that front, but this makes it possible to actually edit the file with vscode still being responsive.

### DIFF
--- a/src/features/changeForwarding.ts
+++ b/src/features/changeForwarding.ts
@@ -6,7 +6,7 @@
 import { Uri, workspace } from 'vscode';
 import { OmniSharpServer } from '../omnisharp/server';
 import * as serverUtils from '../omnisharp/utils';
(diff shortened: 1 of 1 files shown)


333fred:
For reference, a single update with NullableReferenceTypeTests.cs was about 11 MB before this change, and in a 15 second capture of omnisharp output in `trace` logging level, about 30 MB of logs were generated.

PR #3939, opened 2020-07-29
Razor 16.8 Preview 1 Insertion 3

ajaybhargavb:
More bug fixes.

~Working on manually verifying the packages.~
Verified that the packages work as expected.

cc @JoeRobich 

@NTaylorMullen @TanayParikh @ryanbrandenburg

### DIFF
--- a/package.json
+++ b/package.json
@@ -514,17 +514,17 @@
                   "items": {
                     "type": "string"
                   },
-                  "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                  "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
                   "default": []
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -514,17 +514,17 @@
                   "items": {
                     "type": "string"
                   },
-                  "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                  "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
```
ajaybhargavb, line 517: This happened automatically. Should I revert?

JoeRobich: If you could, that would help me spin up a new pre-release. There is a launch configuration `Update Package Dependencies` which I run when I update OmniSharp, but also when I want to add back the zero-length spaces.

gregg-miskelly: If you run `npm run gulp generateOptionsSchema` it should put it back. See src\tools if you need more info.

JoeRobich: Going to put that in my bag of tricks. =)

ajaybhargavb: Done

PR #3994, opened 2020-08-17
Disable Prettier if no prettier config exists

muhammadsammy:
A lot of users of [prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) use it to format files automatically on save leading to unwanted diffs, so I've added an option to disable prettier unless there is a prettier config file.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,7 @@
 {
     "editor.tabSize": 4,
     "editor.insertSpaces": true,
+    "prettier.requireConfig": true,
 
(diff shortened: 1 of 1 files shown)


JoeRobich:
> I've added an option to disable prettier unless there is a prettier config file.

@muhammadsammy Seems reasonable. Thanks for submitting.

JoeRobich:
Passing CI - https://travis-ci.org/github/OmniSharp/omnisharp-vscode/builds/719102977

PR #4076, opened 2020-09-19
Support solution filters (*.slnf)

sharwell:
Addresses https://github.com/dotnet/roslyn/pull/47531#pullrequestreview-485152426

### DIFF
--- a/package.json
+++ b/package.json
@@ -383,10 +383,12 @@
     "workspaceContains:project.json",
     "workspaceContains:*.csproj",
     "workspaceContains:*.sln",
+    "workspaceContains:*.slnf",
     "workspaceContains:*.csx",
     "workspaceContains:*.cake",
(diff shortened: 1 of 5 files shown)


Davilink:
Any update ?

ffMathy:
Any news?

JoeRobich:
@sharwell Thanks Sam! These changes merged as part of #4481

MostHated:
I am not familiar with solution filter, but came across it when vscode omnisharp was not respecting my folder excludes. Is there an example somewhere of how I might use it to filter out a particular folder?

JoeRobich:
@MostHated Here are the docs https://docs.microsoft.com/en-us/visualstudio/msbuild/solution-filters?view=vs-2019. Solution filters are for filtering out projects that would normally be loaded as part of a Solution file (.sln).

PR #4091, opened 2020-09-28
Update Debugger Packages and add Linux ARM/ARM64 Packages

WardenGnaw:
This PR updates to the newest version of the debugger.

Along with this, this PR adds linux ARM and ARM64 packages to package.json.
It also updates platform to check for Linux architecture that have `armv`
and `aarch` in the `uname -m` output then set them as arm or arm64.

Also adds a couple of TODO's once the C# Omnisharp extension is ready to
run on the VS Code ARM/ARM64 builds.

See https://github.com/microsoft/vscode/pull/106289

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@
   * Fully qualify C# Type light bulb ([dotnet/aspnetcore-tooling#24778](https://github.com/dotnet/aspnetcore/issues/24778))
   * Added support for engine logging on .NET process for Blazor WASM apps ([OmniSharp/omnisharp-vscode#4070](https://github.com/OmniSharp/omnisharp-vscode/issues/4070))
   * Fixed bug in clean-up of Blazor WASM debugging session ([OmniSharp/omnisharp-vscode#4056](https://github.com/OmniSharp/omnisharp-vscode/issues/4056))
+  * Debugger Features:
+    * Add support for Function Breakpoints ([#295](https://github.com/OmniSharp/omnisharp-vscode/issues/295))
+  * Debugger Fixes:
+    * [Debugger licensing errors are not reported to the UI ([#3759](https://github.com/OmniSharp/omnisharp-vscode/issues/3759))
+    * [Error processing 'variables' request. Unknown Error: 0x8000211d ([#3926](https://github.com/OmniSharp/omnisharp-vscode/issues/3926))
+    * [Method with a function pointer local breaks variables view and debug console ([#4052](https://github.com/OmniSharp/omnisharp-vscode/issues/4052))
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -215,6 +215,7 @@ function isSupportedPlatform(platform: PlatformInformation): boolean {
     }
 
     if (platform.isLinux()) {
+        // TODO: Add 'arm' and 'arm64'
```
gregg-miskelly, line 218: Returning false here means that we cannot even debug. Is that what we want to do? Or do we want to return true here, but be smart enough to not start the language server and output a big comment to the log indicating why we are doing that?

src/coreclr-debug/activate.ts
```diff
@@ -40,12 +40,14 @@ async function checkForInvalidArchitecture(platformInformation: PlatformInformat
             eventStream.post(new DebuggerPrerequisiteFailure("[ERROR] The debugger cannot be installed. The debugger requires macOS 10.12 (Sierra) or newer."));
             return true;
         }
-        else if (platformInformation.architecture !== "x86_64") {
-            if (platformInformation.isWindows() && platformInformation.architecture === "x86") {
-                eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: x86 Windows is not currently supported by the .NET Core debugger. Debugging will not be available.`));
-            } else {
-                eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: Processor architecture '${platformInformation.architecture}' is not currently supported by the .NET Core debugger. Debugging will not be available.`));
-            }
+        else if (platformInformation.isWindows() && platformInformation.architecture === "x86") {
```
gregg-miskelly, line 43: >platformInformation.architecture === "x86" [](start = 52, length = 42)

Instead how about `!== "x86_64"` and change the 'x86' in the error message to be replaced with `${platformInformation.architecture}`.

src/main.ts
WardenGnaw: Idealy, we would activate when the whole extension is ready to go so users get the full extension feature in VS Code.

@JoeRobich Do you know if the language service would be ready for Linux ARM/ARM64 soon or should only debugger activate for now?

WardenGnaw:
@gregg-miskelly Since we are enabling local debugging for ARM/ARM64, do we want to remove this remote extension error hint?

https://github.com/OmniSharp/omnisharp-vscode/blob/1ab9c4ca09291033ff71716330df9e3032aec79f/src/main.ts#L132-L142

JoeRobich (approved):

gregg-miskelly (approved):

PR #4115, opened 2020-10-15
Update to OmniSharp 1.37.3 and revert useGlobalMono meaning.

JoeRobich:
Due to a performance regression in OmniSharp 1.37.2, we need to update and publish an updated C# extension build. This PR also reverts the meaning of useGlobalMono to "always" since Mono 6.12.0 is stable and shipping with MSBuild 16.8.

CC: @nohwnd, @vzarytovskii

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -139,8 +139,8 @@
                 "updatePackageDependencies"
             ],
             "env": {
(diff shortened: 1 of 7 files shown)


vzarytovskii (approved):

filipw (approved):
LGTM

PR #4141, opened 2020-10-23
Do not send file changed events for .cs files

333fred:
Similar to the rename and codeAction change, this can cause incremental updates to break if the file is updated on disk (not by a vscode edit).

### DIFF
--- a/src/features/changeForwarding.ts
+++ b/src/features/changeForwarding.ts
@@ -50,9 +50,12 @@ function forwardFileChanges(server: OmniSharpServer): IDisposable {
             }
 
             if (changeType === FileChangeType.Change && uri.fsPath.endsWith(".cs")) {
-                // The server watches for file system events from .cs files, if we send
-                // a file changed event as well it will cause a double-apply and potentially
-                // invalidate the file state
+                // When a file changes on disk a FileSystemEvent is generated as well as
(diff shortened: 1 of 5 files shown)


src/features/changeForwarding.ts
```diff
@@ -49,6 +49,13 @@ function forwardFileChanges(server: OmniSharpServer): IDisposable {
                 return;
             }
 
+            if (changeType === FileChangeType.Change && uri.fsPath.endsWith(".cs")) {
+                // The server watches for file system events from .cs files, if we send
+                // a file changed event as well it will cause a double-apply and potentially
+                // invalidate the file state
```
JoeRobich, line 55: How about the following?

```suggestion
                // When a file changes on disk a FileSystemEvent is generated as well as
                // a DidChangeTextDocumentEvent. The OmniSharp server listens for Change events
                // for ".cs" files and reloads their text from disk. This creates a situation where the server
                // may have updated the document to reflect disk and also recieves a set of TextChanges
                // to apply to the document. In order to avoid that situation, we will not send Change events
                // for ".cs" files and instead allow them to be updated via the DidChangeTextDocumentEvent.
```

JoeRobich (approved):
LGTM.

PR #4170, opened 2020-11-02
Do not call updateBuffer if there are no changes.

JoeRobich:
As mentioned in this issue, https://github.com/Microsoft/vscode/issues/50344, the `onDidChangeTextDocument` callback is invoked when a document has changed as well as when its state has changed between 'dirty' and 'clean'.

When it is invoked as a state change there will be no changes and there is no need to call updateBuffer.


filipw (approved):
it's only a hack if we don't call it a workaround 🙂

mholo65 (approved):

PR #4230, opened 2020-11-23
only suppress file changed notifications for C# files

filipw:
This is a follow up to https://github.com/OmniSharp/omnisharp-vscode/pull/4178

We should only suppress events for C# code files, not for files such as `.csproj`, `.editorconfig` or assets files, as it causes OmniSharp to go stale after changes are done in those files.

### DIFF
--- a/src/features/changeForwarding.ts
+++ b/src/features/changeForwarding.ts
@@ -55,7 +55,7 @@ function forwardFileChanges(server: OmniSharpServer): IDisposable {
             }
 
             if (changeType === FileChangeType.Change) {
-                const docs = workspace.textDocuments.filter(doc => doc.uri.fsPath === uri.fsPath);
+                const docs = workspace.textDocuments.filter(doc => doc.uri.fsPath === uri.fsPath && isCSharpCodeFile(doc.uri));
                 if (Array.isArray(docs) && docs.some(doc => !doc.isClosed)) {
                     // When a file changes on disk a FileSystemEvent is generated as well as a
                     // DidChangeTextDocumentEvent.The ordering of these is:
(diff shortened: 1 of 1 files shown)


src/features/changeForwarding.ts
```diff
@@ -82,6 +84,11 @@ function forwardFileChanges(server: OmniSharpServer): IDisposable {
         };
     }
 
+    function isCSharpCodeFile(uri: Uri) : Boolean {
+        let normalized = uri.path.toLocaleLowerCase();
```
333fred, line 88: Could be const

src/features/changeForwarding.ts
```diff
@@ -69,6 +69,8 @@ function forwardFileChanges(server: OmniSharpServer): IDisposable {
                     // being that the file is now in an inconsistent state.
                     // If the document is closed, however, it will no longer be synchronized, so the text change will
                     // not be triggered and we should tell the server to reread from the disk.
+                    // This applies to C# code files only, not other files significant for OmniSharp
+                    // e.g. csproj or editorconfig files
```
333fred, line 73: Do we not get incremental updates for these files as well that we then send to the server?

filipw: at the moment there are quite a few things that need to fire as callbacks in response to file changed events coming via the files changed endpoint (https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.MSBuild/ProjectManager.cs#L379-L477). 
At the moment none of them are arriving

333fred (approved):
LGTM, one const suggestion.

JoeRobich (approved):

Issue #3403, opened 2019-11-13
Generating build/debug assets should not overwrite existing launch and tasks


bwateratmsft:
Currently when generating build/debug assets for .NET Core projects, anything in launch.json and tasks.json will be overwritten. This can cause data loss and user confusion. Instead, the relevant launch and tasks should be added if missing, and if already present, prompted for overwriting.

The vscode-docker extension does this by checking the name of the launch/task. If another launch/task of the same name we're about to create exists, we will prompt if they want to overwrite.

bwateratmsft:
Here's an example scenario for why this is a problem.

1. User creates a .NET Core app, and opens in VSCode
1. User hits F5 without clicking the "Add required assets..." to set up for .NET Core
1. In the "Select environment" pick, user selects the Docker launch configuration, which depends on .NET Core build
1. The Docker extension will abort the debugging attempt and indicate that the user must add Docker support / debug assets. Adding these cannot be done automatically as it requires several inputs.
1. The user adds these assets for Docker debugging. The user hits F5 on the Docker launch profile.
    1. For whatever reason, VSCode does not warn or error if a task's `dependsOn` doesn't exist (in this case, the .NET Core build task).
1. Debugging fails, as the project was never built.

Eventually, suppose they figure out they need to add the .NET Core assets. Upon doing so, the Docker launch and tasks are now overwritten. They can be added back via "Docker: Initialize for debugging" command, but this causes needless confusion and possible data loss.

ucheNkadiCode:
The Docker Tools Team would like to bump this thread and request a fix for the "[.NET Generate Assets for Build and Debug](https://github.com/OmniSharp/omnisharp-vscode/blob/49792876183003d77d5ddf84380fce064bcf5b86/src/assets.ts#L694)" command.

For our .NET customers (our 2nd highest user category), this bug negatively affects our ability to scaffold single dockerfiles and compose files for .NET projects. Also if a user has already created a docker launch configuration, this bug undoes the users work because it wipes their launch configurations when the command is run rather than appending to the file.

[...]

PR #4241, opened 2020-11-26
Merge launchJson and tasksJson when generating assets

nohwnd:
Related #3403


filipw:
ooh that is a pretty bad bug, yes 😄

nohwnd:
@filipw  At the moment I generate the file as we did before when the file does not exist. When it exists I remove the items (tasks / configurations) with the same name as our items and append the new items. 

The prompt I changed in my first commit to be Cancel / No / Yes, turned out to be very confusing. Because we are now trying hard to not overwrite anything, can I just get rid of the prompt, and if there is existing task just rename it to "* - backup"?

This way if someone has their own task with generic name (like build) they won't lose it. Instead it will become "build - Backup"? 

What would be the best way to go about this? I can check for every item name but it's hard to communicate what exactly would be overwritten in that tiny pop up.

ucheNkadiCode:
Adding @bwateratmsft, since he's the Dev for our VS Code Docker Tools and our product is affected by this error.

src/assets.ts
```diff
@@ -541,28 +543,29 @@ export async function addTasksJsonIfNecessary(generator: AssetGenerator, operati
...
+                }
 
-            if (existingTaskConfigs) {
-                tasksJson['tasks'] = tasksJson['tasks'].concat(existingTaskConfigs);
-            }
+                resolve();
+            });
+        }
+        else {
+            // when tasks.json exists just update the tasks node
```
gregg-miskelly, line 561: >// when tasks.json exists just update the tasks node [](start = 12, length = 52)

Shouldn't we check if there are already tasks with the names we expect, and only add whatever tasks are missing?

src/assets.ts
```diff
@@ -576,31 +579,39 @@ async function addLaunchJsonIfNecessary(generator: AssetGenerator, operations: A
...
+                    return reject(err);
+                }
 
-            resolve();
-        });
+                resolve();
+            });
+        } 
+        else {
+            // when launch.json exists append our configurations
```
gregg-miskelly, line 607: >// when launch.json exists append our configurations [](start = 12, length = 52)

I feel like we shouldn't go through the VS Code APIs, as that isn't going to work with our comments.

src/assets.ts
```diff
@@ -716,22 +730,29 @@ export async function generateAssets(server: OmniSharpServer, selectedIndex?: nu
...
-            if (!doGenerateAssets) {
+            let doOverwriteAssets = await shouldOverwriteAssets(generator);
+            if (doOverwriteAssets === 'Cancel') {
                 return; // user cancelled
             }
 
+            let overwrite = doOverwriteAssets === 'Yes';
+
             const operations: AssetOperations = {
                 addLaunchJson: generator.hasExecutableProjects(),
```
gregg-miskelly, line 741: >addLaunchJson [](start = 16, length = 13)

Should this be setting `updateLaunchJson` also? And should `addLaunchJson` be dependent on `overwrite`?

src/assets.ts
```diff
@@ -685,26 +696,29 @@ async function deleteAssets(generator: AssetGenerator) {
...
-                vscode.window.showWarningMessage('Replace existing build and debug assets?', cancelItem, yesItem)
+                vscode.window.showWarningMessage('Replace existing build and debug assets?', cancelItem, noItem, yesItem)
                     .then(selection => {
                         if (selection === yesItem) {
-                            resolve(true);
+                            resolve('Yes');
+                        }
+                        else if (selection === noItem) { 
+                            resolve('No');
                         }
```
gregg-miskelly, line 713: What is your design here? Since you aren't deleting the files anymore (and it doesn't seem like we should), I am not following the goal. Potentially I could imagine this have the behavior of only adding missing sections, but it doesn't seem to be doing this and I am not sure that actually makes sense.

src/assets.ts
nohwnd: What if user has comments in their file? Then we would lose their comments.

src/assets.ts
nohwnd: In my comment above: 

> The prompt I changed in my first commit to be Cancel / No / Yes, turned out to be very confusing. Because we are now trying hard to not overwrite anything, can I just get rid of the prompt, and if there is existing task just rename it to "* - backup"?

This prompt is a leftover from the first concept, right now I was just trying to figure out how to go about this task. If we should append the "- Backup" or prompt to overwrite. In the end think this prompt needs to revert to what it was. 

This would be my current goal: 

Check if there are any configurations or tasks that we would overwrite and get their names. If there are any, prompt with Cancel | Yes.

On Cancel do nothing. 
On Yes, patch the files.

If there are none, patch the files, no prompt.

How exactly the files are patched is a different thing to figure out.

src/assets.ts
nohwnd: Okay updating would lose all comments, ours and theirs. I need a different way.

nohwnd: If you have ideas how to do that please share, I saw that jsonc-parser has quite a lot of capabilities, so maybe it could be done via that, but if there is a known easy way I am all ears.

JoeRobich: Looks like jsonc-parser has utility functions named `modify` and `applyEdits`. You would use `modify` to generate an array of Edits that you could then pass to `applyEdits` in order to get an updated json string to save.

nohwnd: Thanks. I reached out to the maintainer of that package, so I will wait till he gets back to me, maybe there is some example that I can follow.

src/assets.ts
gregg-miskelly: Your suggestion of a backup file certainly sounds reasonable. The other option that occurs to me is to change the prompt to be more clear on what we are overwriting -- "Replace existing build tasks and .NET debug launch configurations?"

bwateratmsft:
Speaking from the perspective of the Docker extension, we depend on the build tasks, but we're reluctant to call the "Generate assets for build and debug..." command automatically because of its behavior of erasing anything existing for tasks.json and launch.json. If it only replaced things of the same name, I think I'd be OK with calling it automatically. A backup file maybe works for us, I'd have to think about it and discuss with @ucheNkadiCode.

nohwnd:
I have a work in progress solution that only overwrites our tasks, and keeps all the comments, but I was on a vacation for a month.

ucheNkadiCode:
> I have a work in progress solution that only overwrites our tasks, and keeps all the comments, but I was on a vacation for a month.

Great job @nohwnd! Thanks for the effort 👍🏽 

I have just a few questions:

- Does the new logic prompt that you are overwriting the existing config as @gregg-miskelly mentioned? 
- If for some reason the user's current .NET debug launch config was named slightly differently, I assume the scaffolding would just toss the config created by "Generate Debug Assets" in without any overwriting correct?
- Will there still be a backup file? I vote that one isn't necessary since the solution is only overwriting one task and keeping the rest. I also believe it's common for most scaffolding commands to overwrite files rather than create a `file.copy.ext`. @bwateratmsft , do you have an opinion here?

bwateratmsft:
I think it's fine to not use a backup file as long as only the C# tasks are overwritten. In the case of the Docker extension, we overwrite our own tasks (with a prompt first), if the name matches. If the name doesn't match, we just add them in, since only the name has to be unique. So, the behavior we are currently doing for Docker sounds exactly the same as what @nohwnd is describing for C#. I like it! (And I'm definitely not biased or anything :smile:)

ucheNkadiCode:
@nohwnd Thanks for all of your work! Let us know when the fix is finalized so we can test it out and make those changes to our Tools 👍🏽

ucheNkadiCode:
Hey @gregg-miskelly, @JoeRobich, and @nohwnd, 

When can we expect this pull request to be reviewed and submitted? The Docker Tools team is patiently awaiting this addition to the codebase so we can take care of one of our top errors mentioned in https://github.com/OmniSharp/omnisharp-vscode/issues/3403. 

Thank you!

nohwnd:
Implemented in a way that keeps any comments that were present in the file already and puts all our comments within the configurations (or tasks) that we add. The comments are written as JSON object properties named OS-COMMENT and then replaced with comment syntax. This approach was recommended by the maintainer of the jsonc-parse library that is used. 

The new behavior is: 
- never delete tasks.json or launch.json
- create the files if they don't exist
- amend the files when they exist by: 
    - replacing tasks with the same name
    - appending tasks if they don't exist
- user is prompted only if there will be overwrites

src/assets.ts
nohwnd: With the current approach we don't lose ours nor the user comments. (Unless they place their comments into the tasks that we overwrite, but they will be prompted to confirm the overwrite).

src/assets.ts
nohwnd: This is outdated, I ended up doing pretty much the same thing as before. It checks if we overwrite any task, and prompts for confirmation. I experimented with printing the names of tasks in the prompt but it was confusing and noisy.

src/assets.ts
nohwnd: Outdated.

src/assets.ts
```diff
@@ -663,32 +701,42 @@ export async function addAssetsIfNecessary(server: OmniSharpServer): Promise<Add
...
+        }
 
-async function deleteAssets(generator: AssetGenerator) {
-    return Promise.all([
-        util.deleteIfExists(generator.launchJsonPath),
-        util.deleteIfExists(generator.tasksJsonPath)
-    ]);
+        if(fs.pathExistsSync(generator.launchJsonPath)) { 
+            const content = fs.readFileSync(generator.launchJsonPath).toString();
+            let configurationNames = [
```
nohwnd, line 719: Not very happy about this list. I tried sharing the names in constants but it was not much better.

JoeRobich:
@nohwnd Could we add tests the merge scenarios in https://github.com/OmniSharp/omnisharp-vscode/blob/master/test/featureTests/assets.test.ts?

src/assets.ts
```diff
@@ -377,22 +391,25 @@ export function createFallbackLaunchConfiguration(): vscode.DebugConfiguration {
...
 // AttachConfiguration
 export function createAttachConfiguration(): string {
-    return `
-{
-    "name": ".NET Core Attach",
-    "type": "coreclr",
-    "request": "attach",
-    "processId": "\${command:pickProcess}"
-}`;
+    const o = {
```
JoeRobich, line 405: Maybe rename these `o` variables to `configuration`.

src/assets.ts
```diff
@@ -581,37 +606,40 @@ async function addLaunchJsonIfNecessary(generator: AssetGenerator, operations: A
 
         const programLaunchType = generator.computeProgramLaunchType();
         const launchJsonConfigurations: string = generator.createLaunchJsonConfigurations(programLaunchType);
+        const formattingOptions: FormattingOptions = {
+            insertSpaces: true,
+            tabSize: 4,
+            eol: '\n'
```
JoeRobich, line 611: Is it possible to pull these formatting options from VS Code?

src/assets.ts
```diff
@@ -730,17 +764,16 @@ export async function generateAssets(server: OmniSharpServer, selectedIndex?: nu
...
 
-            let overwrite = doOverwriteAssets === 'Yes';
-
             const operations: AssetOperations = {
                 addLaunchJson: generator.hasExecutableProjects(),
-                addTasksJson: overwrite,
-                updateTasksJson: !overwrite,
+                // TODO: this distinction does make sense anymore, we either Update, or do nothing.
+                addTasksJson: true,
+                updateTasksJson: false,
```
JoeRobich: Since these are constant now, lets go ahead and remove them.

JoeRobich (commented):
Looking good. Once there is some test coverage, I think it'll be good to merge.

nohwnd:
@JoeRobich the jsonc package needs to be updated to version 3, because the current version does not support merging files. I just changed the version in the file, but there are multiple other automatic changes, that remove zero-width spaces. Are you okay with removing those zero width spaces?

src/assets.ts
nohwnd: I have removed the comment, but kept the values as they were before. There is only addTasksJson as before.

src/assets.ts
nohwnd: I was considering that as well, but the spacing was hardcoded before so I did not go further. I will try it.

src/assets.ts
nohwnd: done

src/assets.ts
nohwnd: done

JoeRobich:
@nohwnd If you run the "Update package dependencies" task it'll fix up the package.json.

JoeRobich (approved):

gregg-miskelly:
@nohwnd I believe your change to createFallbackLaunchConfiguration is going to break the configuration provider support. That is not supposed to return a string. It is used [here](https://github.com/OmniSharp/omnisharp-vscode/blob/c09bf3f65319cc98fbf37f0070b9786ca4b16357/src/configurationProvider.ts#L113).

ucheNkadiCode:
@gregg-miskelly, do you foresee this string issue being a blocker to this merge? As in, we likely cannot rely on the generating of assets until the issue is resolved?

gregg-miskelly:
This PR was already merged. But the fix for the problem is trivial, so we should fix it before we put out the next release.

nohwnd:
I now see how that happened, the other method returns `any` which types the array to be `any[]` and that silently casts to the final type. So that is why I did not get any type error. Fixed in the related PR and tested.

ucheNkadiCode:
Thanks so much @nohwnd! Our team appreciates this fix. @bwateratmsft I'll follow up with you about using this new behavior to mitigate the .NET debug assets issue our team was having. We can absolutely add it to 1.11 for a down the road P2.

Issue #8, opened 2016-01-20
Auto generate triple slash comments with parameters


jrieken:
_From @EmYeuCoGiao on December 25, 2015 3:43_

I use VSCode for unity3d dev, i type /// but this not auto gennerate
<code>
///<summary>

///
///</summary>

///<param name=""></param>
</code>
![capture](https://cloud.githubusercontent.com/assets/14220005/11999614/a40d4bae-aaf4-11e5-96ce-064b8b02bb7e.PNG)

_Copied from original issue: Microsoft/vscode#1644_

jrieken:
_From @bpasero on December 25, 2015 10:17_

This is currently not supported as far as I know.

jrieken:
_From @EmYeuCoGiao on December 25, 2015 10:26_

Thanks you for reply :+1:

DustinCampbell:
This will require a new Roslyn API.

dmccaffery:
There is a third-party extension that (mostly) provides this functionality:

https://marketplace.visualstudio.com/items?itemName=k--kato.docomment

What is required of roslyn to achieve this? The fact that it does not support navigating comments within the document?

DustinCampbell:
I need to expose a new API from Roslyn so that the same feature is not simply re-implemented inside of OmniSharp. Remember, this feature already exists in Roslyn today, and it's a little tricky to get right. The feature is not just generating XML doc comments on ///, but also inserting /// on enter, handling indents properly, getting undo behavior to be correct, not interfering with XML doc comment completion lists, etc.

> What is required of roslyn to achieve this? The fact that it does support navigating comments within the document".

I'm not sure what you mean by this. There's nothing within Roslyn that disallows writing such a thing. Also, it's not clear to me why you would think navigating comments would be related to XML doc comment generation. Could you be a bit more precise?

As I mentioned, the feature already exists in Roslyn. I need to expose that feature as an API so that it can be used in OmniSharp.

dmccaffery:
@DustinCampbell What I meant was, there was an issue sometime ago within Roslyn (probably specifically as it relates to analyzers) whereby navigating the document tree would not expose comment blocks -- it would skip over them. I assumed something similar might be in play here. At any rate, I am no expert when it comes to the roslyn APIs, so I will definitely take your word for it. :)

vanillajonathan:
It would be useful even if it didn't generate the `<param>` tags. Just inserting a blank template with a  `<summary>` tag would be useful.

[...]

PR #4261, opened 2020-12-04
Support auto doc comment generation

333fred:
Adds '\n' and '/' to the on-type format list, enabling vscode support for the newly-added omnisharp feature of auto-generating documentation comments. Closes https://github.com/OmniSharp/omnisharp-vscode/issues/8.

### DIFF
--- a/test/integrationTests/documentationCommentAutoFormatting.integration.test.ts
+++ b/test/integrationTests/documentationCommentAutoFormatting.integration.test.ts
@@ -1,0 +1,61 @@
+import { expect, should } from 'chai';
+import * as vscode from 'vscode';
+import * as path from 'path';
+import { isRazorWorkspace } from './integrationHelpers';
+import testAssetWorkspace from './testAssets/testAssetWorkspace';
+import { skip } from 'rxjs/operators';
+
+const onTypeFormatProviderCommand = 'vscode.executeFormatOnTypeProvider';
+
+function normalizeNewlines(original: string): string {
+    while (original.indexOf('\r\n') != -1) {
+        original = original.replace('\r\n', '\n');
+    }
+
+    return original;
+}
+
+suite(`Documentation Comment Auto Formatting: ${testAssetWorkspace.description}`, function () {
+    let fileUri: vscode.Uri;
+
+    suiteSetup(async function () {
+        should();
+
+        if (isRazorWorkspace(vscode.workspace)) {
+            // The format-on-type provider does not run for razor files.
+            this.skip();
+        }
+
+        const projectDirectory = testAssetWorkspace.projects[0].projectDirectoryPath;
+        const filePath = path.join(projectDirectory, 'DocComments.cs');
+        fileUri = vscode.Uri.file(filePath);
+
+        await vscode.commands.executeCommand('vscode.open', fileUri);
+    });
+
+    suiteTeardown(async () => {
+        await testAssetWorkspace.cleanupWorkspace();
+    });
+
+    test('Triple slash inserts doc comment snippet', async () => {
+        const commentPosition = new vscode.Position(2, 7);
+        const formatEdits = <vscode.TextEdit[]>(await vscode.commands.executeCommand(onTypeFormatProviderCommand, fileUri, commentPosition, '/'));
+        expect(formatEdits).ofSize(1);
+        expect(normalizeNewlines(formatEdits[0].newText)).eq(" <summary>\n    /// \n    /// </summary>\n    /// <param name=\"param1\"></param>\n    /// <param name=\"param2\"></param>\n    /// <returns></returns>");
+        expect(formatEdits[0].range.start.line).eq(commentPosition.line);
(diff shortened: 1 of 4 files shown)


333fred:
@NTaylorMullen, I think the razor plugin will have to update whatever on-type formatting you have to also call on `/` and `\n`, since the C# extension itself does not get called back on razor files.

@kasecato, this exposes the support I recently added to omnisharp that calls the roslyn functionality for this. I expect that either you or we will get issues about duplicate comments being generated: this generation will only work if the user has on-type formatting turned on. I'm not sure what mechanism you're using, so it's possible that you will still have some functionality this does not (in addition to positioning the cursor in a better location, which this cannot do as it's using the on-type formatting mechanism which cannot control the cursor). But I wanted to give you a heads up before it was merged.

test/integrationTests/documentationCommentAutoFormatting.integration.test.ts
```diff
@@ -0,0 +1,61 @@
+import { expect, should } from 'chai';
+import * as vscode from 'vscode';
+import * as path from 'path';
+import { isRazorWorkspace } from './integrationHelpers';
+import testAssetWorkspace from './testAssets/testAssetWorkspace';
+import { skip } from 'rxjs/operators';
+
+const onTypeFormatProviderCommand = 'vscode.executeFormatOnTypeProvider';
+
+function normalizeNewlines(original: string): string {
```
333fred, line 10: node.js does not have `replaceAll`, unfortunately...

kasecato:
@333fred This is a feature to output doccoment from AST, right? My extension is completely regular expression, so it was getting harder and harder to maintain... I've been wanting to get the AST from the roslyn for some time. I'll announce the removal of my extension if necessary.

333fred:
> @333fred This is a feature to output doccoment from AST, right? My extension is completely regular expression, so it was getting harder and harder to maintain... I've been wanting to get the AST from the roslyn for some time. I'll announce the removal of my extension if necessary.

Yep. It's using the same functionality that inserts the doc comments in VS itself.

NTaylorMullen:
> @NTaylorMullen, I think the razor plugin will have to update whatever on-type formatting you have to also call on `/` and `\n`, since the C# extension itself does not get called back on razor files.

We don't currently support OnTypeFormatting in Razor's VSCode implementation so I'd imagine things would continue to work as is. Out of curiosity, how do you all control the cursor when generating doc comments? We've made it work in VS via a custom protocol message; however, that doesn't exist in VSCode AFAIK.

JoeRobich (approved):

333fred:
> Out of curiosity, how do you all control the cursor when generating doc comments? We've made it work in VS via a custom protocol message; however, that doesn't exist in VSCode AFAIK.

That's the secret: we do not. It does mean that the cursor will be placed at the end of the generated comment, but I think that's a worthwhile tradeoff to not having the generation at all.

NTaylorMullen:
> That's the secret: we do not. It does mean that the cursor will be placed at the end of the generated comment, but I think that's a worthwhile tradeoff to not having the generation at all.

Gotcha. Ya that's unfortunate in terms of parity but agree that it's better then nothing at all

333fred:
@JoeRobich do you have any idea what is going on with travis here? I can't actually see any errors?

JoeRobich:
@333fred I hate how the gulp command swallows errors. You can get the error locally by running `tsc -p .`.

```console
~/Source/omnisharp-vscode [auto-doc-gen +0 ~2 -0 !]> tsc -p .
test/integrationTests/documentationCommentAutoFormatting.integration.test.ts:6:1 - error TS6133: 'skip' is declared but its value is never read.

6 import { skip } from 'rxjs/operators';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

333fred:
Alright, now I believe this is just waiting for a new omnisharp-roslyn release.

vchirikov:
It doesn't work on https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.23.8
omnisharp-roslyn: 1.37.6-beta.3
p.s. I have snippets that start with `///`, maybe what's why. Can you try to check with this scenario?
```
"/// <inheritdoc>": {
	"scope": "csharp",
	"prefix": "//i",
	"body": [
		"/// <inheritdoc />",
	],
	"description": "Adds '<inheritdoc />'"
},
```

333fred:
@vchirikov do you have on-type formatting turned on? This support comes via that mechanism.

vchirikov:
@333fred 
No, I have `"editor.formatOnType": false,`. It's not obliviously that this feature included in `formatOnType`, maybe better to change description in changelog at least.
So, I have to use `formatOnType` or use `k--kato.docomment`, isn't it? :(

333fred:
> So, I have to use formatOnType or use k--kato.docomment, isn't it?

Correct.

PR #4300, opened 2020-12-16
Update to use zero based indexes

david-driscoll:
I'm moving forward with the lsp changes, but I figured this should land before that.

OmniSharp has a flag that tells the serializer to serialize with or without zero based indexes.  Internally OmniSharp works purely on zero-based indexes because it makes interacting with Roslyn easier.  However for backward compatibility externally we surface one-based indexes, the `-z` flag tells OmniSharp to return zero-based indexes instead of one based.

### DIFF
--- a/src/features/changeForwarding.ts
+++ b/src/features/changeForwarding.ts
@@ -32,10 +32,10 @@ function forwardDocumentChanges(server: OmniSharpServer): IDisposable {
             const range = change.range;
             return {
                 NewText: change.text,
-                StartLine: range.start.line + 1,
-                StartColumn: range.start.character + 1,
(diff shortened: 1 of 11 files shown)


JoeRobich:
@NTaylorMullen @ryanbrandenburg Would this break the location mapping performed by the Razor extension?

david-driscoll:
Looking at the code it shouldn't all the middleware execution is done after everything is converted to vscode types, which are zero-based already.

JoeRobich (approved):

333fred (commented):
Sorry I never ended up looking at this, I got distracted by a couple video games... LGTM though.

PR #4302, opened 2020-12-18
Add option to organize imports during document formatting.

JoeRobich:
This flag was added to OmniSharp a while back but we never introduced a VS Code setting.


filipw (approved):
yes 👏

Issue #4292, opened 2020-12-12
Intellisense not working


SHARAD2503:
Issue Type: <b>Bug</b>

Intellisense not working

Extension version: 1.23.7
VS Code version: Code 1.52.0 (940b5f4bb5fa47866a54529ed759d95d09ee80be, 2020-12-10T22:45:11.850Z)
OS version: Windows_NT x64 10.0.18363

JoeRobich:
A lot has changed with O# project loading since this issue was opened. If you are still having problems, please open a new issue at https://github.com/omnisharp/omnisharp-roslyn/issues

Issue #4289, opened 2020-12-11
2020.12.12最新版本：智能感知(IntelliSense)简直跟废物一样的慢(F**king slow) 


sakurawarskula:
每次写代码的时候，如果项目非常大，代码在**【实时检查是否有错误的过程】**中**【会导致智能感知失效】**，结果就是大项目的体验极其恶心！！因为智能感知简直慢的让人抓狂！平均每段代码智能感知至少要卡3秒才出来！！几乎每打任何代码都必须等几秒钟

这个问题已经出现了好几个月了至今没修复究竟是为什么？

My English is really bad,so please understand

If the project is very big(more than 1000 scripts)(Unity C#),
When you input code,vscode will Check the code for errors in real time
**But in the process, the Intellisense will fail !!!**

It mean you must wait the check code finish! the Intellisense It's going to show up!!
**This causes a delay of a few seconds in entering any code before intellisense is displayed**

Lucas-M-G:
I have the same issue too. 
Try to downgrade to Version 1.23.1. This was a workaround for me.

sakurawarskula:
> I have the same issue too.
> Try to downgrade to Version 1.23.1. This was a workaround for me.

Unbelieveable !It's working!! thanks!

Lucas-M-G:
Do you have a inner preject reference in your .csproj-File?
Like:
```xml
<ItemGroup>
    <ProjectReference Include="/lib/SomeProject/SomeProject.csproj" />
</ItemGroup>
```

derwaldgeist:
Also, the startup times of OmniSharp are painstakingly slow now. In my Unity installation, it takes minutes (!) until the gazillions of .csproj files are being parsed. A huge degradation of developer performance.

Issue #4080, opened 2020-09-21
Intellisense always Loading...


jkleinecke:
Issue Type: <b>Bug</b>

Intellisense tries to work, but never moves past the Loading... text.  Omnisharp logs do not indicate any errors or on-going activitiy.

Extension version: 1.23.2
VS Code version: Code 1.49.1 (58bb7b2331731bf72587010e943852e13e6fd3cf, 2020-09-16T23:21:17.533Z)
OS version: Darwin x64 19.6.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz (12 x 2200)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|3, 3, 2|
|Memory (System)|32.00GB (3.60GB free)|
|Process Argv|. --crash-reporter-id 21773cd3-da3e-428c-8301-33f6e312bc2f|
|Screen Reader|no|
|VM|0%|
</details>

wendellestradairely:
I've been experiencing this for many months now. Intellisense is very slow compared to Visual Studio IDE.

JoeRobich:
@jkleinecke Sorry you are seeing issues. Can you please share your OmniSharp log as well as the output from running `dotnet --info`?

jkleinecke:
It's worth noting that it's working fine now.  I think it has to do with
opening up a folder that contains many solution files and switching between
the projects.  I've since run dotnet restore on all my projects and I'm no
longer experiencing the issue.

Starting OmniSharp server at 9/22/2020, 1:07:59 PM
    Target:
/Users/james/projects/slb-erc/core-service-layer/src/crew-service/CrewService.sln

OmniSharp server started.
    Path:
/Users/james/.vscode/extensions/ms-dotnettools.csharp-1.23.2/.omnisharp/1.37.1/run
    PID: 32511

[info]: OmniSharp.Stdio.Host
[...]

Issue #4047, opened 2020-09-08
Intellisense not working.


HaykShirinyan:
Issue Type: <b>Bug</b>

It works at first. But after typing a few more lines, the intelisense stops working completely. 

OmniSharp log:

[2020-09-08 00:00:00.901] [renderer1] [error] Problem invoking 'GetCodeActions' on OmniSharp server: Error: Pending request cancelled: /v2/getcodeactions
[2020-09-08 00:00:05.897] [renderer1] [error] Problem invoking 'GetCodeActions' on OmniSharp server: Error: Pending request cancelled: /v2/getcodeactions

Extension version: 1.23.2
VS Code version: Code 1.48.2 (a0479759d6e9ea56afa657e454193f72aef85bd0, 2020-08-25T10:13:11.295Z)
OS version: Windows_NT x64 10.0.18363

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-4210M CPU @ 2.60GHz (4 x 2594)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>protected_video_decode: unavailable_off<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>viz_display_compositor: enabled_on<br>viz_hit_test_surface_layer: disabled_off_ok<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.90GB (8.38GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details>

meysamzare:
I have the same issue too.........

jmarolf:
can you provide a project that this reproduces on ?

julian-vp:
Same issue. Fresh install of VS Code, fresh install of Omnisharp.  Syntax Highlighting, IntelliSense, Go to Definition, Find All References NOT WORKING

Molunerfinn:
me too

matthewjdegarmo:
@HaykShirinyan, Exact same behavior. I get one tab complete and then it's useless.

naster01:
### How to make it crash 
- dotnet tool install -g dotnet-script
- dotnet script init
- Add a nuget package like this :
[...]

PR #4310, opened 2020-12-21
improve queues, (one effect: improve code completion)

an-dr-eas-k:
After updating to the recent 1.23.x versions I had severe performance issues on code completion. 

I investigated that and made the following changes: 
* Completion, CompletionResolve, and AutoComplete should be equally prioritized (add to normal queue)
* remove request from _waiting queue when cancelRequest is called and call onError on the request to reject() any awaited promises.
* add logging (debugMode=true in main.ts) for canceled requests and add logging details to RequestQueue internals.

The second bullet point addresses a deferred queue that stays always full when requests are cancelled. A symptom for that are many log entries of the kind: 
``` 
Problem invoking 'GetCodeActions' on OmniSharp server: Error: Pending request cancelled: /v2/getcodeactions
```

### DIFF
--- a/src/omnisharp/EventType.ts
+++ b/src/omnisharp/EventType.ts
@@ -83,7 +83,7 @@ export enum EventType {
     DotNetTestRunInContextStart = 76,
     DotNetTestDebugInContextStart = 77,
     TelemetryErrorEvent = 78,
-    OmnisharpServerRequestCanceled = 79
+    OmnisharpServerRequestCancelled = 79
 }
 
 //Note that the EventType protocol is shared with Razor.VSCode and the numbers here should not be altered
--- a/src/omnisharp/EventType.ts
+++ b/src/omnisharp/EventType.ts
@@ -83,7 +83,7 @@ export enum EventType {
     DotNetTestRunInContextStart = 76,
     DotNetTestDebugInContextStart = 77,
     TelemetryErrorEvent = 78,
-    OmnisharpServerRequestCancelled = 79
+    OmnisharpServerRequestCancelled = 79,
 }
 
 //Note that the EventType protocol is shared with Razor.VSCode and the numbers here should not be altered
--- a/src/omnisharp/EventType.ts
+++ b/src/omnisharp/EventType.ts
@@ -82,7 +82,8 @@ export enum EventType {
     ProjectDiagnosticStatus = 75,
     DotNetTestRunInContextStart = 76,
     DotNetTestDebugInContextStart = 77,
-    TelemetryErrorEvent = 78
+    TelemetryErrorEvent = 78,
+    OmnisharpServerRequestCanceled = 79
 }
 
 //Note that the EventType protocol is shared with Razor.VSCode and the numbers here should not be altered
--- a/src/omnisharp/loggingEvents.ts
+++ b/src/omnisharp/loggingEvents.ts
@@ -115,16 +115,16 @@ export class OmnisharpServerUnresolvedDependencies implements BaseEvent {
 
 export class OmnisharpServerEnqueueRequest implements BaseEvent {
     type = EventType.OmnisharpServerEnqueueRequest;
-    constructor(public name: string, public command: string) { }
+    constructor(public queueName: string, public command: string) { }
 }
 
 export class OmnisharpServerDequeueRequest implements BaseEvent {
     type = EventType.OmnisharpServerDequeueRequest;
-    constructor(public name: string, public command: string, public id: number) { }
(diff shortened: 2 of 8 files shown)


src/omnisharp/prioritization.ts
```diff
@@ -13,6 +13,8 @@ const priorityCommands = [
 ];
 
 const normalCommands = [
+    protocol.Requests.Completion,
+    protocol.Requests.CompletionResolve,
     protocol.Requests.AutoComplete,
```
filipw, line 18: this makes sense to me, thanks. Also autocomplete is no longer used, so I guess we could clean up all the references to it @JoeRobich @333fred

an-dr-eas-k:
For me, especially the other change, calling dequeue when a request is canceled is most important. Please note the log entries that I entered to debug the behaviour in 954ce95. I found out, the deferred queue is considered full just when two requests are canceled but remain in the _waiting queue forever. Then the queue does no longer accept further requests. This especially had an impact on code completion in the last versions. But also, Unit-test classes loose the "RunTest" links, Code Structure gets lost in the outline panel and so on. 

The code I am working on has lots of automapper statements which made trouble when calling codeactions couple of times that were subsequently canceled, some already _waiting. And suddenly all requests that are prioritized to deferred queue didn't reach out for the omnisharp server.

I am not sure, if adding the id to the requests is proper, but at least solves a quick handling in the dequeue method.

I think this fix addresses at least some of:
#4292 
#4289
#4311 
#4080
#4047
#4072
#3421
#4065

src/omnisharp/server.ts
```diff
@@ -705,7 +705,8 @@ export class OmniSharpServer {
 
     private _makeRequest(request: Request) {
         const id = OmniSharpServer._nextId++;
-
+        request.id = id;
+
```
333fred, line 709: Nit: some extra whitespace.

src/omnisharp/prioritization.ts
333fred: Looks like this is the only remaining reference, it and the constant can be deleted.

333fred (dismissed):
Overall looks good. We can delete the autocomplete entry and the definition in protocol.ts as well.

an-dr-eas-k:
Sure. I needed to do these adjustments because I realized the OmnisharpWorkspaceSymbolProvider (ctrl-t) also needs request.onError to cancel the awaited task when it is in the waiting queue. 

Essentially what changed in the last 4 commits are two things:
1. I added the OmnisharpServerRequestCancelled BaseEvent and did some small changes to other BaseEvents to get more logging for canceled requests in the output panel when debug=true in main.ts.
2. I moved the request.onError call one level higher to be called independent of the queueStatus the request is in. Originally, onError was only called for requests in the _pending queue. 

Cheers.

src/omnisharp/EventType.ts
```diff
@@ -82,7 +82,8 @@ export enum EventType {
     ProjectDiagnosticStatus = 75,
     DotNetTestRunInContextStart = 76,
     DotNetTestDebugInContextStart = 77,
-    TelemetryErrorEvent = 78
+    TelemetryErrorEvent = 78,
+    OmnisharpServerRequestCancelled = 79
```
333fred, line 86: Nit: consider leaving a trailing comma here, so that future PRs that add elements don't have to modify this line.

src/omnisharp/prioritization.ts
333fred: @an-dr-eas-k would you mind remove this reference?

src/omnisharp/loggingEvents.ts
```diff
@@ -115,17 +115,22 @@ export class OmnisharpServerUnresolvedDependencies implements BaseEvent {
...
+
+export class OmnisharpServerRequestCancelled implements BaseEvent {
+    type = EventType.OmnisharpServerRequestCancelled;
+    constructor(public command: string, public id: number) { }
 }
 
 export class OmnisharpServerProcessRequestStart implements BaseEvent {
     type = EventType.OmnisharpServerProcessRequestStart;
-    constructor(public name: string) { }
+    constructor(public name: string, public slots: number) { }
```
333fred, line 133: `slots` isn't a particularly descriptive name. Can you please rename it to something more descriptive?

333fred (changes requested):
Thanks. This is mostly looking good, just a couple of small pieces of feedback.

333fred (approved):
LGTM. @JoeRobich for a second review.

JoeRobich (approved):
Thanks!

PR #4363, opened 2021-01-25
Move CodeActions to LSP to resolve cancellations

333fred:
The tests were failing because the legacy O# routing is timing out when executing code actions. By moving code actions to LSP, it resolves the issue. Also depends on the changes https://github.com/OmniSharp/omnisharp-roslyn/pull/2072 for a fully green run.


src/omnisharp/extension.ts
```diff
@@ -93,11 +93,6 @@ export async function activate(context: vscode.ExtensionContext, packageJSON: an
         localDisposables.add(vscode.languages.registerCompletionItemProvider(documentSelector, new CompletionProvider(server, languageMiddlewareFeature), '.', ' '));
         localDisposables.add(vscode.languages.registerWorkspaceSymbolProvider(new WorkspaceSymbolProvider(server, optionProvider, languageMiddlewareFeature)));
         localDisposables.add(vscode.languages.registerSignatureHelpProvider(documentSelector, new SignatureHelpProvider(server, languageMiddlewareFeature), '(', ','));
-        // Since the CodeActionProvider registers its own commands, we must instantiate it and add it to the localDisposables
```
david-driscoll, line 96: We should wrap this in an if block to only disable this registration when lsp is enabled.

PR #4369, opened 2021-01-31
Add a SyntaxVisualizer panel

333fred:
This implements a syntax visualizer view, which is styled very similarly to the panel that ships in the roslyn-sdk for VS. There are 2 parts to this view:

The top tree view shows the entire syntax tree of the current file, including the type of the node and the span of the node. It also uses the same type icons that vscode uses in other contexts to indicate whether the node element is a class or struct. Selecting a node in the tree highlights the span in the current document that corresponds to that node, and clicking in the document reveals the node in the tree view.

The bottom view is another tree view that contains detailed information about the currently-selected node in the top view. It automatically updates any time the top selection is changed, and contains the node type, the syntaxkind, the results of calling GetTypeInfo/GetConversion, GetSymbolInfo, and GetDeclaredSymbol, and all public properties of the node. The omnisharp-roslyn side of this PR is here: https://github.com/OmniSharp/omnisharp-roslyn/pull/2078.
SyntaxVisualizer


333fred:
I still need to add a setting to turn this panel on/off. My current thought is off-by-default.

JoeRobich:
> I still need to add a setting to turn this panel on/off. My current thought is off-by-default.

I agree with your intuition. =)

ThirdPartyNotices.txt
```diff
@@ -122,3 +123,31 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 ============================================================
 END OF DefinitelyTyped NOTICES AND INFORMATION
+
+%% roslyn-sdk NOTICES AND INFORMATION BEGINS HERE
+============================================================
```
333fred, line 128: This is for the syntax tree icon, which is from https://github.com/dotnet/roslyn-sdk/blob/master/src/VisualStudio.Roslyn.SDK/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Extension/SyntaxTree.ico

333fred:
@filipw @JoeRobich would it be possible to get reviews on this and the corresponding omnisharp-roslyn side PR?

PR #4371, opened 2021-02-01
Re-enable razor tests

333fred:
Re-enabled the razor tests and moved the TFM back. Not sure why the razor error is causing the SDK fail to execute Compile/CoreCompile, but that's unrelated to LSP so reverting for now.


333fred:
Filed https://github.com/OmniSharp/omnisharp-roslyn/issues/2082 to track the bad server behavior during load.

david-driscoll (approved):

PR #4389, opened 2021-02-08
Fix code coverage reporting in PR CI.

JoeRobich:
After the switch to GitHub Actions code coverage was failing to be reported on PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,22 +28,33 @@ jobs:
         node-version: '12.x'
 
     - name: Install dependencies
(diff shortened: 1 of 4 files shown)


JoeRobich:
At first, I was concerned by the decrease in reported coverage until I clicked through to see that with the new configuratin we are finally reporting all the files in our project (See https://codecov.io/gh/OmniSharp/omnisharp-vscode/pull/4389/changes). This decrease is actually a recalibration.

@nohwnd @vzarytovskii Please take a look.

PR #4392, opened 2021-02-10
Prompt user to install Blazor WASM companion extension if needed

captainsafia:
Detects if the companion extension is installed if the user is editing a Blazor WASM project and prompts them to install it if not.﻿

The companion extension is a new requirement for debugging Blazor WASM on VS Code to allow for:

- support for remote debugging
- faster launching for debugging sessions
- ability to debug on apps with windows auth configured

### DIFF
--- a/src/omnisharp/utils.ts
+++ b/src/omnisharp/utils.ts
@@ -240,6 +240,19 @@ async function isBlazorWebAssemblyProject(project: MSBuildProject): Promise<bool
 }
 
 function hasBlazorWebAssemblyDebugPrerequisites() {
+    const companionExtension = vscode.extensions.getExtension('ms-dotnettools.blazorwasm-companion');
+    if (!companionExtension) {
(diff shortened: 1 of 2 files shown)


JoeRobich (approved):
LGTM. If this were a behavior you would like the vendors to test as part of release validation, then you could add a section to the [test-plan](https://github.com/OmniSharp/omnisharp-vscode/blob/master/test-plan.md).

NTaylorMullen (approved):
What's the debug experience if the user doesn't have the companion installed?

captainsafia:
>  If this were a behavior you would like the vendors to test as part of release validation, then you could add a section to the test-plan.

Updated the test plan.

> What's the debug experience if the user doesn't have the companion installed?

The user will be able to run the the app and open it in a standard browser but the debugging browser will not launch.

JoeRobich (approved):

PR #4415, opened 2021-03-01
Activate the SDK extension as part of dependency installation.

JoeRobich:
Possibly resolution to https://github.com/dotnet/vscode-dotnet-runtime/issues/175

cc: @sfoslund, @NTaylorMullen


src/main.ts
```diff
@@ -224,6 +224,22 @@ function isSupportedPlatform(platform: PlatformInformation): boolean {
 }
 
 async function ensureRuntimeDependencies(extension: vscode.Extension<CSharpExtensionExports>, eventStream: EventStream, platformInfo: PlatformInformation, installDependencies: IInstallDependencies): Promise<boolean> {
+    await activateSdkExtension(eventStream);
     return installRuntimeDependencies(extension.packageJSON, extension.extensionPath, installDependencies, eventStream, platformInfo);
```
NTaylorMullen, line 228: Hmm, I believe we'll want to activate the SDK extension after all of the runtime dependencies have been installed but before they've been activated / have the chance to complain about a missing SDK 😄.

Also, since the SDK extension provides a path to dotnet the debugger piece will need to conditionally error because the current VSCode instance wont have the newly acquire dotnet sdk on the path yet since it will be in the midst of being acquired.

gregg-miskelly:
How does the SDK extension help solve https://github.com/dotnet/vscode-dotnet-runtime/issues/175? The problem sounds like it is because PATH hasn't been updated yet. Even without the SDK extension, restarting VS Code should normally fix that. Does the SDK extension have some hack to somehow force that environment variable into all of the VS Code processes?

JoeRobich:
> Does the SDK extension have some hack to somehow force that environment variable into all of the VS Code processes?

Nope, you are right. I got ahead of myself.

gregg-miskelly:
Sorry if this is obvious: I don't think there is a way to actually address the problem as there is too much in VS Code that relies on 'dotnet' being on the PATH. The only think we could maybe do is improve the message in the dialog. The help would already tell folks what the problem is, but maybe that isn't strong enough.

src/main.ts
```diff
@@ -237,7 +238,14 @@ async function activateSdkExtension(eventStream: EventStream) {
         // Ensure SDK is installed.
         eventStream.post(new PackageInstallation(".NET 5.0 SDK"));
         const aquisitionResult = await vscode.commands.executeCommand<{ dotnetPath: string }>('dotnet-sdk.acquire', { version: '5.0' });
-        const result = aquisitionResult.dotnetPath?.length > 0
+
+        const success = aquisitionResult.dotnetPath?.length > 0;
+        if (success) {
+            process.env.PATH += path.delimiter + path.dirname(aquisitionResult.dotnetPath);
```
gregg-miskelly, line 244: I don't think this is going to help. You would need this change also in:
* The process that launches task.json tasks
* The debugger process

JoeRobich: It does work well enough for the processes launched by the C# extension, including OmniSharp, dotnet, and the debugger. It falls apart with the process we do not start such as the terminal and prelaunch tasks.

I am able to debug a pre-built .net core application from my first run of VS Code in the Sandbox.
image

JoeRobich:
@gregg-miskelly you are right. I think the best we can do is check the PATH to see if the extension's sdk install path is present. If missing, we can inform the user that .NET SDK has been installed and that a restart is required.

JoeRobich:
Or maybe I spoke too soon. Investigating https://code.visualstudio.com/api/references/vscode-api#EnvironmentVariableCollection

src/main.ts
```diff
@@ -224,6 +225,29 @@ function isSupportedPlatform(platform: PlatformInformation): boolean {
...
 
+async function activateSdkExtension(eventStream: EventStream, platformInfo: PlatformInformation) {
+    const sdkExtension = vscode.extensions.getExtension("ms-dotnettools.vscode-dotnet-sdk");
+    if (sdkExtension !== undefined) {
+        // Activate extension.
+        await sdkExtension.activate();
+
+        // Ensure SDK is installed.
+        eventStream.post(new PackageInstallation(".NET 5.0 SDK"));
+        const aquisitionResult = await vscode.commands.executeCommand<{ dotnetPath: string }>('dotnet-sdk.acquire', { version: '5.0' });
```
sfoslund, line 240: I believe just calling acquire here will work, do we still need a new command from the SDK extension to check install status? The functional difference would be that the acquire command will cause a SDK to be installed if it hasn't been already, whereas we could implement a new command that will return 'undefined' or something like that if we aren't installing anything to signal that no SDK has been installed.

JoeRobich:
@gregg-miskelly So using the EnvironmentVariableCollection and appending the dotnet folder to the Path allows me to use `dotnet` from the terminal. The build tasks would work too, if they were configured as "shell" instead of "process".

gregg-miskelly:
> @gregg-miskelly So using the EnvironmentVariableCollection and appending the dotnet folder to the Path allows me to use `dotnet` from the terminal. The build tasks would work too, if they were configured as "shell" instead of "process".

Excellent!

package.json
```diff
@@ -400,7 +400,7 @@
     }
   ],
   "engines": {
-    "vscode": "^1.44.0"
+    "vscode": "^1.46.0"
```
JoeRobich, line 403: Need to update the vscode version to get access to EnvironmentVariableCollection.

src/main.ts
JoeRobich: > I believe we'll want to activate the SDK extension after all of the runtime dependencies have been installed but before they've been activated / have the chance to complain about a missing SDK

I don't know about that. We aleady acquire dependencies serially. I think it would be best to just treat the SDK as the most important dependency.

src/main.ts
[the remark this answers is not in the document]
JoeRobich: Updated.

JoeRobich:
Closing this issue as I think bits of these changes need to be spread around to other repos.

src/main.ts
```diff
@@ -223,7 +224,40 @@ function isSupportedPlatform(platform: PlatformInformation): boolean {
...
+        const success = aquisitionResult.dotnetPath?.length > 0;
+        if (success) {
+            const dotnetDirectory = path.dirname(aquisitionResult.dotnetPath);
+
+            // Determine if dotnet path should be added to the Environment Path.
+            if (!process.env.PATH.includes(dotnetDirectory)) {
+                const appendDotnetPath = path.delimiter + dotnetDirectory;
+
+                // Updates VS Code terminal instances and tasks with type 'shell'.
+                environmentVariables.append("PATH", appendDotnetPath);
```
JoeRobich, line 251: This is now performed in the dotnet-sdk extension.

src/main.ts
```diff
@@ -223,7 +224,40 @@ function isSupportedPlatform(platform: PlatformInformation): boolean {
...
 }
 
-async function ensureRuntimeDependencies(extension: vscode.Extension<CSharpExtensionExports>, eventStream: EventStream, platformInfo: PlatformInformation, installDependencies: IInstallDependencies): Promise<boolean> {
+async function ensureRuntimeDependencies(extension: vscode.Extension<CSharpExtensionExports>, environmentVariables: vscode.EnvironmentVariableCollection, eventStream: EventStream, platformInfo: PlatformInformation, installDependencies: IInstallDependencies): Promise<boolean> {
+    await activateSdkExtension(environmentVariables, eventStream);
     return installRuntimeDependencies(extension.packageJSON, extension.extensionPath, installDependencies, eventStream, platformInfo);
 }
 
+async function activateSdkExtension(environmentVariables: vscode.EnvironmentVariableCollection, eventStream: EventStream) {
+    const sdkExtension = vscode.extensions.getExtension("ms-dotnettools.vscode-dotnet-sdk");
```
JoeRobich, line 233: use `"ms-dotnettools.vscode-dotnet-pack" instead.

src/main.ts
```diff
@@ -223,7 +224,40 @@ function isSupportedPlatform(platform: PlatformInformation): boolean {
...
+    if (sdkExtension !== undefined) {
+        // Activate SDK extension.
+        await sdkExtension.activate();
+
+        // Ensure SDK is installed.
+        eventStream.post(new PackageInstallation(".NET 5.0 SDK"));
+        const aquisitionResult = await vscode.commands.executeCommand<{ dotnetPath: string }>('dotnet-sdk.acquire', { version: '5.0', requestingExtensionId: 'ms-dotnettools.csharp' });
```
JoeRobich, line 240: use `"dotnet-sdk. acquireStatus"` instead.

PR #4445, opened 2021-03-18
Add properties to blazorwasm debug configuration.

DL444:
Fixes dotnet/aspnetcore#30977.

Property definition taken directly from https://github.com/dotnet/aspnetcore-tooling/blob/c866a4bce578412c78443988c957fccacf5494cd/src/Razor/src/Microsoft.AspNetCore.Razor.VSCode.Extension/package.json#L68.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3216,54 +3216,6 @@
                 "type": "object",
                 "description": "Environment variables passed to dotnet. Only valid for hosted apps."
               },
-              "logging": {
-                "description": "Optional flags to determine what types of messages should be logged to the output window. Applicable only for the app server of hosted Blazor WASM apps.",
-                "type": "object",
-                "required": [],
-                "default": {},
-                "properties": {
-                  "exceptions": {
-                    "type": "boolean",
-                    "description": "Optional flag to determine whether exception messages should be logged to the output window.",
-                    "default": true
-                  },
-                  "moduleLoad": {
-                    "type": "boolean",
-                    "description": "Optional flag to determine whether module load events should be logged to the output window.",
-                    "default": true
-                  },
-                  "programOutput": {
-                    "type": "boolean",
-                    "description": "Optional flag to determine whether program output should be logged to the output window when not using an external console.",
-                    "default": true
-                  },
-                  "engineLogging": {
-                    "type": "boolean",
-                    "description": "Optional flag to determine whether diagnostic engine logs should be logged to the output window.",
-                    "default": false
-                  },
-                  "browserStdOut": {
-                    "type": "boolean",
-                    "description": "Optional flag to determine if stdout text from the launching the web browser should be logged to the output window.",
-                    "default": true
-                  },
-                  "elapsedTiming": {
-                    "type": "boolean",
-                    "description": "If true, engine logging will include `adapterElapsedTime` and `engineElapsedTime` properties to indicate the amount of time, in microseconds, that a request took.",
-                    "default": false
-                  },
-                  "threadExit": {
-                    "type": "boolean",
-                    "description": "Controls if a message is logged when a thread in the target process exits. Default: `false`.",
-                    "default": false
-                  },
-                  "processExit": {
-                    "type": "boolean",
-                    "description": "Controls if a message is logged when the target process exits, or debugging is stopped. Default: `true`.",
-                    "default": true
-                  }
-                }
-              },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3263,6 +3263,37 @@
                     "default": true
                   }
                 }
+              },
+              "justMyCode": {
```
captainsafia, line 3267: The `sourceFileMap` and `justMyCode` options should be included under the dotnet config flag. E.g. replace this changeset with:

```
                  "dotNetConfig": {
                    "description": "Options passed to the underlying .NET debugger. For more info, see https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger.md.",
                    "type": "object",
                    "required": [],
                    "default": {},
                    "properties": {
                      "justMyCode": {
                        "type": "boolean",
                        "description": "Optional flag to only show user code.",
                        "default": true
                      },
                      "sourceFileMap": {
                        "type": "object",
                        "description": "Optional source file mappings passed to the debug engine. Example: '{ \"C:\\foo\":\"/home/user/foo\" }'",
                        "additionalProperties": {
                          "type": "string"
                        },
                        "default": {
                          "<insert-source-path-here>": "<insert-target-path-here>"
                        }
                      },
                      "logging": {
                        "description": "Optional flags to determine what types of messages should be logged to the output window.",
                        "type": "object",
                        "required": [],
                        "default": {},
                        "properties": {
                          "exceptions": {
                            "type": "boolean",
                            "description": "Optional flag to determine whether exception messages should be logged to the output window.",
                            "default": true
                          },
                          "moduleLoad": {
                            "type": "boolean",
                            "description": "Optional flag to determine whether module load events should be logged to the output window.",
                            "default": true
                          },
                          "programOutput": {
                            "type": "boolean",
                            "description": "Optional flag to determine whether program output should be logged to the output window when not using an external console.",
                            "default": true
                          },
                          "engineLogging": {
                            "type": "boolean",
                            "description": "Optional flag to determine whether diagnostic engine logs should be logged to the output window.",
                            "default": false
                          },
                          "browserStdOut": {
                            "type": "boolean",
                            "description": "Optional flag to determine if stdout text from the launching the web browser should be logged to the output window.",
                            "default": true
                          },
                          "elapsedTiming": {
                            "type": "boolean",
                            "description": "If true, engine logging will include `adapterElapsedTime` and `engineElapsedTime` properties to indicate the amount of time, in microseconds, that a request took.",
                            "default": false
                          },
                          "threadExit": {
                            "type": "boolean",
                            "description": "Controls if a message is logged when a thread in the target process exits. Default: `false`.",
                            "default": false
                          },
                          "processExit": {
                            "type": "boolean",
                            "description": "Controls if a message is logged when the target process exits, or debugging is stopped. Default: `true`.",
                            "default": true
                          }
                        }
                      }
                    }
                  },
                  "browserConfig": {
                    "description": "Options based to the underlying JavaScript debugger. For more info, see https://github.com/microsoft/vscode-js-debug/blob/main/OPTIONS.md.",
                    "type": "object",
                    "required": [],
                    "default": {},
                    "properties": {
                      "outputCapture": {
                        "enum": [
                          "console",
                          "std"
                        ],
                        "description": "From where to capture output messages: the default debug API if set to `console`, or stdout/stderr streams if set to `std`.",
                        "default": "console"
                      }
                    }
                  }
```

DL444: Changed accordingly.

I noticed that `logging` property was originally outside of `dotNetConfig`, while in the original file that property was inside. Is this intentional?

package.json
```diff
@@ -3263,6 +3263,45 @@
                     "default": true
                   }
                 }
+              },
+              "dotNetConfig": {
```
captainsafia, line 3267: And one more thing: the `logging` option from the L3219 above should also be moved inside here.

Almost there!

DL444: The property is now moved.

captainsafia (approved):
Looks great! Thanks for reacting to all the feedback given.

captainsafia:
@JoeRobich Looks like this needs a review from you.

This contributor is submitting a PR to update the extension config with some properties I forgot to update during the last insertion.

JoeRobich (approved):

JoeRobich:
Merging this PR is currently blocked on https://github.com/microsoft/vscode/issues/119822

PR #4459, opened 2021-03-25
Add path to dotnet so child processes can use the CLI

JoeRobich:


JoeRobich:
@nohwnd @vzarytovskii Please review.

PR #4477, opened 2021-04-05
Revert incremental change forwarding

333fred:
While I made this change with good intentions (and the perf benefit when working in Roslyn has been very real), this has proven to be more trouble than it's worth. LSP will be the way to get the perf benefits of incremental changes and should prove to be far more robust. This is a checkout of changeForwarding.ts to d26e995342b8df02135a5cbf0d385a2c02a6dd5b. The only other commits to this file since then have been about 0-based indexes, which no longer matters with this revert.

### DIFF
--- a/src/features/changeForwarding.ts
+++ b/src/features/changeForwarding.ts
@@ -6,7 +6,7 @@
 import { Uri, workspace } from 'vscode';
 import { OmniSharpServer } from '../omnisharp/server';
 import * as serverUtils from '../omnisharp/utils';
-import { FileChangeType, LinePositionSpanTextChange } from '../omnisharp/protocol';
+import { FileChangeType } from '../omnisharp/protocol';
 import { IDisposable } from '../Disposable';
 import CompositeDisposable from '../CompositeDisposable';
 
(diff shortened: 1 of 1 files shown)


JoeRobich (approved):
Thanks @333fred!

src/features/changeForwarding.ts
```diff
@@ -6,40 +6,24 @@
...
+import { FileChangeType } from '../omnisharp/protocol';
 import { IDisposable } from '../Disposable';
 import CompositeDisposable from '../CompositeDisposable';
 
 function forwardDocumentChanges(server: OmniSharpServer): IDisposable {
 
     return workspace.onDidChangeTextDocument(event => {
 
-        let { document, contentChanges } = event;
+        let { document } = event;
```
JoeRobich, line 17:
```suggestion
        let { document, contentChanges } = event;
```

src/features/changeForwarding.ts
```diff
@@ -6,40 +6,24 @@
...
         if (document.isUntitled || document.languageId !== 'csharp' || document.uri.scheme !== 'file') {
             return;
         }
```
JoeRobich, line 21: I think we should restore this change since there are no changes to be synced.

```suggestion

        if (contentChanges.length === 0) {
            // This callback fires with no changes when a document's state changes between "clean" and "dirty".
            return;
        }

```

PR #4498, opened 2021-04-14
Add OSX-ARM64 support

WardenGnaw:
This PR adds the debugger packages for osx-arm64.

### DIFF
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -36,16 +36,18 @@ export async function activate(thisExtension: vscode.Extension<CSharpExtensionEx
 
 async function checkForInvalidArchitecture(platformInformation: PlatformInformation, eventStream: EventStream): Promise<boolean> {
     if (platformInformation) {
-        if (platformInformation.isMacOS() && !CoreClrDebugUtil.isMacOSSupported()) {
-            eventStream.post(new DebuggerPrerequisiteFailure("[ERROR] The debugger cannot be installed. The debugger requires macOS 10.12 (Sierra) or newer."));
-            return true;
+        if (platformInformation.isMacOS()) {
+            // Support x86_64 and arm64
+            // Validate we are on compatiable macOS version if we are x86_64
+            if ((platformInformation.architecture !== "x86_64" && platformInformation.architecture !== "arm64") || 
+                (platformInformation.architecture === "x86_64" && !CoreClrDebugUtil.isMacOSSupported())) {
+                eventStream.post(new DebuggerPrerequisiteFailure("[ERROR] The debugger cannot be installed. The debugger requires macOS 10.12 (Sierra) or newer."));
+                return true;
+            }
         }
         else if (platformInformation.architecture !== "x86_64") {
             if (platformInformation.isWindows() && platformInformation.architecture === "x86") {
                 eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: x86 Windows is not currently supported by the .NET Core debugger. Debugging will not be available.`));
-            } else if (platformInformation.isMacOS() && platformInformation.architecture === "arm64") {
-                eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: arm64 macOS is not officially supported by the .NET Core debugger. You may experience unexpected issues when running in this configuration.`));
-                return false;
             } else {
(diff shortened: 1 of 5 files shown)


src/tools/OptionsSchema.json
```diff
@@ -386,6 +386,10 @@
           "type": "string",
           "description": "When set, text that the target application writes to stdout and stderr (ex: Console.WriteLine) will be saved to the specified file. This option is ignored if console is set to something other than internalConsole.",
           "default": "${workspaceFolder}/out.txt"
+        },
+        "targetArchitecture": {
+          "type": "string",
+          "description": "The architecture of the debuggee. This will automatically be detected unless this parameter is set. Allowed values are x86_64 or arm64."
```
gregg-miskelly, line 392: >arm64 [](start = 155, length = 5)

I would add a note that this is currently only supported for local debugging on macOS.

src/coreclr-debug/activate.ts
```diff
@@ -155,9 +157,16 @@ export class DebugAdapterExecutableFactory implements vscode.DebugAdapterDescrip
 
         // debugger has finished installation, kick off our debugger process
 
+        // Check for targetArchitecture
+        let targetArchitecture = "";
+        if (this.platformInfo.isMacOS())
+        {
+            targetArchitecture = _session?.configuration?.targetArchitecture ?? this.platformInfo.architecture;
```
gregg-miskelly, line 164: >targetArchitecture [](start = 12, length = 18)

Should we fail if target architecture is invalid (maybe just check if the vsdbg-ui exists to validate?)

src/coreclr-debug/activate.ts
```diff
@@ -36,16 +36,18 @@ export async function activate(thisExtension: vscode.Extension<CSharpExtensionEx
...
+                eventStream.post(new DebuggerPrerequisiteFailure("[ERROR] The debugger cannot be installed. The debugger requires macOS 10.12 (Sierra) or newer."));
+                return true;
+            }
         }
         else if (platformInformation.architecture !== "x86_64") {
             if (platformInformation.isWindows() && platformInformation.architecture === "x86") {
                 eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: x86 Windows is not currently supported by the .NET Core debugger. Debugging will not be available.`));
-            } else if (platformInformation.isMacOS() && platformInformation.architecture === "arm64") {
-                eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: arm64 macOS is not officially supported by the .NET Core debugger. You may experience unexpected issues when running in this configuration.`));
-                return false;
```
gregg-miskelly, line 48: Does the user still get a warning on arm64? I think we want one since:
1. The language service isn't downloaded/used I assume? (Could we just use the x64 one?)
2. It requires either the user install the preview of .NET 6 -or- modify their launch.json to target x64
3. The .NET runtime is still very much a preview

src/coreclr-debug/activate.ts
gregg-miskelly, line 164: >this.platformInfo.architecture [](start = 80, length = 30)

Do we want to default to x64 until .NET 6 comes out? Or maybe do the work of running 'dotnet --info' and scaping the RID?

gregg-miskelly (changes requested):
See comments

src/coreclr-debug/util.ts
```diff
@@ -117,3 +118,61 @@ export class CoreClrDebugUtil {
...
+        {
+            if (launchJsonTargetArchitecture !== "x86_64" && launchJsonTargetArchitecture !== "arm64")
+            {
+                throw new Error(`The value '${launchJsonTargetArchitecture}' for 'targetArchitecture' in launch configuraiton is invalid. Expected 'x86_64' or 'arm64'.`);
+            }
+            targetArchitecture = launchJsonTargetArchitecture;
+        }
+        else if (dotnetInfo)
+        {
+            // Find which targetArchitecture to use based on SDK Version or RID.
```
WardenGnaw, line 147: Do we only want to do the `dotnet --info` path if we have a `.dll`? If we are launching an exe do we want to run `file` on the launch.json `program`?

gregg-miskelly: I think that depends on how fancy you want to get (and also how fast `file` is). Personally I think it is reasonable to use the new property you added.

gregg-miskelly (approved):

PR #4509, opened 2021-04-20
Handle ProcessPicker via resolveDebugConfiguration

WardenGnaw:
VS Code commands are limited to only being able to have a single output.
We will handle having an empty processId and show the dialog for
ProcessPicker in resolveDebugConfigurationWithSubstitutedVariables.
We need multiple outputs to handle the latest macOS on Apple M1.

For Apple Silicon M1 (ARM64), we need to determine if we need to use the
x86_64 or arm64 debugger. We are able to detect if the process is using
S_TRANSLATED using the ps commandline with 'flags', if it is set with
0x20000, it is emulated.

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -421,7 +421,7 @@ export function createAttachConfiguration(): string {
         "name": ".NET Core Attach",
         "type": "coreclr",
         "request": "attach",
-        "processId": "\${command:pickProcess}"
+        "processId": ""
     };
 
     return JSON.stringify(configuration);
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -420,8 +420,7 @@ export function createAttachConfiguration(): string {
     const configuration = {
         "name": ".NET Core Attach",
         "type": "coreclr",
-        "request": "attach",
-        "processId": ""
+        "request": "attach"
     };
 
     return JSON.stringify(configuration);
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -13,6 +13,7 @@ import { EventStream } from '../EventStream';
 import CSharpExtensionExports from '../CSharpExtensionExports';
 import { getRuntimeDependencyPackageWithId } from '../tools/RuntimeDependencyPackageUtils';
 import { getDotnetInfo, DotnetInfo } from '../utils/getDotnetInfo';
+import { CoreCLRConfigurationProvider } from './debugConfigurationProvider';
 
 let _debugUtil: CoreClrDebugUtil = null;
 
@@ -30,6 +31,7 @@ export async function activate(thisExtension: vscode.Extension<CSharpExtensionEx
     }
 
     const factory = new DebugAdapterExecutableFactory(platformInformation, eventStream, thisExtension.packageJSON, thisExtension.extensionPath);
+    context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('coreclr', new CoreCLRConfigurationProvider(platformInformation)));
     context.subscriptions.push(vscode.debug.registerDebugAdapterDescriptorFactory('coreclr', factory));
     context.subscriptions.push(vscode.debug.registerDebugAdapterDescriptorFactory('clr', factory));
 }
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -13,7 +13,7 @@ import { EventStream } from '../EventStream';
 import CSharpExtensionExports from '../CSharpExtensionExports';
 import { getRuntimeDependencyPackageWithId } from '../tools/RuntimeDependencyPackageUtils';
 import { getDotnetInfo, DotnetInfo } from '../utils/getDotnetInfo';
-import { CoreCLRConfigurationProvider } from './debugConfigurationProvider';
+import { DotnetDebugConfigurationProvider } from './debugConfigurationProvider';
 
 let _debugUtil: CoreClrDebugUtil = null;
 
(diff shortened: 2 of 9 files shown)


src/features/commands.ts
```diff
@@ -40,10 +40,9 @@ export default function registerCommands(context: vscode.ExtensionContext, serve
     // running the command activates the extension, which is all we need for installation to kickoff
     disposable.add(vscode.commands.registerCommand('csharp.downloadDebugger', () => { }));
 
-    // register process picker for attach
-    let attachItemsProvider = DotNetAttachItemsProviderFactory.Get();
-    let attacher = new AttachPicker(attachItemsProvider);
-    disposable.add(vscode.commands.registerCommand('csharp.listProcess', async () => attacher.ShowAttachEntries()));
+    // register process picker for attach for legacy configurations.
+    disposable.add(vscode.commands.registerCommand('csharp.listProcess', async () => ""));
```
WardenGnaw, line 44: Would we want to search for a special key instead of empty string?

gregg-miskelly: This seems to work pretty well to me. Though can we remove the `async` now? Or is this required to provide an async delegate?

src/coreclr-debug/activate.ts
```diff
@@ -30,6 +31,7 @@ export async function activate(thisExtension: vscode.Extension<CSharpExtensionEx
     }
 
     const factory = new DebugAdapterExecutableFactory(platformInformation, eventStream, thisExtension.packageJSON, thisExtension.extensionPath);
+    context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('coreclr', new CoreCLRConfigurationProvider(platformInformation)));
```
gregg-miskelly, line 34: >coreclr [](http://example.com/codeflow?start=80&length=7)

Since we are getting rid of 'csharp.listProcess', we need to do this for 'clr' also.

src/features/commands.ts
```diff
@@ -40,10 +40,9 @@ export default function registerCommands(context: vscode.ExtensionContext, serve
     // running the command activates the extension, which is all we need for installation to kickoff
     disposable.add(vscode.commands.registerCommand('csharp.downloadDebugger', () => { }));
 
-    // register process picker for attach
-    let attachItemsProvider = DotNetAttachItemsProviderFactory.Get();
-    let attacher = new AttachPicker(attachItemsProvider);
-    disposable.add(vscode.commands.registerCommand('csharp.listProcess', async () => attacher.ShowAttachEntries()));
```
gregg-miskelly, line 46: >csharp.listProcess [](http://example.com/codeflow?start=52&length=18)

Should we remove `"processId": "${command:pickProcess}"` from the template?

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    public async resolveDebugConfigurationWithSubstitutedVariables(folder: vscode.WorkspaceFolder | undefined, debugConfiguration: vscode.DebugConfiguration, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration>
+    {
+        if (!debugConfiguration)
+        {
+            return null;
+        }
+
+        // Process Id is empty, handle Attach to Process Dialog.
+        if (debugConfiguration.request === "attach" && !debugConfiguration.processId)
```
gregg-miskelly, line 22: >) [](http://example.com/codeflow?start=84&length=1)

`&& !debugConfiguration.processName`

gregg-miskelly:
>     disposable.add(vscode.commands.registerCommand('csharp.listRemoteProcess', async (args) => RemoteAttachPicker.ShowAttachEntries(args, platformInfo)));

Feel free to tell me that this is too much of a pain in the but, but would it make sense to make remote work the same way? This would address the work item about substitution variables and make the remote process listing code less hacky since we wouldn't need to open launch.json and try to find the referenced configuration.

---
Refers to: src/features/commands.ts:49 in 4bfd862. [](commit_id = 4bfd862016166c2077e262816a85e5ddfeecc27b, deletion_comment = False)

src/features/processPicker.ts
```diff
@@ -39,6 +40,23 @@ export class AttachPicker {
                     });
             });
     }
```
gregg-miskelly, line 42: Delete?

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+            let attachItemsProvider = DotNetAttachItemsProviderFactory.Get();
+            let attacher = new AttachPicker(attachItemsProvider);
+            const process: AttachItem = await attacher.SelectProcess();
+
+            if (process)
+            {
+                debugConfiguration.processId = process.id;
+                if (this.platformInformation.isMacOS() && this.platformInformation.architecture == 'arm64')
+                {
+                    if (process.flags & 0x20000)
```
gregg-miskelly, line 33: >0x20000 [](http://example.com/codeflow?start=40&length=7)

Probably worth a comment as to what this is

gregg-miskelly (changes requested):
Otherwise LGTM

src/features/commands.ts
gregg-miskelly: If so, we probably should update the documentation

src/coreclr-debug/debugConfigurationProvider.ts
gregg-miskelly, line 22: >if [](http://example.com/codeflow?start=8&length=2)

If we do NOT convert remote to use this model, do we need to ignore remote targets in the case that the user either failed to specify a process id / name, or cancelled the process selection dialog?

WardenGnaw:
Tested:

| Type  | Platform   | Scenario     | Result |
| :------------- | :----------: | -----------: |-----------: |
|  clr | Windows  | Launch  | ✔ |
|  clr | Windows  | Attach | ✔|
|  coreclr | Windows  | Attach | ✔|
|  coreclr | Windows  | Attach | ✔|
|  coreclr | Windows  | Attach to WSL | ✔|
|  coreclr | macOS (ARM64) | Launch x86_64 dotnet | ✔|
|  coreclr | macOS (ARM64) | Attach x86_64 dotnet| ✔|
|  coreclr | macOS (ARM64) | Launch arm64 dotnet| ✔|
|  coreclr | macOS (ARM64) | Attach arm64 dotnet| ✔|

src/assets.ts
```diff
@@ -421,7 +421,7 @@ export function createAttachConfiguration(): string {
         "name": ".NET Core Attach",
         "type": "coreclr",
         "request": "attach",
-        "processId": "\${command:pickProcess}"
+        "processId": ""
```
gregg-miskelly, line 424: >"processId": "" [](http://example.com/codeflow?start=8&length=15)

What do you think about removing it entirely instead?

gregg-miskelly (approved):
Aside from maybe removing 'processId' from the templates looks good to me.

src/assets.ts
WardenGnaw: I also had the same thought, if users wanted to override it themselves then they would most likely investigate the options in `launch.json` and set it themselves.

gregg-miskelly: Sure, but that seems rare, and if you want to do it, seems like `processName` would make more sense than `processId`. The only reason I can think to of using `processId` is if you are on an image without `ps` (or with a `ps` that acts differently than we expect).

gregg-miskelly (approved):

PR #4563, opened 2021-05-23
Add support for InlineValues

333fred:
VSCode 1.56 enables support for inline values in the debugger, with appropriate extension participation. This adds that feature to C#, with the server implementation here: https://github.com/OmniSharp/omnisharp-roslyn/pull/2158.


gregg-miskelly:
If I understand correctly, this will mean that, by default, all properties/fields of all on-screen variables will be evaluated. Correct?

If so:
1. Is this just limited to the method of the current frame?
2. Is there a limit to the number of lines forward/back?
3. Do we want this on-by-default, or should a user need to opt-in? I am worried about perf, especially for evaluating property getters on non-Windows platforms (they can be rather slow).
4. I haven't played around with this feature too much in browser tools, do browsers try to avoid showing stale values (see example below), and if so do you need to handle that in the Roslyn implementation?

Stale value example:
```
int x = GetInitialValue(); // Do we need to avoid showing `x` here since the value has been reassigned
x = GetNextValue(x);
Console.WriteLine(x); // Debugger is stopped here
```

333fred:
> all properties/fields of all on-screen variables will be evaluated

Mostly. Currently, it's just locals, parameters, and properties/fields on `this/T`. I can't really show, say `x.Prop`, because I don't know what the state of `x` is from this provider. It could be `null`, in which case we just cause null reference exceptions. I probably do need to check for the `DebuggerHidden` attribute on properties/fields and avoid generating displays for them, but that's not implemented currently.

> Is this just limited to the method of the current frame?

Yep. While I'd love to show more things, we'd more extensive communication with the debugger to associate frames with methods and do mapping.

> Is there a limit to the number of lines forward/back?

The smaller of the visible lines in the `viewport` or the size of the current method.

> Do we want this on-by-default, or should a user need to opt-in? I am worried about perf, especially for evaluating property getters on non-Windows platforms (they can be rather slow).

VSCode has a global setting for controlling whether inline values are showed or not. I think we should defer to that setting, and similar for any other editor that implements the feature.

> I haven't played around with this feature too much in browser tools, do browsers try to avoid showing stale values (see example below), and if so do you need to handle that in the Roslyn implementation?

I don't view your example as "stale" values. This isn't attempting to be time-travel debugging or anything like that. It's attempting to show the current values of variables, that's it.

333fred:
You can also view a gif of what the feature looks like in the omnisharp-roslyn side PR: https://github.com/OmniSharp/omnisharp-roslyn/pull/2158#issuecomment-846486354.

333fred:
Closing this out as stale. @gregg-miskelly, I do think we should implement this at some point, but perhaps that some point is not today.

Issue #4607, opened 2021-06-11
Remote process picker does not work (attach to Docker container scenario)


karolz-ms:
## Environment data
MacOS BigSur 11.4
.NET SDK 5.0.301
VS Code 1.57.0
C# extension 1.23.12
Docker extension 1.13.0
Docker Desktop for Mac version 3.3.3

## Steps to reproduce
1. Create new web API project: `dotnet new webapi --no-https`
2. Open application folder in VS Code
3. When prompted, add "required build files" to workspace
4. Run `Docker: add Docker files to workspace` command. Select ".NET ASP.NET Core", Linux, port 5000, yes to optional Docker Compose files
5. Change first line of the Dockerfile so that it reads `FROM mcr.microsoft.com/dotnet/aspnet:5.0-focal AS base` (note the image tag is `5.0-focal`, not just `5.0`
6. Right-click Dockerfile and choose "build image"
7. Switch to Docker view (whale icon), right-click the application image (expand to see the "latest" tag), and choose "Run"
8. Open .vscode/launch.json and add the following debug configuration
```json
{
    "name": "Docker .NET Core Attach",
    "type": "docker",
    "request": "attach",
    "platform": "netCore",
    "processId": "${command:pickRemoteProcess}"
}
```
9. Make the added debug configuration active and execute it. Select "Individual Containers" group, then the running application container. When prompted, allow debugger to be copied to the application container.

## Expected  behavior
After the debugger is copied to the application container, the user should be prompted to select the process to attach to.

## Actual behavior
An error message that says "No process with the specified id is currently running"

## Additional information
When Docker extension resolves the debug configuration, it will return the following:
```json
{
  "name": "Docker .NET Core Attach",
  "justMyCode": false,
  "platform": "netCore",
  "processId": "${command:pickRemoteProcess}",
  "processName": undefined,
  "request": "attach",
  "type": "coreclr",
  "sourceFileMap": {
    "/src": "${workspaceFolder}"
  },
  "pipeTransport": {
    "debuggerPath": "/remote_debugger/vsdbg",
    "pipeCwd": "${workspaceFolder}",
    "pipeProgram": "docker",
    "quoteArgs":  false,
    "pipeArgs": [
      "exec",
      "-i",
      "application_container_name"
    ]
  }
}
```
(`application_container_name` is the name of the running application container)

[...]

PR #4617, opened 2021-06-17
Add command 'listRemoteDockerProcess' and variable 'pickRemoteDockerProcess'

WardenGnaw:
This will enable the old RemoteAttachPicker call that will only return
processIds.

Temporary workaround for #4607 until https://github.com/microsoft/vscode/issues/110889 is resolved.

### DIFF
--- a/src/features/commands.ts
+++ b/src/features/commands.ts
@@ -11,6 +11,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as protocol from '../omnisharp/protocol';
 import * as vscode from 'vscode';
+import { RemoteAttachPicker } from './processPicker';
 import { generateAssets } from '../assets';
 import { ShowOmniSharpChannel, CommandDotNetRestoreStart, CommandDotNetRestoreProgress, CommandDotNetRestoreSucceeded, CommandDotNetRestoreFailed } from '../omnisharp/loggingEvents';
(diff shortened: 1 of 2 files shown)


JoeRobich (approved):

src/features/commands.ts
```diff
@@ -42,6 +43,9 @@ export default function registerCommands(context: vscode.ExtensionContext, serve
     disposable.add(vscode.commands.registerCommand('csharp.listProcess', () => ""));
     disposable.add(vscode.commands.registerCommand('csharp.listRemoteProcess', () => ""));
 
+    // List remote processes for docker extension.
+    // Change to return "" when https://github.com/microsoft/vscode/issues/110889 is resolved.
+    disposable.add(vscode.commands.registerCommand('csharp.listRemoteDockerProcess', async (args) => RemoteAttachPicker.ShowAttachEntries(args, platformInfo)));
```
gregg-miskelly, line 48: >RemoteAttachPicker [](http://example.com/codeflow?start=101&length=18)

Shouldn't this be something like?

```
const attachItem = await RemoteAttachPicker(...);
return attachItem.Id;
```

gregg-miskelly (approved):
Otherwise LGTM

gregg-miskelly (approved):

Issue #4523, opened 2021-04-28
Tests debug socket too long on domain users


valters-tomsons:
## Issue Description ##

When logged in as a domain user which results in a longer `$HOME` path, trying to debug any unit tests fails with the following error message:

```
WARNING: Unable to connect to debuggerEventsPipeName '/home/internal.domain.com/name.surname/.vscode/extensions/ms-dotnettools.csharp-1.23.11/.TestDebugEvents-592495'. Socket name '/home/internal.domain.com/name.surname/.vscode/extensions/ms-dotnettools.csharp-1.23.11/.TestDebugEvents-592495' exceeds the maximum length 107. (Parameter 'path')
```

## Steps to Reproduce ##

1. Create a user inside of the domain (I guess a wierdly long username would work too)
2. Try to debug xunit project
3. Debug console will print a warning message

## Expected Behavior ##

Debugger gets attached and tests can be debugged. 
There should be an option to specify a different location for sockets (like `/tmp`)

## Actual Behavior ##

Debugger does not attach, socket name is too long.

## Environment information ##

**VSCode version**: 1.55.2
**C# Extension**: 1.23.11

<details><summary>Mono Information</summary>
        OmniSharp using built-in mono</details>
<details><summary>Dotnet Information</summary>
.NET SDK (reflecting any global.json):
 Version:   5.0.202
 Commit:    db7cc87d51

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  20.10
 OS Platform: Linux
 RID:         ubuntu.20.10-x64
 Base Path:   /usr/share/dotnet/sdk/5.0.202/

Host (useful for support):
  Version: 5.0.5
  Commit:  2f740adc14

.NET SDKs installed:
  3.1.408 [/usr/share/dotnet/sdk]
  5.0.202 [/usr/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.14 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.5 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.14 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.5 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET runtimes or SDKs:
[...]

PR #4637, opened 2021-07-01
Use temporary directory for debug sockets on NIX systems

valters-tomsons:
Fixes: https://github.com/OmniSharp/omnisharp-vscode/issues/4523

This PR stores debug sockets in `/tmp/` or `$TMPDIR` (if set) on Linux & OSX platforms.

### DIFF
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -546,7 +546,7 @@ class DebugEventListener {
         this._server = server;
         this._eventStream = eventStream;
         // NOTE: The max pipe name on OSX is fairly small, so this name shouldn't bee too long.
-        const pipeSuffix = "TestDebugEvents-" + process.pid;
+        const pipeSuffix = "T-" + process.pid;
         if (os.platform() === 'win32') {
             this._pipePath = "\\\\.\\pipe\\Microsoft.VSCode.CSharpExt." + pipeSuffix;
         } else {
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -545,14 +545,13 @@ class DebugEventListener {
         this._fileName = fileName;
         this._server = server;
         this._eventStream = eventStream;
-        const pipeSuffix = "TestDebugEvents-" + process.pid;
 
         if (os.platform() === 'win32') {
-            this._pipePath = "\\\\.\\pipe\\Microsoft.VSCode.CSharpExt." + pipeSuffix;
+            this._pipePath = "\\\\.\\pipe\\Microsoft.VSCode.CSharpExt.TestDebugEvents" + process.pid;
         }
         else {
             let tmpdir = utils.getUnixTempDirectory();
-            this._pipePath = path.join(tmpdir, "." + pipeSuffix);
+            this._pipePath = path.join(tmpdir, "ms-dotnettools.csharp-tde-" + process.pid);
         }
     }
 
--- a/src/features/dotnetTest.ts
+++ b/src/features/dotnetTest.ts
@@ -545,11 +545,18 @@ class DebugEventListener {
         this._fileName = fileName;
         this._server = server;
         this._eventStream = eventStream;
-        // NOTE: The max pipe name on OSX is fairly small, so this name shouldn't bee too long.
-        const pipeSuffix = "T-" + process.pid;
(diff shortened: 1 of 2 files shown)


gregg-miskelly:
I haven't investigated much, but this feels like the wrong fix to me -- in your exact case this is going to get you under the limit, but if your user alias would be a bit longer then it wouldn't help.

Isn't a better fix to move this under $TMP on OSX?

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=4637) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=4637) before we can accept your contribution.<br/><br/>:x: valters-tomsons [sign now](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=4637)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/omnisharp-vscode?pullRequest=4637) it.</sub>

valters-tomsons:
Agreed.

I have added functionality which overrides socket paths to `$TMPDIR` and `/tmp/` for Linux & OSX.

I apologize in advance if code is not clean, I have not worked with typescript much in the past.

src/features/dotnetTest.ts
```diff
@@ -545,11 +545,18 @@ class DebugEventListener {
...
-        const pipeSuffix = "T-" + process.pid;
-        if (os.platform() === 'win32') {
+        const pipeSuffix = "TestDebugEvents-" + process.pid;
+
+        let platform = os.platform();
+        if (platform === 'win32') {
             this._pipePath = "\\\\.\\pipe\\Microsoft.VSCode.CSharpExt." + pipeSuffix;
-        } else {
+        } 
+        else if (platform === 'linux' || platform === 'darwin') {
```
gregg-miskelly, line 554: You can remove the `if (platform === 'linux' || platform === 'darwin')` since you are already in an else block.

src/features/dotnetTest.ts
```diff
@@ -545,11 +545,18 @@ class DebugEventListener {
         this._fileName = fileName;
         this._server = server;
         this._eventStream = eventStream;
-        // NOTE: The max pipe name on OSX is fairly small, so this name shouldn't bee too long.
-        const pipeSuffix = "T-" + process.pid;
-        if (os.platform() === 'win32') {
+        const pipeSuffix = "TestDebugEvents-" + process.pid;
```
gregg-miskelly, line 548: I would suggest getting rid of `pipeSuffix` and using something different on non-Windows. Something like:

```
 if (platform === 'win32') {
            this._pipePath = "\\\\.\\pipe\\Microsoft.VSCode.CSharpExt.TestDebugEvents" + process.pid;
        } else {
            // NOTE: The max pipe name on *NIX is fairly small, so this name shouldn't be too long.
            let tmpdir = utils.getUnixTempDirectory();
            this._pipePath = path.join(tmpdir,"vscode-csharp-tde-" + process.pid); // NOTE: The .NET folks use much longer names than this on Linux, so I assume it is okay there, if this is too long on osx, maybe drop the 'vscode-'??
        }
```

gregg-miskelly (changes requested):
Otherwise LGTM. Thanks for doing this.

src/features/dotnetTest.ts
valters-tomsons: I opted to use `ms-dotnettools.csharp-tde` as similar is used for . A path like `/tmp/ms-dotnettools.csharp-tde-272809` has 37 chars. According to [this](https://serverfault.com/a/641388), there should be plenty of characters left to satisfy OSX's 104 char limit.

gregg-miskelly: To be clear, `$TMPDIR` on OSX (usually? always?) looks like "/var/folders/zz/yyc7kdwj53b927x9n0wngplh0000gq/T/" (49 characters on my mac).

But still, we should be okay. Sounds good to me.

gregg-miskelly (approved):

Issue #4640, opened 2021-07-08
Update to 1.58 and take advantage of the new completion labels


333fred:
https://code.visualstudio.com/updates/v1_58#_detailed-completion-item-labels

Should let us display more info for types of things and the namespace import completions will bring in. Related to https://github.com/microsoft/vscode/issues/39441.

333fred:
@david-driscoll these are also in the 3.17 draft LSP spec, so we could do similar for LSP in general when clients support it.

JoeRobich:
It is neat in the TypeScript completion that you can see the member declaration inlined.
![image](https://user-images.githubusercontent.com/611219/125006161-554bdb80-e012-11eb-9d6b-c2c0d281c503.png)

333fred:
> It is neat in the TypeScript completion that you can see the member declaration inlined.

For this, btw, we'll need new Roslyn APIs. Currently, that's a section in the documentation that's retrieved by the resolve step, and I don't want to try and resolve all documentation for every completion item on completion list pop.

PR #4648, opened 2021-07-11
Use new CompletionItem label API

333fred:
1.58 brought a new completion item API that allows us to specify a description for the completion label. This will give us an experience more akin to VS for import completion, where the namespace of the item is always displayed, right-aligned and slightly faded, in the completion list. Anything else that Roslyn adds to InlineDescription here will automatically show up, but import completion is the only thing that uses it for now.

Closes https://github.com/OmniSharp/omnisharp-vscode/issues/4640.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
         "@types/semver": "5.5.0",
         "@types/tmp": "0.0.33",
         "@types/unzipper": "^0.9.1",
-        "@types/vscode": "1.57.0",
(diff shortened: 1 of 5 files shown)


333fred:
This is what it looks like:
image

/cc @dibarbet, hopefully VS itself will take advantage of the 3.17 LSP API soon so you can replicate over there.

filipw (approved):
very cool! 👏

JoeRobich (approved):

Issue #4696, opened 2021-08-05
Cancelling a .Net Core attach debug request still opens launch.json file


Londovir:
## Issue Description ##

In the most recently updated version of VS Code, and running C# Extension 1.23.14, if my currently selected Debugging launch profile is a .Net Core attach, hitting F5 will spin up the command palette listing of processes I can attach to. However, if I hit ESC to cancel this, a dialog pops up from VS Code saying "No process was selected.", and providing two button options: "Open 'launch.json'" and "Cancel". If you hit "Cancel", the launch.json file is still opened, regardless, rather than silently cancelling the debug launch request.

It's possible this bug is related to [VS Code Issue #54213](https://github.com/microsoft/vscode/issues/54213), which involved what happens when certain return types are given to debug launch actions, but I am not sure.

## Steps to Reproduce ##

Select a .Net Core "attach" debugger profile (ie, type of "coreclr", request of "attach"). Hit F5 while in any code edit window. When the command palette opens, hit ESC to cancel the request. When the dialog opens, click the "Cancel" button.

## Expected Behavior ##

Clicking "Cancel" on a dialog box (as opposed to clicking "Open 'launch.json'" should cancel the operation, and not open the launch.json file.

## Actual Behavior ##

The launch.json file opens regardless of which choice is made in the "No process was selected." dialog.

### OmniSharp log ###
<details>Post the output from Output-->OmniSharp log here</details>

### C# log ###
<details>Post the output from Output-->C# here</details>

## Environment information ##

**VSCode version**: 1.58.2
**C# Extension**: 1.23.14

<details><summary>Dotnet Information</summary>
.NET SDK (reflecting any global.json):
 Version:   5.0.302
 Commit:    c005824e35

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19043
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.302\

Host (useful for support):
  Version: 5.0.8
  Commit:  35964c9215

.NET SDKs installed:
  2.1.816 [C:\Program Files\dotnet\sdk]
  3.1.201 [C:\Program Files\dotnet\sdk]
  3.1.411 [C:\Program Files\dotnet\sdk]
  5.0.104 [C:\Program Files\dotnet\sdk]
  5.0.205 [C:\Program Files\dotnet\sdk]
  5.0.301 [C:\Program Files\dotnet\sdk]
[...]

PR #4699, opened 2021-08-09
Don't create launch.json for no selected process

WardenGnaw:
This PR fixes resolveDebugConfigurationWithSubstitutedVariables to
return undefined when user does not select a process in the process selection
dialog.

This will ensure it does not switch to the `launch.json` tab. 

Related: https://github.com/OmniSharp/omnisharp-vscode/issues/4696

### DIFF
--- a/src/coreclr-debug/debugConfigurationProvider.ts
+++ b/src/coreclr-debug/debugConfigurationProvider.ts
@@ -55,7 +55,8 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
             }
             else
             {
-                throw new Error("No process was selected.");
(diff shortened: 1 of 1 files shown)


WardenGnaw:
@JoeRobich Are the PR builds broken? My changes should not affect any tests since it should only hit during the scenario when not selecting a process in the dialog during a attach debug.

JoeRobich:
> @JoeRobich Are the PR builds broken? My changes should not affect any tests since it should only hit during the scenario when not selecting a process in the dialog during a attach debug.

I'll investigate. Looks like a project selector change may be the cause.

WardenGnaw:
Thank you @JoeRobich!

PR #4780, opened 2021-09-20
Launch with first Folder or Solution target found

JoeRobich:
When individual projects were added to the Project selector we began showing it more often since there were often multiple targets found in the open folder. This change restores the previous behavior of opening the first Solution or Folder when there are multiple potential launch targets.


JoeRobich:
@nohwnd please take a look

PR #4781, opened 2021-09-21
Improvement to offline packaging

gregg-miskelly:
This PR makes various improvements to our offline package support in preparation for switching to that as our primary means of distribution.

More significant changes:
1. Added support for the missing supported platforms
2. Switched to using the same schema as VS Code for the names of platforms
3. Stop dropping the debugger install complete file. The install complete file is used to do some first run OS validation logic, so it should not have been dropped by the offline installer.
4. Lots of error handling improvements:
    * Add a try/catch around `doOfflinePackage` so we could know which package failed
    * Detect and fail if a package download fails
    * Detect and fail if the vsce process exits with a non-zero exit code
    * Detect and fail if the vsce process fails to create the file we expect it to
5. Fix offline package creation on Linux-x64. This would fail due to multiple files with the same name but different casing.
6. Fix incorrect check in debugger activation code

A few minor changes too:
1. Used more async APIs
2. Move the win32 check up to the start instead of failing each package
3. All the debugger references were to the '.NET Core' debugger and I dropped the 'Core' from this.

### DIFF
--- a/src/coreclr-debug/activate.ts
+++ b/src/coreclr-debug/activate.ts
@@ -22,7 +22,9 @@ export async function activate(thisExtension: vscode.Extension<CSharpExtensionEx
 
     if (!CoreClrDebugUtil.existsSync(_debugUtil.debugAdapterDir())) {
         let isValidArchitecture: boolean = await checkIsValidArchitecture(platformInformation, eventStream);
-        if (!isValidArchitecture) {
+        // If this is a valid architecture, we should have had a debugger, so warn if we didn't, otherwise
+        // a warning was already issued, so do nothing.
+        if (isValidArchitecture) {
             eventStream.post(new DebuggerPrerequisiteFailure("[ERROR]: C# Extension failed to install the debugger package."));
             showInstallErrorMessage(eventStream);
         }
@@ -41,7 +43,7 @@ async function checkIsValidArchitecture(platformInformation: PlatformInformation
     if (platformInformation) {
         if (platformInformation.isMacOS()) {
             if (platformInformation.architecture === "arm64") {
-                eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: arm64 macOS is not officially supported by the .NET Core debugger. You may experience unexpected issues when running in this configuration.`));
+                eventStream.post(new DebuggerPrerequisiteWarning(`[WARNING]: arm64 macOS is not officially supported by the .NET debugger. You may experience unexpected issues when running in this configuration.`));
                 return true;
             }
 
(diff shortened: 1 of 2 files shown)


src/coreclr-debug/activate.ts
```diff
@@ -22,7 +22,9 @@ export async function activate(thisExtension: vscode.Extension<CSharpExtensionEx
 
     if (!CoreClrDebugUtil.existsSync(_debugUtil.debugAdapterDir())) {
         let isValidArchitecture: boolean = await checkIsValidArchitecture(platformInformation, eventStream);
-        if (!isValidArchitecture) {
+        // If this is a valid architecture, we should have had a debugger, so warn if we didn't, otherwise
+        // a warning was already issued, so do nothing.
+        if (isValidArchitecture) {
```
WardenGnaw, line 27: Shouldn't this still be `!isValidArchitecture`? `checkIsValidArchitecture` returns true if we support it, else `false`.

WardenGnaw (changes requested):

src/coreclr-debug/activate.ts
gregg-miskelly: No. The goal of this code was to add a warning if this is a valid architecture, but we somehow didn't download a debugger.

gregg-miskelly: This was broken by Austen's ARM64 PR, and we didn't notice because the check is a bit counter intuiative (which is why I added a comment).

WardenGnaw (approved):

tasks/offlinePackagingTasks.ts
```diff
@@ -89,31 +101,52 @@ async function install(platformInfo: PlatformInformation, packageJSON: any) {
...
     let packagesToInstall = await getAbsolutePathPackagesToInstall(runTimeDependencies, platformInfo, codeExtensionPath);
     let provider = () => new NetworkSettings(undefined, undefined);
-    await downloadAndInstallPackages(packagesToInstall, provider, eventStream, isValidDownload);
-    await debugUtil.CoreClrDebugUtil.writeEmptyFile(debuggerUtil.installCompleteFilePath());
+    if (!(await downloadAndInstallPackages(packagesToInstall, provider, eventStream, isValidDownload)))
+    {
+        throw Error("Failed to download package.");
+    }
+
+    // The VSIX Format doesn't allow files that differ only my case. The Linux OmniSharp package had a lowercase version of these files ('.targets') targets from mono,
```
JoeRobich, line 112: nit: type
```suggestion
    // The VSIX Format doesn't allow files that differ only by case. The Linux OmniSharp package had a lowercase version of these files ('.targets') targets from mono,
```

JoeRobich:
CC: @vzarytovskii @nohwnd

PR #4809, opened 2021-10-11
Add option to filter diagnostics out of the FixAll list

JoeRobich:
Addresses https://github.com/OmniSharp/omnisharp-roslyn/issues/2231

Not all Code Action are desired when FixAll is applied. Allow users to configure a set of diagnostic ids to filter out of FixAll operations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -682,6 +682,14 @@
           "default": true,
           "description": "Suppress 'hidden' diagnostics (such as 'unnecessary using directives') from appearing in the editor or the Problems pane."
         },
+        "csharp.fixAll.filteredDiagnostics": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
+          "description": "Array of diagnostic ids which FixAll should ignore."
+        },
         "csharp.referencesCodeLens.enabled": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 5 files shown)


snebjorn:
Isn't the underlying issue that FixAll runs both Analyzers and Refactorings? Refactorings are a lot more invasive. For example a refactoring can convert a `for loop` into a `for each`.
A little more info https://github.com/JosefPihrt/Roslynator/blob/master/docs/AnalyzersVsRefactorings.md

Then again using `for each` over `for loop` could be a style choice. So the real issue is that refactoring IDE0050 can't be turned off via `.editorconfig`. Filtering out diagnostic ids would fix that :)

JoeRobich:
> Isn't the underlying issue that FixAll runs both Analyzers and Refactorings?

The fixall does not run refactorings. The issue is that the fixall services fixes all diagnostics including those reported as hidden. One common reason why a diagnostic would be reported as hidden was to get the behavior of a refactoring, that it is not visible in the editor unless the code actions are requested for the span that it covers. This is an implementation pattern so that an operation typically thought of as a refactoring then has the ability to be applied as a fixall.

So excluding hidden diagnostics should be done. I think additionally having the ability to set a minimum fixall severity (suggestion, warning, error) would beneficial. Although that would not have to be implemented at the same time.

JoeRobich:
Closing this PR. As stated above a better fix would be to enhance the GetFixAll request with a minimumSeverity that when not provided defaults to 'hidden'. The document fixall on save would then pass a minimum severity of 'info' although should probably be configurable.

PR #4872, opened 2021-11-03
Update to 1.37.17 deps

nohwnd:


nohwnd:
@JoeRobich please review in case some of the changes should be removed because they are too small. I tried to clean up the "partial" changes related to new debugger shipping approach.

JoeRobich (approved):
:shipit:

PR #4902, opened 2021-11-18
Improved Regex syntax highlighting

TechnologicalPizza:
Improves semantic highlighting of regex by utilizing Roslyn data that was previously unused.
This "breaks" themes that only style the `regexp` token because new specialized token types were added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3598,6 +3598,38 @@
       {
         "id": "xmlDocCommentText",
         "description": ""
+      },
+      {
+        "id": "regexComment",
+        "description": ""
+      },
+      {
+        "id": "regexCharacterClass",
+        "description": ""
+      },
+      {
+        "id": "regexAnchor",
+        "description": ""
+      },
+      {
+        "id": "regexQuantifier",
+        "description": ""
+      },
+      {
+        "id": "regexGrouping",
+        "description": ""
+      },
+      {
+        "id": "regexAlternation",
+        "description": ""
+      },
+      {
+        "id": "regexSelfEscapedCharacter",
+        "description": ""
+      },
+      {
+        "id": "regexOtherEscape",
+        "description": ""
       }
     ],
     "semanticTokenModifiers": [],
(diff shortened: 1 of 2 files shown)


333fred:
Could you add some screenshots to demonstrate what this looks like?

TechnologicalPizza:
Here are some examples within VSCode with default colors from Visual Studio 2019.

VSCode Light:
image

VS2022 Light:
image

VSCode Dark:
image

VS2022 Dark:
image

JoeRobich:
@TechnologicalPizza Thanks so much for adding these tokens. I think we are missing two things to make this complete. 

1. For the other token types, we maintain entries in the package.json where we map each token type to a textmate scope to aid in theming.

https://github.com/OmniSharp/omnisharp-vscode/blob/675319476db5713da3e45f6879b0eda44b1863a5/package.json#L3664-L3739

2. We also ship semantic enabled Visual Studio 2019 themes that could be updated to better match the VS 2019/2022 regex colors. (see https://github.com/OmniSharp/omnisharp-vscode/tree/master/themes)

If you feel up to working on these two tasks, please let me know. Otherwise, I am happy to take this PR as is.

TechnologicalPizza:
Seems like something interesting I could look at :eyes: 
I'm not really sure what name should go where but I'll investigate a bit first.

JoeRobich:
> I'm not really sure what name should go where but I'll investigate a bit first.

Some are already defined in the themes. Others will have to be a best guess unless you find something definitive.

https://github.com/OmniSharp/omnisharp-vscode/blob/master/themes/vs2019_light.json#L543-L583

TechnologicalPizza:
With commit cba7893 these are the default colors that were already in `vs2019_dark.json`🥳 
image

JoeRobich:
@TechnologicalPizza Sorry I haven't followed up. Were there any required changes to the Light theme?

TechnologicalPizza:
@JoeRobich I don't think so? The images I uploaded were made with stock themes. This PR doesn't change any themes either.
I couldn't really figure out why the Light theme was brighter though. It seemed to use the same color definitions as the Dark theme but it was somehow brightened by VSCode, unless I missed some specific regex stuff in some light theme somewhere.

JoeRobich (approved):

JoeRobich:
Thanks @TechnologicalPizza !

PR #4915, opened 2021-11-29
Send document buffer when semantically highlighting old document versions

JoeRobich:
Enables Semantic Highlighting by default.

Before:
image

After:
image

### DIFF
--- a/src/features/semanticTokensProvider.ts
+++ b/src/features/semanticTokensProvider.ts
@@ -198,6 +198,14 @@ export default class SemanticTokensProvider extends AbstractProvider implements 
         let req = createRequest<protocol.V2.SemanticHighlightRequest>(document, new vscode.Position(0, 0));
         req.Range = range;
 
+
+        // We need to include the document contents in our request when we are highlighting a version of the document other than the current version, such as in the Diff view.
+        const currentDocument = vscode.workspace.textDocuments.find(d => d.fileName === document.fileName);
+        const isCurrentVersion = currentDocument?.version === document.version;
+        if (!isCurrentVersion) {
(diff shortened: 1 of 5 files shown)


JoeRobich:
@nohwnd Please take a look

src/features/semanticTokensProvider.ts
```diff
@@ -187,7 +187,11 @@ export default class SemanticTokensProvider extends AbstractProvider implements
             return null;
         }
 
-        let req = createRequest<protocol.V2.SemanticHighlightRequest>(document, new vscode.Position(0, 0));
+        const currentDocument = vscode.workspace.textDocuments.find(d => d.fileName === document.fileName);
+        const isCurrentVersion = currentDocument?.version === document.version;
+
+        // We need to include the document contents in our request when we are highlighting a version of the document other than the current version, such as in the Diff view.
+        let req = createRequest<protocol.V2.SemanticHighlightRequest>(document, new vscode.Position(0, 0), /* includeBuffer: */ !isCurrentVersion);
```
333fred, line 194: How does this work? AFAICT, the service itself actually doesn't look at the text in the buffer: https://github.com/OmniSharp/omnisharp-roslyn/blob/ff21ac475c/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs#L27

IrishBruse:
+1 this diffs became ugly recently

src/features/semanticTokensProvider.ts
JoeRobich: @333fred Updated to use new omnisharp request property instead of setting the buffer.

PR #4932, opened 2021-12-07
Fix 'watch' Task

gregg-miskelly:
In .NET 6 (and I am not sure how long before?), the 'watch' task wasn't working --
1. It was missing the '--project' arg before the path to the project
2. It included extra args, which don't appear to have an effect, but which caused hot reload to not be enabled


JoeRobich (approved):
:shipit:

PR #4933, opened 2021-12-08
Add access modifiers to semantic highlighting

IrishBruse:
I added access modifiers semantic highlighting as my color theme for vscode would be able to move over to use only semantic highlighting and have textmate as a fall back instead of a weird hybrid between the two.

I didn't make any changes to the theme supplied as visual studio does not by default change anything.
These changes should be fully backward compatible as they are just modifiers onto the plainKeyword token.

This is just an example of what you can do with it with semantic highlighting.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -699,17 +699,17 @@
                   "items": {
                     "type": "string"
                   },
-                  "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                  "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
                   "default": []
                 },
                 "searchMicrosoftSymbolServer": {
                   "type": "boolean",
-                  "description": "If 'true' the Microsoft Symbol server (https\u200b://msdl.microsoft.com\u200b/download/symbols) is added to the symbols search path. If unspecified, this option defaults to 'false'.",
+                  "description": "If 'true' the Microsoft Symbol server (https​://msdl.microsoft.com​/download/symbols) is added to the symbols search path. If unspecified, this option defaults to 'false'.",
                   "default": false
                 },
                 "searchNuGetOrgSymbolServer": {
                   "type": "boolean",
-                  "description": "If 'true' the NuGet.org symbol server (https\u200b://symbols.nuget.org\u200b/download/symbols) is added to the symbols search path. If unspecified, this option defaults to 'false'.",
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=4933) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=4933) before we can accept your contribution.<br/><br/>:x: IrishBruse [sign now](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=4933)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/omnisharp-vscode?pullRequest=4933) it.</sub>

JoeRobich:
@IrishBruse Thanks for submitting this enhancement. So, there are different layers where this change could be implemented and I don't believe the C# extension is the right one.

1. It could be implemented at the lowest layer, in https://github.com/dotnet/roslyn, as part of the syntax highlighting. 
PRO:
- This would be very cheap to do. It works against the abstract syntax tree and the cost of looking up the token string and comparing it has already been paid as part of parsing the source file. 
- The most people get the benefit from this change. Since OmniSharp and many other tools all build on top of Roslyn for their highlighting the entire eco system would improve.
CON:
- The current Roslyn classification system doesn't align with LSP. In LSP you have spans with a single classification and many modifiers. However in Roslyn you have a span with a single classification. Roslyn tries to work around this limitation by producing multiple classified spans for the same token but it is hard to work with.

2. It could be implemented in the middle layer, in https://github.com/OmniSharp/omnisharp-roslyn, as par of the OmniSharp server.
PRO:
- This is still fairly cheap. It would similarly be implemented by processing the classifications produced by Roslyn. We would have access to the document's syntax tree and could avoid looking up the token string by inspecting the token for the Keyword spans.
- Many people will get the benefit of this change. O# powers extensions for VIM and other IDEs.
- O# is already the bridge between the Roslyn and the LSP classification styles so there isn't the mismatch that would exist by trying to implement this in Roslyn.

3. Lastly, it could be implemented here, in the C# extension. 
PRO:
- It is easy to implement.
CON:
- It is the most expensive option. We must look up the token string for each token and then compare it against our list of access modifiers.
- It has the smallest reach. It only benefits VS Code C# users.

So my recommendation would be to submit a change to the O# repo. You change would update this method with the addition logic of checking for Keyword classified spans and added the access modifier as necessary.

https://github.com/OmniSharp/omnisharp-roslyn/blob/8f42f7b95fdc20e0d9ec9dd2c2a65bdde19d17d9/src/OmniSharp.Roslyn.CSharp/Services/SemanticHighlight/SemanticHighlightService.cs#L80-L102

IrishBruse:
Oh ok thanks ill take a look into it thanks for the reply :)

PR #4948, opened 2021-12-10
Add short namespace snippet (.NET 6)

MRmlik12:
When C# 10 released, new cleaner way to implement `namespace` was [added](https://docs.microsoft.com/pl-pl/dotnet/csharp/fundamentals/types/namespaces). To activate this snippet type `snamespace` and press enter.

### DIFF
--- a/snippets/csharp.json
+++ b/snippets/csharp.json
@@ -302,12 +302,12 @@
         ],
         "description": "Namespace"
     },
-    "ShortNamespace": {
-        "prefix": "snamespace",
+    "FileScopedNamespace": {
+        "prefix": "fsnamespace",
         "body": [
             "namespace ${1:Name};"
         ],
-        "description": "Short namespace (.NET 6)"
+        "description": "File-scoped namespace"
     },
     "#if": {
         "prefix": "ifd",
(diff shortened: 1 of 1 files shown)


snippets/csharp.json
```diff
@@ -302,6 +302,13 @@
         ],
         "description": "Namespace"
     },
+    "ShortNamespace": {
+        "prefix": "snamespace",
+        "body": [
+            "namespace ${1:Name};"
+        ],
+        "description": "Short namespace (.NET 6)"
```
JoeRobich, line 310:
```suggestion
        "description": "File-scoped namespace"
```

snippets/csharp.json
```diff
@@ -302,6 +302,13 @@
         ],
         "description": "Namespace"
     },
+    "ShortNamespace": {
+        "prefix": "snamespace",
```
JoeRobich, line 306: I think we should go with either "fsnamespace" or "fsn". What do you think?

JoeRobich (approved):
@MRmlik12 Thanks for contributing this snippet. Just a few suggestions.

snippets/csharp.json
MRmlik12: I think, it was great idea. I change it

PR #4980, opened 2021-12-30
Explicitly install vscode-nls as a dependency

50Wliu:
[features/json/projectJSONContribution.ts](https://github.com/OmniSharp/omnisharp-vscode/blob/5366059b367d918a9d296485a067a8ea45da4f3b/src/features/json/projectJSONContribution.ts#L6) uses the [vscode-nls](https://www.npmjs.com/package/vscode-nls) package; however, up until now, it wasn't actually included as a direct dependency:
```prompt
$ npm ls vscode-nls
+-- microsoft.aspnetcore.razor.vscode@6.0.0-preview.5.21358.6
| `-- vscode-html-languageservice@2.1.7
|   `-- vscode-nls@3.2.5
`-- request-light@0.4.0
  `-- vscode-nls@4.1.2
```

It's probably a good idea to include it so that its version can be controlled :).

(By the way, what's up with the various zero-width breaking spaces in the package.json? npm keeps trying to convert them to...something else.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "vscode-debugprotocol": "1.33.0",
         "vscode-extension-telemetry": "0.1.6",
         "vscode-languageserver-protocol": "3.16.0",
+        "vscode-nls": "^5.0.0",
         "yauzl": "2.10.0"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


JoeRobich:
@50Wliu Thanks for this change.

> (By the way, what's up with the various zero-width breaking spaces in the package.json? npm keeps trying to convert them to...something else.)

The goal is to keep certain example urls from being turned into links when viewed in VS Code. It predates me, so perhaps there is a better way now.

JoeRobich (approved):

PR #4961, opened 2021-12-16
Upgrade to OmniSharp version 1.38.0

JoeRobich:

PR #4989, opened 2022-01-02
Fix OmnisharpDownloader tests

50Wliu:
I'm...quite confused how this didn't fail when testAssets was updated to 4.7.2 in #4961. But, regardless, it's failing now, so here's a fix.


50Wliu:
Ah, it's because it has been failing, but it seems like CI ignores unit test failures:
CI ignores unit test failures

vzarytovskii (approved):

JoeRobich (approved):

PR #5059, opened 2022-02-15
Only semantically highlight documents from uri.scheme 'file'

JoeRobich:
Semantic highlighting is still broken in diff view in some cases. We should adopt a similar approach as TypeScript where we do not semantically highlight those types of documents. We will do this by checking for the 'file' uri.scheme to ensure the file is a local file. Temporary files from GitLens or the VS Code Diff view will use a 'git' uri.scheme. This check is similar to checks in diagnosticProvider and other services.


JoeRobich:
@nohwnd please take a look

filipw (approved):
very good

PR #5088, opened 2022-03-04
Add AnalyzeOpenDocumentsOnly

neoGeneva:
This pull request is exposing the new `AnalyzeOpenDocumentsOnly` feature merged into omnisharp-roslyn via  OmniSharp/omnisharp-roslyn#2346

This pull request does two main things:
- Calls the `fileOpen` and `fileClose` endpoints
- Passes the new `AnalyzeOpenDocumentsOnly` flag as a startup parameter

And one additional weird thing I should point out:
- Calls `filesChanged` endpoint when an active editor changes. I found without this when switching between tabs the diagnostics would be out of date and I'd have to make a modification to force it to refresh the file, which was pretty anoying.

### DIFF
--- a/package.json
+++ b/package.json
@@ -998,7 +998,7 @@
         "omnisharp.analyzeOpenDocumentsOnly": {
           "type": "boolean",
           "default": false,
-          "description": "Will only run analyzers against open files when 'enableRoslynAnalyzers' is true"
+          "description": "Only run analyzers against open files when 'enableRoslynAnalyzers' is true"
         },
         "omnisharp.testRunSettings": {
           "type": [
--- a/package.json
+++ b/package.json
@@ -995,6 +995,11 @@
           "default": false,
           "description": "(EXPERIMENTAL) Enables support for resolving completion edits asynchronously. This can speed up time to show the completion list, particularly override and partial method completion lists, at the cost of slight delays after inserting a completion item. Most completion items will have no noticeable impact with this feature, but typing immediately after inserting an override or partial method completion, before the insert is completed, can have unpredictable results."
         },
+        "omnisharp.analyzeOpenDocumentsOnly": {
+          "type": "boolean",
+          "default": false,
+          "description": "Will only run analyzers against open files when 'enableRoslynAnalyzers' is true"
+        },
         "omnisharp.testRunSettings": {
           "type": [
             "string",
--- a/src/features/fileOpenCloseProvider.ts
+++ b/src/features/fileOpenCloseProvider.ts
@@ -29,7 +29,7 @@ class FileOpenCloseProvider implements IDisposable {
         this._disposable = new CompositeDisposable(this._diagnostics,
             vscode.workspace.onDidOpenTextDocument(this._onDocumentOpen, this),
             vscode.workspace.onDidCloseTextDocument(this._onDocumentClose, this),
-            vscode.window.onDidChangeActiveTextEditor(this._onActiveTextEdtiorChnage, this)
+            vscode.window.onDidChangeActiveTextEditor(this._onActiveTextEditorChnage, this)
         );
     }
 
@@ -49,7 +49,7 @@ class FileOpenCloseProvider implements IDisposable {
         await serverUtils.fileClose(this._server, { FileName: e.fileName });
     }
 
-    private async _onActiveTextEdtiorChnage(e: vscode.TextEditor) {
+    private async _onActiveTextEditorChnage(e: vscode.TextEditor) {
         if (shouldIgnoreDocument(e.document)) {
             return;
         }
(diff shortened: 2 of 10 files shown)


JoeRobich:
@neoGeneva Thanks for the contribution! Since an O# restart is required for this option to take effect, you can add the option name to the list in this file, so that users will be prompted to restart. https://github.com/OmniSharp/omnisharp-vscode/blob/090acf09c8bd9e6802e6c2566de87cbedbc07b5b/src/observers/OptionChangeObserver.ts#L15-L27

neoGeneva:
@JoeRobich Sweet, will do. It looks like `enableRoslynAnalyzers` is missing from that list too, should I add that as well?

JoeRobich:
> It looks like `enableRoslynAnalyzers` is missing from that list too, should I add that as well?

Ah, good catch. Please do.

package.json
```diff
@@ -995,6 +995,11 @@
           "default": false,
           "description": "(EXPERIMENTAL) Enables support for resolving completion edits asynchronously. This can speed up time to show the completion list, particularly override and partial method completion lists, at the cost of slight delays after inserting a completion item. Most completion items will have no noticeable impact with this feature, but typing immediately after inserting an override or partial method completion, before the insert is completed, can have unpredictable results."
         },
+        "omnisharp.analyzeOpenDocumentsOnly": {
+          "type": "boolean",
+          "default": false,
+          "description": "Will only run analyzers against open files when 'enableRoslynAnalyzers' is true"
```
333fred, line 1001:
```suggestion
          "description": "Only run analyzers against open files when 'enableRoslynAnalyzers' is true"
```

src/features/fileOpenCloseProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+    dispose = () => this._disposable.dispose();
+}
+
+function shouldIgnoreDocument(document: vscode.TextDocument) {
+    if (document.languageId !== 'csharp') {
+        return true;
+    }
+
+    if (document.uri.scheme !== 'file' &&
+        !isVirtualCSharpDocument(document)) {
```
333fred, line 69: This probably should check source-generated files as well?

neoGeneva: So it does check those files, I think that this just reads a little weird (it's the same logic that in `diagnosticsProvider.ts`). So if the scheme is file or it is a virtual document then it won't be ignored (though it's phrased as the inverse.)

333fred: I don't believe so. Virtual documents have to do with razor remapping, and I don't think we want to ignore roslyn analyzers for razor files. cc @NTaylorMullen for input.

neoGeneva: Yep, what I'm saying is that it isn't ignoring the razor virtual files:
image

So that check is the same as the one in `diagnosticsProvider.ts`, and the open / close is currently to control how analyzers run, so open/close should be ignoring the same files that `diagnosticsProvider.ts` is, right?

333fred: I think I misread this condition the first time. This is fine.

src/features/fileOpenCloseProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    private async _onDocumentClose(e: vscode.TextDocument) {
+        if (shouldIgnoreDocument(e)) {
+            return;
+        }
+
+        await serverUtils.fileClose(this._server, { FileName: e.fileName });
+    }
+
+    private async _onActiveTextEditorChnage(e: vscode.TextEditor) {
```
333fred, line 52:
```suggestion
    private async _onActiveTextEditorChange(e: vscode.TextEditor) {
```

src/features/fileOpenCloseProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+                let document = editor.document;
+
+                await this._onDocumentOpen(document);
+            }
+        }, 0);
+
+        this._disposable = new CompositeDisposable(this._diagnostics,
+            vscode.workspace.onDidOpenTextDocument(this._onDocumentOpen, this),
+            vscode.workspace.onDidCloseTextDocument(this._onDocumentClose, this),
+            vscode.window.onDidChangeActiveTextEditor(this._onActiveTextEditorChnage, this)
```
333fred, line 32:
```suggestion
            vscode.window.onDidChangeActiveTextEditor(this._onActiveTextEditorChange, this)
```

333fred:
Other than the typos, this is looking good.

neoGeneva:
Sweet, thanks @333fred, all fixed.

JoeRobich (approved):

neoGeneva:
Woohoo, awesome! Thanks everybody 🙂

JoeRobich:
@neoGeneva Thanks for the contribution! I spun up a prerelease for anyone interested in trying it out https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.24.2-beta1

lonix1:
@neoGeneva How do we use this new feature - is it on by default or is there a config option we must set?

Issue #5085, opened 2022-03-03
Higher ambient resource use after upgrade to 1.24.1


lonix1:
## Environment data
`dotnet --info` output: 6.0.102
VS Code version: 1.64.2
C# Extension version: 1.24.1
linux

## OmniSharp log
nothing interesting

## Steps to reproduce
start vscode

## Expected  behavior
Load all projects, then keep quiet.

## Actual behavior
After upgrading to v1.24.1, vscode loads all projects as usual, and much faster because it seems to be doing it on multiple threads.

But after loading, code/omnisharp doesn't stop working, and the machine's fan never shuts up (it slows down but I hear it). `htop` shows that `code` and `dotnet ... ms-dotnettools.csharp` processes are always doing something, even if I'm not using vscode.

I have no idea how to diagnose this, or whether "it's just me". I do know that it was not light on resources before, but it wasn't heavy either. Now it's quite heavy. `htop` shows activity all the time. Nothing has changed other than the upgrade to 1.24.1.

Has anyone else experienced this, and do you have any tips to share? If "it's just me" then please close the issue, but maybe if others document their experiences below we can discover if there's a problem?

lonix1:
Perhaps related:

I have endless streams of these errors:
````
Received response for /findusages but could not find request.
Received response for /v2/blockstructure but could not find request.
Received response for /v2/getcodeactions but could not find request.
Received response for /v2/getcodeactions but could not find request.
Received response for /quickinfo but could not find request.
Received response for /v2/getcodeactions but could not find request.
Received response for /completion/resolve but could not find request.
Received response for /v2/getcodeactions but could not find request.
Received response for /v2/getcodeactions but could not find request.
````

I used to have the same in older versions too, but in this new version the problem got much worse.

Also, hovering / Intellisense, autocompletion/ etc. are VERY slow, sometimes they do nothing. If I observer the log window, I see the above errors.

Sometimes I need to <kbd>CTRL+.</kbd> multiple times for the autotcomplete to kick in.

Hope that helps diagnose the problem? Any more info needed please let me know?

lonix1:
On second thought, I'm not sure the two are related. As I wrote in the original post, vscode + o# consume resources even when I'm working in a different window - if vscode is running, it's doing something heavy. It could be a naughty extension, but I didn't have this issue in the previous version.

valderrey:
I have the same problem that you. It's very slow after upgrade. If I disable the extension works fine

[...]

PR #5089, opened 2022-03-05
Debounce diagnostic requests

JoeRobich:
May contribute to #5085

### DIFF
--- a/src/features/diagnosticsProvider.ts
+++ b/src/features/diagnosticsProvider.ts
@@ -15,7 +15,7 @@ import { isVirtualCSharpDocument } from './virtualDocumentTracker';
 import { TextDocument } from '../vscodeAdapter';
 import OptionProvider from '../observers/OptionProvider';
 import { Subject, Subscription } from 'rxjs';
-import { throttleTime } from 'rxjs/operators';
+import { debounceTime } from 'rxjs/operators';
 import { DiagnosticStatus } from '../omnisharp/protocol';
 import { LanguageMiddlewareFeature } from '../omnisharp/LanguageMiddlewareFeature';
 
@@ -137,12 +137,12 @@ class DiagnosticsProvider extends AbstractSupport {
 
         this._subscriptions.push(this._validateCurrentDocumentPipe
             .asObservable()
-            .pipe(throttleTime(750))
+            .pipe(debounceTime(750))
             .subscribe(async x => await this._validateDocument(x)));
 
         this._subscriptions.push(this._validateAllPipe
             .asObservable()
-            .pipe(throttleTime(3000))
+            .pipe(debounceTime(3000))
             .subscribe(async () => {
                 if (this._validationAdvisor.shouldValidateAll()) {
                     await this._validateEntireWorkspace();
--- a/src/features/diagnosticsProvider.ts
+++ b/src/features/diagnosticsProvider.ts
@@ -122,7 +122,7 @@ class DiagnosticsProvider extends AbstractSupport {
     private _disposable: CompositeDisposable;
     private _diagnostics: vscode.DiagnosticCollection;
     private _validateCurrentDocumentPipe = new Subject<vscode.TextDocument>();
-    private _validateAllPipe = new Subject();
+    private _validateAllPipe = new Subject<string>();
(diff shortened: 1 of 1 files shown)


src/features/diagnosticsProvider.ts
```diff
@@ -280,41 +275,39 @@ class DiagnosticsProvider extends AbstractSupport {
             .filter(diagnosticInFile => diagnosticInFile !== undefined);
     }
 
-    private _validateEntireWorkspace(): NodeJS.Timeout {
-        return setTimeout(async () => {
-            let value = await serverUtils.codeCheck(this._server, { FileName: null }, new vscode.CancellationTokenSource().token);
+    private async _validateEntireWorkspace() {
+        let value = await serverUtils.codeCheck(this._server, { FileName: null }, new vscode.CancellationTokenSource().token);
```
JoeRobich, line 279: No need to use set Timeouts the debounce has already waited to execute this code.

src/features/diagnosticsProvider.ts
```diff
@@ -208,12 +206,13 @@ class DiagnosticsProvider extends AbstractSupport {
...
         if (!this._analyzersEnabled) {
-            this._validateAllPipe.next();
+            this._validateAllPipe.next("onDocumentOpenOrChange");
         }
     }
 
     private _onProjectAnalysis(event: protocol.ProjectDiagnosticStatus) {
-        if (event.Status == DiagnosticStatus.Ready) {
+        if (event.Status == DiagnosticStatus.Ready &&
+            event.ProjectFilePath === "(100 %)") {
```
JoeRobich, line 215: We are currently getting Ready status updates from O# with "(0 %)" regularly. This was causing us to continually request updated diagnostics.

src/features/diagnosticsProvider.ts
```diff
@@ -136,26 +136,24 @@ class DiagnosticsProvider extends AbstractSupport {
...
-            .pipe(throttleTime(750))
-            .subscribe(async x => await this._validateDocument(x)));
+            .pipe(debounceTime(750))
+            .subscribe(x => this._validateDocument(x)));
 
         this._subscriptions.push(this._validateAllPipe
-            .asObservable()
-            .pipe(throttleTime(3000))
-            .subscribe(async () => {
+            .pipe(debounceTime(3000))
```
JoeRobich, line 143: throttle and debounce work similarly. Throttle fires on first request then ignores requests for a set amount of time. Debounce will wait for incoming requests to settle down before firing. This is an important distinction because we would want to wait for a users to be finished editing a document before requesting diagnostics.

src/features/diagnosticsProvider.ts
```diff
@@ -122,7 +122,7 @@ class DiagnosticsProvider extends AbstractSupport {
     private _disposable: CompositeDisposable;
     private _diagnostics: vscode.DiagnosticCollection;
     private _validateCurrentDocumentPipe = new Subject<vscode.TextDocument>();
-    private _validateAllPipe = new Subject();
+    private _validateAllPipe = new Subject<string>();
```
JoeRobich, line 125: This was useful for debugging to see what event was causing all diagnostics to be requested.

JoeRobich:
Please take a look @nohwnd

333fred (approved):

nohwnd (approved):

lonix1:
The latest version is completely unusable as everything is so slow.

I assume this fix will only be released next month. Until then, is there a prerelease version we can use? We'll provide feedback if it solves the problem.

333fred:
> Until then, is there a prerelease version we can use?

@lonix1 if you set your o# version to latest, it should pick up the prerelease build with this change as soon as it finishes building.

filipw:
> > Until then, is there a prerelease version we can use?
> 
> @lonix1 if you set your o# version to latest, it should pick up the prerelease build with this change as soon as it finishes building.

This change will not show up when you set `"omnisharp.path":"latest"` since it's an extension-level change. It will require a new prerelease VSIX to be published here first https://github.com/OmniSharp/omnisharp-vscode/releases

333fred:
Oh right, I forgot which repo I was on 😅.

Issue #4251, opened 2020-11-30
"Go to type definition" and other navigations not working


macon:
## Issue Description ##
"Go to Type Definition" and other navigations are no longer working for C#.

I've tried reinstalling, disabling all extensions, trying vscode-insiders.

"Go to Definition" (F12) works fine but is disabled in the main menu (as are the other navigations). Setting various key bindings for "editor.action.goToTypeDefinition" don't help.

## Steps to Reproduce ##
On any C# project, place cursor on a variable of one of your types. Go to menu `Go` and see `Go to Type Definition` greyed out. Alternatively, right click the variable and I don't see `Go to Type Definition`. Also using the keybinding for `Go to Type Definition` does nothing.

## Expected Behavior ##
All these standard C# navigation features used to work.

## Actual Behavior ##
Many standard menu navigations are greyed-out/disabled, none of the keybindings work.

### OmniSharp log ###
<details>Starting OmniSharp server at 11/30/2020, 9:42:25 AM
    Target: /Users/mcotton/git/ob-certificate-service/Secret.Service.sln

OmniSharp server started.
    Path: /Users/mcotton/.vscode/extensions/ms-dotnettools.csharp-1.23.6/.omnisharp/1.37.3/run
    PID: 13453

[info]: OmniSharp.Stdio.Host
        Starting OmniSharp on MacOS 10.15.7 (x64)
[info]: OmniSharp.Services.DotNetCliService
        DotNetPath set to dotnet
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        Located 1 MSBuild instance(s)
            1: StandAlone 16.8.0 - "/Users/mcotton/.vscode/extensions/ms-dotnettools.csharp-1.23.6/.omnisharp/1.37.3/omnisharp/.msbuild/Current/Bin"
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        MSBUILD_EXE_PATH environment variable set to '/Users/mcotton/.vscode/extensions/ms-dotnettools.csharp-1.23.6/.omnisharp/1.37.3/omnisharp/.msbuild/Current/Bin/MSBuild.exe'
[info]: OmniSharp.MSBuild.Discovery.MSBuildLocator
        Registered MSBuild instance: StandAlone 16.8.0 - "/Users/mcotton/.vscode/extensions/ms-dotnettools.csharp-1.23.6/.omnisharp/1.37.3/omnisharp/.msbuild/Current/Bin"
[...]

PR #5094, opened 2022-03-06
Add GoToTypeDefinition provider

jtschuster:
PR https://github.com/OmniSharp/omnisharp-roslyn/pull/2315 add a `/gototypedefinition` endpoint. This PR adds a VS Code provider that utilizes the endpoint.

Fixes https://github.com/OmniSharp/omnisharp-vscode/issues/4251


333fred:
I'll try to take a look at this over the next couple of nights.

src/features/typeDefinitionProvider.ts
```diff
@@ -0,0 +1,98 @@
...
+        let req = <GoToTypeDefinitionRequest>createRequest(document, position);
+        req.WantMetadata = true;
+
+        const locations: Location[] = [];
+        try {
+            const goToTypeDefinitionResponse = await serverUtils.goToTypeDefinition(this._server, req, token);
+            // the defintion is in source
+            if (goToTypeDefinitionResponse && goToTypeDefinitionResponse.Definitions) {
+
+                for (const definition of goToTypeDefinitionResponse.Definitions) {
```
333fred, line 36: This logic should be sharable with https://github.com/OmniSharp/omnisharp-vscode/blob/090acf09c8/src/features/definitionProvider.ts#L36, pretty much in its entirety. Let's refactor these so they don't have to be maintained separately.

333fred (changes requested):

src/features/typeDefinitionProvider.ts
jtschuster: I've updated to use the same `CSharpDefinitionProvider` class to implement both `DefinitionProvider` and `TypeDefinitionProvider`, and put as much as I could into a shared function.

333fred (approved):
LGTM (commit 6). @JoeRobich for another review.

jtschuster:
Forgot I was using a local build of Omnisharp with the new endpoint for local testing. The "/gototypedefinition" endpoint was just added recently, so this will fail in CI until it uses the next Omnisharp release that includes it.

JoeRobich:
Thanks @jtschuster!

crawlchange:
F12 is still not working. "Go to Type Definition" or "Go to Definition" are still not working. Does OmniSharp do ANYTHING at all?

PR #5129, opened 2022-03-28
Update README.MD's .NET 6 SDK section

eriawan:
The .NET 6 SDK requires version 17.0.x of MSBuild instead of version 16.10 as stated in README.MD.

This is based on the requirements in .NET 6 SDK download page:
https://dotnet.microsoft.com/en-us/download/dotnet/6.0
image

Even since its Preview version, .NET 6 SDK requires at least MSBuild 17.0.x.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ You can also use the .NET 6 build of OmniSharp which runs on the .NET 6 SDK. See
 
 ### Note about using .NET 6 SDKs
 
-The .NET 6 SDK requires version 16.10 of MSBuild.
+The .NET 6 SDK requires version 17.0 of MSBuild.
 
 For Windows users who have Visual Studio installed, this means you will need to have Visual Studio 16.11 or newer installed.
 


JoeRobich:
@eriawan Thanks for the contribution.

So while there is no .NET 6 support built-in to VS2019, to determine what the minimum version of MSBuild which is required to use the SDK you have to look in the SDK itself. In the 6.0.100 SDK is a file named minimumMSBuildVersion and it contains the version number 16.10.0. This is also the minimum version required for 6.0.201.

image

This is not to say that the SDK won't warn you to use VS2022, but everything, as far as O# is concerned, should work as expected with MSBuild 16.10 installed. This has been proven out by Mono users, who do not have access to a 17.0 MSBuild, being able to develop with .NET 6. 

If you have a scenario where you are seeing errors using MSBuild 16.0 with the .NET 6 SDK, please open a new issue for us to investigate. For now I will close this PR.

Issue #5150, opened 2022-04-03
OmniSharp fails to load when the path provided contains spaces.


usrDottik:
## Environment data
`dotnet --info` output:
```
.NET SDK (reflecting any global.json):
 Version:   6.0.201     
 Commit:    ef40e6aa06  

Runtime Environment:    
 OS Name:     Windows   
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.201\

Host (useful for support):
  Version: 6.0.3
  Commit:  c24d9a9c91

.NET SDKs installed:
  6.0.200 [C:\Program Files\dotnet\sdk]
  6.0.201 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.3 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```
VS Code version:
VSCode 1.66.0

C# Extension version:
Current latest as 4/3/22 (UTC -0300), that is **v1.24.3**

## OmniSharp log
```
Starting OmniSharp server at 4/3/2022, 7:10:00 PM
    Target: c:\Users\Dottik\source\repos\Test Project\Test Project.sln

OmniSharp server started with .NET 6.0.201
.
    Path: c:\Users\Dottik\.vscode\extensions\ms-dotnettools.csharp-1.24.3-win32-x64\.omnisharp\1.38.2-net6.0\OmniSharp.dll
    PID: 7896

System.ArgumentException: OmniSharp only supports being launched with a directory path or a path to a solution (.sln, .slnf) file. (Parameter 'path')
[...]

PR #5154, opened 2022-04-04
Quote arguments containing spaces when launching OmniSharp.

JoeRobich:
Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/5150

Tested locally with a Folder name containing a space. Failed to launch prior to this change and opens as expected with the change.

### DIFF
--- a/src/omnisharp/launcher.ts
+++ b/src/omnisharp/launcher.ts
@@ -3,7 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { spawn, ChildProcess } from 'child_process';
+import { spawn } from 'cross-spawn';
+import { ChildProcess } from 'child_process';
 
 import { PlatformInformation } from '../platform';
 import * as path from 'path';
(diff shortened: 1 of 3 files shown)


src/omnisharp/launcher.ts
```diff
@@ -339,10 +339,16 @@ function getConfigurationValue(globalConfig: vscode.WorkspaceConfiguration, csha
 }
 
 async function launchDotnet(launchInfo: LaunchInfo, cwd: string, args: string[], platformInfo: PlatformInformation, options: Options, dotnetResolver: IHostExecutableResolver): Promise<LaunchResult> {
+    function escapeIfNeeded(arg: string) {
+        const hasSpaceWithoutQuotes = /^[^"].* .*[^"]/;
```
333fred, line 343: What does this regex actually mean? A space not surrounded by quotes?

JoeRobich: yeah stole it from the launchWindows method. Could just check startsWith('"') and contains(' ') instead.

333fred: I am slightly worried about quotes in the middle of arguments: should I be? Or will that break things elsewhere?

333fred: > Could just check startsWith('"')

Make sure you account for leading/trailing spaces that don't matter to the command.

nohwnd:
https://github.com/nodejs/node/issues/7367#issuecomment-315210235 Recommends using cross-spawn package which implements this correctly without the need to manually escape, or at least that is what they say. Might be worth a try.

nohwnd:
https://www.npmjs.com/package/cross-spawn

JoeRobich:
Updated to use cross-spawn package. Need to verify Windows launch behavior.

333fred (approved):
LGTM as long as we test the behavior.

JoeRobich:
Confirmed works with spaces in workspace path on windows with both "omnisharp.useModernNet": "true" & "false"

PR #5172, opened 2022-04-15
Fix .net6 OmniSharp acquisition on Linux arm64

JoeRobich:
The arm64 .net6.0 OmniSharp package was misconfigured as the Full Framework package.


filipw (approved):
oops

PR #5177, opened 2022-04-19
Add support for InlayHint.TextEdits

JoeRobich:
This cannot merge until we take a new release of O# that includes https://github.com/OmniSharp/omnisharp-roslyn/pull/2385.

This PR adds support for InlayHint.TextEdits which allows the hint to be double clicked in VS Code. The TextEdits are then applied to the file removing the need for that particular hint.

https://user-images.githubusercontent.com/611219/164500269-06b16a14-8fdd-411b-bb44-45e6a5ecce9b.mov


src/omnisharp/typeConversion.ts
```diff
@@ -62,7 +62,28 @@ export function toVSCodePosition(point: protocol.V2.Point): vscode.Position {
     return new vscode.Position(point.Line, point.Column);
 }
 
-export function createRequest<T extends protocol.Request>(document: vscode.TextDocument, where: vscode.Position, includeBuffer: boolean = false): T {
+export function toVSCodeTextEdit(textChange: protocol.LinePositionSpanTextChange): vscode.TextEdit {
+    return new vscode.TextEdit(toVSCodeRange(textChange), textChange.NewText);
+
+    function toVSCodeRange(textChange: protocol.LinePositionSpanTextChange): vscode.Range {
```
333fred, line 68: Consider just using `toRange2`.

333fred:
What are the errors in the example gif? Import errors? If so, is there any way we can have the needed imports added to the text edits?

src/omnisharp/typeConversion.ts
JoeRobich: not sure how I could have missed such a well named helper. Will update.

JoeRobich:
> What are the errors in the example gif? Import errors? If so, is there any way we can have the needed imports added to the text edits?

In hindsight it was a bad idea to use dotnet/format as an example because the .editorconfig requires the use of `var`. It isn't import related, although now I wonder if I could get that to fail.

JoeRobich:
@333fred Opened https://github.com/dotnet/roslyn/issues/61438 to track adding imports when necessary.

333fred (approved):

Issue #5120, opened 2022-03-19
Announcement: Planned removal of the included Mono & MSBuild tools


JoeRobich:
**Background:**

The included Mono and MSBuild tools were added to support easy development of C# using .NET Core. We have frequently updated these build tools to match the current shipping SDK tooling. 

On Windows we did not need to ship a copy of Mono because we could rely on the .NET Framework being installed. For MacOS and Linux we shipped a minimal Mono that would support running MSBuild and OmniSharp. With these tools, the .NET SDK installed, and the dotnet CLI discoverable from the environment PATH, developers would have good support when developing against the newer SDK-style projects. 

There was some support for Full Framework projects with this configuration, but that was considered a bonus. Our most common Full Framework projects (Unity) were unsupported with this configuration because the .NET 4.7.1 targeting pack was not included.

In the future .NET Framework OmniSharp builds will not ship with Mono or the MSBuild tooling (See announcement [omnisharp-roslyn#2339](https://github.com/OmniSharp/omnisharp-roslyn/issues/2339), See PR [omnisharp-roslyn#2360](https://github.com/OmniSharp/omnisharp-roslyn/pull/2360)).

**Proposal:**

To ensure that the C# extension remains usable out of the box for .NET SDK projects, we will be changing the default value of `omnisharp.useModernNet` to `true`. This means by default OmniSharp will run on the .NET SDK and .NET Framework projects will likely not load properly. It will require the .NET 6 SDK be installed and the dotnet CLI to be discoverable from the environment PATH.

Users who need .NET Framework support will have to set `omnisharp.useModernNet` to `false` to download the .NET Framework build of OmniSharp.  Linux and MacOS users will need to have a complete system install of Mono including MSBuild ([Download](https://www.mono-project.com/download/stable/)). Windows users will need to have the Visual Studio MSBuild Tools or a full Visual Studio installed ([Download](https://visualstudio.microsoft.com/downloads/)).

<img width="800" alt="Screenshot of the Visual Studio download page." src="https://user-images.githubusercontent.com/611219/159102533-19525e64-d06c-4d85-8c7c-ed0b716aef2a.png">

The `omnisharp.useGlobalMono` setting will be removed as there is no longer an included Mono and we will always use the system installation of Mono.

HaraldKorneliussen:
I guess the removal of bundled msbuild is the reason omnisharp suddenly fails on my projects with

[...]

PR #5186, opened 2022-04-24
Low-hanging nullable fruit

50Wliu:
There's a bunch of low-hanging fruit to get variables properly classified as nullable (and null-checked).

This pass gets most of src/features strict-clean, with the exception of server.ts and anything that depends on IHostExecutableResolver.

I'm hoping most of the changes should be self-explanatory, let me know if anything looks suspicious. Any changes to functionality are unintentional.

Also let me know if you'd prefer not to receive these types of PRs, because I get that they change a lot of files and reviewing is probably a bit of a small headache.

### DIFF
--- a/src/omnisharp/OmnisharpManager.ts
+++ b/src/omnisharp/OmnisharpManager.ts
@@ -22,10 +22,10 @@ export class OmnisharpManager {
         private platformInfo: PlatformInformation) {
     }
 
-    public async GetOmniSharpLaunchInfo(defaultOmnisharpVersion: string, omnisharpPath: string, useFramework: boolean, serverUrl: string, latestVersionFileServerPath: string, installPath: string, extensionPath: string): Promise<LaunchInfo> {
-        if (!omnisharpPath) {
+    public async GetOmniSharpLaunchInfo(defaultOmnisharpVersion: string, omnisharpPath: string | undefined, useFramework: boolean, serverUrl: string, latestVersionFileServerPath: string, installPath: string, extensionPath: string): Promise<LaunchInfo> {
+        if (omnisharpPath === undefined) {
             // If omnisharpPath was not specified, return the default path.
-            let basePath = path.resolve(extensionPath, '.omnisharp', defaultOmnisharpVersion + (useFramework ? '' : `-net${modernNetVersion}`));
+            const basePath = path.resolve(extensionPath, '.omnisharp', defaultOmnisharpVersion + (useFramework ? '' : `-net${modernNetVersion}`));
             return this.GetLaunchInfo(this.platformInfo, useFramework, basePath);
         }
 
--- a/src/omnisharp/options.ts
+++ b/src/omnisharp/options.ts
@@ -7,7 +7,7 @@ import { vscode, WorkspaceConfiguration } from '../vscodeAdapter';
 
 export class Options {
     constructor(
-        public path: string,
+        public path: string | undefined,
         public useModernNet: boolean,
         public useGlobalMono: string,
         public waitForDebugger: boolean,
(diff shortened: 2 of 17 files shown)


src/omnisharp/OmnisharpManager.ts
```diff
@@ -22,10 +22,10 @@ export class OmnisharpManager {
         private platformInfo: PlatformInformation) {
     }
 
-    public async GetOmniSharpLaunchInfo(defaultOmnisharpVersion: string, omnisharpPath: string, useFramework: boolean, serverUrl: string, latestVersionFileServerPath: string, installPath: string, extensionPath: string): Promise<LaunchInfo> {
-        if (!omnisharpPath) {
+    public async GetOmniSharpLaunchInfo(defaultOmnisharpVersion: string, omnisharpPath: string | undefined, useFramework: boolean, serverUrl: string, latestVersionFileServerPath: string, installPath: string, extensionPath: string): Promise<LaunchInfo> {
+        if (omnisharpPath === undefined) {
```
JoeRobich, line 26:
```suggestion
        if (omnisharpPath === undefined || omnisharpPath.length === 0) {
```

JoeRobich (approved):
Thanks for these changes. Enabling more of the TS compilers strict checking is certainly a goal for us. As well as moving from tslint to eslint.

There are a few failing unit tests that have checks like `expect(options.path).to.be.null` that will need to be updated.

test/unitTests/OmnisharpManager.test.ts
```diff
@@ -136,7 +136,7 @@ suite(OmnisharpManager.name, () => {
             });
 
             test('Returns the default path if the omnisharp path is not set', async () => {
-                let launchInfo = await manager.GetOmniSharpLaunchInfo(defaultVersion, "", useFramework, server.baseUrl, latestfilePath, installPath, extensionPath);
+                const launchInfo = await manager.GetOmniSharpLaunchInfo(defaultVersion, undefined, useFramework, server.baseUrl, latestfilePath, installPath, extensionPath);
```
JoeRobich, line 139: I think we should add an additional test here for cases when is set to empty string. Users are very likely to leave `omnisharp.path: ""` in their settings.json when trying to switch from "latest" back to release and the empty string case should be treated like `undefined`.

src/omnisharp/OmnisharpManager.ts
50Wliu: Good catch, you're right. The integration tests also exposed a related configuration problem: the defaults for string-related options are all set to `null` at the moment. Going to push up a test commit so I can run the tests using a Codespace that actually has .NET installed.

50Wliu: Right, it's because VS Code coerces undefined values for known settings into preset default values: https://github.com/microsoft/vscode/blob/6f26fa19db7e4cd2af57f52fd7009e715c3d7aab/src/vs/platform/configuration/common/configurationRegistry.ts#L578

Of course, this isn't documented anywhere I could find in their API documentation. :)

src/omnisharp/options.ts
```diff
@@ -207,53 +209,44 @@ export class Options {
...
+        function toUseGlobalMonoValue(value?: boolean): string | undefined {
+            if (value === undefined) {
+                return undefined;
+            }
+
             // True means 'always' and false means 'auto'.
             return value ? "always" : "auto";
         }
 
-        if (omnisharpConfig.has('useGlobalMono')) {
```
50Wliu, line 242: I'm not confident this fallback logic ever worked...I'm running the `singleCsproj` integration test suite right now which doesn't have `useGlobalMono` defined, but this still returns true (with `get` returning "auto", the default).

I suppose the upside is this whole section is going away soon, so I don't suppose it matters too much.

50Wliu (commented):
Ready for review again. Changed all the string-related settings to default to the empty string rather than having to deal with TypeScript warning that it _could_ be undefined. The only time it could be undefined is if the config is removed from package.json, which I think is highly unlikely without corresponding code changes...

src/omnisharp/options.ts
JoeRobich: This bit of code will be removed prior to the next release to address https://github.com/OmniSharp/omnisharp-vscode/issues/5120

JoeRobich (approved):
Latest round of changes look good. Thanks!

PR #5199, opened 2022-05-02
Clear nullability warnings in server/omnisharp.ts

50Wliu:
Mostly, this adds a mini state machine to server.ts that controls which variables can be accessed.

I believe there's the following behavior changes to watch out for:
* `_makeRequest` now errors out if a request is sent when the server isn't running. (This would throw an uncaught exception anyway because of `this._serverProcess` being undefined, now it's just explicit.)
* `stop` now short-circuits if the server is already stopped.
* `autoStart` now calls `_start` directly rather than going through `restart`. Given how it's called autoStart, I feel like this makes sense.

Potential future improvements:
* Make launchTarget non-nullable. This is more difficult because `o.restart` currently tries to restart by implicitly using the server's last-known launch target. However, it could be possible if we...
* Make the state machine into actual classes that inherit from each other. This would provide even stronger safety guarantees because now we can hide entire methods instead of just properties. And this enables the first bullet point by statefully exposing the last launch target for `o.restart` to target.

My naming scheme for ServerState and State is terrible. Please advise on better names for them :).


src/omnisharp/server.ts
```diff
@@ -769,7 +775,11 @@ export class OmniSharpServer {
         }
     }
 
-    private _makeRequest(request: Request) {
+    private _makeRequest(request: Request): number {
+        if (this._state.status !== ServerState.Started) {
+            throw new Error("Tried to make a request when the OmniSharp server wasn't running");
```
50Wliu, line 780: Should this post to the EventStream instead?

JoeRobich (approved):
Thanks!

src/omnisharp/server.ts
JoeRobich: I could see it emitting an event. `OmnisharpFailure` isn't what we want because it focuses the OmniSharp Log in the Output pane. It likely needs a new Event type which will be logged by the `OmnisharpDebugModeLoggerObserver`. This could certainly come as a follow up.

50Wliu: Okay, I looked into what it would take to make this emit an event instead. The event emitting itself is easy; the problem is that `_makeRequest` needs to return an ID that's used by RequestQueue: https://github.com/OmniSharp/omnisharp-vscode/blob/f1f38b774243526a798a16922d3b8c2f2008a84b/src/omnisharp/requestQueue.ts#L95-L99

So, two options:
1. Keep the `throw` to stop execution.
2. Change `_makeRequest` to return `number | undefined`, emit the event, and don't add the request to the queue if it's undefined.

Issue #4096, opened 2020-09-30
Make csharp.listRemoteProcess support images without ps


gregg-miskelly:
## Issue Description ##
The `csharp.listRemoteProcess` command would be helpful in Docker extension scenarios, because our debugging all happens through a pipe program (namely, `docker exec`). However, the implementation of it doesn't work in the Debian .NET base image because `ps` isn't present.

## Steps to Reproduce ##
1. Scaffold a .NET (Core) app
1. With the C# extension, add the build and debug assets
1. With the Docker extension, do `Add Dockerfiles...`. Say "Yes" to Compose files.
1. Right click `docker-compose.yaml` and do Compose Up.
1. Go to the Debug tab. Add a debug configuration for "Docker .NET Core Attach (Preview)". Add `"processId": "${command:csharp.listRemoteProcess}"` to that config. (Ordinarily this isn't needed since only one `dotnet` process exists in the container, but in some scenarios--like `dotnet watch`--more than one exists)
1. F5 that config

## Expected Behavior ##
Remote processes are listed to pick from

## Actual Behavior ##
Error popup reading "See remote-attach output".

On a Debian image:
```
Executing: docker exec -i netcore2020_netcore2020_1 sh -s < "c:\Users\bwater\.vscode\extensions\ms-dotnettools.csharp-1.23.1\scripts\remoteProcessPickerScript"
Linux
stderr: sh: 1: ps: not found
Error Message: Command failed: docker exec -i netcore2020_netcore2020_1 sh -s < "c:\Users\bwater\.vscode\extensions\ms-dotnettools.csharp-1.23.1\scripts\remoteProcessPickerScript"
sh: 1: ps: not found
```

Changing the implementation of this to avoid `ps`, or adding an alternative command that uses OS APIs*, would enable this scenario.

*Note: VSDBG is present in the container so scripts are not the only option

## Work around ##
Modify your docker file to add the `procps` package.

ssaraswati:
This is implemented now in 1.25.0 and should work in both Alpine and Debian based Docker images

JoeRobich:
Thanks for confirming @ssaraswati. Closing this issue.

PR #5207, opened 2022-05-13
Add a fallback for ps in remoteProcessPickerScript 

ssaraswati:
Addresses #4096 by providing a fallback to ps when ps is not available or when ps doesn't support all options.

This solves connecting to Alpine hosts that don't support the flags output option and Ubuntu docker images that don't have ps installed by default.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=5207) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=5207) before we can accept your contribution.<br/><br/>:x: ssaraswati [sign now](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=5207)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/omnisharp-vscode?pullRequest=5207) it.</sub>

scripts/remoteProcessPickerScript
```diff
@@ -1 +1,24 @@
...
+        # Try to mimic the output of ps -axww -o pid=,flags=,comm=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,args= from /proc/$pid/ data
+        echo "        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        for pid in `cd /proc && ls -d [0-9]*`
+            do {
+                if [ -e /proc/$pid/stat ]
+                then
+                    flags=`awk '{print $9}' /proc/$pid/stat`;
+                    command=`awk '{print $2}' /proc/$pid/stat | tr -d '()'`
+                    args=`xargs -0 < /proc/$pid/cmdline | awk '{print $2}'`;
+                    printf "%5s %.1s %-49s %-27s \n\r" $pid $flags $command $args
```
JoeRobich, line 18: Could you add comments for what each of these lines is accomplishing. I am not an expert with linux utils and it would be helpful when referencing it in the future.

JoeRobich:
This seems similar to a script that VS Code ships with - https://github.com/microsoft/vscode/blob/main/src/vs/base/node/ps.sh

scripts/remoteProcessPickerScript
ssaraswati: Thanks for feedback, I have updated it with explanation of each line

ssaraswati:
> This seems similar to a script that VS Code ships with - https://github.com/microsoft/vscode/blob/main/src/vs/base/node/ps.sh

Yes very similar from the looks of it, only difference being the columns its outputting really, I am not familiar with vs code extension development, would it be possible to use their script instead of maintaining this?

JoeRobich (approved):
Thanks!

JoeRobich:
> would it be possible to use their script instead of maintaining this?

It doesn't appear that they expose this as an API and I don't think we would have access to it.

PR #5221, opened 2022-05-23
Use MSBuild info when computing program to launch

### LINKED ISSUE #3535: launch.json is generated with the wrong `program` path in projects that define `RuntimeIdentifier`
**Edit by @gregg-miskelly: See [this comment](https://github.com/OmniSharp/omnisharp-vscode/issues/3535#issuecomment-854172590) for more detailed steps to reproduce this problem.**

When my Project has a <RuntimeIdentifier> tag in my .csproj file the .launch.json of VS Code don't recognize the correctly path of the .dll of project.

For example:
The correct path is: "${workspaceFolder}/MyProject/bin/Debug/netcoreapp3.1/win-x86/MyProject.dll"

But the .launch.json has this path
"${workspaceFolder}/MyProject/bin/Debug/netcoreapp3.1/MyProject.dll"

I need to change this manually to work

50Wliu:
While working on nullability improvements in the root src folder, I came across this bit of code that didn't have an easy nullability fix. After investigating a bit, I realized it was attempting to form the program path manually, when MSBuild already supplies the correct paths.
I left the whitespace changes in because that seems like good code hygiene to cleanup. Actual changes are only L164-165 in assets.ts and all changes in protocol.ts.

### Details
`computeProgramPath` takes in a project (via `this.startupProject`) and returns the path to the entry assembly DLL.
The problem is that it tries to do this manually with a bit of path manipulation and assumes the output path is always the default `<project>/bin/<config>/<tfm>/<assembly>.dll`. **This is a bad assumption** as the output path can be changed via the MSBuild `OutputPath` property. omnisharp-roslyn itself does this! https://github.com/OmniSharp/omnisharp-roslyn/blob/c6f1e848202d7fff16d52b81496b6980b436952d/build/Settings.props#L20

Luckily, omnisharp-roslyn already computes the absolute assembly path for us via `TargetPath`. So just...use that. Nice. And add some documentation to MSBuildProject to help others along in the future.
TargetPath contains correct path

(It would be good to add an integration test for this. And I don't know how the Configuration is determined, since I don't have VS installed on this laptop to try toggling it in VS and seeing if it changes in VS Code.)

Fixes #3535

### DIFF
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -162,6 +162,11 @@ export class AssetGenerator {
         }
 
         const relativeTargetPath = path.relative(this.workspaceFolder.uri.fsPath, this.startupProject.TargetPath);
+        if (relativeTargetPath === this.startupProject.TargetPath) {
+            // This can happen if, for example, the workspace folder and the target path
+            // are on completely different drives.
+            return this.startupProject.TargetPath;
+        }
         return path.join('${workspaceFolder}', relativeTargetPath);
     }
 
@@ -170,6 +175,8 @@ export class AssetGenerator {
             throw new Error("Startup project not set");
         }
 
+        // Startup project will always be a child of the workspace folder,
+        // so the special check above isn't necessary.
         const relativeProjectPath = path.relative(this.workspaceFolder.uri.fsPath, this.startupProject.Path);
         return path.join('${workspaceFolder}', path.dirname(relativeProjectPath));
     }
--- a/src/assets.ts
+++ b/src/assets.ts
@@ -5,7 +5,7 @@
 
 import * as fs from 'fs-extra';
 import * as jsonc from 'jsonc-parser';
-import { FormattingOptions, ModificationOptions } from 'jsonc-parser'; 
+import { FormattingOptions, ModificationOptions } from 'jsonc-parser';
 import * as os from 'os';
 import * as path from 'path';
 import * as protocol from './omnisharp/protocol';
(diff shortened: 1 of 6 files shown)


src/assets.ts
```diff
@@ -161,12 +161,8 @@ export class AssetGenerator {
             throw new Error("Startup project not set");
         }
 
-        const startupProjectDir = path.dirname(this.startupProject.Path);
-        const relativeProjectDir = path.join('${workspaceFolder}', path.relative(this.workspaceFolder.uri.fsPath, startupProjectDir));
-        const configurationName = 'Debug';
```
50Wliu, line 166: Not sure how MSBuild determines what the current Configuration is...but also, Debug might not exist since that's user-editable. So might be best to just stick with TargetPath because that's guaranteed to exist.

JoeRobich: The concern that I have is that this is generating the tasks and launch json files and they need to be in sync such that we will be launching the app that was built by the tasks.json build task. 

We aren't passing a specific configuration argument in the build task when building prior to launch or in O# when loading projects, so this might be a safe assumption, but would like a second opinion. @gregg-miskelly what do you think?

src/assets.ts
```diff
@@ -156,27 +156,22 @@ export class AssetGenerator {
...
-        const targetFramework = protocol.findNetCoreTargetFramework(this.startupProject);
-        const result = path.join(relativeProjectDir, `bin/${configurationName}/${targetFramework.ShortName}/${this.startupProject.AssemblyName}.dll`);
-        return result;
+        const relativeTargetPath = path.relative(this.workspaceFolder.uri.fsPath, this.startupProject.TargetPath);
```
gregg-miskelly, line 164: >const relativeTargetPath [](http://example.com/codeflow?start=8&length=24)

Feels like this should really do something like:

```
let targetPath = this.startupProject.TargetPath;
let workspaceRoot = this.workspaceFolder.uri.fsPath;
// TODO: normalize both paths to have the correct directory separator character (forward or backslash), and ensure that workspaceRoot ends in a directory separator
if (targetPath.startsWith(workspaceRoot))
{
   return path.join('${workspaceFolder}', targetPath.Substring(workspaceRoot.Length));
}
else
{
    return targetPath;
}
```

Otherwise I have no idea what this would do if the user modified their output directory to point at a different drive.

src/assets.ts
gregg-miskelly: I am assuming that neither OmniSharp or the default tasks.json task will specify a configuration. So it will use whatever the msbuild project/props/targets files decide is the default. It seems like in most cases the build and OmniSharp will wind up coming to the same default. I am sure there is some case where they could work out something different, but its not like the old code was exactly bullet proof -- it would have come up with the wrong answer any time that the user overwrote nearly any msbuild property.

gregg-miskelly (approved):

JoeRobich (approved):

PR #5236, opened 2022-05-27
Clear nullability warnings in src

50Wliu:
I'll do my best to call out suspicious changes here:
* Extension activation now throws if on an unsupported platform
* ID_LIKE removed from LinuxDistribution FromReleaseInfo

I think that's actually it? The rest of the changes look more-or-less straightforward.

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -19,8 +19,7 @@ const unknown = 'unknown';
 export class LinuxDistribution {
     public constructor(
         public name: string,
-        public version: string,
-        public idLike?: string[]) { }
+        public version: string) { }
 
     public static async GetCurrent(): Promise<LinuxDistribution> {
         // Try /etc/os-release and fallback to /usr/lib/os-release per the synopsis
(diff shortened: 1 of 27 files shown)


src/platform.ts
```diff
@@ -134,100 +129,61 @@ export class PlatformInformation {
...
-    private static async GetIsMusl() {
-        let output;
+    private static async GetIsMusl(): Promise<boolean> {
         try {
-            output = await util.execChildProcess('ldd --version');
+            const output = await util.execChildProcess('ldd --version');
+            return output.includes('musl');
         } catch (err) {
-            output = err.message;
+            return false;
```
JoeRobich, line 185: The goal was to replicate this bit of code from the [dotnet-install.sh](https://github.com/dotnet/install-scripts/blob/3c6cc06b5054af7afcc5cbb4e659b551cc03e236/src/dotnet-install.sh#L187-L189) which treats stderr the same as stdout.

```sh
is_musl_based_distro() {
  (ldd --version 2>&1 || true) | grep -q musl
}
```

```suggestion
            return err.message.includes('musl');
```

JoeRobich (approved):
Thanks @50Wliu! It is great to see the progress you are making.

src/platform.ts
50Wliu: Thanks for the context, took suggestion & updated it to work with err typed as unknown.

Issue #5227, opened 2022-05-25
Show error message or set default when Max Project Results is 0


JacobDeuchert:
## Is your feature request related to a problem? Please describe.

For some reason i had the _Max Project Results_ setting set to 0. With the previous versions this setting worked fine with but with the latest 1.25.0 version omnisharp did not start or logged any kind of output.

## Describe the solution you would like
I would be nice for other users which accidentally set the Max Project Results to 0 to either show a message in the omnisharp output log like "No launchTarges found. Make sure your workspace contains a .sln/.csproj file and your Max Project Results is set bigger 0"

Another solution would be to use  1 or 250 instead when set to 0.

## Applicable Scenarios
User accidentally sets the Max Project Results to 0 and has a hard time finding the reason why omnisharp is not working and completly silent.

PR #5241, opened 2022-05-28
Return all launch targets when `maxProjectResults` is set to 0.

JoeRobich:
Resolves https://github.com/OmniSharp/omnisharp-vscode/issues/5227


filipw (approved):
makes sense to me!

PR #5255, opened 2022-06-05
Package manager nullability fixes

50Wliu:
Covers:
* src/packageManager/
* src/razor/
* src/tools/

Notable changes:
* Added missing `architectures` to macOS x64 OmniSharp download (I hope this wasn't intentional to allow M1 Macs to download both ARM64 + x64 OmniSharps? `architectures` is marked as non-null in IPackage).
* Remove all guards that checked for non-null packages, as there is a root check for that when initially reading `runtimeDependencies` from package.json.
  * This is _slightly_ risky as it's possible I missed a root caller that doesn't check that runtimeDependencies from an `any`-typed package.json is non-null - but I mean, with the previous behavior we would just _silently refuse to download OmniSharp_. At least now there will be an obvious error thrown somewhere.
* Similarly, removed all version checks as there's a root `semver.valid(version)` check.
* `downloadAndInstallPackages` is only ever called with a list of validated OmniSharp packages (function should be renamed to reflect that...), so removed some unnecessary checks around that.

While we're here, I'd like to understand the current state of how runtimeDependencies are defined/parsed, so that I have the context to create a followup PR cleaning up some of the associated logic.
* Is `installTestPath` used for anything? There's 0 references from AbsolutePathPackage and I don't see any meaningful reads from IPackage.
* Why do the packages in package.json specify a URL + installPath, if they're just overriden in OmnisharpPackageCreator (albeit, to the same thing) from constants in server.ts?

### DIFF
--- a/src/packageManager/downloadAndInstallPackages.ts
+++ b/src/packageManager/downloadAndInstallPackages.ts
@@ -17,64 +17,50 @@ import { PackageInstallStart } from "../omnisharp/loggingEvents";
 import { DownloadValidator } from './isValidDownload';
 
 export async function downloadAndInstallPackages(packages: AbsolutePathPackage[], provider: NetworkSettingsProvider, eventStream: EventStream, downloadValidator: DownloadValidator, useFramework: boolean): Promise<boolean> {
-    if (packages) {
-        eventStream.post(new PackageInstallStart());
-        for (let pkg of packages) {
-            let installationStage = "touchBeginFile";
-
-            if (pkg.id === "OmniSharp") {
-                if (pkg.isFramework !== useFramework) {
-                    continue;
-                }
-
-                if (pkg.url === null) {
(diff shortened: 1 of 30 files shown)


src/packageManager/downloadAndInstallPackages.ts
```diff
@@ -17,64 +17,50 @@ import { PackageInstallStart } from "../omnisharp/loggingEvents";
...
+                if (downloadValidator(buffer, pkg.integrity, eventStream)) {
+                    installationStage = "installPackage";
+                    await InstallZip(buffer, pkg.description, pkg.installPath, pkg.binaries, eventStream);
+                    installationStage = 'touchLockFile';
+                    await touchInstallFile(pkg.installPath, InstallFileType.Lock);
+                    break;
                 }
-
-                if (pkg.url === null) {
-                    eventStream.post(new InstallationFailure(installationStage, new Error("A release package of OmniSharp does not exist for this platform. Set \"omnisharp.path\" to \"latest\" in Settings to use an experimental build.")));
```
JoeRobich, line 31: Thanks for removing this.

JoeRobich (approved):
Thanks @50Wliu! Sorry it took me so long to review.

PR #5338, opened 2022-08-12
Provide source generated file info for workspace symbols

333fred:
VSCode side of https://github.com/OmniSharp/omnisharp-roslyn/pull/2431. I also updated the source generated doc provider to have a lazy-fetch option, as we don't want to fetch and cache generated files for every single symbol that gets returned in a workspace symbol request.


333fred:
@JoeRobich @filipw for review.

filipw (approved):
LGTM!

PR #5339, opened 2022-08-13
Support generated files in referenceProvider

333fred:
Note that FAR _in_ a source generated file is not currently supported. But now, when a symbol location for a generated file is returned by FAR, we will ensure that the generated info is cached for use when the user attempts to navigate to it.


333fred:
@JoeRobich @filipw for review.

JoeRobich (approved):

PR #5349, opened 2022-08-19
Observer nullability fixes

50Wliu:
Changes are pretty straightforward. No behavioral changes (except possibly returning the empty string instead of undefined for SolutionId in telemetry).

Brings the number of strict errors down from 403 to 378.


JoeRobich (approved):
Thanks!

PR #5359, opened 2022-08-29
Add Ardalis specification class snippet

Typhon0:
Hello, I don't know if it's within the scope of this extension so feel free to reject my pull request 😄 

The change is just a snippet to create a specification class which is pattern that you can use with this package ▶️  https://github.com/ardalis/Specification.

### DIFF
--- a/snippets/csharp.json
+++ b/snippets/csharp.json
@@ -489,5 +489,16 @@
             "}$0"
         ],
         "description": "create xunit test method"
-    }
+    },
+    "Specification Class": {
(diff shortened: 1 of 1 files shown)


heartacker:
you can creat user snippet in stead of this

JoeRobich:
Closing as we do not accept snippets for 3rd party libraries. Creating an extension to ship this snippet on its own would not be difficult ([see guide](https://code.visualstudio.com/api/language-extensions/snippet-guide)).

PR #5382, opened 2022-09-23
Add projectFilesIncludePattern & projectFilesExcludePattern options.

xieyubo:
Add two options instead of hardcoded string, so that user have chance to change it.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=5382) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=5382) before we can accept your contribution.<br/><br/>:x: xieyubo [sign now](https://cla.dotnetfoundation.org/OmniSharp/omnisharp-vscode?pullRequest=5382)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OmniSharp/omnisharp-vscode?pullRequest=5382) it.</sub>

JoeRobich:
So I have a few thoughts about this, 
1) The include pattern is not something that should be adjusted. We have to specify what activates our extension at publish time and changes to this setting would not have an effect.
https://github.com/OmniSharp/omnisharp-vscode/blob/55bc0546da2a3b233133d1dca3eee2e1e964f59f/package.json#L603
Also, we can't introduce "new" project or solution types to load as we specifically look for the extensions being searched for a nothing else (see `resourcesAndFolderMapToLaunchTargets` method in launcher.ts).
2) The excludes pattern may not be necessary either. It may be more correct for us to union our exclude patterns with those configured in `files.exclude`. Would this achieve the result you are looking for?

xieyubo:
The use case: our project contains a lot of *.csproj files (no .sln file, we use dirs.proj to manage). We also have a git submodule which comes from a third-party repository contains a lot of *.sln files. Now, when we launch vscode, it will automatically load the projects from the third-party repository. We want to exclude the third-party folder totally. Is there anyway to do?

JoeRobich:
@xieyubo Ah, in that case you would still want to see the files from the submodule in VS Code, but simply not have them be considered as a project location. If you remove the include pattern setting, I think we could take this PR.

xieyubo:
Updated.

JoeRobich (approved):
Looks good. Thanks @xieyubo!

PR #5397, opened 2022-10-08
Add prerequisite check for running OmniSharp.

JoeRobich:
Add actionable prompts when .NET SDK or Mono is missing but required for O#.

### DIFF
--- a/package.json
+++ b/package.json
@@ -231,7 +231,8 @@
         "darwin"
       ],
       "architectures": [
-        "x86_64"
+        "x86_64",
+        "arm64"
       ],
       "binaries": [
         "./mono.osx",
@@ -333,7 +334,7 @@
     },
     {
       "id": "OmniSharp",
-      "description": "OmniSharp for Linux (Framework / arm64)",
+      "description": "OmniSharp for Linux (Mono / arm64)",
       "url": "https://roslynomnisharp.blob.core.windows.net/releases/1.39.1/omnisharp-linux-arm64-1.39.1.zip",
       "installPath": ".omnisharp/1.39.1",
       "platforms": [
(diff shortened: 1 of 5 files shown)


src/utils/getDotnetInfo.ts
```diff
@@ -20,8 +20,8 @@ export async function getDotnetInfo(dotNetCliPaths: string[]): Promise<DotnetInf
         return _dotnetInfo;
     }
 
-    let dotnetExeName = CoreClrDebugUtil.getPlatformExeExtension();
-    let dotnetExecutablePath = undefined;
+    let dotnetExeName = join('dotnet', CoreClrDebugUtil.getPlatformExeExtension());
```
JoeRobich, line 23: This is currently broken in main.

package.json
```diff
@@ -231,7 +231,8 @@
         "darwin"
       ],
       "architectures": [
-        "x86_64"
+        "x86_64",
+        "arm64"
```
JoeRobich, line 235: Allow x64 Mono build of O# to install on Apple Silicon.

src/omnisharp/requirementCheck.ts
```diff
@@ -0,0 +1,128 @@
...
+    No
+}
+
+interface PromptItem extends vscode.MessageItem {
+    result: PromptResult;
+}
+
+async function promptToDownloadDotNetSDK() {
+    return new Promise<PromptResult>((resolve, reject) => {
+        const message = 'OmniSharp requires an install of the .NET SDK to provide language services when `omnisharp.useModernNet` is enabled in Settings. Please install the latest .NET SDK and restart.';
```
333fred, line 104:
```suggestion
        const message = 'OmniSharp requires an install of the .NET SDK to provide language services when `omnisharp.useModernNet` is enabled in Settings. Please install the latest .NET SDK and restart vscode. If you see this error after installing and restarting vscode, you may need to log out and log back in or restart your system for changes to the PATH to take effect.';
```

333fred (approved):

PR #5405, opened 2022-10-12
coreclr-debug nullability

50Wliu:
This handles the coreclr-debug folder. This is the last src folder that needs to be converted; one more PR will follow to address recent changes to the other folders & enable strict mode for all of src.

Behavior changes:
* DotNetCliError has been removed. We now show the complete error string in both the notification and the output channel. (This isn't strictly necessary - I can always add the error back, just didn't think the overhead of a custom error type was worth it here.)
* getDotnetInfo now throws if version info can't be parsed.
* ❗Fixed dotnetExeName using `path.join` instead of just string concatenation, resulting in a Windows executable of `dotnet\.exe` rather than `dotnet.exe`.
* Converted DotnetInfo from a class to an interface and removed the unused OsVersion property.

Impact:
22 strict violations in src, down from 43 in master. (Fun fact: including test, we're now at 198 strict violations. Below 200!)

### DIFF
--- a/src/coreclr-debug/ParsedEnvironmentFile.ts
+++ b/src/coreclr-debug/ParsedEnvironmentFile.ts
@@ -3,13 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as fs from 'fs-extra';
+import * as fs from 'fs';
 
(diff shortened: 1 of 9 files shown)


gregg-miskelly (approved):

JoeRobich (approved):
Thanks @50Wliu!

PR #5536, opened 2023-01-16
Improve outline

trolleyman:
This abbreviates the names of namespaces, delegates, structs, enums, classes and interfaces when in the outliner. This especially helps the repetitiveness of the breadcrumbs at the top of the editor, making it much easier to use (as it uses up much less screen space).

#### Breadcrumbs before & after
image

image

| Outline Before | Outline After |
|-------|-------|
| image | image |

#### Code
```csharp
namespace OmnisharpTest
{
    namespace Foo
    {
        namespace Bar.Baz
        {
            public class BazClass
            {
                public class InnerBazClass
                {
                    public enum InnerEnum
                    {
                        A,
                        B,
                        C,
                    }

                    public interface IFoo
                    {
                        public void Method1(string many1, int many2, BazClass many3, InnerEnum many4, string many5);
                    }
                }

                public const string ConstantString = "ABCDEF";
                public static readonly OtherClass? OtherClassInstance = null;

                public string Field;
                public int Property { get; set; }

                public int this[int index] { get => 1; }

                public delegate int TestDelegate(string arg);

                public event TestDelegate? TestEvent;

                public BazClass(string field)
                {
                    Field = field;
                }

                ~BazClass()
                {
                }

                public void Method2(string argument1, int argument2, BazClass argument3, string argument4, string argument5)
                {
                    const string SomeConstant = "ABCDEF";

                    Console.WriteLine(SomeConstant);

                    void TestFunction(string testArg2)
                    {
                        ;
                    }
                    TestFunction("a");
                }
            }

            public class OtherClass
            {

            }

            public readonly struct WrappedInt
            {
                public static WrappedInt operator +(WrappedInt a) => a;
                public static WrappedInt operator -(WrappedInt a) => new WrappedInt(-a._inner);
                public static WrappedInt operator +(WrappedInt a, WrappedInt b) => new WrappedInt(a._inner + b._inner);
                public static WrappedInt operator -(WrappedInt a, WrappedInt b) => new WrappedInt(a._inner - b._inner);

                private readonly int _inner;

                public WrappedInt(int inner)
                {
                    _inner = inner;
                }

                public override string ToString() => _inner.ToString();
            }
        }
    }
}

namespace OmnisharpTest.Foo
{
    namespace OmnisharpTest.Foo
    {
        namespace Test123
        {

        }
    }
}
```


trolleyman:
@microsoft-github-policy-service agree

trolleyman:
Not sure how to get this reviewed - @dibarbet?

src/features/documentSymbolProvider.ts
```diff
@@ -46,11 +46,49 @@ function createSymbols(elements: Structure.CodeElement[]): vscode.DocumentSymbol
...
+        case SymbolKinds.Struct:
+        case SymbolKinds.Namespace:
+            if (typeof parentElement === 'undefined') {
+                return { name: element.DisplayName, details: '' };
+            }
+            const prefix = `${parentElement.DisplayName}.`;
+            if (!element.DisplayName.startsWith(prefix)) {
+                return { name: element.DisplayName, details: '' };
+            }
+            const name = element.DisplayName.slice(prefix.length);
```
dibarbet, line 64: I think that in most cases the `element.Name` property should be the non-fully qualified name of the class/type/interace/etc.

e.g.
image


So I think we might not need to do any slicing here, instead just pass in the 'name' as the as the name and the displayName as the details

dibarbet:
@trolleyman apologies for the delay - if you are ever waiting for a reply for more than a day or two, please ping me directly, sometimes I miss the original notification emails.

src/features/documentSymbolProvider.ts
trolleyman: Thanks for the review, nw about the delay :)

The problem is with namespace elements - without the slice the namespace `Foo.Bar.Baz` (without any parents) will resolve as `Baz`, which causes some confusion with the breadcrumbs, making it seem like the namespace is just Baz.

dibarbet: I see what you mean, looks like the server just populates the name with the last identifier.  

Can we do this:
1.  Limit the slicing to only happen for namespace case
2.  For the rest, use the name and display name as-is.

We could change the server side to fix this, but seems like overkill for such a small change, so I'd rather not.

trolleyman: Hey - I've done that, with a slight tweak - I've used `DisplayName` for the name for some types, and `Name` for others, so that more information is given - for example, below each function's `Name` is just `OverloadMethod`, which is a bit worse IMO in the outliner on small sizes (and gives less information in the breadcrumb). Having it be the `DisplayName` is how it is currently. It's up to you though, if you think the `Name` version is better I can change it to that.

### `DisplayName`
image

### `Name`
image

dibarbet: Ah absolutely, using DisplayName there looks better - good call.

dibarbet (approved):

dibarbet:
thanks for the contribution!

trolleyman:
@dibarbet Thanks for the review! :)

PR #5523, opened 2023-01-10
Update Razor to 7.0.0-preview.23053.8

allisonchou:
Updates Razor to 7.0.0-preview.23053.8

PR #5544, opened 2023-01-19
Update Omnisharp-roslyn

dibarbet:


src/tools/UpdatePackageDependencies.ts
```diff
@@ -239,21 +238,22 @@ function getLowercaseFileNameFromUrl(url: string): string {
     let index = url.lastIndexOf("/");
     let fileName = url.substr(index + 1).toLowerCase();
 
-    // With Razor putting two version numbers into their filename we need to split up the name to
-    // correctly identify the version part of the filename.
-    let nameParts = fileName.split('-');
-    let potentialVersionPart = nameParts[nameParts.length - 1];
-    let versions = findVersions(potentialVersionPart, { loose: true });
-    if (!versions || versions.length == 0) {
+    if (fileName.startsWith("omnisharp")) {
```
dibarbet, line 241: the old code was super busted, trying to convert non semvar versions to semvar.  When the Razor versions changed in https://github.com/OmniSharp/omnisharp-vscode/pull/5523/files (to remove the preview number in their version) the findVersions package started adding an extraneous '.0' at the end.  This broke the update script.

I changed this to instead define explicitly what its looking for so at least its more clear on what is going wrong when version numbers change.

The whole format in the package.json should probably change so we don't have to parse URLs with random versions for file names, but I don't have time for that at the moment.

filipw (approved):

src/tools/UpdatePackageDependencies.ts
333fred: When you say semvar, do you mean semver?

333fred (approved):

PR #5589, opened 2023-02-21
Carlosguerravz/devcerts check

### LINKED ISSUE #3187: Automatically prompt to create and trust ASP.NET Core HTTPS development certificate
Currently, ASP.NET Core web apps created with "dotnet new" have HTTPS enabled by default. The very first time you debug (or, after `dotnet dev-certs https --clean`), when no cert is created, debugging will fail, telling the user to run `dotnet dev-certs https`, optionally with `--trust` on Mac/Windows.

Similarly to how this extension suggests to create the tasks.json/launch.json when opening a C# project, I think it should prompt to do the above `dotnet dev-certs https [--trust]` when opening an ASP.NET Core project when HTTPS is enabled on it, if no cert is already trusted.

On Windows, a popup asks if the user wants to trust the certificate; on Mac it requires password input, so it is still user-interactive, and the user needs to know to expect the popup if they say "Yes" to creating/trusting the cert. (On Linux there's no centralized store for trusted certs, so this is moot there.)

CarlosGuerraVz:
This commit fixes #3187 by implementing the proposal in the [comments](https://github.com/OmniSharp/omnisharp-vscode/issues/3187#issuecomment-1402914194).

- Added launch option `checkForDevCert` and documented it on `debugger-launchjson.md`.
- When debugging, if not on Linux, on remote, or on a browser, the extension runs `dotnet dev-certs https --check` to check for dev-certs.
- If not found, it displays the following message:
image

- Selecting yes will run `dotnet dev-certs https --trust`
- More information will redirect to the page containing `the description for the `checkForDevCert` launch option.
- If certificate creation fails it will log the failure to the C# output and will display a message offering to show the error output.
- Tested on Windows and MacOS

### DIFF
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -338,3 +338,16 @@ Example:
 ```json
     "targetArchitecture": "arm64"
 ```
+
+## Check for DevCert
+If unspecified, it will be enabled when `serverReadyAction` is set AND when `pipeTransport` is NOT set.
+
+When `true` and if VS is runing on Windows or MacOS, the extension will try to find if your system contains development certificates by running `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating them. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create self signed certificates.
+
+You can override this behavior by setting `checkForDevCert` in your `launch.json`.
+
+Example:
+
+```json
+    "checkForDevCert": "false"
+```
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -342,9 +342,9 @@ Example:
 ## Check for DevCert
 If unspecified, it will be enabled when `serverReadyAction` is set AND when `pipeTransport` is NOT set.
 
-When `true` and if VS is runing on Windows or MacOS, the extension will try to find if your system contains development certificates by running `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating them. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create self signed certificates.
+When `true` and if VS is runing on Windows or MacOS, the extension will try to find if your system contains a development certificate by running `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating them. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create self signed certificates.
 
-You can override this behavior by setting `checkForDevCert` in your `launch.json`.
+You can override this behavior by setting `checkForDevCert` to false in your `launch.json`.
 
 Example:
 
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -341,7 +341,7 @@ Example:
 
 ## Check for DevCert
 
-This option controls if, on launch, the the debugger should check if the computer has a self-signed HTTPS certificate used to develop web projects running on https endpoints. For this it will try to run `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating one. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create a trusted self-signed certificate.
+This option controls if, on launch, the the debugger should check if the computer has a self-signed HTTPS certificate used to develop web projects running on https endpoints. For this it will try to run `dotnet dev-certs https --check --trust`, if no certs are found it will prompt the user to suggest creating one. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create a trusted self-signed certificate.
 
 If unspecified, defaults to true when `serverReadyAction` is set.
 This option does nothing on Linux, VS Code remote, and VS Code Web UI scenarios.
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -341,7 +341,7 @@ Example:
 
 ## Check for DevCert
 
-When `true` and if Visual Studio Code is runing on Windows or MacOS, the the debugger will check if the computer has a self-signed HTTPS certificate used to develop web servers running on https endpoints, it will try to run `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating one. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create a trusted self-signed certificate.
+This option controls if, on launch, the the debugger should check if the computer has a self-signed HTTPS certificate used to develop web projects running on https endpoints. For this it will try to run `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating one. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create a trusted self-signed certificate.
 
 If unspecified, defaults to true when `serverReadyAction` is set.
 This option does nothing on Linux, VS Code remote, and VS Code Web UI scenarios.
--- a/debugger-launchjson.md
+++ b/debugger-launchjson.md
@@ -340,9 +340,11 @@ Example:
 ```
 
 ## Check for DevCert
-If unspecified, it will be enabled when `serverReadyAction` is set AND when `pipeTransport` is NOT set.
 
-When `true` and if VS is runing on Windows or MacOS, the extension will try to find if your system contains a development certificate by running `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating them. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create self signed certificates.
+When `true` and if Visual Studio Code is runing on Windows or MacOS, the the debugger will check if the computer has a self-signed HTTPS certificate used to develop web servers running on https endpoints, it will try to run `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating one. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create a trusted self-signed certificate.
+
+If unspecified, defaults to true when `serverReadyAction` is set.
+This option does nothing on Linux, VS Code remote, and VS Code Web UI scenarios.
 
 You can override this behavior by setting `checkForDevCert` to false in your `launch.json`.
 
--- a/src/coreclr-debug/debugConfigurationProvider.ts
+++ b/src/coreclr-debug/debugConfigurationProvider.ts
@@ -9,9 +9,11 @@ import { RemoteAttachPicker, DotNetAttachItemsProviderFactory, AttachPicker, Att
 import { Options } from '../omnisharp/options';
 import { PlatformInformation } from '../platform';
 import { hasDotnetDevCertsHttps, createSelfSignedCert } from '../utils/DotnetDevCertsHttps';
+import { EventStream } from '../EventStream';
+import { DevCertCreationFailure } from '../omnisharp/loggingEvents';
  
 export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
-    constructor(public platformInformation: PlatformInformation, private options: Options) {}
+    constructor(public platformInformation: PlatformInformation, private readonly eventStream: EventStream, private options: Options) {}
 
     public async resolveDebugConfigurationWithSubstitutedVariables(folder: vscode.WorkspaceFolder | undefined, debugConfiguration: vscode.DebugConfiguration, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration | null | undefined>
     {
@@ -75,7 +77,7 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
 
             if (debugConfiguration.checkForDevCert)
             {
-                checkForDevCerts(this.options.dotNetCliPaths);
+                checkForDevCerts(this.options.dotNetCliPaths, this.eventStream);
             }
         }
 
@@ -83,7 +85,7 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
     }
 }
 
-function checkForDevCerts(dotNetCliPaths: string[]){
+function checkForDevCerts(dotNetCliPaths: string[], eventStream: EventStream){
     hasDotnetDevCertsHttps(dotNetCliPaths).then(async (returnData) => {
         if(returnData.error) //if the prcess returns 0 error is null, otherwise the return code can ba acessed in returnData.error.code
         {
@@ -93,25 +95,31 @@ function checkForDevCerts(dotNetCliPaths: string[]){
 
             const result = await vscode.window.showInformationMessage(
                 "The selected launch configuration is configured to launch a web browser but no trusted development certificate was found. Create a trusted self-signed certificate?", 
-                { title:labelYes}, { title:labelNotNow, isCloseAffordance: true }, { title:labelMoreInfo}
+                { title:labelYes }, { title:labelNotNow, isCloseAffordance: true }, { title:labelMoreInfo }
                 ); 
             if (result?.title === labelYes)
             {
                 let returnData = await createSelfSignedCert(dotNetCliPaths);
                 if (returnData.error === null)
                 {
-                    vscode.window.showInformationMessage('Self-signed certificate sucessfully created');
+                    vscode.window.showInformationMessage('Self-signed certificate sucessfully created.');
                 }
                 else
                 {
-                    vscode.window.showWarningMessage("Couldn't create self-signed certificate");
+                    eventStream.post(new DevCertCreationFailure(`${returnData.error.message}\ncode: ${returnData.error.code}\nstdout: ${returnData.stdout}`));
+
+                    const labelShowOutput: string = "Show Output";
+                    const result = await vscode.window.showWarningMessage("Couldn't create self-signed certificate. See for more information.", labelShowOutput);
+                    if (result === labelShowOutput){
+                        vscode.commands.executeCommand("workbench.action.output.show.extension-output-ms-dotnettools.csharp-#3-C#");
+                    }
                 }
             }
             if (result?.title === labelMoreInfo)
             {
                 const launchjsonDescriptionURL = 'https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md#check-for-devcert';
                 vscode.env.openExternal(vscode.Uri.parse(launchjsonDescriptionURL));
-                checkForDevCerts(dotNetCliPaths);
+                checkForDevCerts(dotNetCliPaths, eventStream);
             }
         }
     });
--- a/src/coreclr-debug/debugConfigurationProvider.ts
+++ b/src/coreclr-debug/debugConfigurationProvider.ts
@@ -75,16 +75,17 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
 
             if (debugConfiguration.checkForDevCert)
             {
-                hasDotnetDevCertsHttps(this.options.dotNetCliPaths).then(async (hasDevCert) => {
-                    if(!hasDevCert)
+                hasDotnetDevCertsHttps(this.options.dotNetCliPaths).then(async (returnData) => {
+                    if(returnData.error) //if the prcess returns 0 error is null, otherwise the return code can ba acessed in returnData.error.code
                     {
                         const result = await vscode.window.showInformationMessage(
                             "The selected launch configuration is configured to launch a web browser but no trusted development certificate was found. Create a trusted self-signed certificate?", 
-                            'Yes', 'Not Now', "Don't Ask Again", 
+                            { title:"Yes"}, { title:'Not Now', isCloseAffordance: true }, { title:"Don't Ask Again"}
                             ); 
-                        if (result === 'Yes')
+                        if (result?.title === 'Yes')
                         {
-                            if (await createSelfSignedCert(this.options.dotNetCliPaths))
+                            let returnData = await createSelfSignedCert(this.options.dotNetCliPaths);
+                            if (!returnData.error)
                             {
                                 vscode.window.showInformationMessage('Self-signed certificate sucessfully created');
                             }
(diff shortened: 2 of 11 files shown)


src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,42 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+            }
+
+            if (debugConfiguration.checkForDevCert)
+            {
+                hasDotnetDevCertsHttps(this.options.dotNetCliPaths).then(async (hasDevCert) => {
+                    if(!hasDevCert)
+                    {
+                        const result = await vscode.window.showInformationMessage(
+                            "The selected launch configuration is configured to launch a web browser but no trusted development certificate was found. Create a trusted self-signed certificate?", 
+                            "Don't Ask Again", 'Not Now', 'Yes'
```
gregg-miskelly, line 79: >"Don't Ask Again", 'Not Now', 'Yes' [](http://example.com/codeflow?start=28&length=35)

Is "Don't Ask Again" the default value? If so, I would reverse the order so "Yes" is the default.

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,42 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
                 return undefined;
             }
         }
+        
+        if (!this.platformInformation.isLinux() && !vscode.env.remoteName && vscode.env.uiKind != vscode.UIKind.Web)
```
gregg-miskelly, line 65: >if [](http://example.com/codeflow?start=8&length=2)

Probably worth a comment explaining the reason for some of these checks. Something like:

// We want to ask the user if we should run `dotnet  dev-certs https --trust`, but this doesn't work in a few cases --
// Linux -- not supported by the .NET CLI as there isn't a single root cert store
// VS Code remoting/Web UI -- the trusted cert work would need to happen on the client machine, but we don't have a way to run code there currently
// `pipeTransport` -- the dev cert on the server will be different from the client

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: On other info messages "don't ask again" is usually the default, but I can see why "Yes" would be preferred.
image

src/utils/DotnetDevCertsHttps.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CoreClrDebugUtil } from "../coreclr-debug/util";
+
+// Will return true if `dotnet dev-certs https --check` succesfully finds a development certificate. 
+export async function hasDotnetDevCertsHttps(dotNetCliPaths: string[]): Promise<Boolean> {
+
+    let dotnetExecutablePath = getDotNetExecutablePath(dotNetCliPaths);
+
+    try {
+        const data = await execChildProcess(`${dotnetExecutablePath ?? 'dotnet'} dev-certs https --check`, process.cwd(), process.env);
+        return (/A valid certificate was found/.test(data));
```
gregg-miskelly, line 17: >(/A valid certificate was found/.test(data)); [](http://example.com/codeflow?start=15&length=45)

The output might not come back as ENU, so I would just rely on the exit code.

src/utils/DotnetDevCertsHttps.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Will run `dotnet dev-certs https --trust` to prompt the user to create self signed certificates. Retruns true if sucessfull.
+export async function createSelfSignedCert(dotNetCliPaths: string[]): Promise<Boolean> {
+    
+    let dotnetExecutablePath = getDotNetExecutablePath(dotNetCliPaths);
+
+    try {
+        const data = await execChildProcess(`${dotnetExecutablePath ?? 'dotnet'} dev-certs https --trust`, process.cwd(), process.env);
+        return (/Successfully created and trusted a new HTTPS certificate/.test(data));
```
gregg-miskelly, line 31: >(/Successfully created and trusted a new HTTPS certificate/.test(data)); [](http://example.com/codeflow?start=15&length=72)

Same

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,42 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                            }
+                            else
+                            {
+                                vscode.window.showWarningMessage("Couldn't create self-signed certificate");
+                            }
+                        }
+                        else if (result === "Don't Ask Again")
+                        {
+                            await vscode.window.showInformationMessage("To don't see this message again set launch option `checkForDevCert` to `false` in launch.json", { modal: true });
+                        }
```
gregg-miskelly, line 99: Rather than having a button for this, I would just stick this as part of the message (unless VS Code has a max length).

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: Changed to 
image

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: The problem is VS Code messages [don't allow formatting or break lines in them](https://github.com/microsoft/vscode/issues/48900), so it makes it not very readable to add that that information in one big block of text.

src/tools/OptionsSchema.json
```diff
@@ -407,6 +407,11 @@
         "targetArchitecture": {
           "type": "string",
           "description": "[Only supported in local macOS debugging] The architecture of the debuggee. This will automatically be detected unless this parameter is set. Allowed values are x86_64 or arm64."
+        },
+        "checkForDevCert": {
+          "type": "boolean",
+          "description": "When true, the presence of dev certificates will be checked by runing dotnet `dev-certs https --check`, and if that fails the user will be prompted to create dev certs by running `dotnet dev-certs https --trust`",
```
gregg-miskelly, line 413: >description [](http://example.com/codeflow?start=11&length=11)

How about:

When true, the debugger will check if the computer has a self-signed HTTPS certificate used to develop web servers running on https endpoints. This option does nothing on Linux, VS Code remote, and VS Code Web UI scenarios. If the HTTPS certificate is not found, the user will be prompted to install it.

src/utils/DotnetDevCertsHttps.ts
```diff
@@ -0,0 +1,50 @@
...
+        const data = await execChildProcess(`${dotnetExecutablePath ?? 'dotnet'} dev-certs https --check`, process.cwd(), process.env);
+        return (/A valid certificate was found/.test(data));
+    }
+    catch (err) {
+        return false;
+    }
+}
+
+// Will run `dotnet dev-certs https --trust` to prompt the user to create self signed certificates. Retruns true if sucessfull.
+export async function createSelfSignedCert(dotNetCliPaths: string[]): Promise<Boolean> {
```
gregg-miskelly, line 25: >Boolean [](http://example.com/codeflow?start=78&length=7)

I think I would instead suggest returning an error string if this fails rather than just a `Boolean`. I would suggest including the stderr and exit code.

src/coreclr-debug/debugConfigurationProvider.ts
gregg-miskelly: Got it. Do you know if we do anything similar anywhere else? I wonder if we want a 'More information' button instead, and we could take you to the debugger-launchjson.md section for this option.

CC @sayedihashimi in case he has any thoughts

debugger-launchjson.md
```diff
@@ -338,3 +338,16 @@ Example:
 ```json
     "targetArchitecture": "arm64"
 ```
+
+## Check for DevCert
+If unspecified, it will be enabled when `serverReadyAction` is set AND when `pipeTransport` is NOT set.
+
+When `true` and if VS is runing on Windows or MacOS, the extension will try to find if your system contains a development certificate by running `dotnet dev-certs https --check`, if no certs are found it will prompt the user to suggest creating them. If approved by the user, the extension will run `dotnet dev-certs https --trust` to create self signed certificates.
```
gregg-miskelly, line 345: >VS [](http://example.com/codeflow?start=19&length=2)

"Visual Studio Code"

debugger-launchjson.md
```diff
@@ -338,3 +338,16 @@ Example:
 ```json
     "targetArchitecture": "arm64"
 ```
+
+## Check for DevCert
+If unspecified, it will be enabled when `serverReadyAction` is set AND when `pipeTransport` is NOT set.
```
gregg-miskelly, line 343: >If [](http://example.com/codeflow?start=0&length=2)

I like the text you have, but I think we need start with an explanation of what development certificates are.

Maybe something similar to the text I suggested for OptionsSchema.json?

This option controls if, on launch, the the debugger should check if the computer has a self-signed HTTPS certificate used to develop web projects running on https endpoints.

gregg-miskelly, line 343: >will be enabled [](http://example.com/codeflow?start=19&length=15)

To be more precise, how about "defaults to `true`"

src/tools/OptionsSchema.json
gregg-miskelly: Probably also need a sentence about what the default is (maybe copy from debugger-launchjson.md)?

src/coreclr-debug/debugConfigurationProvider.ts
gregg-miskelly: I should add, the reason I am suggesting this is that I feel like have the button be called "Don't Ask Again" rather over promises.

src/coreclr-debug/debugConfigurationProvider.ts
gregg-miskelly: We could use the `detail` property to provide this information: [https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L1974](https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L1974)

src/coreclr-debug/debugConfigurationProvider.ts
gregg-miskelly: We could use the [`detail`](https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L1974) property to provide this information.

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,46 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+            }
+
+            if (debugConfiguration.checkForDevCert)
+            {
+                hasDotnetDevCertsHttps(this.options.dotNetCliPaths).then(async (hasDevCert) => {
+                    if(!hasDevCert)
+                    {
+                        const result = await vscode.window.showInformationMessage(
+                            "The selected launch configuration is configured to launch a web browser but no trusted development certificate was found. Create a trusted self-signed certificate?", 
+                            'Yes', "Don't Ask Again", 'Not Now'
```
gregg-miskelly: >'Not Now' [](http://example.com/codeflow?start=54&length=9)

If I am reading the API docs correctly, I think you want to make at least this one a [`MessageItem`](https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L1939) so you can make the escape key be the equivalent to this.

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: > We could use the [`detail`](https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L1974) property to provide this information.

For this the main info message would need be to be a modal which might be to obstructing. 

> Got it. Do you know if we do anything similar anywhere else? I wonder if we want a 'More information' button instead, and we could take you to the debugger-launchjson.md section for this option.
> 
> CC @sayedihashimi in case he has any thoughts

I'm liking this idea.

src/utils/DotnetDevCertsHttps.ts
CarlosGuerraVz: Should I show the error string in the warning message when it fails?

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,47 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+            }
+
+            if (debugConfiguration.checkForDevCert)
+            {
+                hasDotnetDevCertsHttps(this.options.dotNetCliPaths).then(async (returnData) => {
+                    if(returnData.error) //if the prcess returns 0 error is null, otherwise the return code can ba acessed in returnData.error.code
+                    {
+                        const result = await vscode.window.showInformationMessage(
+                            "The selected launch configuration is configured to launch a web browser but no trusted development certificate was found. Create a trusted self-signed certificate?", 
+                            { title:"Yes"}, { title:'Not Now', isCloseAffordance: true }, { title:"Don't Ask Again"}
```
WardenGnaw, line 83: nit: `Yes`, `Not Now`, and `Don't Ask Again` should be saved as string variables and compare against them.

See https://github.com/microsoft/vscode-cpptools/blob/48653c0854c2b923da7d149505f62a0f1967c2b3/Extension/src/Debugger/extension.ts#L213-L218

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,47 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                                vscode.window.showInformationMessage('Self-signed certificate sucessfully created');
+                            }
+                            else
+                            {
+                                vscode.window.showWarningMessage("Couldn't create self-signed certificate");
+                            }
+                        }
+                        if (result?.title === "Don't Ask Again") //TODO: change option to More information and redirect to debugger-launchjson.md
+                        {
+                            await vscode.window.showInformationMessage("To don't see this message again set launch option 'checkForDevCert' to 'false' in launch.json", { modal: true });
```
gregg-miskelly, line 99: >await [](http://example.com/codeflow?start=28&length=5)

Do we need to re-ask the user if the click 'More information'? Or is there a way to have a button that doesn't close the dialog?

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,6 +61,47 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                        if (result?.title === 'Yes')
+                        {
+                            let returnData = await createSelfSignedCert(this.options.dotNetCliPaths);
+                            if (!returnData.error)
+                            {
+                                vscode.window.showInformationMessage('Self-signed certificate sucessfully created');
+                            }
+                            else
+                            {
+                                vscode.window.showWarningMessage("Couldn't create self-signed certificate");
```
gregg-miskelly, line 94: >") [](http://example.com/codeflow?start=105&length=2)

You want to do something with the failure stuff you got. Probably, to write it to the C# log, and updated the message to be something like "Couldn't create self-signed certificate. See <something> for more information."

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: There doesn't seem to be a way of not closing the dialog.
Just re-asking seems to work fine.

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: > Got it. Do you know if we do anything similar anywhere else? I wonder if we want a 'More information' button instead, and we could take you to the debugger-launchjson.md section for this option.

Implemented this

src/utils/DotnetDevCertsHttps.ts
```diff
@@ -0,0 +1,53 @@
...
+        let dotnetFullPath = join(dotnetPath, dotnetExeName);
+        if (CoreClrDebugUtil.existsSync(dotnetFullPath)) {
+            dotnetExecutablePath = dotnetFullPath;
+            break;
+        }
+    }
+    return dotnetExecutablePath;
+}
+
+async function execChildProcess(command: string, workingDirectory: string = getExtensionPath(), env: NodeJS.ProcessEnv = {}): Promise<ExecReturnData> {
```
WardenGnaw, line 41: Can we not reuse https://github.com/OmniSharp/omnisharp-vscode/blob/414ae389c2bdfc6b425289f7bb73a7d7ed7d04f2/src/common.ts#L59?

src/utils/DotnetDevCertsHttps.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Will run `dotnet dev-certs https --trust` to prompt the user to create self signed certificates. Retruns true if sucessfull.
+export async function createSelfSignedCert(dotNetCliPaths: string[]): Promise<ExecReturnData> {
+
+    let dotnetExecutablePath = getDotNetExecutablePath(dotNetCliPaths);
+
+    return await execChildProcess(`${dotnetExecutablePath ?? 'dotnet'} dev-certs https --trust`, process.cwd(), process.env);
+}
+
+function getDotNetExecutablePath(dotNetCliPaths: string[]): string | undefined{
```
WardenGnaw, line 27: Can you move this to getDotnetInfo and import it here?
https://github.com/OmniSharp/omnisharp-vscode/blob/e3960a0d41155502c2cdbb427573f913ab38f075/src/utils/getDotnetInfo.ts#L18-L27

src/utils/DotnetDevCertsHttps.ts
CarlosGuerraVz: initially I was using it, but it returns the error information in a string instead of it being structured, so I preferred to use my own implementation.

src/utils/DotnetDevCertsHttps.ts
CarlosGuerraVz: done

src/coreclr-debug/debugConfigurationProvider.ts
sayedihashimi: I like the idea where we show a More Info link.

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,7 +63,64 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                    vscode.window.showInformationMessage('Self-signed certificate sucessfully created.');
+                }
+                else
+                {
+                    eventStream.post(new DevCertCreationFailure(`${returnData.error.message}\ncode: ${returnData.error.code}\nstdout: ${returnData.stdout}`));
+
+                    const labelShowOutput: string = "Show Output";
+                    const result = await vscode.window.showWarningMessage("Couldn't create self-signed certificate. See output for more information.", labelShowOutput);
+                    if (result === labelShowOutput){
+                        vscode.commands.executeCommand("workbench.action.output.show.extension-output-ms-dotnettools.csharp-#3-C#");
```
gregg-miskelly, line 114: >workbench.action.output.show.extension-output-ms-dotnettools.csharp-#3-C# [](http://example.com/codeflow?start=56&length=73)

I feel like this command name is not stable.

gregg-miskelly: I think you want to use `.show` on the output channel: [https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L6448](https://github.com/microsoft/vscode/blob/c83eff40dd2c2115f5fd5e67a6ce304fbc6fdefa/src/vscode-dts/vscode.d.ts#L6448)

src/coreclr-debug/debugConfigurationProvider.ts
```diff
@@ -59,7 +63,64 @@ export class DotnetDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                checkForDevCerts(this.options.dotNetCliPaths, this.eventStream);
+            }
+        }
 
         return debugConfiguration;
     }
 }
+
+function checkForDevCerts(dotNetCliPaths: string[], eventStream: EventStream){
+    hasDotnetDevCertsHttps(dotNetCliPaths).then(async (returnData) => {
```
gregg-miskelly, line 89: >hasDotnetDevCertsHttps [](http://example.com/codeflow?start=4&length=22)

Will this throw if `dotnet` is not found?

gregg-miskelly (commented):
Otherwise LGTM

src/coreclr-debug/debugConfigurationProvider.ts
CarlosGuerraVz: it won't throw as execChildProcess() will just include the error in the object it returns, but because we only check for the presence of returnData.error we will still show the "dev certificate not found" message. I think it should check for the exact error code (6), and ignore otherwise.

src/utils/DotnetDevCertsHttps.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as cp from 'child_process';
+import { getExtensionPath } from "../common";
+import { getDotNetExecutablePath } from "./getDotnetInfo";
+
+// Will return true if `dotnet dev-certs https --check` succesfully finds a development certificate. 
+export async function hasDotnetDevCertsHttps(dotNetCliPaths: string[]): Promise<ExecReturnData> {
+
+    let dotnetExecutablePath = getDotNetExecutablePath(dotNetCliPaths);
+
+    return await execChildProcess(`${dotnetExecutablePath ?? 'dotnet'} dev-certs https --check`, process.cwd(), process.env);
```
CarlosGuerraVz, line 15: Should we also use `--trust` to also check if the certificate is trusted?
VS also shows a message if the certificate exists but is not trusted.

WardenGnaw (approved):

Issue #5417, opened 2022-10-22
Security: HTTPS profile should be selected, not HTTP


Rick-Anderson:
related: #5416

VSC should follow security best practices and launch the the `HTTPS` profile like VS does.

VS output has: 
```
Using launch settings from 'C:\tmp\TodoApi\Properties\launchSettings.json' [Profile 'http']...
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5072
warn: Microsoft.AspNetCore.HttpsPolicy.HttpsRedirectionMiddleware[3]
      Failed to determine the https port for redirect.
```

Follow the instructions at the [Microsoft Web API tutorial-Create a web project](https://learn.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-6.0&tabs=visual-studio-code#create-a-web-project-1)

```
dotnet new webapi -o TodoApi
cd TodoApi
code -r ../TodoApi
```

^F5 and it launches a browser to `http:localhost::<port>` not  `HTTPS` from `launchSettings.json`

gregg-miskelly:
I am not able to reproduce this, but it doesn't sound like an issue with VS Code or this extension. 

When I try to launch the app I see the following in the debug console:

```
-------------------------------------------------------------------
You may only use the Microsoft .NET Core Debugger (vsdbg) with
Visual Studio Code, Visual Studio or Visual Studio for Mac software
to help you develop and test your applications.
-------------------------------------------------------------------
Using launch settings from 'C:\proj\net6\TodoApi\Properties\launchSettings.json' [Profile 'TodoApi']...
...
Microsoft.Hosting.Lifetime: Information: Now listening on: https://localhost:7038
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: https://localhost:7038
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5255
Microsoft.Hosting.Lifetime: Information: Now listening on: http://localhost:5255
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
```

It binds to both http and https because this is what launchSettings.json says to do:

```
...
  "profiles": {
    "TodoApi": {
      "applicationUrl": "https://localhost:7038;http://localhost:5255",
...
```

[...]

PR #5602, opened 2023-02-27
Default profile HTTPS

CarlosGuerraVz:
This commit fixes Issue #5417 for selecting https profile by default on MacOS and Windows, better matching VS behavior.

- On starting debug, the extension now looks for the first profile in `launchSettings.json` that contains a https url


gregg-miskelly:
I think you want to fix this in vsdbg-ui instead of here.

Issue #5004, opened 2022-01-13
Debug Windows x86_64 on ARM64


cwensley:
## Issue Description ##
Hi, I'm trying to use VSCode to debug an x64 app with .NET Framework on a Windows ARM64 device.  Is this possible?  on MacOS, you can specify `"targetArchitecture": "x86_64"` for the launch config, but this doesn't appear to work on Windows.

I have tried both the x64 and ARM64 version of VS Code, and neither work now.  This worked at one point using the x64 version of VS Code but then stopped working recently, and it would be nice if this worked again. 

## Steps to Reproduce ##

Add a launch.json that uses the "clr" debug type on Windows 11 ARM to an executable that is x86_64 running in emulation on Windows 11 ARM. 

## Expected Behavior ##

Should be able to debug

## Actual Behavior ##

Cannot debug, gives the message:

```
Error processing 'configurationDone' request. The remote computer does not have a CLR version which is compatible with the remote debugging components. To install a compatible CLR version, see the instructions in the 'Remote Components Setup' page on the Visual Studio CD.
```

### OmniSharp log ###
<details>Post the output from Output-->OmniSharp log here</details>

### C# log ###
<details>Post the output from Output-->C# here</details>

## Environment information ##

**VSCode version**: 1.63.2
**C# Extension**: 1.23.17

<details><summary>Dotnet Information</summary>
.NET SDK (reflecting any global.json):
 Version:   6.0.100
 Commit:    9e8b04bbff

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-arm64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.100\

Host (useful for support):
  Version: 6.0.0
  Commit:  4822e3c3aa

.NET SDKs installed:
  5.0.403 [C:\Program Files\dotnet\sdk]
  5.0.404 [C:\Program Files\dotnet\sdk]
  6.0.100 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 5.0.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #5646, opened 2023-04-05
Implement support for switching targetArchitecture on Windows ARM64

hajekj:
**Changes:**
* Enable use of `targetArchitecture` property on Windows
* Update `targetArchitecture` to be an enum with properties `arm64` and `x86_64`
* Download both ARM64 and X86_64 `vsdbg-ui` when on Windows ARM64
* Update tests

Addresses #5004

Tested manually both on Windows X64 and ARM64. Can provide an ARM64 Windows VM for validation if needed.

### DIFF
--- a/src/tools/OptionsSchema.json
+++ b/src/tools/OptionsSchema.json
@@ -410,7 +410,7 @@
             "x86_64",
             "arm64"
           ],
-          "description": "The architecture of the debuggee. This will automatically be detected unless this parameter is set. Allowed values are x86_64 or arm64."
+          "description": "The architecture of the debuggee. This will automatically be detected unless this parameter is set. Allowed values are x86_64 or arm64. This value is ignored on Linux."
         },
         "checkForDevCert": {
           "type": "boolean",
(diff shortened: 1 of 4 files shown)


src/tools/OptionsSchema.json
```diff
@@ -406,7 +406,11 @@
         },
         "targetArchitecture": {
           "type": "string",
-          "description": "[Only supported in local macOS debugging] The architecture of the debuggee. This will automatically be detected unless this parameter is set. Allowed values are x86_64 or arm64."
+          "enum": [
+            "x86_64",
+            "arm64"
+          ],
+          "description": "The architecture of the debuggee. This will automatically be detected unless this parameter is set. Allowed values are x86_64 or arm64."
```
gregg-miskelly, line 413: I would suggest also adding: `This value is ignored on Linux.` at the end of the description.

gregg-miskelly (changes requested):
Otherwise looks good to me. Thanks for taking the time!

src/tools/OptionsSchema.json
WardenGnaw: Changes to `OptionsSchema.json` also require you to run `gulp generateOptionsSchema`.

See https://github.com/OmniSharp/omnisharp-vscode/blob/master/src/tools/README.md

hajekj: I have added `This value is ignored on Linux.` to the end of the description and re-generated the schema.

WardenGnaw (approved):

gregg-miskelly (approved):

gregg-miskelly:
@hajekj Thanks for taking the time!

PR #5666, opened 2023-05-03
Combine test compile step into normal compile

dibarbet:
This makes it so compile also compiles tests.  There were lots of nullable warnings which I either fixed or suppressed when makes sense.

Unfortunately it looks like chai/mocha methods don't tell the TS compiler when things have been asserted to be non-null, so there are lots of `!`.


test/unitTests/logging/TelemetryObserver.test.ts
```diff
@@ -27,20 +27,20 @@ suite('TelemetryReporterObserver', () => {
...
+                measure.push(measures!);
+                errorProp.push(errorProps!);
             },
         };
     }, useModernNet);
 
     setup(() => {
         name = "";
-        property = null;
+        property = undefined;
```
JoeRobich, line 43: Could we have just set property to `{}` and avoided introducing `undefined`?

test/unitTests/OmnisharpPackageCreator.test.ts
```diff
@@ -31,66 +31,66 @@ suite("GetOmnisharpPackage : Output package depends on the input package and oth
 
         test(`Architectures, binaries and platforms do not change useFramework: ${useFramework}`, () => {
             let testPackage = inputPackages.find(element => (element.platformId && element.platformId == "os-architecture"));
-            let resultPackage = SetBinaryAndGetPackage(testPackage, useFramework, serverUrl, version, installPath);
+            let resultPackage = SetBinaryAndGetPackage(testPackage!, useFramework, serverUrl, version, installPath);
```
JoeRobich, line 34: Could add a helper such as the one in this [TS playground](https://www.typescriptlang.org/play?#code/DYUwLgBAbghsCuIIC4IDt4FsBGIBOEAPuvMMERPGgCYgBmAlmiNRALwQDMA3AFAMBnAHIB7MENLAAFLAQgAlH14BjEWgEjQAOmAiA5lIBEsxBEEQB4AISHFvXiaQcqtRs2pLBo8ZJlxEdrx0VMpgDGpmwmISZAA8ACoAfH5yqPHyKBAwApZ4YALQ-kjmomgxwDDYoAmJEADevBBNZnQQKaZsnSRkRMSO7F0u9EwsGQ3NExBgABZ4IgDu6CCLAKJ4c3hG-eYYPSIEQ24sWrZ8EwC+vOdAA), then the compiler would be happy.

```typescript
function isNotNull<T>(value: T) : asserts value is NonNullable<T> {
    if (value === null || value === undefined) {
        throw new Error("value is null or undefined.");
    }
}
```

tsconfig.json
```diff
@@ -18,13 +18,7 @@
         "gulpfile.ts", // tslint gets quite angry if we don't do this
         "src",
         "tasks",
-        // "test", // Uses its own tsconfig.json until all nullability issues are fixed.
+        "test",
```
JoeRobich, line 21: Does this mean the tests will packaged into the distributed extension?

JoeRobich (approved):
Thanks so much for taking on the job of fixing up the tests. Had a few suggestions but I would be fine if things merged as is.

tsconfig.json
dibarbet: I didn't see any test files included in the extension.js dist folder, so I think we're good there.

test/unitTests/logging/TelemetryObserver.test.ts
dibarbet: In some cases it works, in others the code relies on undefined.  I'll check this one

test/unitTests/OmnisharpPackageCreator.test.ts
dibarbet: I added this, and updated all the places that I thought made sense to use it
1.  Places that were asserting not null using the non-annotated APIs were replaced with this function
2.  Places where the nullable item was used multiple times (and would have required multiple `!`).

Otherwise, I left the `!` usage when the nullable item was used only once

JoeRobich (approved):
Latest changes look good!

PR #5754, opened 2023-06-08
Merge internal fork of LSP tools host back

dibarbet:
Merges the preview version of the extension code back into this repo.  

TODO
1.  ~~Update server publishing to publish to a public feed and consume it here - https://github.com/dotnet/roslyn/pull/68558~~
2.  Create CI pipeline in dnceng-public

In separate PRs:
1.  Create mirror in dnceng/internal
2.  Create official build in dnceng/internal
3.  Create release pipeline in dnceng/internal
4.  Create devkit insertion pipeline in internal


heartacker:
💯

dibarbet:
Need to setup the dnceng pipeline, but I think this might need to merge first

package.json
```diff
@@ -9,11 +9,11 @@
   "icon": "images/csharpIcon.png",
   "preview": false,
   "bugs": {
-    "url": "https://github.com/microsoft/vscode-dotnettools/issues"
+    "url": "https://github.com/dotnet/vscode-csharp"
```
jasonmalinowski, line 12: Was the intention though that the vscode-dotnettools repo is still the primary place to file bugs (and we'll move them around from there?)

package.json
```diff
@@ -9,11 +9,11 @@
...
   },
   "repository": {
     "type": "git",
-    "url": "https://devdiv@dev.azure.com/devdiv/DevDiv/_git/vscode-csharp-next"
+    "url": "https://github.com/dotnet/vscode-csharp"
```
jasonmalinowski, line 16: If this is supposed to be a clone URL, github puts a .git on the end. I think the regular URL works for clones but I'm not sure if that's something they don't want people doing.

server/NuGet.config
```diff
@@ -3,7 +3,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="VS" value="https://devdiv.pkgs.visualstudio.com/_packaging/VS/nuget/v3/index.json" />
+    <add key="vs-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" />
```
jasonmalinowski, line 6: vs-impl is being used due the other workaround due to the signing issues, or is this actually the right feed in this case?

jasonmalinowski (approved):

package.json
dibarbet: No - when you report an issue you select an extension to report against.  If someone selects the C# extension, bugs should go here.

package.json
dibarbet: Nope - this should just be a link to the repo in the marketplace page.

server/NuGet.config
dibarbet: This seems like the right feed for the moment.  However it would be harder to change until I fix signing, so if you think there is a better one, I'm happy to change it later.

jasonmalinowski: It seems a bit strange that we're using vs-impl since I usually imagine that as a feed we consume that contains packages produced by the VS team where in this case we're wearing our .NET hat, not VS hat. Although maybe we already publish other packages there too?

dibarbet: We do indeed, for example all our external access packages.  This seemed to fit with those since its purpose is to be consumed by another one of our components.  Other potential feeds are dotnet-eng or dotnet-tools.  But publishing to any of those needs to wait for my other changes.

dibarbet: Discussed offline- we're going to try and switch to the [msft_consumption ](https://dev.azure.com/azure-public/vside/_artifacts/feed/msft_consumption)feed to ensure that the versions we insert here are retained

PR #5814, opened 2023-06-15
Ship prerelease builds out of main instead of separate branch

dibarbet:
Aligns to the new publishing workflow where we publish weekly builds out of main instead of from a separate prerelease branch.

This removes the automatic CI for the prerelease branch and updates main to produce correctly versioned vsix's.  Once we do next weeks publishing, I'll delete the prerelease branch itself.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -2,7 +2,6 @@ trigger:
   branches:
     include:
     - main
-    - prerelease
 pr: none
 
 variables:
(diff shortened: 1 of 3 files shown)


arkalyanms:
Lets hold off on this PR. Until we can get integration tests health up, I believe the ask is still to stick to prerelease temporarily. I am working with Manish, Michael and Richa to put it down formally.

dibarbet:
> Lets hold off on this PR. Until we can get integration tests health up, I believe the ask is still to stick to prerelease temporarily. I am working with Manish, Michael and Richa to put it down formally.

Ah gotcha, will move to draft.

PR #5834, opened 2023-06-19
Fix issue with Razor commands being registered twice

dibarbet:
Was debugging something else locally and ended up needing to fix this
Resolves https://github.com/dotnet/vscode-csharp/issues/5779

### DIFF
--- a/src/lsptoolshost/roslynLanguageClient.ts
+++ b/src/lsptoolshost/roslynLanguageClient.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { LanguageClient, LanguageClientOptions, ServerOptions } from "vscode-languageclient/node";
+import { LanguageClient, LanguageClientOptions, ServerOptions, Trace } from "vscode-languageclient/node";
 import CompositeDisposable from "../CompositeDisposable";
 import { IDisposable } from "../Disposable";
 
@@ -18,6 +18,7 @@ export class RoslynLanguageClientInstance extends LanguageClient {
         clientOptions: LanguageClientOptions,
         forceDebug?: boolean) {
             super(id, name, serverOptions, clientOptions, forceDebug);
+
             this._disposables = new CompositeDisposable();
     }
 
--- a/src/lsptoolshost/roslynLanguageClient.ts
+++ b/src/lsptoolshost/roslynLanguageClient.ts
@@ -8,7 +8,7 @@ import CompositeDisposable from "../CompositeDisposable";
 import { IDisposable } from "../Disposable";
 
 /**
- * Implementation of the base LangaugeClient type that allows for additional items to be disposed of
+ * Implementation of the base LanguageClient type that allows for additional items to be disposed of
  * when the base LanguageClient instance is disposed.
  */
 export class RoslynLanguageClient extends LanguageClient {
--- a/src/lsptoolshost/roslynLanguageClient.ts
+++ b/src/lsptoolshost/roslynLanguageClient.ts
@@ -3,11 +3,15 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { LanguageClient, LanguageClientOptions, ServerOptions, Trace } from "vscode-languageclient/node";
+import { LanguageClient, LanguageClientOptions, ServerOptions } from "vscode-languageclient/node";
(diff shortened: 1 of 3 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -472,40 +442,75 @@ export class RoslynLanguageServer {
 
         // Razor will call into us (via command) for generated file didChange/didClose notifications. We'll then forward these
         // notifications along to Roslyn. didOpen notifications are handled separately via the vscode.openTextDocument method.
-        vscode.commands.registerCommand(RoslynLanguageServer.roslynDidChangeCommand, (notification: DidChangeTextDocumentParams) => {
+        client.addDisposable(vscode.commands.registerCommand(RoslynLanguageServer.roslynDidChangeCommand, (notification: DidChangeTextDocumentParams) => {
```
dibarbet, line 445: before, on restarts we were trying to register the command again.  Now we register the command as a disposable that will get disposed of when the server shuts down, so we can register it again on the new instance.

davidwengier (approved):
Surely this fixes _all_ commands not being registered twice. Razor is being unfairly singled out! 😛

allisonchou (approved):

src/lsptoolshost/roslynLanguageClient.ts
```diff
@@ -0,0 +1,40 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { LanguageClient, LanguageClientOptions, ServerOptions } from "vscode-languageclient/node";
+import CompositeDisposable from "../CompositeDisposable";
+import { IDisposable } from "../Disposable";
+
+/**
+ * Implementation of the base LangaugeClient type that allows for additional items to be disposed of
```
jasonmalinowski, line 11:
```suggestion
 * Implementation of the base LanguageClient type that allows for additional items to be disposed of
```

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: Is there a reason we're not just registering these during extension activation (assuming we're not in Omnisharp mode) and instead having some other indirection around ensuring they're hitting the current client?

src/utils/getCSharpDevKit.ts
```diff
@@ -7,6 +7,8 @@ import * as vscode from 'vscode';
 import { CSharpDevKitExports } from "../CSharpDevKitExports";
 
 export const csharpDevkitExtensionId = "ms-dotnettools.csdevkit";
+export const csharpDevkitIntelliCodeExtensionId = "ms-dotnettools.vscodeintellicode-csharp";
+
```
jasonmalinowski, line 11: Extra blank line.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -472,40 +442,75 @@ export class RoslynLanguageServer {
...
+                let message = `Detected installation of ${csharpDevkitExtensionId}. Would you like to relaunch the language server for added features?`;
+                ShowInformationMessage(vscode, message, { title, command });
+            } else {
+                // Any other change to extensions is irrelevant - an uninstall requires a reload of the window
+                // which will automatically restart this extension too.
+            }
+        }));
+    }
+
+    private registerTelemtryChanged(languageClient: RoslynLanguageClient) {
```
jasonmalinowski, line 506:
```suggestion
    private registerTelemetryChanged(languageClient: RoslynLanguageClient) {
```

jasonmalinowski (approved):
Seems reasonable, although I still wonder if we should just be registering the commands once and that's it.

src/lsptoolshost/roslynLanguageServer.ts
dibarbet: @allisonchou @davidwengier do you know?  Do these need to be registered after the server is created, or can I move them?

dibarbet: I'll move just the command registration earlier.  I don't think it can hurt if the commands are being registered earlier (if they were being called they would throw with missing command before, now they'll throw with server not started instead)

davidwengier: No specific knowledge from me.

allisonchou: Yep, I think that change should be fine!

ryzngard (approved):

PR #5837, opened 2023-06-20
Implement report issue command for prerelease extension

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/5747


src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -0,0 +1,102 @@
...
+export const DotNetRuntimeVersion = '7.0';
+
+interface IDotnetAcquireResult {
+    dotnetPath: string;
+}
+
+/**
+ * Resolves the dotnet runtime from given options and the dotnet runtime VSCode extension.
+ */
+export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
```
dibarbet, line 25: moved the runtime resolution into the `IHostExecutableResolver` so we can get this information in other places (and share more code with O#).

dibarbet: this is all essentially just moved

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -641,6 +602,39 @@ export async function activateRoslynLanguageServer(context: vscode.ExtensionCont
     _languageServer.start();
 }
 
+export function getServerPath(options: Options, platformInfo: PlatformInformation) {
```
dibarbet, line 605: moved to re-use

src/observers/OpenURLObserver.ts
```diff
@@ -1,23 +0,0 @@
...
-import { vscode } from "../vscodeAdapter";
-import { BaseEvent, OpenURL } from "../omnisharp/loggingEvents";
-import { EventType } from "../omnisharp/EventType";
-
-export class OpenURLObserver {
-
-    constructor(private vscode: vscode) {
-    }
-
-    public post = (event: BaseEvent) => {
```
dibarbet, line 15: this seemed entirely unnecessary - an event + observer just for a single location to open a URL

src/shared/reportIssue.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+    catch (error) {
+        monoVersion = `There is a problem with running OmniSharp on mono: ${error}`;
+    }
+
+    return `<details><summary>Mono Information</summary>
+    ${monoVersion}</details>`;
+}
+
+function getLogInfo(useOmnisharp: boolean): string {
```
dibarbet, line 115: this is a new bit

src/shared/reportIssue.ts
```diff
@@ -0,0 +1,142 @@
...
+    return 0;
+}
+
+function generateExtensionTable(extensions: Extension<any>[]) {
+    if (extensions.length <= 0) {
+        return "none";
+    }
+
+    const tableHeader = `|Extension|Author|Version|Folder Name|\n|---|---|---|---|`;
+    const table = extensions.map((e) => `|${e.packageJSON.name}|${e.packageJSON.publisher}|${e.packageJSON.version}|${basename(e.extensionPath)}|`).join("\n");
```
dibarbet, line 92: this is all the same, except I added the extension folder name.  This is the only place where it will show which architecture the extension comes from.  Useful because we've had issues with the marketplace installing the wrong architecture version.

src/shared/reportIssue.ts
```diff
@@ -0,0 +1,142 @@
...
+${monoInfo}
+<details><summary>Dotnet Information</summary>
+${fullDotnetInfo}</details>
+<details><summary>Visual Studio Code Extensions</summary>
+${generateExtensionTable(extensions)}
+</details>
+`;
+
+    const queryStringPrefix: string = "?";
+    const issueDefault = "Please paste the output from your clipboard";
```
jasonmalinowski, line 69: Do I dare ask why we don't just fill in the body directly since we have a way to do that?

src/shared/reportIssue.ts
```diff
@@ -0,0 +1,142 @@
...
+
+${monoInfo}
+<details><summary>Dotnet Information</summary>
+${fullDotnetInfo}</details>
+<details><summary>Visual Studio Code Extensions</summary>
+${generateExtensionTable(extensions)}
+</details>
+`;
+
+    const queryStringPrefix: string = "?";
```
jasonmalinowski, line 68: Why not just inline this?

src/shared/reportIssue.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+async function getMonoIfPlatformValid(options: Options, monoResolver: IHostExecutableResolver): Promise<string> {
+    let monoVersion = "Unknown Mono version";
+    try {
+        const monoInfo = await monoResolver.getHostExecutableInfo(options);
+        monoVersion = `OmniSharp using mono: ${monoInfo.version}`;
+    }
+    catch (error) {
+        monoVersion = `There is a problem with running OmniSharp on mono: ${error}`;
```
jasonmalinowski, line 108: The other place we're catching exceptions we have a different approach for converting it to a string (the instanceof check). Should we make them the same?

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -0,0 +1,102 @@
...
+    async getHostExecutableInfo(options: Options): Promise<HostExecutableInformation> {
+        let dotnetRuntimePath = options.commonOptions.dotnetPath;
+        const serverPath = getServerPath(options, this.platformInfo);
+        if (!dotnetRuntimePath)
+        {
+            let dotnetInfo = await this.acquireDotNetProcessDependencies(serverPath);
+            dotnetRuntimePath = path.dirname(dotnetInfo.path);
+        }
+
+        // TODO need exe?
```
jasonmalinowski, line 39: Not sure what this TODO means.

jasonmalinowski (approved):

src/shared/reportIssue.ts
dibarbet: Just the name of the immediate parent - I didn't include the entire path for that exact reason :)

test/unitTests/features/reportIssue.test.ts
```diff
@@ -37,19 +36,29 @@ suite(`${reportIssue.name}`, () => {
...
-    let eventStream: EventStream;
-    let eventBus: TestEventBus;
-    let getDotnetInfo = FakeGetDotnetInfo;
+    let getDotnetInfo = async () => Promise.resolve(fakeDotnetInfo);
     let options: Options;
     let issueBody: string;
 
     setup(() => {
         vscode = getFakeVsCode();
+        // @ts-ignore
```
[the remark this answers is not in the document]
dibarbet: It needs to return a vscode.Uri, but the only way I know how to get one is to call vscode.Uri.Parse... which we're implementing here.  Since the result of this call doesn't matter for the unit test anyway, I return the same input value.  Since it isn't a vscode.Uri type though I need to suppress the TS warnings.

When we get integration tests I want to delete these unit tests in favor of integration tests anyway.

src/shared/reportIssue.ts
dibarbet: I think it is because of potential URI escaping issues (the body gets put into the actual URI).  Might be possible still, I can investigate it later.  For now I'll leave it since this is the existing behavior.

src/shared/reportIssue.ts
dibarbet: I didn't change this part (despite what git says).  I can make this better though.

src/shared/reportIssue.ts
dibarbet: Potentially - again I didn't modify this either (even though git says I did).  I'll fix it up though.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
dibarbet: was copy / pasted.  I'll remove.

test/unitTests/features/reportIssue.test.ts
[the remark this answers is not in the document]
jasonmalinowski: @dibarbet The VSCode URI package is referencable separately: https://github.com/microsoft/vscode-uri. So it's pretty easy to add a reference to that and then have a "real" parse method in play.

src/shared/reportIssue.ts
jasonmalinowski: Yeah, I don't know why we can't just call an escape method here. :-/

dibarbet: its not as simple as doing encodeURIComponent and passing it along, I did try that, but you get this:
image

So will dig into this more later.

dibarbet:
@WardenGnaw mind taking a quick look at the debugger change?

package.json
```diff
@@ -856,17 +857,17 @@
                     "items": {
                       "type": "string"
                     },
-                    "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                    "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
```
WardenGnaw, line 860: Please run `gulp generateOptionsSchema` to fix the example urls in `package.json`

src/coreclr-debug/util.ts
```diff
@@ -127,11 +127,11 @@ export function getTargetArchitecture(platformInfo: PlatformInformation, launchJ
     }
 
     // Otherwise, look at the runtime ID.
-    if (dotnetInfo.RuntimeId.includes('arm64'))
+    if (dotnetInfo.RuntimeId?.includes('arm64'))
```
WardenGnaw, line 130: If we are starting to debug but only have a runtime-only dotnet, should we have an error message to install an SDK runtime?

@gregg-miskelly

test/unitTests/features/reportIssue.test.ts
```diff
@@ -37,19 +37,28 @@ suite(`${reportIssue.name}`, () => {
...
         },
-        id: "id2"
+        id: "id2",
+        extensionPath: "c:/extensions/xyz-x64"
+    };
+
+    const fakeDotnetInfo: DotnetInfo = {
+        FullInfo: "myDotnetInfo",
+        Version: "1.0.x",
+        RuntimeId: "1.1.x"
```
WardenGnaw, line 47: Should this be a real fake value like `win10-x64`?

src/coreclr-debug/util.ts
gregg-miskelly: I don't know of any reason why we couldn't debug without an SDK.

WardenGnaw: What about:
```
if (!dotnetInfo.RuntimeId) {
   throw new Error(`Unable to determine RuntimeId. Please set 'targetArchitecture' in your launch.json configuration.`)
} else {
   if (dotnetInfo.RuntimeId.Contains....)

   ....

   throw new Error...
}
```

test/unitTests/features/reportIssue.test.ts
dibarbet: For this particular test it doesn't make a difference, since report issue only cares about the FullInfo.  But I'll change it to something more real

package.json
dibarbet: I wish we could fix this.  It gets automatically changed every time you install or update a dependency.

gregg-miskelly: FYI the update dependencies script takes care of fixing up package.json.

src/shared/reportIssue.ts
dibarbet: Great suggestion.  I'm using that
image

WardenGnaw (approved):
:shipit:

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -0,0 +1,101 @@
...
+export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
+    constructor(private platformInfo: PlatformInformation) { }
+
+    private hostInfo: HostExecutableInformation | undefined;
+
+    async getHostExecutableInfo(options: Options): Promise<HostExecutableInformation> {
+        let dotnetRuntimePath = options.commonOptions.dotnetPath;
+        const serverPath = getServerPath(options, this.platformInfo);
+        if (!dotnetRuntimePath)
+        {
```
jaredpar, line 34:
```suggestion
        if (!dotnetRuntimePath) {
```

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Resolves the dotnet runtime from given options and the dotnet runtime VSCode extension.
+ */
+export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
+    constructor(private platformInfo: PlatformInformation) { }
+
+    private hostInfo: HostExecutableInformation | undefined;
+
+    async getHostExecutableInfo(options: Options): Promise<HostExecutableInformation> {
+        let dotnetRuntimePath = options.commonOptions.dotnetPath;
+        const serverPath = getServerPath(options, this.platformInfo);
```
jaredpar, line 32: This seems wrong in the current form. This is a generic .NET resolver but it's calling into roslyn's specific `getServerPath` method. Shouldn't we make this fully generic and have the paths / file namse passed in so we can share with razor?

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -0,0 +1,101 @@
...
+import { IHostExecutableResolver } from "../shared/constants/IHostExecutableResolver";
+import { getServerPath } from "../lsptoolshost/roslynLanguageServer";
+import { PlatformInformation } from "../shared/platform";
+import { Options } from "../shared/options";
+import { existsSync } from 'fs';
+import { CSharpExtensionId } from '../constants/CSharpExtensionId';
+
+export const DotNetRuntimeVersion = '7.0';
+
+interface IDotnetAcquireResult {
```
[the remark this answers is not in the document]
dibarbet: The dotnet runtime installer API returns an object of string, not a string itself, so this definition needs to match.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
dibarbet: Yeah - when I originally wrote it, I hadn't thought about sharing it with Razor yet.  I'll update this.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Resolves the dotnet runtime for a server executable from given options and the dotnet runtime VSCode extension.
+ */
+export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
+    constructor(
+        private platformInfo: PlatformInformation,
+        /**
+         * This is a function instead of a string because the server path can change while the extension is active (when the option changes).
+         */
+        private getServerPath: (options: Options, platform: PlatformInformation) => string) { }
```
dibarbet, line 30: A couple reasons for doing it this way
1.  As mentioned in the comment it needs to be a function for when the server path changes (from the option).  One `IHostExecutableResolver` is created during activation for roslyn, and re-use across server restarts.
2.  It didn't seem appropriate to update `getHostExecutableInfo` because not all implementations of `IHostExecutableResolver` require a server path (for example used in O# with dotnet and mono as well).  Its more of a property of this specific resolver.

src/shared/reportIssue.ts
```diff
@@ -65,11 +64,10 @@ ${generateExtensionTable(extensions)}
...
 
-    const queryStringPrefix: string = "?";
-    const issueDefault = "Please paste the output from your clipboard";
-    const fullUrl = `${issuesUrl}${queryStringPrefix}body=${issueDefault}`;
-    await vscode.env.clipboard.writeText(body);
-    vscode.env.openExternal(vscode.Uri.parse(fullUrl));
+    await vscode.commands.executeCommand("workbench.action.openIssueReporter", {
+        extensionId: CSharpExtensionId,
+        issueBody: body
+    });
```
jasonmalinowski, line 70: Nifty!

jasonmalinowski (approved):

PR #5839, opened 2023-06-21
Fix Razor completion tooltip documentation

allisonchou:
Razor completion tooltips were previously missing for C# items and HTML snippets.
The tooltips are sometimes cut off (see screenshot), but this is a problem with C# files as well and is likely an editor issue.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1831344/

Before:
image

After:
image

### DIFF
--- a/src/razor/src/Completion/RazorCompletionItemProvider.ts
+++ b/src/razor/src/Completion/RazorCompletionItemProvider.ts
@@ -13,7 +13,7 @@ import { getUriPath } from '../UriPaths';
 import { ProvisionalCompletionOrchestrator } from './ProvisionalCompletionOrchestrator';
 import { LanguageKind } from '../RPC/LanguageKind';
 import { RoslynLanguageServer } from '../../../lsptoolshost/roslynLanguageServer';
-import { CompletionItem, CompletionParams, CompletionTriggerKind } from 'vscode-languageclient';
+import { CompletionItem, CompletionList, CompletionParams, CompletionTriggerKind } from 'vscode-languageclient';
 import { UriConverter } from '../../../lsptoolshost/uriConverter';
 import * as RazorConventions from '../RazorConventions';
 import { MappingHelpers } from '../Mapping/MappingHelpers';
@@ -35,11 +35,6 @@ export class RazorCompletionItemProvider
 
             let completions: vscode.CompletionList | vscode.CompletionItem[];
 
-            // For CSharp, completions need to keep the "data" field
-            // on the completion item for lazily resolving the edits in
-            // the resolveCompletionItem step. Using the vs code command
-            // drops that field because it doesn't exist in the declared vs code
-            // CompletionItem type.
             if (language === LanguageKind.CSharp) {
                 const params: CompletionParams = {
                     context: {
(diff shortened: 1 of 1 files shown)


src/razor/src/Completion/RazorCompletionItemProvider.ts
```diff
@@ -71,6 +66,13 @@ export class RazorCompletionItemProvider
                 completions instanceof Array ? completions  // was vscode.CompletionItem[]
                     : completions ? completions.items       // was vscode.CompletionList
                         : [];
+            
+            // For CSharp, completions need to keep the "data" field
```
allisonchou, line 70: Comment moved down - I feel like it makes more sense here.

src/razor/src/Completion/RazorCompletionItemProvider.ts
```diff
@@ -128,6 +130,8 @@ export class RazorCompletionItemProvider
                         completionItem.insertText = intellicodeCompletion.textEditText;
                     }
                 }
+
+                (<CompletionItem>completionItem).data = data;
```
davidwengier, line 134: Does this override the data with the default in all cases, even if the completion item has provided its own?

src/razor/src/Completion/RazorCompletionItemProvider.ts
ryzngard: Thank you :)

ryzngard: Oh actually, this is subtle... it might make sense to put before the call to `RoslynLanguageServer.provideCompletionsCommand`. It's the reason that command exists

src/razor/src/Completion/RazorCompletionItemProvider.ts
ryzngard: afaik we shouldn't be modifying this

src/razor/src/Completion/RazorCompletionItemProvider.ts
```diff
@@ -204,6 +208,13 @@ export class RazorCompletionItemProvider
 
             item = newItem;
 
+            // The documentation object Roslyn returns can have a different
+            // shape than what the client expects, so we do a conversion here.
+            const markdownString = <vscode.MarkdownString>(item.documentation);
```
ryzngard, line 213: This is weird... and probably against my understanding of typescript. We're casting to `vscode.MarkdownString` to make a new one? Does something magic happen in the constructor?

allisonchou: Typescript seems to be weird in that we can have a `vscode.MarkdownString` but the properties on the type may not be what is expected. 

For example, at the beginning of this method we get passed something that _technically_ passes as a `vscode.MarkdownString`, but has totally different properties:
image

After the cast above and the creation of a new `MarkdownString` in the lines below, we're able to return the shape VS Code expects:
image

davidwengier: I suspect this is actually wrong, but not in a way that makes a difference. I would guess that Roslyn LSP is returning the LSP type [MarkupContent](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition), which has a kind and a value. This code is casting it to a `vscode.MarkdownString` which also has a value, so it can read the value property without a compilation error. Since types, and therefore casts, don't really exist in the compiled JavaScript this works fine. The creation of the new markdown string is then really a conversion from the LSP type to the VS Code type.

davidwengier (commented):
I think there are probably issues with the types used in this provider as a whole, but the completion resolve stuff was always a hack, so happy that this fixes a bug and moves us forward.

davidwengier (approved):

src/razor/src/Completion/RazorCompletionItemProvider.ts
allisonchou: Appreciate the insight! That would make sense actually. I pushed a change that checks for `MarkupContent` instead, it's a little hacky since we have to convert to `unknown` first, but it's probably better than how it was before.

PR #5850, opened 2023-06-22
Pass the configured VSCode language to the server

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/5716

Now uses VSCode's configured language instead of the OS language when launching the language server.

For example, in the following screenshot my OS is configured in English, but I've set VSCode's display language to Japanese.  Previously we would have output strings in the OS language (English), but now we output them in VSCode's display language.
image

Note that this is not full localization support - strings defined on the client extension side are not localized yet.  This just fixes the server outputting the wrong localization.

Requires a server update with https://github.com/dotnet/roslyn/pull/68739


dibarbet:
Converting to draft - switching to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams

dibarbet:
Not needed, LSP provides this info to the server.

PR #5855, opened 2023-06-23
Allow rzls to launch using standard dotnet

jaredpar:
This PR has two purposes:

1. Carving a path out where by we can ship rzls in a non-self contained fashion.
2. Unify the `dotnet` acquisition and launch process between the razor and roslyn servers

This is my first every TypeScript PR so please be gentle :smile:

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -38,7 +38,7 @@ import {
     CompletionItem,
 } from 'vscode-languageclient/node';
 import { PlatformInformation } from '../shared/platform';
-import { acquireDotNetProcessDependencies } from './dotnetRuntime';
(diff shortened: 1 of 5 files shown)


dibarbet (changes requested):
I apologize, I'm about to change this code in https://github.com/dotnet/vscode-csharp/pull/5837

It should however have a re-usable piece for you to use in Razor (a dotnetHostExecutableResolver)

dibarbet:
This is the bit I think you could pass into Razor and re-use
https://github.com/dotnet/vscode-csharp/pull/5837/files#diff-d5522fe2afc29ab947fe42d4a313c64bbbeb4cd1d3c105dbbe25a287eafbc6a6

jaredpar:
Well at least it looks like I was on the right track here.

PR #5857, opened 2023-06-23
Fix debugging O# integration tests on preview version

dibarbet:
The launch.json configurations for debugging O# integration tests were not working well for a couple reasons
1.  It had disable all extensions, meaning the dotnet runtime extension was not present and therefore C# failed to activate.
2.  Source maps weren't being included for the test files.

This isn't where we want to be long term for integration tests, but its a bit better.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -41,7 +41,9 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
-                "--disable-extensions",
+                // Create a temp profile that has no extensions / user settings.
+                // This allows us to only have the C# extension + the dotnet runtime installer extension dependency.
+                "--profile-temp",
                 "--extensionDevelopmentPath=${workspaceRoot}",
                 "--extensionTestsPath=${workspaceRoot}/out/test/featureTests"
             ],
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -41,7 +41,9 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
-                "--disable-extensions",
+                // Create a temp profile that has no extensions / user settings.
+                // This allows us to only have the C# extension + the dotnet runtime installer extension dependency.
+                "--profile-temp",
```
dibarbet, line 46: This launches vscode with an empty profile with no user settings / extensions.  This is better than --disable-extensions as it 
1.  Resets settings
2.  Allows us to install dependent extensions if needed.

I would have preferred if we could launch into a specific integration test profile, but that requires the profile to be imported first via a UI wizard.  There's no way to launch and import a profile via the vscode cli that I could find.

See https://github.com/microsoft/vscode/issues/159572#issuecomment-1231859949

arkalyanms (approved):

akhera99 (approved):

PR #5858, opened 2023-06-24
Support opening of source-generated files over LSP

jasonmalinowski:
This depends on https://github.com/dotnet/roslyn/pull/68771 for the server-side piece. But, if you don't have the server side piece, we won't return a URI that will trigger this code, so it's fine to have it in now.

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -61,6 +61,7 @@ import { IDisposable } from '../disposable';
 import { registerNestedCodeActionCommands } from './nestedCodeAction';
 import { registerRestoreCommands } from './restore';
 import { BuildDiagnosticsService } from './buildDiagnosticsService';
+import { registerSourceGeneratedFilesContentProvider } from './sourceGeneratedFilesContentProvider';
 
 let _channel: vscode.OutputChannel;
 let _traceChannel: vscode.OutputChannel;
@@ -899,6 +900,8 @@ export async function activateRoslynLanguageServer(
     registerDebugger(context, languageServer, languageServerEvents, platformInfo, _channel);
 
     registerRestoreCommands(context, languageServer, dotnetChannel);
+
+    registerSourceGeneratedFilesContentProvider(context, languageServer);
 
     registerOnAutoInsert(languageServer);
 
--- a/src/lsptoolshost/roslynProtocol.ts
+++ b/src/lsptoolshost/roslynProtocol.ts
@@ -170,6 +170,14 @@ export interface ProjectNeedsRestoreName {
     projectFilePaths: string[];
 }
 
+export interface SourceGeneratorGetRequestParams {
+    textDocument: lsp.TextDocumentIdentifier;
+}
+
+export interface SourceGeneratedDocumentText {
+    text: string;
+}
+
 export namespace WorkspaceDebugConfigurationRequest {
     export const method = 'workspace/debugConfiguration';
     export const messageDirection: lsp.MessageDirection = lsp.MessageDirection.clientToServer;
@@ -272,4 +280,10 @@ export namespace ProjectNeedsRestoreRequest {
     export const method = 'workspace/_roslyn_projectNeedsRestore';
     export const messageDirection: lsp.MessageDirection = lsp.MessageDirection.serverToClient;
     export const type = new lsp.RequestType<ProjectNeedsRestoreName, void, void>(method);
+}
+
+export namespace SourceGeneratorGetTextRequest {
+    export const method = 'sourceGeneratedFile/_roslyn_getText';
+    export const messageDirection: lsp.MessageDirection = lsp.MessageDirection.clientToServer;
+    export const type = new lsp.RequestType<SourceGeneratorGetRequestParams, SourceGeneratedDocumentText, void>(method);
 }
--- a/src/lsptoolshost/sourceGeneratedFilesContentProvider.ts
+++ b/src/lsptoolshost/sourceGeneratedFilesContentProvider.ts
@@ -1,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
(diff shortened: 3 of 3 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -662,6 +663,17 @@ function registerRazorCommands(context: vscode.ExtensionContext, languageServer:
     }));
 }
 
+function registerSourceGeneratedFilesContentProvider(context: vscode.ExtensionContext, languageServer: RoslynLanguageServer) {
+    context.subscriptions.push(vscode.workspace.registerTextDocumentContentProvider('source-generated', new class implements vscode.TextDocumentContentProvider {
+        // TODO: also implement eventing for when the file changes
```
jasonmalinowski, line 668: For now I'm just doing a small implementation of TextDocumentContentProvider here; the Omnisharp side has a fuller implementation that does support eventing but will require a bit of refactoring to be able to reuse.

CyrusNajmabadi:
Will review tomorrow!

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -734,6 +736,27 @@ function registerRazorCommands(context: vscode.ExtensionContext, languageServer:
...
+            new (class implements vscode.TextDocumentContentProvider {
+                async provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): Promise<string> {
+                    const result = await languageServer.sendRequest(
+                        RoslynProtocol.SourceGeneratorGetTextRequest.type,
+                        { textDocument: { uri: UriConverter.serialize(uri) } },
+                        token
+                    );
+                    return result.text;
+                }
+            })()
```
CyrusNajmabadi, line 755: so i personally really dont' like these sorts of anonymous inner classes.  my pref is a distinct class with clear cpatures.  but i don't feel strongly enough to block.  if you feel like this is ok, then i'm ok with it.

src/lsptoolshost/roslynProtocol.ts
```diff
@@ -100,3 +108,9 @@ export namespace ProjectInitializationCompleteNotification {
     export const messageDirection: MessageDirection = MessageDirection.serverToClient;
     export const type = new NotificationType(method);
 }
+
+export namespace SourceGeneratorGetTextRequest {
+    export const method = 'sourceGenerator/getText';
```
CyrusNajmabadi, line 113: this is roslyn-specific right?  curious if it might make sense to add some sort of roslyn_ prefix on things specific to us.

this is a non-blocking question for this PR.  

@dibarbet for his thoughts too.

.vscode/settings.json
```diff
@@ -16,5 +16,6 @@
     "mocha.enabled": true,
     "omnisharp.autoStart": false,
     "editor.formatOnSave": false,
-    "eslint.lintTask.enable": true
+    "eslint.lintTask.enable": true,
+    "dotnet.defaultSolution": "disable"
```
CyrusNajmabadi, line 20: what's this about?

CyrusNajmabadi (approved):

src/lsptoolshost/roslynProtocol.ts
dibarbet: Agree, we should prefix roslyn methods.  Maybe similar to the VS protocol does `_vs_` we could do `_roslyn_`

.vscode/settings.json
dibarbet: Ah this is useful - prevents devkit from trying to activate on the random test assets we hgave.

dibarbet (approved):

.vscode/settings.json
jasonmalinowski: Yep, there's no "normal" C# solution in this repository, but a bunch of test projects. So without this any time you open the repository Dev Kit will prompt encouraging to open one of those, which you never need to do.

src/lsptoolshost/roslynProtocol.ts
jasonmalinowski: @dibarbet Should I start with this one then and we go back to the other ones later?

dibarbet: Yeah lets start here, we can fix other ones in separate PRs

CyrusNajmabadi: love it. :)

.vscode/settings.json
CyrusNajmabadi: thaks!

moander:
Any chance this will get merged soon?

pragone:
I thought I'd leave a comment here thanking you for the great work, and also echoing the ask to merge this soon.
This is making development in projects with source generated code quite difficult. Because as soon as you hit this bug the best thing to do is to restart VSCode.
Thanks again

jasonmalinowski:
@glen-84 Thanks for the friendly ping! Hopefully can get this in a week or two -- I recognize that's what I said a few weeks go but that prior work turned into more than expected. But that work is now cleared so this is towards the top.

.vscode/settings.json
jasonmalinowski: This bit ultimately got taken as another PR around that time.

src/lsptoolshost/roslynProtocol.ts
jasonmalinowski: Rebased and this is now called sourceGeneratedFile/_roslyn_getText, which matches how we're naming other things.

moander:
@jasonmalinowski any updates on this one? I'm really looking forward to this patch. Have a nice weekend! :)

jasonmalinowski:
@moander I _think_ I'm going to have the Roslyn side of the change today, assuming no other random fires come up. (They've been coming up a lot recently. 😢 )

jacobjmarks:
Bump on this; I believe @jasonmalinowski has gone on leave. Are we able to reassign?

zaevi:
Is there any progress?

Delsin-Yu:
Will there be any further progress on this? cc @dibarbet

penev92:
Bump!
This is proving to be quite the blocker for us at https://github.com/OpenRA/OpenRA and for many others, I'm sure. Is @jasonmalinowski able to finish it? Is someone able to take over? Can someone lay out a plan for what else is needed so someone else can take over?

jasonmalinowski:
Closing in favor of https://github.com/dotnet/vscode-csharp/pull/7581 which merged this change.

PR #5867, opened 2023-06-26
Scope down files watched in workspace

arkalyanms:
Perhaps the per language files could even be skipped - cs, razor, cshtml.


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -137,8 +137,18 @@ export class RoslynLanguageServer {
             // Register the server for plain csharp documents
             documentSelector: documentSelector,
             synchronize: {
-                // Notify the server about file changes to '.clientrc files contain in the workspace
-                fileEvents: vscode.workspace.createFileSystemWatcher('**/*.*')
+                // Notify the server about file changes to all supported files contained in the workspace
+                fileEvents: [
+                    vscode.workspace.createFileSystemWatcher('**/*.sln'),
```
dibarbet, line 142: I don't actually think we need any of these to be honest.  We already should be watching the relevant csharp files based on the `documentSelector` above.  For the sln / csproj files, I think the server should be registering them itself (instead of hardcoding this on the client).  See https://github.com/dotnet/roslyn/blob/main/src/Features/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/FileWatching/LspFileChangeWatcher.cs#L124

However, would like @jasonmalinowski to confirm that he's not expecting automatic registration for these files

jasonmalinowski: Hmm, I actually didn't realize this existed -- so does this mean we get LSP notifications for these files even if we didn't request it? So this might explain then why it seemed to be working despite the bug that was fixed in https://github.com/dotnet/roslyn/pull/68741 -- my requests for things to watch wasn't doing anything, but this meant the notifications were sent either way.

But assuming everything is working correctly now, yes, I wouldn't expect us to have anything here -- we'll request what we need with better granularity, and we'll also handle things that aren't included here (like additional files).

arkalyanms: @jasonmalinowski I believe we tested this offline to confirm the docselector provides the needed watchers. Can this be merged now?

arkalyanms: Auto merge button has been hit. Please feel free to sign off when ready.

dibarbet (approved):

PR #5868, opened 2023-06-27
Enable full solution background analysis

mavasani:
Closes https://github.com/dotnet/roslyn/issues/68553

- Adds 2 new background analysis scope options: one for compiler diagnostics and one for analyzer diagnostics (options match VS)
- Both the options allow the same set of values as in VS: Open documents, Entire solution, Current document and None (disabled). Currently, `Current document` option is not yet supported in LSP as we do not have a way to know which is the current active document in LSP server.
- Builds on top of https://github.com/dotnet/roslyn/pull/68799
- Needs couple more Roslyn side changes as a follow-up: https://github.com/dotnet/roslyn/issues/68797 and ~~https://github.com/dotnet/roslyn/issues/68798~~


package.json
```diff
@@ -1274,6 +1274,42 @@
             "description": "Show regular expressions in completion list.",
             "order": 20
           },
+          "dotnet.backgroundAnalysis.solutionCrawlerBackgroundAnalysisScope": {
+            "type": "string",
+            "enum": [
+              "activeFile",
+              "openFiles",
+              "fullSolution",
+              "none"
```
mavasani, line 1283: Enum values match Roslyn side enum names: https://github.com/dotnet/roslyn/blob/bdbf1cb87abc8f9554ee331a1113d533ae5ec5fb/src/Workspaces/Core/Portable/SolutionCrawler/BackgroundAnalysisScope.cs#L7-L34

package.json
```diff
@@ -1274,6 +1274,42 @@
...
+            ],
+            "default": "openFiles",
```
mavasani, line 1285: This should change to `activeFile` once the LSP server can support it.

package.json
```diff
@@ -1274,6 +1274,42 @@
...
+            "enumDescriptions": [
+              "Current document (not yet supported)",
+              "Open documents",
+              "Entire solution",
+              "None"
+            ],
+            "description": "Run background code analysis for:",
```
mavasani, line 1292: Descriptions match VS side descriptions in Tools Options dialog

package.json
```diff
@@ -1274,6 +1274,42 @@
             "description": "Show regular expressions in completion list.",
             "order": 20
           },
+          "dotnet.backgroundAnalysis.solutionCrawlerBackgroundAnalysisScope": {
+            "type": "string",
+            "enum": [
+              "activeFile",
```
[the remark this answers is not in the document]
mavasani: @dibarbet Removing active file worked just fine. I have done the change. I'll let @Cosifne answer your other query about renaming the option name completely.

CyrusNajmabadi:
I agree that we should not talk about solution crawling.

test/unitTests/lspToolHost/configurationMiddleware.test.ts
```diff
@@ -45,6 +45,8 @@ const testData = [
     { serverOption: "csharp|code_style.formatting.indentation_and_spacing.indent_style", vsCodeConfiguration: "codeStyle.formatting.indentationAndSpacing.indentStyle", declareInPackageJson: false},
     { serverOption: "csharp|code_style.formatting.new_line.end_of_line", vsCodeConfiguration: "codeStyle.formatting.newLine.endOfLine", declareInPackageJson: false},
     { serverOption: "code_style.formatting.new_line.insert_final_newline", vsCodeConfiguration: "codeStyle.formatting.newLine.insertFinalNewline", declareInPackageJson: false},
+    { serverOption: "mystery_language|background_analysis.dotnet_solution_crawler_background_analysis_scope", vsCodeConfiguration: "dotnet.backgroundAnalysis.solutionCrawlerBackgroundAnalysisScope", declareInPackageJson: true},
```
Cosifne, line 48: Each option should have two tests (because it is a per-lang option)
`csharp|background_analysis.dotnet_solution_crawler_background_analysis_scope` -> `dotnet.backgroundAnalysis.solutionCrawlerBackgroundAnalysisScope`
and
`mystery_language|background_analysis.dotnet_solution_crawler_background_analysis_scope` -> `null`

Only C# option should be marked as `declareInPackageJson: true`

It is in this way because for per-language option the server would ask the option value for all the supported languages.
And our client would only reply C# related options.

mavasani: Thanks @Cosifne - fixed now.

package.json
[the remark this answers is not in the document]
Cosifne: Ah, I forget to send my comments.
The general rule for naming is to "match" what's on the server. Like here, if you don't like `SolutionCrawler` then it's better to remove the name from the server.
In theory, we can do whatever we like in the client by creating a hashmap, and mapping from the server name to the client-side name, but it should be avoided when possible.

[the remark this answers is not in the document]
mavasani: > Like here, if you don't like SolutionCrawler then it's better to remove the name from the server

@Cosifne - wouldn't that cause a breaking change in VS though? Renaming the option on server means now any user on VS who had changed their option value to something else will lose their changed settings.

src/lsptoolshost/optionNameConverter.ts
```diff
@@ -29,7 +29,13 @@ export function convertServerOptionNameToClientConfigurationName(section: string
...
         const optionNamePrefix = getPrefix(optionName, prefixes);
 
-        const featureName = optionNamePrefix == '' ? optionName : optionName.substring(optionNamePrefix.length + 1);
+        let featureName = optionNamePrefix == '' ? optionName: optionName.substring(optionNamePrefix.length + 1);
+        
+        // Handle special case for background analysis option: we want to avoid
+        // 'solution crawler' terminology in the user facing VSCode option.
+        if (featureName =="solution_crawler_background_analysis_scope") {
+            featureName = "analyzer_diagnostics_scope";
+        }
```
mavasani, line 38: I added this special handling for mapping the name for this option, but if we feel it is fine to rename the server option name, I can do so as well. I am concerned about this leading to breaking VS users though..

package.json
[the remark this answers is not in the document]
Cosifne: I am pretty sure I have done the rename before. See here https://github.com/dotnet/roslyn/pull/66899
The name of the option is just an internal string, it will be mapped to VS storage [here](https://github.com/dotnet/roslyn/blob/575bc42589145ba18b4f1cc2267d02695f861d8f/src/VisualStudio/Core/Def/Options/VisualStudioOptionStorage.cs#L116) (There is a similar one for VS mac, you can find it by search the name)

These strings would be persisted by VSCode, we can do whatever we want now since it is still in preview.
But once the extension is shipped, it would be hard to change these values.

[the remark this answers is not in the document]
mavasani: Nice - let me change the server option name then.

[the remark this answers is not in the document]
mavasani: @Cosifne - https://github.com/dotnet/roslyn/pull/68810

mavasani:
> I agree that we should not talk about solution crawling.

This has now been fixed.

arkalyanms (approved):

dibarbet:
@Cosifne if this merges without the server side change to enable the option, will that cause errors?

Cosifne:
> @Cosifne if this merges without the server side change to enable the option, will that cause errors?

If client change goes in before server change.
It will act like the user can specify a value in the client, but there is no effect. 

If server change goes in before client change.
Server would keep asking the client the value of the option, but client would always return string.Empty to server.

Cosifne (approved):

dibarbet (approved):

mavasani:
@Cosifne @dibarbet are we okay to merge this?

dibarbet:
@mavasani since it looks like the main roslyn side PRs are merged, would you mind updating the roslyn version as well in the PR, so that it all works end to end?  Instructions here - https://github.com/dotnet/vscode-csharp/blob/main/CONTRIBUTING.md#updating-the-roslyn-server-version (let me know if they're good instructions)

No worries if not, Shen mentioned nothing should break if the roslyn version isn't updated, and I can update later this week

mavasani:
> @mavasani since it looks like the main roslyn side PRs are merged, would you mind updating the roslyn version as well in the PR, so that it all works end to end? Instructions here - https://github.com/dotnet/vscode-csharp/blob/main/CONTRIBUTING.md#updating-the-roslyn-server-version (let me know if they're good instructions)
> 
> No worries if not, Shen mentioned nothing should break if the roslyn version isn't updated, and I can update later this week

Thanks Dave. I tried those steps but install still fails to find the new package (I tried a signed build from 24 hours ago, still no luck):

```
C:\Program Files\dotnet\sdk\8.0.100-preview.5.23303.2\NuGet.targets(156,5): error : Failed to download package 'Microso
ft.CodeAnalysis.LanguageServer.4.7.0-3.23328.5' from 'https://pkgs.dev.azure.com/azure-public/3ccf6661-f8ce-4e8a-bb2e-e
ff943ddd3c7/_packaging/36a629e1-6c5b-4bcd-aa2e-6018802d6b99/nuget/v3/flat2/microsoft.codeanalysis.languageserver/4.7.0-
3.23328.5/microsoft.codeanalysis.languageserver.4.7.0-3.23328.5.nupkg'. [c:\vscode-csharp\server\ServerDownload.csproj]
C:\Program Files\dotnet\sdk\8.0.100-preview.5.23303.2\NuGet.targets(156,5): error : Response status code does not indic
ate success: 401 (Unauthorized - No local versions of package 'microsoft.codeanalysis.languageserver'; please provide a
uthentication to access versions from upstream that have not yet been saved to your feed. (DevOps Activity ID: DA658997
-7077-4FEC-BA50-3469AEDE9730)). [c:\vscode-csharp\server\ServerDownload.csproj]
C:\Program Files\dotnet\sdk\8.0.100-preview.5.23303.2\NuGet.targets(156,5): error : Central Directory corrupt. [c:\vsco
de-csharp\server\ServerDownload.csproj]
C:\Program Files\dotnet\sdk\8.0.100-preview.5.23303.2\NuGet.targets(156,5): error :   The parameter is incorrect. : 'c:
\vscode-csharp\out\.nuget\microsoft.codeanalysis.languageserver\4.7.0-3.23328.5\fzqkzsy1.yj1' [c:\vscode-csharp\server\
ServerDownload.csproj]
[08:33:12] 'installDependencies' errored after 1.3 min
[08:33:12] Error: Failed to download nuget package microsoft.codeanalysis.languageserver.4.7.0-3.23328.5
    at ChildProcess.<anonymous> (c:\vscode-csharp\tasks\offlinePackagingTasks.ts:163:23)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:552:15)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at Process.callbackTrampoline (node:internal/async_hooks:130:17)
```

Kimj2539 (approved):

Kimj2539:
Done

package.json
```diff
@@ -1285,6 +1285,38 @@
...
+            "enum": [
+              "openFiles",
+              "fullSolution",
+              "none"
+            ],
+            "default": "openFiles",
+            "enumDescriptions": [
+              "Open documents",
+              "Entire solution",
+              "None"
```
CyrusNajmabadi, line 1299: would have loved the descriptions and the enum member names to match :)

mavasani: These match the enum field names and user facing descriptions in VS. I think it would be reasonable to change the enum field names in Roslyn, and once the new values flow into this repo, we can rename the field names here as well. However, these are just internal implementation details, so not sure if it critical to make these changes.

Kimj2539 (commented):
Done

PR #5870, opened 2023-06-27
Switch to eslint + prettier

dibarbet:
TSLint is deprecated and we're lacking a consistent style throughout the repo

This PR adds eslint + typescript eslint + prettier with the generally recommended settings or with settings that were previously configured by TSLint.

Also importantly changes the case of files to be camelCase.  Interfaces are currently ignored - those need to be renamed to not be `IInterface`.  The typical ts way is to just name them as standard types with regular names.

There are likely more rules we can turn on, but I would do those in a followup PR.

I think commit at a time is the best way to look at this.

ab9ed4a939e844ba88a5bd65821cfeae15c23a61 look at the .eslintrc.js file for the config and skim the changes to see if anything looks out of place
57f79a1cf61f05c3e33affb27decf8abef85f704 finishes addressing the errors introduced by the initial config in the commit above.
f7296dadf74b17eaba381241f9c0e20d09c71cf2 adds file name rules to eslint.  The rest is basically just file renames and can probably be skimmed. 
033b0aa6ccdc313379884e093e5f894056661fe4 removes TSLint and adds the remaining tslint rules to the eslint config + adds prettier config rules.
01ab588e7a7726d450af2e2ddcdb39d6dc5b2ccf is the result of running the automatic prettier fixer.  Look at the previous commit for the rules. 
079859067015bece640235b625132bbbd02f5a6c fixes a couple more outstanding issues
And then the rest are more small fixes that got left out / mixed in the first set of changes.

### NOTE
This introduces a new warning when running the build, specifically
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.1.0

YOUR TYPESCRIPT VERSION: 5.1.3

Please only submit bug reports when using the officially supported version.

=============
```

This is a known issue and most likely not a problem for us, see https://github.com/typescript-eslint/typescript-eslint/issues/6934
We should update when there is a version with the fix.


.prettierrc
```diff
@@ -1,3 +1,7 @@
 {
-  "singleQuote": true
+  "trailingComma": "es5",
```
JoeRobich, line 2: I would imagine that you could get away with `all` here.

src/json.ts
```diff
@@ -3,14 +3,19 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
```
JoeRobich, line 10:
```suggestion
```

JoeRobich (commented):
Love this work.

.prettierrc
dibarbet: It will introduce 1k more problems.  And IMHO the problems aren't really problems - its just requiring a comma at the end of every list

arkalyanms (approved):

dibarbet:
@davidwengier / @allisonchou mind taking a look from the Razor side?
and @WardenGnaw for the debugger side?

Would like to get this in since it affects almost every file in the repo.

davidwengier (approved):

WardenGnaw (approved):

PR #5879, opened 2023-06-29
Bump Razor to 7.0.0-preview.23301.2

davidwengier:


.vscode/launch.json
```diff
@@ -285,6 +285,21 @@
...
...-b59f-d4d6c9a96743/c204acfa0fd9d0e06fbcc3aac3ba846e/razorlanguageserver-linux-musl-x64-7.0.0-preview.23328.2.zip,https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/68986857aef04be58efab25ffd338a93/razorlanguageserver-linux-x64-7.0.0-preview.23328.2.zip,https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/a31700a64049f7c770a525662ef74f3c/razorlanguageserver-osx-arm64-7.0.0-preview.23328.2.zip,https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/d7017a3e07d15f057353ab21b8710714/razorlanguageserver-osx-x64-7.0.0-preview.23328.2.zip,https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/b2aa92ccbdf03761b6783bf123bfbe22/razorlanguageserver-win-arm64-7.0.0-preview.23328.2.zip,https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/3b7fc7c2f06e48d78bec2125adcc41aa/razorlanguageserver-win-x64-7.0.0-preview.23328.2.zip,https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/d1b37bd821529147e7ffec08f98be17c/razorlanguageserver-win-x86-7.0.0-preview.23328.2.zip",
```
dibarbet, line 299: does the guid in the URL change?  For O# there is not a guid and the only thing that changes is the version so I've been thinking about removing the NEW_DEPS_URLS for O# building the url from that version number instead.

package.json
```diff
@@ -589,43 +589,46 @@
     {
       "id": "Razor",
       "description": "Razor Language Server (Windows / x64)",
-      "url": "https://download.visualstudio.microsoft.com/download/pr/badc9874-13de-4e0c-bad3-ea51383f5a51/477b855ceecec09efb2045364090418d/razorlanguageserver-win-x64-7.0.0-preview.23275.2.zip",
+      "url": "https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/3b7fc7c2f06e48d78bec2125adcc41aa/razorlanguageserver-win-x64-7.0.0-preview.23328.2.zip",
```
dibarbet, line 592: nvm, I see the guid changes.  But maybe we could just specify the guid and the version number in the launch.json...
Anyway, no need to do now just thinking about later improvements.

dibarbet (approved):

Kimj2539 (approved):

Kimj2539 (approved):

package.json
davidwengier: There are actually two guids here, one is for the folder which is the same for each insertion, but one is unique per file, so we would still need some part of each url.

Kimj2539 (approved):

Kimj2539 (approved):

Kimj2539 (approved):

package.json
davidwengier: Having said that, as discussed previously, if this isn't going to be a runtime dependency, we can upload it to the same blob that omnisharp goes to if that makes it easier.

.vscode/launch.json
WardenGnaw: FYI The debugger packages update only the version number, so we use `NEW_DEPS_ID=Debugger` and `NEW_DEPS_VERSION=1.2.3`

src/tools/updatePackageDependencies.ts
```diff
@@ -20,7 +20,7 @@ interface PackageJSONFile {
     };
 }
 
-const dottedVersionRegExp = /[0-9]+\.[0-9]+\.[0-9]+/g;
+const dottedVersionRegExp = /[0-9]+\.[0-9]+\.[0-9]+[-a-z0-9.]*/g;
```
WardenGnaw, line 23: nit: prerelease semver also allows A-Z

PR #5884, opened 2023-07-01
Prototype extension API support

333fred:
Adds a new experimental API to the C# extension that allows other extensions to communicate with the C# language server. This API is not yet finalized and may change in the future. Relates to https://github.com/dotnet/roslyn/issues/68696.


src/csharpExtensionExports.ts
```diff
@@ -20,4 +22,10 @@ export interface CSharpExtensionExports {
     initializationFinished: () => Promise<void>;
     logDirectory: string;
     profferBrokeredServices: (container: GlobalBrokeredServiceContainer) => void;
+    sendRequest: <Params, Response, Error>(
```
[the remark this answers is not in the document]
333fred: I could certainly do this, perhaps have an `experimental` object that is the thing that defines these methods?

src/csharpExtensionExports.ts
```diff
@@ -20,4 +22,10 @@ export interface CSharpExtensionExports {
...
+        type: RequestType<Params, Response, Error>,
+        params: Params,
+        token: vscode.CancellationToken
+    ) => Promise<Response>;
+    serverExecutablePath: () => Promise<string | undefined>;
```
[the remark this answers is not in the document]
333fred: > unless you reload / restart the server

That is exactly what I do.

> Is there a way to check if another extension has a specific export function without activating the other extension itself?

I'm unsure, but I doubt it. These export objects are the returns from `activate`.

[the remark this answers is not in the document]
jasonmalinowski: > Was wondering if it would be possible for other extensions to export a well known function / property that tells us where we can go look for additional parts to load.

We could absolutely have extensions stick something in their package.json which lets us know they'd be exporting such a function. In @333fred's case though, I really would hope he doesn't need to write a custom function, as opposed to simply having static metadata that lets us discover his DLL and pull it into the composition. I'm not saying the custom function may not need to exist, but unless @333fred is trying to do something far more complex than I remember, we should be able to keep a simple path.

src/csharpExtensionExports.ts
jasonmalinowski, line 25: Do we also need something for notifications, since I think we eventually learned (the hard way) that a request w/o a response and a notification aren't technically the same thing?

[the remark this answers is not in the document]
jasonmalinowski: An experimental object seems reasonable. Although this method doesn't worry me too much since I think this is something we'd have long term in it's current form; I'd absolutely expect the server path function to go away though once we create a proper way to contribute stuff since we obviously shouldn't be copying DLLs long term.

jasonmalinowski, line 25:
```suggestion
    sendServerRequest: <Params, Response, Error>(
```

Or maybe something a bit more verbose, if only because I can imagine we might also have other places to send requests too? (the Razor LSP says hi...)

src/csharpExtensionExports.ts
```diff
@@ -20,4 +22,10 @@ export interface CSharpExtensionExports {
     initializationFinished: () => Promise<void>;
     logDirectory: string;
     profferBrokeredServices: (container: GlobalBrokeredServiceContainer) => void;
+    sendRequest: <Params, Response, Error>(
+        type: RequestType<Params, Response, Error>,
+        params: Params,
+        token: vscode.CancellationToken
+    ) => Promise<Response>;
```
jasonmalinowski, line 29: What behavior do you expect if the server hasn't started yet? For this to error or just block until it's started?

src/csharpExtensionExports.ts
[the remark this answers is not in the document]
333fred: Static metadata would work.

src/csharpExtensionExports.ts
333fred: I haven't needed it so far. I have no doubt someone might need it, but what I implemented was what I needed.

src/csharpExtensionExports.ts
333fred: Block is what I intended.

333fred:
@jasonmalinowski @dibarbet any further feedback?

src/csharpExtensionExports.ts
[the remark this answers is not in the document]
333fred: Updated to use this approach.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -575,6 +585,10 @@ export class RoslynLanguageServer {
                 throw new Error(`Invalid log level ${logLevel}`);
         }
     }
+
+    public getServerExecutablePath(): string | undefined {
+        return this._serverExecutablePath;
+    }
```
333fred, line 591:
```suggestion
```

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -412,6 +413,10 @@ export class RoslynLanguageServer {
             args.push('--logLevel', logLevel);
         }
 
+        if (this.additionalExtensionPaths.length > 0) {
+            args.push('--additionalExtensions', `"${this.additionalExtensionPaths.join(',')}"`);
```
dibarbet, line 417: I think this is fine to merge, even if the server doesn't have this property, right?  Since it'll only activate if there's another extension that provides this path.

dibarbet (approved):

src/lsptoolshost/roslynLanguageServer.ts
333fred: That's what I expect. I just want to make sure that this is the flag and package.json names we want.

jasonmalinowski, line 417: Why --additionalExtensions? I'd expect the existing --extension flag already works, right? If these have to be different for some reason then we should figure out better naming than "extension" and "additionalExtension".

jasonmalinowski, line 417: Why comma delimited? I'd expect each switch passed separately, since this will break with paths that contain commas.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -672,6 +685,30 @@ export async function activateRoslynLanguageServer(
...
     // Start the language server.
     _languageServer.start();
+
+    return _languageServer;
+
+    function scanExtensionPlugins(): string[] {
+        return vscode.extensions.all.flatMap((extension) => {
+            const loadPaths = extension.packageJSON.contributes?.['csharpExtensionLoadPaths'];
+            if (loadPaths === undefined || loadPaths === null) {
+                _traceChannel.appendLine(`Extension ${extension.id} does not contribute csharpExtensionLoadPaths`);
```
jasonmalinowski, line 695: Given this is the common case, just skip the logging here?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -672,6 +685,30 @@ export async function activateRoslynLanguageServer(
...
+                return [];
+            }
+
+            if (!Array.isArray(loadPaths) || loadPaths.some((loadPath) => typeof loadPath !== 'string')) {
+                _channel.appendLine(
+                    `Extension ${
+                        extension.id
+                    } has invalid csharpExtensionLoadPaths. Expected string array, found ${typeof loadPaths}`
```
jasonmalinowski, line 703: Just print loadPaths in the JSON, so easier to debug? (I've seen too many "expected Object" that becomes impossible to debug...)

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -672,6 +685,30 @@ export async function activateRoslynLanguageServer(
...
+                );
+                return [];
+            }
+
+            _traceChannel.appendLine(`Extension ${extension.id} contributes csharpExtensionLoadPaths: ${loadPaths}`);
+            return loadPaths;
```
jasonmalinowski, line 709: We don't have to join the path with the extension path to create a full path?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -672,6 +685,30 @@ export async function activateRoslynLanguageServer(
 
     // Start the language server.
     _languageServer.start();
+
+    return _languageServer;
+
+    function scanExtensionPlugins(): string[] {
```
jasonmalinowski, line 691: I know we had some concerns about performance, how expensive was this in the end?

src/main.ts
```diff
@@ -306,6 +311,9 @@ export async function activate(
     if (!useOmnisharpServer) {
         tryGetCSharpDevKitExtensionExports(csharpLogObserver);
 
+        const languageServerExport = new RoslynLanguageServerExport(
+            <Promise<RoslynLanguageServer>>roslynLanguageServerPromise
```
jasonmalinowski, line 315:
```suggestion
            roslynLanguageServerPromise!
```

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator implies that works like in C#, right?

src/lsptoolshost/roslynLanguageServer.ts
333fred: I have no idea what existing flags exist :slightly_smiling_face:.

333fred: I can do that instead

src/lsptoolshost/roslynLanguageServer.ts
333fred: I figure that's what trace logging is for: the common case that someone is trying to debug what's happening.

src/lsptoolshost/roslynLanguageServer.ts
333fred: Ended up not contributing noticeably in my (relatively unscientific) benchmarks.

src/lsptoolshost/roslynLanguageServer.ts
333fred: Sure

src/lsptoolshost/roslynLanguageServer.ts
333fred: Good point, thanks.

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: Is the _traceChannel though the LSP channel? @dibarbet?

jasonmalinowski (approved):

spouliot:
@333fred just to let you know that I can see/access the experimental API from the latest released extensions. Looking forward to documentation on how we can use it to replace our O# hooks 😄

333fred:
@spouliot that's fine. Just note that any api in there will definitely change, even if we ship it for real as is, since we will need to move it from the `experimental` accessor to the real accessor.

PR #5895, opened 2023-07-06
Read and react to O# option changes only when devkit isnt installed

arunchndr:
Address https://github.com/dotnet/vscode-csharp/issues/5774

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -85,10 +85,8 @@ export async function activate(
     const razorOptions = optionProvider.GetLatestOptions().razorOptions;
     requiredPackageIds.push('Razor');
 
-    const csharpDevkitExtension = vscode.extensions.getExtension(csharpDevkitExtensionId);
-    const useOmnisharpServer =
-        !csharpDevkitExtension && optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer;
-    if (useOmnisharpServer) {
+    const useOmnisharpServer = optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer;
+    if (optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer) {
         requiredPackageIds.push('OmniSharp');
     }
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -85,8 +85,10 @@ export async function activate(
     const razorOptions = optionProvider.GetLatestOptions().razorOptions;
     requiredPackageIds.push('Razor');
 
-    const useOmnisharpServer = optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer;
-    if (optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer) {
+    const csharpDevkitExtension = vscode.extensions.getExtension(csharpDevkitExtensionId);
+    const useOmnisharpServer =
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -85,10 +85,8 @@ export async function activate(
     const razorOptions = optionProvider.GetLatestOptions().razorOptions;
     requiredPackageIds.push('Razor');
 
-    const csharpDevkitExtension = vscode.extensions.getExtension(csharpDevkitExtensionId);
-    const useOmnisharpServer =
-        !csharpDevkitExtension && optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer;
-    if (useOmnisharpServer) {
+    const useOmnisharpServer = optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer;
+    if (optionProvider.GetLatestOptions().commonOptions.useOmnisharpServer) {
```
dibarbet, line 89: I think we still need to check for devkit here - we don't want to download the O# package if we're not using O#.

src/shared/observables/createOptionStream.ts
```diff
@@ -7,14 +7,15 @@ import { Options } from '../options';
...
+import { csharpDevkitExtensionId } from '../../utils/getCSharpDevKit';
 
 export default function createOptionStream(vscode: vscode): Observable<Options> {
     return Observable.create((observer: Observer<Options>) => {
         const disposable = vscode.workspace.onDidChangeConfiguration((e) => {
             //if the omnisharp or csharp configuration are affected only then read the options
             if (
                 e.affectsConfiguration('dotnet') ||
-                e.affectsConfiguration('omnisharp') ||
+                (!vscode.extensions.getExtension(csharpDevkitExtensionId) && e.affectsConfiguration('omnisharp')) ||
```
sharwell, line 18: 💡 Unless we know `getExtension` is very efficient, consider reversing this condition

src/shared/observables/createOptionStream.ts
```diff
@@ -7,14 +7,15 @@ import { Options } from '../options';
...
+import { csharpDevkitExtensionId } from '../../utils/getCSharpDevKit';
 
 export default function createOptionStream(vscode: vscode): Observable<Options> {
     return Observable.create((observer: Observer<Options>) => {
         const disposable = vscode.workspace.onDidChangeConfiguration((e) => {
             //if the omnisharp or csharp configuration are affected only then read the options
             if (
                 e.affectsConfiguration('dotnet') ||
-                e.affectsConfiguration('omnisharp') ||
+                (e.affectsConfiguration('omnisharp') && !vscode.extensions.getExtension(csharpDevkitExtensionId)) ||
```
jasonmalinowski, line 18: Should this logic instead move to somewhere like

https://github.com/dotnet/vscode-csharp/blob/f01115f20d428a704df42a55ab8ea31816beb2d8/src/shared/options.ts#L349

It'll probably work either way, but I'm somewhat surprised already that the Omnisharp option isn't being ignored by that.

jasonmalinowski (approved):
This will work, although I'm surprised it doesn't already given the other code that's doing similar filtering.

JoeRobich:
@arunchndr Is this change still needed?

PR #5903, opened 2023-07-07
Release notes configuration

arkalyanms:


.github/release.yml
```diff
@@ -0,0 +1,22 @@
+# .github/release.yml
+
+changelog:
```
dibarbet, line 3: How does this work exactly?  Does this update a github release somewhere?  We haven't been doing github releases for the prerelease version of the extension (though we could).

arkalyanms: yes this will hook up to the github release note generation with the new labels where we can set the release to be pre-release until GA.

.github/release.yml
```diff
@@ -0,0 +1,22 @@
...
+    labels:
+      - Infrastructure
+  categories:
+    - title: Exciting New Features 🎉
+      labels:
+        - Feature Request
+    - title: Bug fixes 🛠️
+      labels:
+        - bug
+    - title: C# on LSP (Omnisharp Parity)
```
WardenGnaw, line 14: Does this need to be wrapped in `''` as '#' means comment in YAML?
```suggestion
    - title: 'C# on LSP (Omnisharp Parity)'
```

jasonmalinowski (approved):
This is nifty!

PR #5931, opened 2023-07-19
Fix '"file" argument must be of type string' error

jasonmalinowski:
Fixes https://github.com/dotnet/vscode-csharp/issues/5920
Fixes https://github.com/dotnet/vscode-dotnet-runtime/issues/1007

(and many, many more duplicates)

### DIFF
--- a/src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
+++ b/src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
@@ -88,18 +88,17 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
     /**
      * Acquires the .NET runtime and any other dependencies required to spawn a particular .NET executable.
      * @param path The path to the entrypoint assembly. Typically a .dll.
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -88,18 +88,17 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
     /**
      * Acquires the .NET runtime and any other dependencies required to spawn a particular .NET executable.
      * @param path The path to the entrypoint assembly. Typically a .dll.
-     * @returns The path to the `dotnet` command to use to spawn the process.
      */
-    private async acquireDotNetProcessDependencies(path: string) {
-        const dotnetPath = await this.acquireRuntime();
+    private async acquireDotNetProcessDependencies(path: string): Promise<HostExecutableInformation> {
```
jasonmalinowski, line 92: The implicit return type didn't match the docs, but this is what it's returning. (And that's what other code is expecting.)

arkalyanms (approved):

PR #5932, opened 2023-07-19
Separate Razor server and Typescript versions (O# and non-O#)

allisonchou:
**Problem:** We have been recently unable to update Razor's language server version in this repo to the newest bits due to incompatibilities between O# and rzls.

**Solution:** This PR enables O# users to utilize older versions of the TS and rzls code, while the normal C# extension can utilize newer versions. The O# rzls bits will only be downloaded at runtime if the user explicitly selects the O# option, so it should hopefully not add too much overhead for the majority of users. We do automatically download the TS O# bits, but that seems unavoidable.

**Follow-up:** In the next few days, I plan to remove `OmnisharpPlugin` from the main branch of the Razor repo entirely. It is unnecessary since we have a separate O# branch in the Razor repo (`release/omnisharp-vscode`) where we can maintain both `OmnisharpPlugin` and an older version of rzls for O# compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "defaults": {
     "roslyn": "4.8.0-1.23367.7",
     "omniSharp": "1.39.7",
-    "razor": "7.0.0-preview.23301.2"
+    "razor": "7.0.0-preview.23328.2"
   },
   "main": "./dist/extension",
   "brokeredServices": [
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -38,7 +38,8 @@
   "defaults": {
     "roslyn": "4.8.0-1.23367.7",
     "omniSharp": "1.39.7",
-    "razor": "7.0.0-preview.23301.2"
+    "razor": "7.0.0-preview.23328.2",
```
allisonchou, line 41: I updated Razor to the original version from @davidwengier's PR: https://github.com/dotnet/vscode-csharp/pull/5879/commits/dd29e3c4f5c46c5f0ea59822082459b3593a56f0

package.json
```diff
@@ -38,7 +38,8 @@
...
+    "razorOmnisharp": "7.0.0-preview.23363.1"
```
allisonchou, line 42: This looks like a newer version compared to above, but is actually built off an older version of the Razor language server (https://github.com/dotnet/razor/pull/8951).

package.json
```diff
@@ -80,6 +81,7 @@
     "http-proxy-agent": "4.0.1",
     "https-proxy-agent": "5.0.0",
     "jsonc-parser": "3.0.0",
+    "microsoft.aspnetcore.razor.vscode": "https://download.visualstudio.microsoft.com/download/pr/aee63398-023f-48db-bba2-30162c68f0c4/0af42abab690d5de903a4a814d6aedc1/microsoft.aspnetcore.razor.vscode-7.0.0-preview.23363.1.tgz",
```
allisonchou, line 84: Added this bit since we might want to have separate TS logic between O# and the regular C# extension

package.json
```diff
@@ -589,46 +591,43 @@
...
-      "url": "https://download.visualstudio.microsoft.com/download/pr/fddc6f7e-083a-44bd-9a8d-200a4146b13b/c51fc9fae3f6fa2b1992df6dd16f50d2/razorlanguageserver-win-x64-7.0.0-preview.23301.2.zip",
+      "url": "https://download.visualstudio.microsoft.com/download/pr/51d2bf3c-cb3c-4385-b59f-d4d6c9a96743/3b7fc7c2f06e48d78bec2125adcc41aa/razorlanguageserver-win-x64-7.0.0-preview.23328.2.zip",
       "installPath": ".razor",
       "platforms": [
         "win32"
       ],
       "architectures": [
         "x86_64"
-      ],
-      "integrity": "BE0001FDAEB075CE4F9148308C09CF97E85DB3BCBFBE0C2A3D70987D8D265047"
```
allisonchou, line 600: The gulp task to automatically update dependencies wasn't working for me, so I ended up doing this manually (and getting rid of the auto-generated integrity property since it's non-mandatory)

src/main.ts
```diff
@@ -239,7 +245,24 @@ export async function activate(
...
+            // Download Razor O# server
+            const razorOmnisharpDownloader = new RazorOmnisharpDownloader(
+                networkSettingsProvider,
+                eventStream,
+                context.extension.packageJSON,
+                platformInfo,
+                context.extension.extensionPath
+            );
+
+            await razorOmnisharpDownloader.DownloadAndInstallRazorOmnisharp(
```
allisonchou, line 257: Razor O# server bits will only be downloaded at runtime. We do, however, download the O# Razor TS bits by default which adds a bit of size to the extension, but there doesn't seem to be a good alternative.

davidwengier (approved):
This is better than I thought we'd get to, so I'm over the moon, but also unqualified to make a real determination :)

davidwengier:
> In the next few days, I plan to remove OmnisharpPlugin from the main branch of the Razor repo entirely.

Presumably the same can be said for the .ts files in the main branch? At some point we should move our textmate grammar over here too, since that would then be the only node dependency in our repo (I think!)

package.json
dibarbet: +1 to what David said - integrity is useful when its getting downloaded at runtime.  I wouldn't remove it entirel.y

src/razor/razor.ts
```diff
@@ -12,15 +13,30 @@ import { EventStream } from '../eventStream';
...
 ) {
     const razorConfig = vscode.workspace.getConfiguration('razor');
     const configuredLanguageServerDir = razorConfig.get<string>('languageServer.directory', '');
     const languageServerDir =
-        configuredLanguageServerDir.length > 0 ? configuredLanguageServerDir : path.join(extensionPath, '.razor');
+        configuredLanguageServerDir.length > 0
+            ? configuredLanguageServerDir
+            : useOmnisharpServer
+            ? path.join(extensionPath, '.razor/omnisharp')
+            : path.join(extensionPath, '.razor');
```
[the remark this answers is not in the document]
allisonchou: Good point - updated to `.razoromnisharp`, thanks!

package.json
allisonchou: Gotcha - I can address this in a follow-up PR as DavidW suggested. Just need to figure out why the script isn't working for me 😅

allisonchou:
> Presumably the same can be said for the .ts files in the main branch? At some point we should move our textmate grammar over here too, since that would then be the only node dependency in our repo (I think!)

Yep, the *.ts files in the main branch can go as well. Agree on moving the Textmate grammar over, we should definitely do that in a follow-up as well

PR #5940, opened 2023-07-19
Fix path to server when running platform neutral vsix

dibarbet:
Won't fully work until https://github.com/dotnet/roslyn/pull/69110 is in

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -644,12 +644,10 @@ export async function activateRoslynLanguageServer(
 }
 
 function getServerPath(options: Options, platformInfo: PlatformInformation) {
-    const clientRoot = __dirname;
-
     let serverPath = options.commonOptions.serverPath;
     if (!serverPath) {
         // Option not set, use the path from the extension.
-        serverPath = path.join(clientRoot, '..', '.roslyn', getServerFileName(platformInfo));
+        serverPath = getInstalledServerPath(platformInfo);
     }
 
     if (!fs.existsSync(serverPath)) {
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -659,8 +657,10 @@ function getServerPath(options: Options, platformInfo: PlatformInformation) {
     return serverPath;
 }
 
-function getServerFileName(platformInfo: PlatformInformation) {
-    const serverFileName = 'Microsoft.CodeAnalysis.LanguageServer';
+function getInstalledServerPath(platformInfo: PlatformInformation) {
```
jasonmalinowski, line 660: Are we moving to always stating return types for functions or not?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -659,8 +657,10 @@ function getServerPath(options: Options, platformInfo: PlatformInformation) {
...
+    const clientRoot = __dirname;
+    const serverFilePath = path.join(clientRoot, '..', '.roslyn', 'Microsoft.CodeAnalysis.LanguageServer');
+
     let extension = '';
     if (platformInfo.isWindows()) {
```
jasonmalinowski, line 665: Not related to your change but it bugs me the isMacOS isn't an else-if....

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -673,7 +673,13 @@ function getServerFileName(platformInfo: PlatformInformation) {
         extension = '.dll';
     }
 
-    return `${serverFileName}${extension}`;
+    let pathWithExtension = `${serverFilePath}${extension}`;
+    if (!fs.existsSync(pathWithExtension)) {
```
jasonmalinowski, line 677: So on Linux we don't set extension, expect this to fail and then take the if block?

jasonmalinowski (approved):

src/lsptoolshost/roslynLanguageServer.ts
dibarbet: we should, but I haven't enabled the rule, hence me making bad changes :) ill fix

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: Oh right, I forgot there'd be no extension. It _might_ make sense to put a comment there that the lack of setting that is intentional, but...

jasonmalinowski (approved):

PR #5945, opened 2023-07-21
On Auto Insert - Fix lack of newline support

akhera99:
Fixes [#68589](https://github.com/dotnet/roslyn/issues/68589)

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -660,7 +660,9 @@ export async function activateRoslynLanguageServer(
         const capabilities = await _languageServer.getServerCapabilities();
 
         if (capabilities._vs_onAutoInsertProvider) {
-            if (!capabilities._vs_onAutoInsertProvider._vs_triggerCharacters.includes(change.text)) {
+            const changeTrimmed = change.text.replace(/(?!^)\s+/g, '');
+
+            if (!capabilities._vs_onAutoInsertProvider._vs_triggerCharacters.includes(changeTrimmed)) {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -660,7 +660,9 @@ export async function activateRoslynLanguageServer(
         const capabilities = await _languageServer.getServerCapabilities();
 
         if (capabilities._vs_onAutoInsertProvider) {
-            if (!capabilities._vs_onAutoInsertProvider._vs_triggerCharacters.includes(change.text)) {
+            const changeTrimmed = change.text.replace(/(?!^)\s+/g, '');
```
dibarbet, line 663: Why does whitespace in the input text cause indentation to not get properly applied, and why does removing the whitespace fix it?

akhera99: The trigger character is '\n' so it doesn't match otherwise. Also the server side expects the text to match exactly.

CyrusNajmabadi: this needs comments explaining what the heck is going on (including exaplining the regex :)).

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -794,10 +796,12 @@ async function applyAutoInsertEdit(e: vscode.TextDocumentChangeEvent, token: vsc
     const uri = UriConverter.serialize(e.document.uri);
     const textDocument = TextDocumentIdentifier.create(uri);
     const formattingOptions = getFormattingOptions();
+    const changeTrimmed = change.text.replace(/(?!^)\s+/g, '');
```
CyrusNajmabadi, line 799: looks like a repeat of the mutation that happened elsewhere. unify?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -797,9 +804,10 @@ async function applyAutoInsertEdit(e: vscode.TextDocumentChangeEvent, token: vsc
     const request: RoslynProtocol.OnAutoInsertParams = {
         _vs_textDocument: textDocument,
         _vs_position: position,
-        _vs_ch: change.text,
+        _vs_ch: changeTrimmed,
```
dibarbet, line 807: so we remove whitespace, but I think I still don't understand why we need to remove whitespace.  Does the summary text not get formatted properly if the line is already indented?

akhera99: Correct. The trigger characters looks for a new line character and the change.Text is something akin to "\n   ", which I presume is the spaces needed to bring the cursor to line up with the comment. So we need to remove whitespace to match the trigger character and, in the OnAutoInsertHandler, it again looks to see the changed text exactly matches the new line character.

dibarbet (approved):

PR #5965, opened 2023-07-28
Implement client side of test debugging

dibarbet:
Client side part of test debugging

Part of https://github.com/dotnet/vscode-csharp/issues/5719
debug_test_commands

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -207,6 +207,8 @@ export class RoslynLanguageServer {
 
         // Register Razor dynamic file info handling
         this.registerDynamicFileInfo(this._languageClient);
+
+        this.registerDebuggerAttach(this._languageClient);
     }
 
     public async stop(): Promise<void> {
@@ -485,6 +487,25 @@ export class RoslynLanguageServer {
         );
         client.onNotification(RoslynLanguageServer.removeRazorDynamicFileInfoMethodName, async (notification) =>
             vscode.commands.executeCommand(DynamicFileInfoHandler.removeDynamicFileInfoCommand, notification)
+        );
+    }
+
+    private registerDebuggerAttach(client: RoslynLanguageClient) {
+        client.onRequest<RoslynProtocol.DebugAttachParams, RoslynProtocol.DebugAttachResult, void>(
+            RoslynProtocol.DebugAttachRequest.type,
+            async (request) => {
+                const debugConfiguration: vscode.DebugConfiguration = {
+                    name: '.NET Core Attach',
+                    type: 'coreclr',
+                    request: 'attach',
+                    processId: request.processId,
+                };
+
+                const result = await vscode.debug.startDebugging(undefined, debugConfiguration, undefined);
+                return {
+                    didAttach: result,
+                };
+            }
         );
     }
 
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     }
   },
   "defaults": {
-    "roslyn": "4.8.0-1.23372.1",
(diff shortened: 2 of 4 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -488,6 +490,25 @@ export class RoslynLanguageServer {
...
+            async (request) => {
+                const debugConfiguration: vscode.DebugConfiguration = {
+                    name: '.NET Core Attach',
+                    type: 'coreclr',
+                    request: 'attach',
+                    processId: request.processId,
+                };
+
+                const result = await vscode.debug.startDebugging(undefined, debugConfiguration, undefined);
+                return {
```
gregg-miskelly, line 505: Does something inside the target process block waiting for `Debugger.IsAttached` to return true? If not, you likely need a listener to know when the attach is complete (and we might need a debugger change to raise a new event).

dibarbet: Yeah - the server side blocks and the tests won't start running until it gets this response from the client.

gregg-miskelly: Right, that isn't good enough -- you are going to return before the real attach is complete

dibarbet: Ah the vscode API returns before the attach is complete?  That is interesting.  So on the server side I could wait for `Debugger.IsAttached`?  Is there anything on the client side that would have the same behavior that I could wait for?

dibarbet: Ah this is hard because I'm not actually attaching to the server process, but the vs test process.  So I can't check `Debugger.IsAttached` in the server, that check has to be done in the test host.

dibarbet: > you likely need a listener to know when the attach is complete (and we might need a debugger change to raise a new event).

Do you have any more details on how to do this?

Also wondering if any of the other vscode debug events might be able to let me know whenthe debugger is actually attached.
```
/**
         * An {@link Event} which fires when the {@link debug.activeDebugSession active debug session}
         * has changed. *Note* that the event also fires when the active debug session changes
         * to `undefined`.
         */
        export const onDidChangeActiveDebugSession: Event<DebugSession | undefined>;

        /**
         * An {@link Event} which fires when a new {@link DebugSession debug session} has been started.
         */
        export const onDidStartDebugSession: Event<DebugSession>;

        /**
         * An {@link Event} which fires when a custom DAP event is received from the {@link DebugSession debug session}.
         */
        export const onDidReceiveDebugSessionCustomEvent: Event<DebugSessionCustomEvent>;
```

gregg-miskelly: If it isn't easy for the test host to call `Debugger.IsAttached` I think what we should do is have the debugger raise a custom `AttachComplete` event that we should be able to receive through `onDidReceiveDebugSessionCustomEvent`. If that sounds good to you, I can open an issue.

dibarbet: Yeah I think having the debugger raise a complete event sounds perfect.  That would then work no matter what process we're trying to attach to.

gregg-miskelly: I opened #5968. I would suggest completing this PR when you are ready and I can wire things up when we insert a debugger with the fix.

dibarbet: Perfect thanks!  It will likely take a day or two for this PR to go in as the server side changes also need to be reviewed and merged before this goes in.

Cosifne (approved):

PR #5982, opened 2023-08-01
Route razor/simplifyTypeNames to Roslyn

lonitra:
Draft until https://github.com/dotnet/razor/pull/9070 and https://github.com/dotnet/roslyn/pull/69323 are merged as roslyn/razor versions need to be bumped

This PR adds logic to route LSP call `razor/simplifyTypeNames` to Roslyn's `textDocument/simplifyTypeNames` (new LSP method added by ) so that 'Generate EventHandler' code action will behave the same as VS and will generate an event handler with the simplified type names.


davidwengier (approved):

ryzngard (approved):

src/razor/src/simplify/serializableDelegatedSimplifyMethodParams.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { SerializableTextDocumentIdentifierAndVersion } from './serializableTextDocumentIdentifierAndVersion';
+import { TextEdit } from 'vscode-html-languageservice';
+
+export interface SerializableDelegatedSimplifyMethodParams {
+    identifier: SerializableTextDocumentIdentifierAndVersion;
```
davidwengier, line 10: I forgot we changed to use `TextDocumentAndIdentifier`. @ryzngard do we need to update other types in the extension to this, or is the serialization compatible with whatever they currently have?

PR #5984, opened 2023-08-02
Download Razor before start OmniSharp

davidwengier:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1864007

Now that Razor is back to being a runtime dependency, we were failing to load on a clean machine because we were initializing OmniSharp, and telling it about the Razor OmniSharpPlugin.dll, before we had even started trying to download it. This change frontloads that download, and waits for it to finish before activating O#.

Worth noting, if we don't take this PR then whilst the first run of VS Code is broken for Razor, subsequent runs (or window reloads) work fine. The Razor download is currently sitting at 70mb. I leave it up to you to decide if the current behavior is a bug, or by design :)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -225,6 +225,24 @@ export async function activate(
             eventStream.subscribe(omnisharpDebugModeLoggerObserver.post);
         }
 
+        const razorObserver = new RazorLoggerObserver(csharpChannel);
+        eventStream.subscribe(razorObserver.post);
+
+        if (!razorOptions.razorDevMode) {
+            // Download Razor O# server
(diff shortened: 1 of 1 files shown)


ryzngard (approved):

arkalyanms:
@dibarbet fyi. Tagging as a potential release candidate we can pick once razor signs off on the build. 

Sidenote: Debugger issue might be same.

dibarbet:
@arkalyanms I think the debugger issue might be different - because we use the same version of the debugger for O# and for the prerelease version.  So it should already be included and not have to download it

PR #5990, opened 2023-08-02
Onboard Localization pipeline

Cosifne:
Sample run https://github.com/dotnet/vscode-csharp/pull/5987
https://dev.azure.com/dnceng/internal/_build/results?buildId=2234885&view=results

This PR contains two localization file changes. One in code and one in package.json as the test sample.
In code, it uses vs-l10n to do localization.

The pipeline works like this:
1. Extract localization strings to bundle.l10n.json. (l10n-dev)
2. Convert them into .xlf file. (l10-dev)
3. Use OneLocBulid task to generate translated .xlf file (OneLocBuild)
4. Import all the translated file back to json under `l10n` folder. (l10n-dev)
5. Use dotnet-bot to submit a PR back to `main` on GitHub.

I feel it is enough to always translate our main branch here. If there is any problem we can address it in the future.

After this, two more things are needed.
1. Ask the LOC team to onboard our repo to their list, so that all things are get translated.
2. Extract all the strings, and package.json to use l10n, this might be in another PR.


azure-pipelines/loc.yml
```diff
@@ -0,0 +1,73 @@
...
+  - job: Localization
+    pool:
+      name: NetCore1ESPool-Internal
+      demands: ImageOverride -equals windows.vs2022preview.amd64
+    steps:
+    - task: NodeTool@0
+      displayName: 'Install Node.js 18.x'
+      inputs:
+        # Octokit client needs 18.x to have 'fetch' function.
+        versionSpec: '18.x'
```
Cosifne, line 26: GitHub client needs this

package.json
```diff
@@ -996,17 +1003,17 @@
                     "items": {
                       "type": "string"
                     },
-                    "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                    "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
```
Cosifne, line 1006: I think all these changes are generated by l10n-dev. If it is not good I can bring this back

package.json
```diff
@@ -1322,7 +1329,7 @@
         "properties": {
           "dotnet.defaultSolution": {
             "type": "string",
-            "description": "The path of the default solution to be opened in the workspace, or set to 'disable' to skip it.",
+            "description": "%configuration.dotnet.defaultSolution.description%",
```
Cosifne, line 1332: The sample used in this PR

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -494,7 +494,7 @@ export class RoslynLanguageServer {
                     return;
                 }
 
-                const title = 'Restart Language Server';
+                const title = vscode.l10n.t('Restart Language Server');
```
Cosifne, line 497: The other sample used in this PR

loc/LocProject.json
```diff
@@ -0,0 +1,14 @@
+{
+    "Projects": [
+        {
+            "LanguageSet": "VS_Main_Languages",
+            "LocItems": [
+                {
+                    "SourceFile": "./loc/vscode-csharp.xlf",
```
Cosifne, line 7: `vscode-csharp.xlf` is a generated file from `l10nDevGenerateXlf` task in package.json.
It is passed to the localization build.

./loc folder is used to stage all the generated files.

package.json
```diff
@@ -65,13 +66,17 @@
     "test:integration:slnFilterWithCsproj": "tsc -p ./ && gulp test:integration:slnFilterWithCsproj",
     "test:artifacts": "tsc -p ./ && mocha out/test/artifactTests/**/*.test.js",
     "unpackage:vsix": "gulp vsix:release:unpackage",
-    "updatePackageDependencies": "gulp updatePackageDependencies"
+    "updatePackageDependencies": "gulp updatePackageDependencies",
+    "l10nDevGenerateLocalizationBundle": "npx @vscode/l10n-dev export --outDir ./l10n ./src",
```
Cosifne, line 70: These three tasks would
1. Generate bundle.l10n.json from code.
2. Export `bundle.l10n.json` and `package.nls.json` to `vscode-csharp.xlf`
3. When `vscode-csharp.xlf` get translated, import them back to `bundle.l10.<lang>.json` and `package.nls.<lang>.json`

tasks/localizationTasks.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    const stdout = git.stdout.toString();
+    if (printCommand) {
+        console.log(stdout);
+    }
+    return stdout;
+}
+
+gulp.task('publish localization content', async () => {
```
Cosifne, line 98: This task is basically performing
`git add -A` (Stage all the changed files)
`git diff HEAD` (Find all the changed file, find 'bundle.l10n.<lang>.json', 'package.nls.<lang>.json')
`git reset` (Clean all staged files)
`git add bundle.l10n.<lang>.json' package.nls.<lang>.json `

Use bot account to create a PR in GitHub

package.json
```diff
@@ -996,17 +1003,17 @@
...
+                    "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
                     "default": []
                   },
                   "searchMicrosoftSymbolServer": {
                     "type": "boolean",
-                    "description": "If 'true' the Microsoft Symbol server (https\u200b://msdl.microsoft.com\u200b/download/symbols) is added to the symbols search path. If unspecified, this option defaults to 'false'.",
+                    "description": "If 'true' the Microsoft Symbol server (https​://msdl.microsoft.com​/download/symbols) is added to the symbols search path. If unspecified, this option defaults to 'false'.",
```
dibarbet, line 1011: these changes need to be reverted, think you can run `gulp generateOptionsSchema` to fix

tasks/localizationTasks.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    return changedLocFilesOrDirectory;
+}
+
+function getAllPossibleLocalizationFileNames(): string[] {
+    const files = [];
+    for (const lang of localizationLanguages) {
+        for (const file of locFiles) {
+            files.push(util.format(file, lang));
```
dibarbet, line 59: should we verify that all the files exist here and throw if not?  I think it would be a problem if one of them doesn't exist right?

tasks/localizationTasks.ts
dibarbet: I'm probably missing something - why do we need the first 3 steps?  Is it not possible to always just do a `git add` on all of the possible localization files?  I think git should ignore ones trhat aren't changed

tasks/localizationTasks.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    const localizationChanges = getGeneratedLocalizationChanges(diffResults);
+    if (localizationChanges.length == 0) {
+        console.log('No localization file changed');
+        return;
+    }
+    await git(['reset']);
+    const newBranchName = `localization/${parsedArgs.commitSha}`;
+    await git(['config', '--local', 'user.name', parsedArgs.userName]);
```
dibarbet, line 114: we could potentially make these optional so when you run this locally you don't have to provide them (but we still provide them in CI)

tasks/localizationTasks.ts
```diff
@@ -0,0 +1,174 @@
...
+        ],
+        // Note: don't print PAT to console
+        false
+    );
+    await git(['fetch', remoteRepoAlias]);
+
+    const lsRemote = await git(['ls-remote', remoteRepoAlias, 'refs/head/' + newBranchName]);
+    if (lsRemote.trim() !== '') {
+        // If the localization branch of this commit already exists, don't try to create another one.
+        console.log(`${newBranchName} already exists in ${parsedArgs.targetRemoteRepo}. Skip pushing.`);
```
dibarbet, line 143: is this expected to happen?  Should it throw?

tasks/localizationTasks.ts
```diff
@@ -0,0 +1,174 @@
...
+    const result = await git(['diff'].concat(args));
+    // Line ending from the stdout of git is '\n' even on Windows.
+    return result
+        .replaceAll('\n', EOL)
+        .split(EOL)
+        .map((fileName, _) => fileName.trim())
+        .filter((fileName) => fileName.length !== 0);
+}
+
+async function git(args: string[], printCommand = true): Promise<string> {
```
dibarbet, line 77: It seems like we always print the command, so maybe removable?

tasks/localizationTasks.ts
```diff
@@ -0,0 +1,174 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import * as util from 'node:util';
+import { EOL } from 'node:os';
+import { Octokit } from '@octokit/rest';
+
+type Options = {
+    userName: string;
+    email: string;
+    owner: string;
```
dibarbet, line 19: I think owner is the fork owner right?
Is it necessary to support forks?  IMHO I'd prefer keeping the number of options down and just always push to the main repo unless there is a really good reason not to

package.json
dibarbet: So when I add a new localizable string - do I need to manually add it to a json file somewhere?  Or is that taken care of by 1)?  If so I think we should run 1) as part of the regular build step

tasks/localizationTasks.ts
Cosifne: No, the loc task would generate a lot of other stuff, see here https://dev.azure.com/dnceng/internal/_build/results?buildId=2234885&view=artifacts&pathAsName=false&type=publishedArtifacts

package.json
Cosifne: OK

package.json
Cosifne: yeah sounds like 1 need to be added to the compile step

tasks/localizationTasks.ts
Cosifne: No, when we call `git add remote url`, PAT is encoded like this
`https://${parsedArgs.userName}:${pat}@github.com/${parsedArgs.owner}/${parsedArgs.targetRemoteRepo}.git`
I don't want to reveal the PAT in this case

tasks/localizationTasks.ts
Cosifne: Not sure if I understand your comment.
This one has just generated all the possible files here. e.g.
a combination of `['bundle.l10n.%s.json', 'package.nls.%s.json']` with `['cs', 'de', 'es', 'fr', 'it', 'ja', 'ko', 'pl', 'pt-br', 'ru', 'tr', 'zh-cn', 'zh-tw']`
Later the real file path comes from the result of `git diff --name-only` so the file should exist.

tasks/localizationTasks.ts
Cosifne: Ah you mean hardcode `dotnet` here? Sure it can be done

tasks/localizationTasks.ts
Cosifne: I suppose this might happen, in case the pipeline run is canceled when branch is pushed but no PR is created

tasks/localizationTasks.ts
dibarbet: What I mean is we know what files we want to keep - we could always run a git add on all the files we know we want, and don't add everything else (its left as unstaged).

Basically kind of just hardcode the files we know we want
`git add xyz-cs.json`
`git add xyz-de.json`
etc.

then commit and push

Then we dont need to diff / reset or anything

package.json
Cosifne: Done, added to compile and compileDev

Cosifne:
Because Phil merged this https://github.com/dotnet/vscode-csharp/pull/5962 so there are more strings need localization : )

dibarbet (commented):
Mostly good now - I think still just pending if we can avoid the diff reset in https://github.com/dotnet/vscode-csharp/pull/5990/files#r1282385281

tasks/localizationTasks.ts
Cosifne: Done :)

tasks/localizationTasks.ts
dibarbet: Lets log this as an actual pipeline error
https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=powershell#logissue-log-an-error-or-warning

dibarbet (approved):
1 minor comment

phil-allen-msft (approved):

l10n/bundle.l10n.json
```diff
@@ -0,0 +1,85 @@
+{
+  "Cannot load Razor language server because the directory was not found: '{0}'": "Cannot load Razor language server because the directory was not found: '{0}'",
```
WardenGnaw, line 2: @Cosifne The `keys` of this file can be used to help organize who owns the strings / help what the usage should be. 

See https://github.com/microsoft/vscode-cpptools/blob/ae904ec3306d563db44445dac90728ec706424d3/Extension/package.nls.json

Cosifne: This key is auto-generated by the l10n tool, let me see if this can be optimized or not

WardenGnaw: @Cosifne Sorry I'm incorrect here. I think this is correct for the `l10n.json` file. I got this confused with the `nls.json` file.

PR #5991, opened 2023-08-02
Add back accidentally excluded changelog file from vsix

dibarbet:


.vscodeignore
```diff
@@ -30,7 +30,7 @@ gulpfile.ts
 !install.Lock
 ISSUE_TEMPLATE
 .mocharc.jsonc
-*.md
+CONTRIBUTING.md
```
arkalyanms, line 33: do we any other md except the changelog?

arkalyanms (approved):

.vscodeignore
dibarbet: And the readme (though luckily that doesn't get excluded even if we exclude it)

dibarbet: Ah I just saw the test-plan as well.   Let me edit this ....

PR #6058, opened 2023-08-07
Respect the newly added Background analysis scope option in OmniSharp

### LINKED ISSUE #5998: Roslyn Analyzer works even though I disabled It
I don't like analyzing my code while Editing So I Disabled the Roslyn Analyzer through  ```Enable support for roslyn analyzers, code fixes and rulesets``` setting

But It still works even after restarting language server and restarting vscode

I switched to version v1.26.0 and Everything is working fine

mavasani:
Closes #5998

This ensures that the Background analysis scope option becomes the source of truth for both C# extension and OmniSharp. We should drive the users to use the below two options even when OmniSharp server is enabled. They are the replacement for the existing options `Enable support for roslyn analyzers, code fixes and rulesets` and `Analyze open documents only` and enable configuring these separately for compiler and analyzer diagnostics.

image

image

NOTE: I ran into https://github.com/dotnet/vscode-csharp/issues/6057 quite often while playing around with this space, I'll send a separate Roslyn PR for fixing that issue.

### DIFF
--- a/src/shared/options.ts
+++ b/src/shared/options.ts
@@ -122,11 +122,9 @@ export class Options {
         const diagnosticAnalysisScope = Options.readOption<string>(
             config,
             'dotnet.backgroundAnalysis.analyzerDiagnosticsScope',
-            'openFiles'
+            useOmnisharpServer ? 'none' : 'openFiles'
         );
-        const enableRoslynAnalyzers =
-            Options.readOption<boolean>(config, 'omnisharp.enableRoslynAnalyzers', false) &&
-            diagnosticAnalysisScope != 'none';
+        const enableRoslynAnalyzers = diagnosticAnalysisScope != 'none';
         const enableEditorConfigSupport = Options.readOption<boolean>(
             config,
             'omnisharp.enableEditorConfigSupport',
@@ -145,9 +143,7 @@ export class Options {
             'omnisharp.enableImportCompletion'
         );
         const enableAsyncCompletion = Options.readOption<boolean>(config, 'omnisharp.enableAsyncCompletion', false);
-        const analyzeOpenDocumentsOnly =
-            Options.readOption<boolean>(config, 'omnisharp.analyzeOpenDocumentsOnly', false) ||
-            diagnosticAnalysisScope == 'openFiles';
+        const analyzeOpenDocumentsOnly = diagnosticAnalysisScope == 'openFiles';
         const organizeImportsOnFormat = Options.readOption<boolean>(config, 'omnisharp.organizeImportsOnFormat', false);
         const disableMSBuildDiagnosticWarning = Options.readOption<boolean>(
             config,
--- a/src/shared/options.ts
+++ b/src/shared/options.ts
@@ -119,12 +119,14 @@ export class Options {
             'omnisharp.useEditorFormattingSettings',
             true
         );
+        const enableRoslynAnalyzersLegacyOption = Options.readOption<boolean>(config, 'omnisharp.enableRoslynAnalyzers', false);
         const diagnosticAnalysisScope = Options.readOption<string>(
             config,
             'dotnet.backgroundAnalysis.analyzerDiagnosticsScope',
             useOmnisharpServer ? 'none' : 'openFiles'
         );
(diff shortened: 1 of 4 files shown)


test/unitTests/options.test.ts
```diff
@@ -37,7 +37,7 @@ suite('Options tests', () => {
         options.omnisharpOptions.enableDecompilationSupport.should.equal(false);
         options.omnisharpOptions.enableImportCompletion.should.equal(false);
         options.omnisharpOptions.enableAsyncCompletion.should.equal(false);
-        options.omnisharpOptions.analyzeOpenDocumentsOnly.should.equal(false);
+        options.omnisharpOptions.analyzeOpenDocumentsOnly.should.equal(true);
```
dibarbet, line 40: is it possible to avoid changing the default in O#?

src/shared/options.ts
```diff
@@ -119,7 +119,14 @@ export class Options {
             'omnisharp.useEditorFormattingSettings',
             true
         );
-        const enableRoslynAnalyzers = Options.readOption<boolean>(config, 'omnisharp.enableRoslynAnalyzers', false);
+        const diagnosticAnalysisScope = Options.readOption<string>(
+            config,
+            'dotnet.backgroundAnalysis.analyzerDiagnosticsScope',
+            'openFiles'
+        );
+        const enableRoslynAnalyzers =
```
dibarbet, line 127: Instead of having a separate value for `diagnosticAnalysisScope` and the O# ones (`enableRoslynAnalyzers`, `analyzeOpenDocumentsOnly`)  - can we convert the O# ones into a `diagnosticAnalysisScope` value?

So if there is no value set for any of the options we respect the default value of `diagnosticAnalysisScope`
If there is a value set for `diagnosticAnalysisScope` we use that.
If there is no value for `diagnosticAnalysisScope` but there are values for O# options, we convert them to `diagnosticAnalysisScope` and set that option value.

And then if we add some code on migrateOptions, we could even remove the O# option from the package.json, leaving us with 1 unified option.

src/shared/options.ts
```diff
@@ -119,7 +119,12 @@ export class Options {
             'omnisharp.useEditorFormattingSettings',
             true
         );
-        const enableRoslynAnalyzers = Options.readOption<boolean>(config, 'omnisharp.enableRoslynAnalyzers', false);
+        const diagnosticAnalysisScope = Options.readOption<string>(
+            config,
+            'dotnet.backgroundAnalysis.analyzerDiagnosticsScope',
+            useOmnisharpServer ? 'none' : 'openFiles'
+        );
+        const enableRoslynAnalyzers = diagnosticAnalysisScope != 'none';
```
dibarbet, line 127: This looks good but I think we need to support the case where someone has specified a custom value for O# in their settings.json.  We probably still need to read the O# option names (even though they aren't defined in the package.json) and convert them to the new option.  

For other options that have the same value this is easier (see an example on line 52), but for this one we'll have to have some custom code to convert the values.

Not required for this PR as well - but we also migrate the old O# option names to the new dotnet names in here - https://github.com/dotnet/vscode-csharp/blob/main/src/shared/migrateOptions.ts#L9
Again it might be a bit more complicated for this option since we need to convert two options into one, but something to consider.

mavasani: Thanks David. As you mentioned, given that the new and legacy options are different data types, its a bit more trickier to handle the migration. I have added back the logic to respect the legacy options, though my only concern is that if that forces us back to have the UX for the same. For example, if user had enabled analyzers with legacy option, and now they want to disable it, we only allow them to change the new option value and because we implicitly respect the old option value, they are stuck with not being able to disable analyzers.

dibarbet: Yeah - lets take this, and at some point later on we can update the migration code to migrate the old option values to the new ones so we can avoid that issue.

dibarbet (approved):

src/shared/options.ts
mavasani: Filed https://github.com/dotnet/vscode-csharp/issues/6090 to track this work

PR #6060, opened 2023-08-07
Show toast when project loading fails

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/6038

image

Server side - https://github.com/dotnet/roslyn/pull/69424


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -198,6 +199,25 @@ export class RoslynLanguageServer {
...
+                modal: false,
+                detail: failure.message,
+            };
+            const selection = isError
+                ? await vscode.window.showErrorMessage(title, options, showLogs)
+                : await vscode.window.showWarningMessage(title, options, showLogs);
+            if (selection === showLogs) {
+                _channel.show();
+            }
+        });
```
jasonmalinowski, line 219: Do we just want to make this be a more general endpoint for any sort of failure, not just project loads?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -198,6 +199,25 @@ export class RoslynLanguageServer {
             );
         });
 
+        this._languageClient.onNotification(RoslynProtocol.ProjectLoadFailureNotification.type, async (failure) => {
```
jasonmalinowski, line 202: If we have a bunch of notifications at once are we going to spam the user? Or does VS Code throttle this in some way for us?

arkalyanms: +1. Can we kick off an array of projects and 1 collated exception that failed load instead?

dibarbet: from testing it looks like only 1 shows at once

arkalyanms (approved):

jasonmalinowski (approved):
It'd be nice if LSP supported this automatically!

I left one comment on the Roslyn side wondering if we should be using a prefix for the notification name, which would apply to this PR.

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: @dibarbet Once that's dismissed though, does the next one appear? Or the other are just ignored?

dibarbet: the others are ignored.  I was OK with that because these just point to the logs which will show all the errors.

PR #6062, opened 2023-08-07
Tell the language server to open projects if no solution exists

jasonmalinowski:
Fixes https://github.com/dotnet/vscode-csharp/issues/5923


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -349,8 +365,19 @@ export class RoslynLanguageServer {
...
-                    this.openSolution(solutionUris[0]);
+                if (solutionUris) {
+                    if (solutionUris.length === 1) {
+                        this.openSolution(solutionUris[0]);
+                    } else if (solutionUris.length === 0) {
+                        // We have no solutions, so we'll enumerate what project files we have and just use those.
+                        const projectUris = await vscode.workspace.findFiles(
+                            '**/*.csproj',
+                            '**/node_modules/**',
+                            options.omnisharpOptions.maxProjectResults
```
jasonmalinowski, line 376: This is an omnisharp option that I'm consuming here even on the non-omnisharp side, in case there is a user out there where this limit is significant. Having this not be a dotnet option is a bit of a violation of naming though, but I'm not sure I want to create a new option and setup the migration at this moment until we have a chance to spec it a bit more. Put another way, I'm OK with this bug at the moment.

arkalyanms:
Signing off on commit 2.

arkalyanms (approved):

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -114,6 +115,9 @@ export class RoslynLanguageServer {
      */
     private _solutionFile: vscode.Uri | undefined;
 
+    /** The project files previously opened; we hold onto this for the same reason as _solutionFile. */
+    private _projectFiles: vscode.Uri[] = new Array<vscode.Uri>();
```
arkalyanms, line 119: I understand the solution file is also a moving target to cache. But the contents were cache agnostic. The project file list on the other hand as a cache seems more volatile if cache and not  tracked continuously.

arkalyanms (approved):

src/lsptoolshost/roslynLanguageServer.ts
arkalyanms: Not a blocker, just want to get on the same  page

src/lsptoolshost/roslynLanguageServer.ts
dibarbet: Fine with as-is - lets have an issue to track.  But it should be pretty straightforward to migrate the option name if you wanted.  We just put the O# name as the fallback, then add a value to https://github.com/dotnet/vscode-csharp/blob/main/src/shared/migrateOptions.ts

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -320,21 +324,33 @@ export class RoslynLanguageServer {
...
 
-    private async sendOpenSolutionNotification(): Promise<void> {
-        if (
-            this._solutionFile !== undefined &&
-            this._languageClient !== undefined &&
-            this._languageClient.isRunning()
-        ) {
-            const protocolUri = this._languageClient.clientOptions.uriConverters!.code2Protocol(this._solutionFile);
-            await this._languageClient.sendNotification('solution/open', new OpenSolutionParams(protocolUri));
+    public async openProjects(projectFiles: vscode.Uri[]): Promise<void> {
```
dibarbet, line 331: do we have a picker for projects like we do slns?

dibarbet: I realize after I write this, a project picker for projects may not make sense, especially in huge workspaces.  Though I am curious - what happens if we load a project that depends on another project via project reference (and the dependency is not loaded)?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -320,21 +324,33 @@ export class RoslynLanguageServer {
...
+        this._solutionFile = undefined;
+        this._projectFiles = projectFiles;
+        await this.sendOpenSolutionAndProjectsNotifications();
+    }
+
+    private async sendOpenSolutionAndProjectsNotifications(): Promise<void> {
+        if (this._languageClient !== undefined && this._languageClient.isRunning()) {
+            if (this._solutionFile !== undefined) {
+                const protocolUri = this._languageClient.clientOptions.uriConverters!.code2Protocol(this._solutionFile);
+                await this._languageClient.sendNotification('solution/open', new OpenSolutionParams(protocolUri));
```
dibarbet, line 341: should this explicitly return so we don't attempt to load both a sln and projects?

src/lsptoolshost/roslynLanguageServer.ts
JoeRobich: Array constructor could have been monkey patched. Prefer `[]` since it is guaranteed to construct an array.

```suggestion
    private _projectFiles: vscode.Uri[] = [];
```

src/lsptoolshost/openProjectParams.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DocumentUri } from 'vscode-languageclient/node';
+
+export class OpenProjectParams {
```
dibarbet, line 8: I think these can go into roslynProtocol.ts, and can just be an interface (doesn't need to be a class)

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: @dibarbet I'm not including a picker _in this PR_. Omnisharp did have one but in the case of a large workspace I'm not entirely sure how useful that experience would have been, and at least one customer so far has pointed out they were switching projects when they probably didn't mean to.

If the dependency isn't there, we'll just treat it as a metadata reference and load the DLL.

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: Both shouldn't be set, but if it was there's no harm: we'll say to open the solution and then also give a set of projects (and in the end we will have loaded the union of everything).

src/lsptoolshost/openProjectParams.ts
jasonmalinowski: This is the same as openSolutionParams.ts but I agree it can be cleaned up. Do you want me to clean up both?

dibarbet: Yes please!

src/lsptoolshost/roslynLanguageServer.ts
dibarbet: Gotcha.  Lets make sure that experience is tracked as well.  I agree it could be complicated - for example it would be nice if you once you select the projects to load you can save it somewhere (e.g. similar to default solution).  Or maybe instead we just implement load projects on demand... Not sure either.

dibarbet (approved):

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: I'd hope projects-on-demand meets many needs, or yes, some other way to specify the list of projects (probably via globbing.)

src/lsptoolshost/openProjectParams.ts
jasonmalinowski: Change pushed.

dibarbet (approved):

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: To echo something I mentioned to @dibarbet privately: making the fallback work is easy, but the option is a bit funky in this case since the filtering isn't exactly deterministic in what it will filter out: it prevents certain large repositories from causing pathological problems, but it's not ideal.

jasonmalinowski:
@CyrusNajmabadi This behavior only changes the behavior for the base C# extension. DevKit has separate support for dealing with a missing sln (right now it generates one for you.) The Dev Kit owners are definitely working on fixing that and we'll probably have some more work here to better unify it.

PR #6064, opened 2023-08-08
Add localization to debugger components

WardenGnaw:
This PR adds in localization via vscode.l10n.t(...) calls to the strings
used in the debugger components of vscode-csharp.

This also moves the tests that now call vscode.l10n.t to be integration tests.
Also refactored DotnetInfo to its own file to avoid pulling in vscode for unit tests.


WardenGnaw:
Whats the best way to localize this block?
https://github.com/dotnet/vscode-csharp/blob/621a36248184c59b386b9bffaa3bc42878a9da80/src/shared/assets.ts#L453-L469

gregg-miskelly:
> Whats the best way to localize this block?
> 
> https://github.com/dotnet/vscode-csharp/blob/621a36248184c59b386b9bffaa3bc42878a9da80/src/shared/assets.ts#L453-L469

My thought:
* Turn the text into three or four localizable strings (ex: 'The C# extension was unable to automatically decode projects in the current workspace to create a runnable launch.json file. A template launch.json file has been created as a placeholder....') -- one per paragraph, and maybe one for each list item, and probably one for the word "WARNING".
* Find and/or implement a simple word wrap algorithm.
* Turn the text into the comment.

src/coreclrDebug/activate.ts
```diff
@@ -128,7 +132,9 @@ async function checkIsValidArchitecture(
             if (platformInformation.architecture === 'x86') {
                 eventStream.post(
                     new DebuggerPrerequisiteWarning(
-                        `[WARNING]: x86 Windows is not supported by the .NET debugger. Debugging will not be available.`
+                        vscode.l10n.t(
+                            `[WARNING]: x86 Windows is not supported by the .NET debugger. Debugging will not be available.`
```
gregg-miskelly, line 136: >` [](http://example.com/codeflow?start=28&length=1)

Should this be using `'` instead of back ticks?

gregg-miskelly: There are a few others like this too.

gregg-miskelly (approved):
Otherwise LGTM

WardenGnaw:
> > Whats the best way to localize this block?
> > https://github.com/dotnet/vscode-csharp/blob/621a36248184c59b386b9bffaa3bc42878a9da80/src/shared/assets.ts#L453-L469
> 
> My thought:
> 
> * Turn the text into three or four localizable strings (ex: 'The C# extension was unable to automatically decode projects in the current workspace to create a runnable launch.json file. A template launch.json file has been created as a placeholder....') -- one per paragraph, and maybe one for each list item, and probably one for the word "WARNING".
> * Find and/or implement a simple word wrap algorithm.
> * Turn the text into the comment.

Heres an example of the `wordWrapString` method.
image

gregg-miskelly (approved):

Cosifne (approved):

test/integrationTests/features/reportIssue.test.ts
```diff
@@ -3,15 +3,15 @@
...
+import { getFakeVsCode } from '../../unitTests/testAssets/fakes';
 import reportIssue from '../../../src/shared/reportIssue';
 import { expect } from 'chai';
 import { vscode } from '../../../src/vscodeAdapter';
 import { Options } from '../../../src/shared/options';
-import { FakeMonoResolver, fakeMonoInfo } from '../fakes/fakeMonoResolver';
-import { FakeDotnetResolver } from '../fakes/fakeDotnetResolver';
+import { FakeMonoResolver, fakeMonoInfo } from '../../unitTests/fakes/fakeMonoResolver';
+import { FakeDotnetResolver } from '../../unitTests/fakes/fakeDotnetResolver';
 import { DotnetInfo } from '../../../src/shared/utils/getDotnetInfo';
```
WardenGnaw, line 13: This file is already faking a bunch of objects and does not use the real getDotnetInfo. Should I just define the object here and move this file back to unit test?

PR #6074, opened 2023-08-08
Try to find a valid dotnet version from the path before falling back to runtime extension

dibarbet:
We've gotten lots of reports of issues with the .net runtime downloader extension, e.g. https://github.com/dotnet/vscode-csharp/issues/6004

While we wait for those to be fixed, we can workaround the issue by seeing if there is a valid dotnet version on the path that we can use to start the server instead.  This will not fix issues in devkit.

 TODO - verify on linux/mac.


src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -101,4 +121,55 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
+            if (result.stderr) {
+                throw new Error(`Unable to read dotnet version information. Error ${result.stderr}`);
+            }
+
+            const dotnetVersion = semver.parse(result.stdout.trimEnd());
+            if (!dotnetVersion) {
+                throw new Error(`Unknown result output from 'dotnet --version'. Received ${result.stdout}`);
+            }
+
+            if (semver.lt(dotnetVersion, this.minimumDotnetVersion)) {
```
dibarbet, line 144: everything before here is the exact code O# uses which has been working

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -101,4 +121,55 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
+     * Checks dotnet --version to see if the value on the path is greater than the minimum required version.
+     * This is adapated from similar O# server logic and should be removed when we have a stable acquisition extension.
+     * @returns true if the dotnet version is greater than the minimum required version, false otherwise.
+     */
+    private async findDotnetFromPath(): Promise<string | undefined> {
+        try {
+            // Run dotnet version to see if there is a valid dotnet on the path with a high enough version.
+            const result = await promisify(exec)(`dotnet --version`);
+
+            if (result.stderr) {
```
gregg-miskelly, line 135: Do you want to also check that it is the right architecture? https://github.com/dotnet/vscode-csharp/blob/d35d2744cb4d965f333d5cff2f30360c74461ba7/src/shared/utils/getDotnetInfo.ts#L13

dibarbet: I think technically it should but these are the reasons I have for not doing it
1.  O# has been using this code for quite a while to start its server successfully (without checking platform as best I can tell).
2.  Checking the platform could be challenging - we can easily enough find the dotnet platform by parsing `dotnet --info`.  We can also find the OS platform from `PlatformInformation`.  But I'm not as sure how I can find the platform that the server executable is targeting (which could potentially be different from the OS platform, e.g. we have an arm MacOS + arm dotnet + x64 executable).  If you know of an easy crossplatform way to check the dll, I can do it, but I don't know of one off top of my head.
3.  I want to remove this anyway as soon as the SDK extension works better and supports global installs and we have 0 of these problems.

Let me know what you think.

gregg-miskelly: > 1. O# has been using this code for quite a while to start its server successfully (without checking platform as best I can tell).

Based on the number of bugs that have been reported over the years, I would quibble with how "successful" OmniSharp is :). I haven't counted up the votes, but I would be willing to put money on this being the most reported problem in the history of this extension.

> 2. Checking the platform could be challenging

Good question. I *think* `PlatformInformation` will work, as this extension downloads the version of Rosyln (and all other dependencies) that match that. But I can try some experimentation.

> 3.  I want to remove this anyway as soon as the SDK

I think we are still going to find a check like this useful to stop having the mountains of bugs from folks having x64, even if we just port that check to the dotnet Aquisition extension.

gregg-miskelly: RE: Checking the platform --
To check the platform when we are installed as platform dependent, I think our best option is top open the `<ExtensionRoot>\.vsixmanifest` file and look at `PackageManifest.Metadata.Identity.TargetPlatform` (see below).

Otherwise, `PlatformInformation` will give us the right value since this must be we downloaded on our own.

Example manifest:
```
<?xml version="1.0" encoding="utf-8"?>
	<PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
		<Metadata>
			<Identity Language="en-US" Id="csharp" Version="2.0.328" Publisher="ms-dotnettools" TargetPlatform="win32-x64"/>
```

dibarbet: > I would quibble with how "successful" OmniSharp is :)

Fair - but at least more successful than the current version of the extension that uses the runtime downloader...

> even if we just port that check to the dotnet Aquisition extension.

Yeah ideally the API is similar to the runtime one today - we pass a exe / dll, they handle all these weird cases to acquire the correct .net

> To check the platform when we are installed as platform dependent, I think our best option is top open the <ExtensionRoot>\.vsixmanifest file and look at PackageManifest.Metadata.Identity.TargetPlatform (see below).

Hmm I can try this.

gregg-miskelly: I am not sure if we want to use this, but it looks like there is also `____metadata.targetPlatform` in the package json content.

JoeRobich: `dotnet --info` gives good information on architectures. Probably would have been smart for O# to have taken advantage of it.

```sh
~/Source/omnisharp-roslyn [master ≡]> dotnet --info
.NET SDK:
 Version:   8.0.100-preview.4.23260.5
 Commit:    2268e7b15c

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  13.5
 OS Platform: Darwin
 RID:         osx.13-arm64
 Base Path:   /usr/local/share/dotnet/sdk/8.0.100-preview.4.23260.5/

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.0-preview.6.23329.7
  Architecture: arm64
  Commit:       5340be2ccc

.NET SDKs installed:
  6.0.203 [/usr/local/share/dotnet/sdk]
  6.0.400 [/usr/local/share/dotnet/sdk]
  7.0.203 [/usr/local/share/dotnet/sdk]
  7.0.302 [/usr/local/share/dotnet/sdk]
  7.0.306 [/usr/local/share/dotnet/sdk]
  8.0.100-preview.1.23115.2 [/usr/local/share/dotnet/sdk]
  8.0.100-preview.3.23178.7 [/usr/local/share/dotnet/sdk]
  8.0.100-preview.4.23260.5 [/usr/local/share/dotnet/sdk]
  8.0.100-preview.6.23330.14 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.5 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.8 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.5 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.9 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-preview.1.23112.2 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-preview.3.23177.8 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-preview.4.23260.4 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-preview.6.23329.11 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.5 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.8 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.5 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.9 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.1.23110.8 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.3.23174.8 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.4.23259.5 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.6.23329.7 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

Other architectures found:
  x64   [/usr/local/share/dotnet/x64]
    registered at [/etc/dotnet/install_location_x64]

Environment variables:
  Not set

global.json file:
  /Users/joeyrobichaud/Source/omnisharp-roslyn/global.json

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
  https://aka.ms/dotnet/download
```

dibarbet: It looks like the target platform is exposed directly in the packagejson via the extension API as well - see https://github.com/microsoft/vscode/issues/161286#issuecomment-1288115901.  Going to check if this works.

dibarbet: > It looks like the target platform is exposed directly in the packagejson via the extension API as well - see [microsoft/vscode#161286 (comment)](https://github.com/microsoft/vscode/issues/161286#issuecomment-1288115901). Going to check if this works.

Nevermind - switching to the .vsixmanifest.  The package.json value does technically work, but it seems to only ever be defined if the extension is installed from the marketplace.  Installing via a vsix locally leaves it as undefined.  However there is a .vsixmanifest with the correct architecture in that case so it seems a bit better.

dibarbet: Updated!

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -101,4 +123,103 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
+                );
+            }
+
+            const dotnetVersion = semver.parse(dotnetVersionStr);
+            if (!dotnetVersion) {
+                throw new Error(`Unknown result output from 'dotnet --version'. Received ${dotnetVersionStr}`);
+            }
+
+            if (semver.lt(dotnetVersion, this.minimumDotnetVersion)) {
+                throw new Error(
```
gregg-miskelly, line 154: Would it make sense to issue a warning and fall back to the dotnet acquisition extension in this case?

gregg-miskelly: Nevermind. I see.

gregg-miskelly (approved):

src/shared/utils/getDotnetInfo.ts
```diff
@@ -25,6 +25,7 @@ export async function getDotnetInfo(dotNetCliPaths: string[]): Promise<DotnetInf
```
marcpopMSFT, line 25: This file uses --info. --info is not meant for automation and does not have a guaranteed format. The recommendation is to use --list-runtimes and --list-sdks. If there is data you need that is not avaialble in those existing commands, please let us know and we can look into additional stable, machine readable options that are not --info.

CC @elinor-fung @vitek-karas since the work would come out of the host to add new information/commands potentially.

dibarbet: I can probably replace the version lookup using the --list-sdks command with some work.  However we also read the runtimeid (for the debugger) and the architecture (for this PR) which don't appear to be in either of those commands.

Generally I would like to avoid needing to parse dotnet info here as well - and once the new version of the sdk acquisition extension is available I will likely be able to remove the part of the code I'm adding here.  I'm not so sure about the other use cases like the debugger (cc @gregg-miskelly).

gregg-miskelly: For the debugger, we are trying to check the version and architecture of the globally installed .NET, as we need the one the application is going to run on. If we want to switch the target app to run on the version of .NET that the acquisition extension downloads, we could probably get rid of the check. But `dotnet --list-runtimes` and `dotnet --list-sdks` don't have architecture information, so those commands aren't useful enough.

gregg-miskelly: Thinking about it more: as long as we provide a way to specify a .NET SDK in the extension, we probably are going to need a way to check the architecture.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -101,4 +123,103 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
+            const command = this.platformInfo.isWindows() ? 'where' : 'which';
+            const whereOutput = await promisify(exec)(`${command} dotnet`);
+            if (!whereOutput.stdout) {
+                throw new Error(`Unable to find dotnet from ${command}.`);
+            }
+
+            const path = whereOutput.stdout.trim();
+            if (!existsSync(path)) {
+                throw new Error(`dotnet path does not exist: ${path}`);
+            }
```
AArnott, line 169: I reused this code for C# Dev Kit, but it failed in AzP because `where` returns one line for each dotnet found on the PATH, and this code assumes exactly one line will be produced. I'll fix my copy. But I suspect your code here will fail on any customer machine that has dotnet.exe installed in two places.

dibarbet: Ah - thanks for the info.  I'll take a look.

Issue #5731, opened 2023-06-08
Theme / color coding does not render correctly using LSP / pre release version


fuzzlebuck:
## Issue Description ##

After switching to the new pre release v2.0.206 the color formatting / theming does not appear correctly.

## Steps to Reproduce ##

Open up any example C# unit, switch to the pre release version of the extension and reload the ide.

Note that colors do not appear correctly, most of the color definition is missing, everything looks similar.

Switch between the preview version and the release version to see the tokens with different colors.

It's more subtle on the default themes like 'Dark modern' but much more obvious on other themes like the one I use 'Monokai Pro'

Seems that different tokens / elements are not being identified correctly, so everything looks the same color.

## Expected Behavior ##

Code should appear using the same colors in the current release version.

[How Monokai Pro should look](https://i.imgur.com/BOwf2D2.png)
[How Dark Modern should look](https://i.imgur.com/MzZBzId.png)

## Actual Behavior ##

Colors are not right.

[How Monokai Pro looks with the LSP / pre release extension](https://i.imgur.com/CRtFnQ8.png)
[How Dark Modern looks with the LSP / pre release extension](https://i.imgur.com/QYMFJfZ.png)

## Environment information ##

Have confirmed this on both 1.79.0-insider and 1.78.2

**VSCode version**: 1.79.0-insider
**C# Extension**: 1.25.9

<details><summary>Dotnet Information</summary>
.NET SDK:
 Version:   7.0.105
 Commit:    e1bc5e001c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.105\

Host:
  Version:      7.0.5
  Architecture: x64
  Commit:       8042d61b17

.NET SDKs installed:
  3.1.426 [C:\Program Files\dotnet\sdk]
  5.0.301 [C:\Program Files\dotnet\sdk]
  5.0.408 [C:\Program Files\dotnet\sdk]
  5.0.411 [C:\Program Files\dotnet\sdk]
  6.0.203 [C:\Program Files\dotnet\sdk]
  7.0.100 [C:\Program Files\dotnet\sdk]
  7.0.105 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #6094, opened 2023-08-10
Adjusts csharp semantic token scopes to match 1.26 outputs

fuzzlebuck:
This PR adjusts the semantic token scopes to match those in earlier versions, 1.26 etc.

Directly relates to Issue #5731

I've compared the foreground outputs using the Inspect Editor Tokens in version 1.26.0 and in this version an confirmed the primary foreground now matches for all of the updated scopes in this PR.

For example the two listed in the original issue:

**Namespace in 1.26.0**:

image

**Namespace in latest**:

image

**Parameter in 1.26.0**:

image

**Parameter in latest**:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -4736,6 +4736,12 @@
       {
         "language": "csharp",
         "scopes": {
+          "typeParameter": [
+            "entity.name.type.type-parameter"
+          ],
+          "keyword": [
+            "keyword.cs"
+          ],
           "excludedCode": [
             "support.other.excluded.cs"
           ],
--- a/package.json
+++ b/package.json
@@ -4737,40 +4737,40 @@
         "language": "csharp",
         "scopes": {
           "namespace": [
-            "entity.name.type.namespace.cs"
+            "entity.name.namespace"
           ],
           "type": [
             "entity.name.type"
           ],
           "class": [
-            "entity.name.type.class.cs"
+            "entity.name.type.class"
           ],
           "enum": [
-            "entity.name.type.enum.cs"
+            "entity.name.type.enum"
           ],
           "interface": [
-            "entity.name.type.interface.cs"
+            "entity.name.type.interface"
           ],
           "struct": [
-            "entity.name.type.struct.cs"
+            "entity.name.type.struct"
           ],
           "typeParameter": [
             "entity.name.type.type-parameter.cs"
           ],
           "parameter": [
-            "entity.name.variable.parameter.cs"
+            "variable.parameter"
           ],
           "variable": [
             "entity.name.variable.local.cs"
           ],
           "property": [
-            "variable.other.property.cs"
+            "variable.other.property"
           ],
           "enumMember": [
-            "variable.other.enummember.cs"
+            "variable.other.enummember"
           ],
           "method": [
-            "entity.name.function.member.cs"
+            "entity.name.function.member"
           ],
           "macro": [
             "keyword.preprocessor.cs"
@@ -4809,7 +4809,7 @@
             "entity.name.variable.field.cs"
           ],
           "constant": [
-            "variable.other.constant.cs"
+            "variable.other.constant"
           ],
           "extensionMethod": [
             "entity.name.function.extension.cs"
--- a/package.json
+++ b/package.json
@@ -4736,48 +4736,6 @@
       {
         "language": "csharp",
         "scopes": {
-          "namespace": [
-            "entity.name.namespace"
-          ],
-          "type": [
-            "entity.name.type"
-          ],
-          "class": [
-            "entity.name.type.class"
-          ],
-          "enum": [
-            "entity.name.type.enum"
-          ],
-          "interface": [
-            "entity.name.type.interface"
-          ],
-          "struct": [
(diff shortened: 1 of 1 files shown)


fuzzlebuck:
@dotnet-policy-service agree

fuzzlebuck:
I believe the tests are failing for an unrelated reason, and I'm unable to restart the pipeline due to lack of access if someone could retry them for me?

package.json
```diff
@@ -4737,40 +4737,40 @@
         "language": "csharp",
         "scopes": {
           "namespace": [
-            "entity.name.type.namespace.cs"
+            "entity.name.namespace"
```
dibarbet, line 4740: So looking at the 1.x version, it doesn't specify any fallback textmate scopes for semantic tokens.  Which I believe means that it uses the default fallbacks as defined here - https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#predefined-textmate-scope-mappings .

I think removing the fallbacks for the core LSP semantic token names (aka anything in that table) would have the same affect as what you've done here.  Would you be able to try that?  If it works I think that is the direction we should take.

dibarbet (commented):
Thanks for the contribution - one suggestion to try out.

package.json
fuzzlebuck: Makes sense, thank you, I've made the change and it looks good!

It's also resolved a 1.x issue I had noticed where local variables had the .cs namespace so looked incorrect in some themes.

dibarbet:
So trying it out locally - it definitely doesn't look right (mainly keywords I think?) on Visual Studio 2019 Dark
image

What the colors should look like:
image

It looks like its getting the `keyword.control` scope which is purple in the theme.  But these aren't control keywords.  I do see that there isn't a fallback scope listed for `keyword` in the table - we might need to keep that one?

cc @JoeRobich in case he knows of another idea

package.json
```diff
@@ -4736,48 +4736,6 @@
...
-          "enum": [
-            "entity.name.type.enum.cs"
-          ],
-          "interface": [
-            "entity.name.type.interface.cs"
-          ],
-          "struct": [
-            "entity.name.type.struct.cs"
-          ],
-          "typeParameter": [
```
dibarbet, line 4757: `typeParameter doesn't list a fallback in the table, we probably need to keep this one

package.json
```diff
@@ -4736,48 +4736,6 @@
...
-          "enumMember": [
-            "variable.other.enummember.cs"
-          ],
-          "method": [
-            "entity.name.function.member.cs"
-          ],
-          "macro": [
-            "keyword.preprocessor.cs"
-          ],
-          "keyword": [
```
dibarbet, line 4778: same with keyword here.

dibarbet: If `keyword.cs` doesn't work on the other themes - try just 'keyword'.  It looks like it should have the correct color in the theme json
https://github.com/dotnet/vscode-csharp/blob/main/themes/vs2019_dark.json#L311

fuzzlebuck: Okay, thank you, I will update and will double check them one by one and compare the outputs.

package.json
fuzzlebuck: This one is interesting, in 1.x it maps as **entity.name.type.parameter**
image

Putting it back in the scopes list for 2.x it comes out as expected **entity.name.type.type-parameter.cs**, but should it be the more generic one or the .cs specific one?
image

fuzzlebuck: Actually ignore that I can see in the VS theme's it's using the full .cs version.

fuzzlebuck: I think we should set it to "entity.name.type.type-parameter" in the scopes

dibarbet: > This one is interesting, in 1.x it maps as entity.name.type.parameter

Interesting - I don't know where that is coming from.  

> I think we should set it to "entity.name.type.type-parameter" in the scopes

Yup, I'm good with that.  If though `entity.name.type.parameter` looks better on other themes, and it works with the VS theme, I'm also OK with setting to that.  But `entity.name.type.type-parameter` seems to be what O# used.

fuzzlebuck:
Updated with changes and compared each token that was removed against v 1.26.0:

1.26.0:
image

Latest update:
image

A couple of small discrepancies to mention:

* The preprocessor has a different color now due to it being mapped to **entity.name.function.preprocessor** whereas in 1.26.0 it was mapped to **keyword.preprocessor.cs**, let me know if you think I should change it to the .cs one.

* The type parameter for generics is now mapping to **entity.name.type.type-parameter** whereas in 1.26.0 it was mapped to **entity.name.type.parameter** although in the themes I looked at it made no difference aesthetically.

* I couldn't work out how to match a plain **entity.name.type** in code.

package.json
fuzzlebuck: I checked this one on a couple of themes and it looks ok to me.

package.json
fuzzlebuck: Added this back in, looks ok now, please see my latest comment with comparisons.

fuzzlebuck:
Happy to help, and thank you!

PR #6103, opened 2023-08-10
Add Razor grammar tests

allisonchou:
Moves Razor grammar tests from `dotnet/razor` to `dotnet/vscode-csharp` so we can completely get rid of our ts and node dependencies in the Razor repo. The Razor grammar tests run via jest instead of mocha, so I added it as a dev dependency in package.json.


package.json
```diff
@@ -1013,17 +1020,17 @@
                     "items": {
                       "type": "string"
                     },
-                    "description": "Array of symbol server URLs (example: http\u200b://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
+                    "description": "Array of symbol server URLs (example: http​://MyExampleSymbolServer) or directories (example: /build/symbols) to search for .pdb files. These directories will be searched in addition to the default locations -- next to the module and the path where the pdb was originally dropped to.",
```
dibarbet, line 1023: these changes need to be reverted

package.json
```diff
@@ -117,6 +118,7 @@
     "@types/del": "3.0.1",
     "@types/fs-extra": "5.0.4",
     "@types/gulp": "4.0.5",
+    "@types/jest": "^24.0.25",
```
dibarbet, line 121: I am very very wary about being in the business of managing yet another test frameworks.  If we're going to take ownership of these in this repo I think we need to convert these tests.

test/razorTests/Microsoft.AspNetCore.Razor.VSCode.Grammar.Test/README.md
```diff
@@ -0,0 +1,50 @@
+﻿# VSCode TextMate Grammar Tests
+
+[Jest](https://jestjs.io/en) is used for VS Code TextMate Grammar unit tests.
+
+## Setup
+
+```bash
+yarn install
```
dibarbet, line 8: Is this applicable?  Do these actually require yarn?  We definitely should not be introducing another build system when we're already using npm/gulp.

dibarbet (changes requested):

davidwengier:
Isn't this missing the actual Razor grammar? Or does that come from a package from the Razor repo still?

allisonchou:
> Isn't this missing the actual Razor grammar? Or does that come from a package from the Razor repo still?

The Razor grammar was previously added to this repo and can be found here: https://github.com/dotnet/vscode-csharp/blob/main/src/razor/syntaxes/aspnetcorerazor.tmLanguage.json

package.json
dibarbet: can also fix it using the `gulp generateOptionsSchema` commmand

davidwengier:
> The Razor grammar was previously added to this repo and can be found here: [`main`/src/razor/syntaxes/aspnetcorerazor.tmLanguage.json](https://github.com/dotnet/vscode-csharp/blob/main/src/razor/syntaxes/aspnetcorerazor.tmLanguage.json?rgh-link-date=2023-08-10T22%3A34%3A32Z)

Oh okay, cool. Though that's the compiled grammar, and the [yaml file](https://github.com/dotnet/razor/blob/main/src/Razor/src/Microsoft.AspNetCore.Razor.VSCode.Extension/syntaxes/aspnetcorerazor.tmLanguage.yml) is the source for it. Is the plan for that to move in a subsequent step?

allisonchou:
> Oh okay, cool. Though that's the compiled grammar, and the [yaml file](https://github.com/dotnet/razor/blob/main/src/Razor/src/Microsoft.AspNetCore.Razor.VSCode.Extension/syntaxes/aspnetcorerazor.tmLanguage.yml) is the source for it. Is the plan for that to move in a subsequent step?

Ah, good call on the missing yml file. Just added it!

davidwengier:
Can we also add tasks/scripts/whatever to compile it? Also might want to consider having something to pull in the latest csharp, html, css and js grammars like we have here: https://github.com/dotnet/razor/blob/main/eng/scripts/UpdateTextMate.ps1
(and when we remove everything from the Razor repo we can also update that script to pull the razor grammar from this repo)

package.json
allisonchou: (Resolved offline - the vscode-csharp repo now contains jest)

package.json
```diff
@@ -55,12 +55,14 @@
...
     "l10nDevGenerateLocalizationBundle": "npx @vscode/l10n-dev export --outDir ./l10n ./src",
-    "compile": "tsc -p ./ && npx eslint ./ && npm run l10nDevGenerateLocalizationBundle",
+    "compile": "tsc -p ./ && npx eslint ./ && npm run l10nDevGenerateLocalizationBundle && npm run compile:razorTextMate",
+    "compile:razorTextMate": "npx js-yaml src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml > src/razor/syntaxes/aspnetcorerazor.tmLanguage.json",
     "compileDev": "tsc -p ./ && npx eslint ./ && webpack --mode development && npm run l10nDevGenerateLocalizationBundle",
     "watch": "tsc -watch -p ./",
     "tdd": "mocha --config ./.mocharc.jsonc --watch --watch-extensions ts omnisharptest/omnisharpUnitTests/**/*.test.ts*",
     "test": "tsc -p ./ && gulp test",
     "test:integration": "tsc -p ./ && gulp test:integration",
+    "test:razor": "tsc -p ./ && gulp test:razor",
```
davidwengier, line 65: Should this run `compile:razorTextMate` too?

davidwengier (approved):

package.json
allisonchou: Thanks, good catch!

scripts/UpdateTextMate.ps1
```diff
@@ -0,0 +1,40 @@
+# This script updates our *.tmLanguage.json files from the latest copies in the VS Code repo.
```
dibarbet, line 1: The scripts folder contains scripts that ship in the extension VSIX and are used by the extension at runtime.  I don't think this is needed at runtime right?  This should move somewhere else.

If its not too hard, the best place for this is a gulp task (if this ever needs to run in CI I think this needs to be gulp, not sure if all the linux machines have powershell).

scripts/UpdateTextMate.ps1
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    # tmLanguage.json
+    $url = "https://raw.githubusercontent.com/microsoft/vscode/$sha/extensions/$lang/syntaxes/$filename"
+    Write-Host "Downloading $url"
+
+    $content = Invoke-WebRequest -Uri $url
+
+    # Copy to grammar tests
+    $content.content | Out-File -FilePath "../test/razorTests/Microsoft.AspNetCore.Razor.VSCode.Grammar.Test/embeddedGrammars/$fileName" -Encoding ascii
```
dibarbet, line 28: It looks like this only updates grammars for tests, is that correct?

dibarbet (commented):
Just need to figure out what to do with UpdateTextMate.ps1.  I didn't review the tests individually (just the infra)

scripts/UpdateTextMate.ps1
allisonchou: Ah, didn't realize the scripts folder is shipped. Yeah, UpdateTextMate.ps1 isn't needed at runtime. I moved it to the Razor tests folder since it's only used for tests at the moment. Currently it doesn't run in Razor's CI so I think it's fine to leave it as a Powershell script for now and convert it to a gulp task later if we decide we do want it in CI

scripts/UpdateTextMate.ps1
allisonchou: Yep, only grammars for tests

dibarbet (approved):

PR #6110, opened 2023-08-11
Support unit test debugging options when attaching to unit test process

dibarbet:
Part of https://github.com/dotnet/vscode-csharp/issues/5719#issuecomment-1674076732

Tested that various logging, symbol options, etc are able to be modified and respected when debugging unit tests.


dibarbet:
Actually - need to move the options into the dotnet section.  Converting to draft.

package.json
```diff
@@ -1555,6 +1346,215 @@
             "description": "Search symbols in reference assemblies. It affects features requires symbol searching, such as add imports.",
             "order": 80
           },
+          "dotnet.unitTestDebuggingOptions": {
```
dibarbet, line 1349: moved from the O# option section and renamed, let me know if I should keep the same name

WardenGnaw (approved):

gregg-miskelly (approved):

package.json
Cosifne: I am fine the name if debugger team like this

Cosifne (approved):

PR #6118, opened 2023-08-11
Move package.nls.*.json to root

Cosifne:
Looks like vs code can only load package.nls.*json from root dir.
Verified:
image


package.json
```diff
@@ -70,7 +70,7 @@
     "unpackage:vsix": "gulp vsix:release:unpackage",
     "updatePackageDependencies": "gulp updatePackageDependencies",
     "l10nDevGenerateXlf": "npx @vscode/l10n-dev generate-xlf ./package.nls.json ./l10n/bundle.l10n.json --outFile ./loc/vscode-csharp.xlf",
-    "l10nDevImportXlf": "npx @vscode/l10n-dev import-xlf ./loc/vscode-csharp.*.xlf --outDir ./l10n"
+    "l10nDevImportXlf": "npx @vscode/l10n-dev import-xlf ./loc/vscode-csharp.*.xlf --outDir ./l10n && move /l10n/package.nls.*.json ."
```
Cosifne, line 73: 'move' can't be used on bash. (mv should be there)
But for package.nls.json, if a contributor makes a change in package.json and creates a reference to it. That's everything. 
This task should not be run on dev's machine. It should only be run on CI (windows environment, because localization tasks can only be run there)

Cosifne: I can't find anything about @vscode/l10n-dev import-xlf to support multiple output-dir. Seems like it can only have one out dir
image

dibarbet: CI is currently linux based I believe

dibarbet: Can we just specify '.' as the output dir so we don't have to move it?

Cosifne: No, localization CI is windows see https://github.com/dotnet/vscode-csharp/blob/aab28042db1b880823ddd85b3b0fea3f8632df3e/azure-pipelines/loc.yml#L20

Cosifne: > Can we just specify '.' as the output dir so we don't have to move it?

Yeah but in that case we need to move `bundle.l10.*.json` to `l10` folder

Cosifne: Or it's can be something like this, we put all `bundle.l10.*.json`  and `package.nls.*.json` to root. And change this https://github.com/dotnet/vscode-csharp/blob/aab28042db1b880823ddd85b3b0fea3f8632df3e/package.json#L46

I just feel it's too many things under the root

dibarbet: > Yeah but in that case we need to move bundle.l10.*.json to l10 folder

Oooh ok yeah can leave it there and move it

> OneLocBuild Task is required to run on windows

Ah this runs in the separate pipeline - my bad.

dibarbet (approved):

PR #6119, opened 2023-08-11
Support prerelease from main instead of prerelease

dibarbet:
Example - https://dnceng.visualstudio.com/internal/_build/results?buildId=2241794&view=results

Now you have to manually select the build to release

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -1,16 +1,16 @@
 trigger: none
 pr: none
 
+resources:
+  pipelines:
+  - pipeline: officialBuildCI
+    source: dotnet-vscode-csharp
+    branch: main
+
 parameters:
   - name: test
     type: boolean
     default: true
-  - name: branch
-    type: string
-    default: prerelease
-    values:
-    - prerelease
-    - release
 
 variables:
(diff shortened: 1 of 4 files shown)


arkalyanms (approved):

azure-pipelines/release.yml
```diff
@@ -1,16 +1,16 @@
 trigger: none
 pr: none
 
+resources:
```
Cosifne, line 4: Seems like you are suppose to create release from 'release' branch, instead always from main, is that correct?

dibarbet: This is just sets the default branch to pick from, you can actually pick any branch and build you want in the resources screen.

Running this on a main build will upload a prerelease, running on a release branch build will upload a release

dibarbet: image

Cosifne: Yeah I am fine with it but just want to check (because now we run everything from main)

dibarbet: oh sorry I misunderstood - no the pipeline yaml comes from main still.  Even if the pipeline branch is main you can pick a resource from any branch

PR #6132, opened 2023-08-15
Show a prompt if we have more than one solution file

jasonmalinowski:
This prompt is pretty close to what DevKit does in this same case.


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -425,6 +425,23 @@ export class RoslynLanguageServer {
...
+                            vscode.l10n.t(
+                                'Your workspace has multiple Visual Studio Solution files; please select one to get full IntelliSense.'
+                            ),
+                            { title: vscode.l10n.t('Choose solution'), open: true },
+                            { title: vscode.l10n.t('Do not load solutions'), open: false }
+                        );
+
+                        if (chosen) {
+                            if (chosen.open) {
+                                vscode.commands.executeCommand('dotnet.openSolution');
```
jasonmalinowski, line 440: Do we also want to update the property or no like Dev Kit? We've been seeing feedback where people do have multiple solutions and might be confused if this does change it.

dibarbet (approved):

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: @dibarbet had the idea of just adding a third button. With some wordsmithing I think I got them all to fit.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -425,6 +425,32 @@ export class RoslynLanguageServer {
...
                         this.openSolution(solutionUris[0]);
+                    } else if (solutionUris.length > 1) {
+                        // We have more than one solution, so we'll prompt the user to use the picker.
+                        const chosen = await vscode.window.showInformationMessage(
+                            vscode.l10n.t(
+                                'Your workspace has multiple Visual Studio Solution files; please select one to get full IntelliSense.'
+                            ),
+                            { title: vscode.l10n.t('Choose'), action: 'open' },
+                            { title: vscode.l10n.t('Choose and set default'), action: 'openAndSetDefault' },
+                            { title: vscode.l10n.t('Do not load any'), action: 'disable' }
```
dibarbet, line 436: Other option:
'Disable load'

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -425,6 +425,32 @@ export class RoslynLanguageServer {
...
                     if (solutionUris.length === 1) {
                         this.openSolution(solutionUris[0]);
+                    } else if (solutionUris.length > 1) {
+                        // We have more than one solution, so we'll prompt the user to use the picker.
+                        const chosen = await vscode.window.showInformationMessage(
+                            vscode.l10n.t(
+                                'Your workspace has multiple Visual Studio Solution files; please select one to get full IntelliSense.'
+                            ),
+                            { title: vscode.l10n.t('Choose'), action: 'open' },
+                            { title: vscode.l10n.t('Choose and set default'), action: 'openAndSetDefault' },
```
dibarbet, line 435: other option:
'Choose default'

dibarbet (approved):

src/lsptoolshost/roslynLanguageServer.ts
jasonmalinowski: I thought about that but decided to make it a bit clearer the second button is first button plus something more. But if we have to save room this would proably be where I'd do it.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -425,6 +425,32 @@ export class RoslynLanguageServer {
                 if (solutionUris) {
                     if (solutionUris.length === 1) {
                         this.openSolution(solutionUris[0]);
```
genlu, line 427: Sorry I'm a bit late to the party. Since we are trying to improve this area, could we also show the solution picker if default solution can't be opened and there's other solution files in the hierachy?

> [Error - 1:58:27 PM] [LanguageServerHost]Microsoft.Build.Exceptions.InvalidProjectFileException: The project file could not be loaded. Could not find file 'c:\Users\gel\source\repos\ConsoleApp16\ConsoleApp16.sln'.  c:\Users\gel\source\repos\ConsoleApp16\ConsoleApp16.sln

jasonmalinowski: @genlu We _could_, although I'm still curious how you got an error like that in the first place. :smile:

PR #6139, opened 2023-08-16
Use deployment stage to require a manual approval to run the pipeline

dibarbet:
Looks like
image

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -17,64 +17,70 @@ variables:
 - group: vscode-csharp release secrets
 
 jobs:
-- job: PublishToMarketplace
+- deployment: PublishToMarketplace
+  displayName: PublishToMarketplace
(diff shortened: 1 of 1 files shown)


azure-pipelines/release.yml
```diff
@@ -17,64 +17,70 @@ variables:
 - group: vscode-csharp release secrets
 
 jobs:
-- job: PublishToMarketplace
+- deployment: PublishToMarketplace
```
dibarbet, line 20: review with whitespace hidden.

dibarbet, line 20: Using a deployment job with an environment that requires approval seems to be the only way to enable manual approvals in a yaml pipeline.  See https://stackoverflow.com/questions/61656077/implementing-gates-in-azure-yaml-pipelines

arkalyanms (approved):

PR #6140, opened 2023-08-16
Allow release pipeline to automatically create tag in Roslyn and vscode-csharp

Cosifne:
Add another stage in release pipeline so whenever a release is created, dotnet-bot will create release tag for us in roslyn and vscode-csharp repo.

When you select the correct resources in the pipeline, the stage would
1. Checkout the package.json and nuget.config from the resources commit. 
2. Find the roslyn version from package.json and nuget server from nuget.config.
3. Download the nuspec file from the server. And find the commit from there.
4. Use dotnet-bot to create a release tag back to roslyn repo using the commit.
5. Use dotnet-bot to create a release tag in vscode-csharp using the resource commit.

Note: this rely on the commit number we shipped in nuget package.
image

I have tested it in my own folk https://github.com/Cosifne/roslyn/releases/tag/v0.0.6.


tasks/createTagsTasks.ts
```diff
@@ -0,0 +1,169 @@
...
+    const nugetConfig = await xml2js.parseStringPromise(nugetFileString);
+    const nugetServerIndex = nugetConfig.configuration.packageSources[0].add[0].$.value;
+    if (!nugetServerIndex) {
+        logError("Can't find nuget server index.");
+        return null;
+    }
+
+    console.log(`Nuget server index is ${nugetServerIndex}`);
+    // Find the matching commit from nuget server
+    // 1. Ask the PackageBaseAddress from nuget server index.
```
Cosifne, line 80: How it works:
1. Parse the package.json to get roslyn version.
2. Parse the nuget.config to get server index.
3. Send http request to nuget server index, it would gives something like this:
image
This is required to be implemented by nuget server so we should be able to get it.
4. Use that to get nuspec file https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource#download-package-manifest-nuspec
5. From nuspec file, find the commit associated with the nuget package.

tasks/createTagsTasks.ts
```diff
@@ -0,0 +1,169 @@
...
+): Promise<boolean> {
+    const pat = process.env['GitHubPAT'];
+    if (!pat) {
+        throw 'No GitHub Pat found.';
+    }
+
+    console.log(`Start to tag ${owner}/${repo}. Commit: ${commit}, tag: ${releaseTag}, message: ${tagMessage}`);
+    const octokit = new Octokit({ auth: pat });
+    await octokit.auth();
+    const createTagResponse = await octokit.request(`POST /repos/${owner}/${repo}/git/tags`, {
```
Cosifne, line 123: GitHub requires to create `tag` first, then create the reference.
See https://docs.github.com/en/rest/git/tags?apiVersion=2022-11-28#create-a-tag-object

azure-pipelines/release.yml
```diff
@@ -15,66 +15,100 @@ parameters:
...
-  - pwsh: |
-      # Our build pipeline would generated build based on attempt number. Publishing the latest attempt.
-      $allArtifacts = Get-ChildItem -Path "VSIXs - Attempt*" | Sort-Object -Descending
-      if ($allArtifacts.Length -eq 0) {
-        throw "No Artifacts is downloaded."
-      }
+stages:
+- stage: PublishStage
+  jobs:
+  - job: PublishToMarketplace
```
dibarbet, line 24: Apologies I just merged my approval change which created merge conflicts.  This will need to take the change from my PR to turn this into a deployment job - https://github.com/dotnet/vscode-csharp/pull/6139

azure-pipelines/release.yml
```diff
@@ -15,66 +15,100 @@ parameters:
...
+    - pwsh: |
+        # Need these two files to find the roslyn commit for the given branch
+        git checkout $(resources.pipeline.officialBuildCI.sourceCommit) -- ./package.json ./server/NuGet.config
+      displayName: 'Checkout package.json and NuGetConfig from targeting branch'
+    - pwsh: |
+        npm ci
+        npm install
+        npm install -g gulp
+        # build number is generated by nerdbank git versioning and set to AzureDev
+        gulp createTags --releaseVersion $(resources.pipeline.officialBuildCI.runName) --releaseCommit $(resources.pipeline.officialBuildCI.sourceCommit)
```
dibarbet, line 111: Should we skip this (or have some kind of dry run) if this is a test run?

tasks/createTagsTasks.ts
dibarbet: So web requests are kind of flaky and I'm hesitent to have the release CI fail because of them.  

I think a more reliable approach would be to write out the commit to some file in the .roslyn directory when we download the package.  Because when we download the package we have the .nuspec file on disk already - we can just read it and extract the commit to a file inside the .roslyn directory.
image

tasks/createTagsTasks.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    const lowerCaseLanguageServerPackageName = 'Microsoft.CodeAnalysis.LanguageServer'.toLocaleLowerCase();
+    const nuspecGetUrl = `${id}${lowerCaseLanguageServerPackageName}/${roslynVersion}/${lowerCaseLanguageServerPackageName}.nuspec`;
+    console.log(`Url to get nuspec file is ${nuspecGetUrl}`);
+    const nuspecResponse = await sendHttpsGetRequestAsync(nuspecGetUrl);
+    if (!nuspecResponse) {
+        return null;
+    }
+
+    const nuspecFile = await xml2js.parseStringPromise(nuspecResponse.data);
```
dibarbet, line 102: Instead of adding this whole dependency, we can just use a regex to find the correct line - I've done this for an xml case already here for example - https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/dotnetRuntimeExtensionResolver.ts#L196

package.json
```diff
@@ -125,11 +126,13 @@
     "@types/unzipper": "^0.9.1",
     "@types/uuid": "^9.0.1",
     "@types/vscode": "1.73.0",
+    "@types/xml2js": "^0.4.11",
```
dibarbet, line 129: see other comments, I think we can avoid this dependency

package.json
```diff
@@ -110,6 +110,7 @@
   },
   "devDependencies": {
     "@types/archiver": "5.1.0",
+    "@types/axios": "^0.14.0",
```
dibarbet, line 113: See other comments - I think we can avoid this dependency if we write out the commit information when downloading the nuget package.

tasks/createTagsTasks.ts
Cosifne: Done

tasks/createTagsTasks.ts
Cosifne: Fixed

azure-pipelines/release.yml
```diff
@@ -15,66 +15,100 @@ parameters:
...
+        Else {
+          vsce @additionalPublishArgs
+        }
+      displayName: 🚀 Publish to Marketplace
+      workingDirectory: $(Pipeline.Workspace)
+      env:
+        VSCE_PAT: $(VSCodeMarketplacePAT)
+- stage: 'TagReleaseNumberToRoslyn'
+  displayName: 'Tag release number of vscode-csharp to Roslyn'
+  dependsOn: 'PublishStage'
```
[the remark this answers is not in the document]
Cosifne: image
The doc said it will run if the prev steps/job succeed, but here it depends on the stage. To be safe I added the `succeeded('PublishStage')`

azure-pipelines/release.yml
Cosifne: I passed the `${{ paramters.test }}` into the task and skip the tagging. (Only print the commit found with the nuget packages)

tasks/createTagsTasks.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    const roslynCommit = await findRoslynCommitAsync();
+    if (!roslynCommit) {
+        logError('Failed to find roslyn commit.');
+        return;
+    }
+
+    if (options.dryRun) {
+        console.log('Tagging is skipped in dry run mode.');
+    } else {
```
dibarbet, line 30: nit:
```
if (options.dryRun) {
    log...
    return;
}

<continue tagging>
```

dibarbet (approved):

PR #6145, opened 2023-08-16
Set text to lf for bundle.l10.json

Cosifne:
According to https://github.com/microsoft/vscode-l10n/issues/129, there is no way to config bundle.l10n.json's line ending.
In order not to get the git warning again, set the line ending of bundle.l10n.json to `lf` explicitly.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# This is a generated file from vscode-l10n tool, and it always uses 'lf' and line ending. Set it to 'lf' so we don't get warning every time when localization file get changed.
+l10n/bundle.l10n.json text eol=lf


dibarbet (approved):

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# This is a generated file from vscode-l10n tool, and it always uses 'lf' and line ending. Set it to 'lf' so we don't get warning every time when localization file get changed.
+l10n/bundle.l10n.json text eol=lf
```
jasonmalinowski, line 2: Do we need to do this for the other files too?

jasonmalinowski (approved):

.gitattributes
Cosifne: I only see this file shows up in the git warning. If you see anything else maybe add them in future?

PR #6155, opened 2023-08-16
Add the configuration option for creating binlogs

jasonmalinowski:


package.json
```diff
@@ -1156,6 +1156,12 @@
             "default": null,
             "description": "Override for path to langauge server --extension arguments."
           },
+          "dotnet.projectLoading.binaryLogPath": {
```
dibarbet, line 1159: Instead of requiring the user come up with a path, we could make this a boolean and thent dump the log in the extension log folder.  We already pass that to the language server and use it for detailed logs from the test host for example.  
https://github.com/dotnet/roslyn/blob/main/src/Features/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs#L198

Then just write out that location

jasonmalinowski: I guess if we have the option either way, this is just more flexible?

dibarbet: Main reason is to just reduce the amount of work / possible failures.  Easier to just check a box than write/ copy a path to a (writeable) directory.  Not a strong opinion here just wanted to suggest as an option

dibarbet: I assume new bin logs overwrite previous?

jasonmalinowski: > I assume new bin logs overwrite previous?

I generate a GUID per session and then include that GUID in each of the binlogs. That way if you forget you had set this (or the language server restarts) you don't immediately lose your logs. That said, it might be good for people to know where they put the logs, so that way if it's growing they're aware of it, and also they don't lose them on upgrade if it's wiped.

dibarbet (approved):

PR #6174, opened 2023-08-19
Update option changes toast and switch their tests to jest

dibarbet:
It's possible to review commit at a time (a lil messy) or all at once.

This updates option toasts to reload the window (for C#), localize them, and enable them to be tested with jest.

Jest provides the following benefits:

1.  Actively maintained, including mocking tools.
2.  VSCode extension that includes a test window where we can run/debug single tests.  Previously we had to hand write custom launch.json files and couldn't run single tests.
3.  Easily able to mock the 'vscode' library for unit tests.  Previously we had to always pass around the 'vscodeAdapter' abstraction in production code to allow unit tests to work.  With jest, we can just use the regular 'vscode' library as normal in production code.
    a.  This is essentially a requirement as we add more localization strings - in order for automatic generation of the json files to work, we must use the real 'vscode' library in-place and not the vscodeAdapter.

Commits:

1.  f80420df219f2021ab4af80e2461a5f87682f917 - Updates the option toast to reload the window for the C# case to workaround https://github.com/dotnet/vscode-csharp/issues/5882
5.  4737540b1dc8ee65d13e12d56edb3ecba87b3fed - Localize options toasts.
6.  70b7a37dba926a0be99b318329a83fb4b6c429fb - Add support for jest tests and migrate options tests over to jest. 
7.  f4b9617758be3ebdaf1d05893a2b73f0188c5eb9 - Merge Ankita's test refactoring in.
8.  0532fa93c38b8e4430970faf0171d464d8381afa - Move non-o# tests out of O# folder and add non-O# leg.


omnisharptest/omnisharpJestTests/optionChangeObserver.test.ts
```diff
@@ -0,0 +1,164 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { timeout } from 'rxjs/operators';
+import { from as observableFrom, Subject, BehaviorSubject } from 'rxjs';
+import { Options } from '../../src/shared/options';
+import { registerOmnisharpOptionChanges } from '../../src/omnisharp/omnisharpOptionChanges';
+
+import * as jestLib from '@jest/globals';
```
dibarbet, line 11: It is possible to globally import jest type definitions so you can just use 'describe' - however the functions will conflict with mocha ones.  So for now we manually import the type definitions.

We also need to name it something not 'jest' as it will throw errors because jest (the runner) is already imported.

omnisharptest/omnisharpJestTests/optionChangeObserver.test.ts
```diff
@@ -0,0 +1,164 @@
...
+import * as vscode from 'vscode';
+import { getVSCodeWithConfig, updateConfig } from '../../test/unitTests/fakes';
+
+jestLib.describe('OmniSharpConfigChangeObserver', () => {
```
dibarbet, line 15: these are essentially the same tests, just migrated to jest.

test/unitTests/fakes.ts
```diff
@@ -221,6 +214,48 @@ export function getFakeVsCode(): vscode.vscode {
...
+            t: (
+                options:
+                    | string
+                    | {
+                          message: string;
+                          args?: Array<string | number | boolean> | Record<string, any>;
+                          comment: string | string[];
+                      },
+                ...args: (string | number | boolean)[] | Record<string, any>[]
+            ) => {
```
dibarbet, line 227: Yes - apparently this is ho9w you have to implement an interface that defines overloads - define one function that handles every possible variation on the input.  Quite insane really.

.vscode/extensions.json
```diff
@@ -1,6 +1,7 @@
 {
     "recommendations": [
         "dbaeumer.vscode-eslint",
-        "esbenp.prettier-vscode"
+        "esbenp.prettier-vscode",
+        "orta.vscode-jest"
```
dibarbet, line 5: this adds the jest test explorer - which is how you run and debug tests without manually configuring launch.json

src/vscodeAdapter.ts
```diff
@@ -999,4 +999,16 @@ export interface vscode {
         sessionId: string;
         openExternal(target: Uri): Thenable<boolean>;
     };
+
+    l10n: {
```
dibarbet, line 1003: We still need this type-
1.  The old mocha tests (and the production code being tested) still require it to run.
2.  The new jest tests currently require it to define the fake 'vscode' to export.  We can potentially replace this fake with real mocks once everything has migrated over.

azure-pipelines.yml
```diff
@@ -39,8 +39,25 @@ stages:
...
           demandsName: ImageOverride -equals Build.Ubuntu.2204.Amd64.Open
+        windows:
+          demandsName: ImageOverride -equals 1es-windows-2022-open
     pool:
       name: NetCore-Public
       demands: $(demandsName)
     steps:
     - template: azure-pipelines/test.yml
+
+- stage: Test_Omnisharp
```
dibarbet, line 50: Added a separate leg for O# vs blue here - we actually had a few unit tests for blue, but they were running incorrectly under the O# folder.  I moved them to the regular tests folder and added a leg to run them in CI as well

akhera99 (approved):
Added CI test leg looks good to me!

dibarbet:
going to merge, will take feedback in a followup if any.

PR #6175, opened 2023-08-19
Use dotnet --list-runtimes instead of version from dotnet --info

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/6167

The version that the old O# was picking up from dotnet --info actually was the runtime version.  This happened to work for many of us because we had .net 8 installed which is higher than .net 7.  

Instead of using SDK version and the version output by dotnet --info, we should actually just look at `dotnet --list-runtimes` (which is meant to be machine readable) to find a runtime higher than our desired runtime.  Adapted from https://devdiv.visualstudio.com/DevDiv/_git/vs-green/pullrequest/491545


arkalyanms (approved):

JoeRobich:
I see this changed O# tests. Has this change been tested with O#?

dibarbet:
> I see this changed O# tests. Has this change been tested with O#?

Only via the O# integration tests.  What broke?

JoeRobich:
> Only via the O# integration tests. What broke?

This is what I get for skimming a PR. Changes look good. Sorry for the interruption.

PR #6188, opened 2023-08-22
Integration Tests - Add gulp tasks for lsp integration tests

akhera99:
Also explicitly provides workspaces for lsp vs non lsp omnisharp test paths


test/runIntegrationTests.ts
```diff
@@ -0,0 +1,74 @@
...
+function getSln(workspacePath: string): string | undefined {
+    if (workspacePath.endsWith('slnWithGenerator')) {
+        return 'slnWithGenerator.sln';
+    }
+    return undefined;
+}
+
+async function main() {
+    try {
+        const vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
```
dibarbet, line 20: we should share this code with the O# version since this should be basically the same (even with jest).  We can probably keep only the copy in the 'test' folder.  We will need to switch certain variables based on which test is running (e.g. 'extensionTestsPath')

tasks/testTasks.ts
```diff
@@ -86,24 +91,63 @@ gulp.task(
     gulp.series('omnisharp:jest:test', 'omnisharptest:feature', 'omnisharptest:unit', 'omnisharptest:integration:stdio')
 );
 
+gulp.task('test:integration:slnWithCsproj', async () => runIntegrationTest('slnWithCsproj'));
+
 gulp.task('test:unit', async () => {
     runJestTest(/unitTests.*\.ts/);
 });
 
 gulp.task('test', gulp.series('test:unit'));
```
dibarbet, line 100: In order to enable this for CI it needs to be added here I believe

dibarbet (approved):
Approving, assuming the intent is to add to CI in a followup PR.  If not it needs to update the gulp 'test' task to include the integation task

PR #6223, opened 2023-08-25
Implement project load telemetry support

dibarbet:
Implements the same event that O# already reports but from Roslyn LSP.

Server side - https://github.com/dotnet/roslyn/pull/69714

image


src/shared/projectConfiguration.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export function reportProjectConfigurationEvent(
+    reporter: ITelemetryReporter,
+    projectConfig: ProjectConfigurationMessage,
+    platformInfo: PlatformInformation,
+    dotnetInfo: DotnetInfo | undefined,
+    solutionPath?: string,
+    useModernNet?: boolean
+) {
+    let solutionId = '';
```
sharwell, line 32: ❗ Return early without any calculations if telemetry is disabled.

src/shared/projectConfiguration.ts
```diff
@@ -0,0 +1,67 @@
...
+    telemetryProps['FileExtensions'] = projectConfig.FileExtensions.join('|');
+    telemetryProps['FileCounts'] = projectConfig.FileCounts?.join('|') ?? '';
+    telemetryProps['NetSdkVersion'] = dotnetInfo?.Version ?? '';
+    telemetryProps['sdkStyleProject'] = projectConfig.SdkStyleProject.toString();
+
+    if (useModernNet) {
+        telemetryProps['useModernNet'] = useModernNet.toString();
+    }
+
+    reporter.sendTelemetryEvent('ProjectConfiguration', telemetryProps);
```
sharwell, line 54: 💡 It would be good to update `reporter` to throw unconditionally when telemetry is disabled so these wasted CPU cycles are harder to introduce

sharwell (changes requested):

src/shared/projectConfiguration.ts
jasonmalinowski: I'd imagine that'll mean we just ship a bunch of bugs every time we aren't testing that way. But maybe have sendTelemetryEvent take a lambda which returns the item, and we skip that if we aren't using it?

src/shared/projectConfiguration.ts
jasonmalinowski: Although that skip should be happening in the LSP server side -- we shouldn't be sending notification only to ignore them _here_.

src/shared/projectConfiguration.ts
```diff
@@ -0,0 +1,67 @@
...
+    telemetryProps['TargetFrameworks'] = projectConfig.TargetFrameworks.join('|');
+    telemetryProps['References'] = projectConfig.References.join('|');
+    telemetryProps['FileExtensions'] = projectConfig.FileExtensions.join('|');
+    telemetryProps['FileCounts'] = projectConfig.FileCounts?.join('|') ?? '';
+    telemetryProps['NetSdkVersion'] = dotnetInfo?.Version ?? '';
+    telemetryProps['sdkStyleProject'] = projectConfig.SdkStyleProject.toString();
+
+    if (useModernNet) {
+        telemetryProps['useModernNet'] = useModernNet.toString();
+    }
```
jasonmalinowski, line 52: Just to note this flag is being obsoleted with the OOP work. It's still useful for Omnisharp telemetry maybe but isn't useful for the LSP case. Just something to call out if anybody is actually using this.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -206,6 +208,19 @@ export class RoslynLanguageServer {
...
+        this._languageClient.onNotification(RoslynProtocol.ProjectConfigurationNotification.type, (params) => {
+            reportProjectConfigurationEvent(
+                this.telemetryReporter,
+                params,
+                this.platformInfo,
+                dotnetInfo,
+                this._solutionFile?.fsPath,
+                true
+            );
+        });
```
jasonmalinowski, line 222: I agree we don't want to be running it all the time, but can we just run it the first time we get a message async so we're not holding up the LSP server launch? Or just include the version number in the message itself?

src/shared/projectConfiguration.ts
dibarbet: Resolved on the server side to not send the event at all.

src/shared/projectConfiguration.ts
dibarbet: It doesn't appear to be used in the dotnet data, so think we're good to remove if you need to for OOP

src/shared/projectConfiguration.ts
dibarbet: I will add something here in a followup - needs some work and it is an existing issue with O#.  Potentially we could make the roslyn lsp one throw (or log an error) and skip in the O# case.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -206,6 +209,22 @@ export class RoslynLanguageServer {
             );
         });
 
+        // Retrieve the dotnet info outside of the notification so we're not running dotnet --info every time the project changes.
```
jasonmalinowski, line 212: Stale comment now.

jasonmalinowski (approved):

dibarbet:
Merging so that this makes it in for tomorrows validation.

PR #6228, opened 2023-08-25
Fix quoting bug

333fred:
The server was not expecting paths to be quoted, but we were doing so on the client. I've also added a trace-level print of the actual arguments we will use to start the server, which helped me figure this out.

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -527,7 +527,7 @@ export class RoslynLanguageServer {
         }
 
         for (const extensionPath of this.additionalExtensionPaths) {
-            args.push('--extension', `${extensionPath}`);
+            args.push('--extension', extensionPath);
         }
 
         // Get the brokered service pipe name from C# Dev Kit (if installed).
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -527,7 +527,7 @@ export class RoslynLanguageServer {
         }
 
         for (const extensionPath of this.additionalExtensionPaths) {
-            args.push('--extension', `"${extensionPath}"`);
+            args.push('--extension', `${extensionPath}`);
```
jasonmalinowski, line 530:
```suggestion
            args.push('--extension', extensionPath);
```

333fred, line 530:
```suggestion
            args.push('--extension', extensionPath);
```

333fred:
With this fix, the compiler developer extension is alive!

genlu (approved):

genlu:
> the compiler developer extension

what is this?

333fred:
@genlu https://github.com/333fred/compiler-developer-sdk. We're using this as a testing ground for C# LSP extensibility.

PR #6236, opened 2023-08-26
Fix Razor formatting bug

allisonchou:
Fixes https://github.com/microsoft/vscode-dotnettools/issues/218

Short PR, but a hard one to investigate. The reason why nothing was happening was because the HTML formatter wasn't respecting the `html.format.preserveNewLines` setting, which in turn made some of our code go from:
```Razor
@model Test

Text
```
to
```Razor
@model Test Text
```
Since the above is invalid Razor syntax, it caused Razor's `FormattingDiagnosticValidationPass` to fail, resulting in our formatter doing nothing.

While the HTML formatter failed to respect the `html.format.preserveNewLines` setting, the HTML _range_ formatter respected it. I was confused as to why this was the case, and started digging into the HTML code. I eventually filed https://github.com/microsoft/vscode/issues/191395 which I believe is the root cause. (Tl;dr HTML isn't explicitly registering a document formatter, which I believe they need to do in order to pick up settings.) Once the HTML-side bug is fixed, Razor should hypothetically be able to remove the workaround in this PR.

Razor formatting will still be broken in the case where the user turns `html.format.preserveNewLines` off (it's on by default, so hopefully this case is rare), since vscode's FormattingOptions doesn't take custom formatter settings. The real long-term fix here is probably to complete https://github.com/dotnet/razor/issues/7564 on Razor's end, but that seems to be a ways off.

### DIFF
--- a/src/razor/src/formatting/formattingHandler.ts
+++ b/src/razor/src/formatting/formattingHandler.ts
@@ -60,7 +60,7 @@ export class FormattingHandler {
             const textDocument = await vscode.workspace.openTextDocument(razorDocumentUri);
             const synchronized = await this.documentSynchronizer.trySynchronizeProjectedDocument(
                 textDocument,
-                razorDocument.csharpDocument,
+                razorDocument.htmlDocument,
                 formattingParams.hostDocumentVersion,
                 cancellationToken
             );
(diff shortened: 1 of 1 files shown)


davidwengier (approved):
Wow, nice find!

One question, and maybe you don't know the answer which is cool, but if HTML isn't registering a document formatter, then why would it be called at all when we executed the formatting command? Essentially, I wonder if the response will be that this is by design, and the workaround will therefore be permanent. Fine by me if it is, just thinking out loud :)

allisonchou:
> Wow, nice find!
> 
> 
> 
> One question, and maybe you don't know the answer which is cool, but if HTML isn't registering a document formatter, then why would it be called at all when we executed the formatting command? Essentially, I wonder if the response will be that this is by design, and the workaround will therefore be permanent. Fine by me if it is, just thinking out loud :)

Good question, to be honest I'm not quite sure. My theory is that there's some fallback way they call the formatter, but I couldn't find the exact path through my digging. Will be curious to see how the VS Code HTML team responds. 😊

PR #6249, opened 2023-08-29
Convert new integration tests to jest and add to CI

dibarbet:
This converts the new integration tests to use jest.

Note - these tests do not work in the jest explorer (the test needs to launch vscode which launches the tests).  It is likely possible to do with a custom jest test runner (install vscode, launch, run tests, get output json, report back), but it isn't extremely simple.  So for now we just rely on the gulp commands or launch.json to run the tests.

image


test/integrationTests/index.ts
```diff
@@ -3,15 +3,47 @@
...
+import * as tty from 'tty';
 import * as path from 'path';
-import * as testRunner from '../testRunner';
+import { jestIntegrationTestProjectName } from './jest.config';
+
+// Linux: prevent a weird NPE when mocha on Linux requires the window size from the TTY
+// Since we are not running in a tty environment, we just implement the method statically
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+tty.getWindowSize = function () {
```
dibarbet, line 16: todo - remove, not using mocha

jest.config.ts
```diff
@@ -5,17 +5,11 @@
...
-    testEnvironment: 'node',
-    transformIgnorePatterns: ['/dist/.+\\.js'],
-    // We need to explicity ignore the out directory for modules - otherwise we'll get duplicate vscode module,
-    // the TS version from the __mocks__ directory and the compiled js version from the out directory.
-    modulePathIgnorePatterns: ['out'],
-    // Specify jest to only run tests in jest folders.
-    // We also have to include the __mocks__ folder.  That folder must be next to node_modules so we can't move it,
-    // but if we specify roots, jest won't automatically pick it up.  So we have to specify it here.
-    roots: ['<rootDir>/test/unitTests', '<rootDir>/omnisharptest/omnisharpJestTests', '<rootDir>/__mocks__'],
+    projects: [
```
dibarbet, line 8: we define separate jest 'projects' for our test runs.  this allows us to easily select which set to run and have different configurations for each.

azure-pipelines.yml
```diff
@@ -38,26 +38,28 @@ stages:
     strategy:
       matrix:
         linux:
-          demandsName: ImageOverride -equals Build.Ubuntu.2204.Amd64.Open
+          demandsName: 1es-ubuntu-2004-open
```
dibarbet, line 41: the demands weren't being respected for some reason when we put the entire string in here.  So everything was actually just running on windows...

Now we actually run on the correct platform.

dibarbet: This also uses 1es-ubuntu because it has more of the packages we need (libgtk, libgbm, xvfb)

akhera99 (approved):

PR #6264, opened 2023-08-31
Consume per-runtime language servers

jasonmalinowski:
To keep our package size more manageable, we now create a separate language server package per runtime.

This consumes https://github.com/dotnet/roslyn/pull/69786.

### DIFF
--- a/tasks/offlinePackagingTasks.ts
+++ b/tasks/offlinePackagingTasks.ts
@@ -73,26 +73,25 @@ gulp.task('installDependencies', async () => {
     }
 });
 
+gulp.task(
+    'updateRoslynVersion',
+    // Run the fetch of all packages, and then also installDependencies after
+    gulp.series(async () => {
+        const packageJSON = getPackageJSON();
+
+        // Fetch the neutral package that we don't otherwise have in our platform list
+        await acquireRoslyn(packageJSON, undefined, true);
+
+        // And now fetch each platform specific
+        for (const p of platformSpecificPackages) {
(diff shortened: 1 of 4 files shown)


dibarbet (approved):

tasks/offlinePackagingTasks.ts
```diff
@@ -73,22 +73,21 @@ gulp.task('installDependencies', async () => {
     }
 });
 
-// Install Tasks
-async function installRoslyn(packageJSON: any, platformInfo?: PlatformInformation) {
-    const roslynVersion = packageJSON.defaults.roslyn;
-    const packagePath = await acquireNugetPackage('Microsoft.CodeAnalysis.LanguageServer', roslynVersion);
+gulp.task('fetchAllRoslynPlatformPackages', async () => {
```
dibarbet, line 76: Can you update the contributing docs to mention that you should run this task to update roslyn instead of the 'installDependencies' task?

dibarbet (approved):

tasks/offlinePackagingTasks.ts
jasonmalinowski: Ah yes, right.

jasonmalinowski: Not only did I update the documentation, I also updated this to run installDependencies after. And renamed it to "updateRoslynVersion" which is just a better name, honestly.

dibarbet (approved):

PR #6266, opened 2023-09-01
Add the 1.x settings name to migrated settings descriptions.

JoeRobich:
This would allow for users accustomed to the old setting name to find the new setting in the Settings page.


JoeRobich:
I could add them for the inlay settings, but they are so similar I assume they wouldn't be a problem.

dibarbet (approved):

PR #6269, opened 2023-09-01
Fix Razor browser discovery issues on Mac and Linux

allisonchou:
Fixes
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1842267/ and
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1865790/

Marked as dupes but they were actually two separate issues:
- Mac: `@vscode/js-debug-browsers`, which we use to determine which browsers the user has installed, expected an `execa` object to be passed into one of their methods we were calling. However, they only expected this on Macs (which is probably why this wasn't caught sooner).
- Linux: We needed to upgrade `@vscode/js-debug-browsers` to a newer version, since the old version contained buggy code that resulted in one of the filed bugs above.

Other notes:
- I also tried exploring adding Firefox support, but unfortunately it appears VS Code [only supports](https://code.visualstudio.com/docs/nodejs/browser-debugging) Edge/Chrome as built-in debug browsers.


src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -180,28 +181,33 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
             );
-            return BlazorDebugConfigurationProvider.edgeBrowserType;
+
+            return BlazorDebugConfigurationProvider.chromeBrowserType;
         }
 
-        showErrorMessage(vscode, vscode.l10n.t('Run and Debug: A valid browser is not installed'));
+        showErrorMessage(
+            vscode,
+            vscode.l10n.t('Run and Debug: A valid browser is not installed. Please install Edge or Chrome.')
```
allisonchou, line 209: Updated error message to explicitly say which browsers we support.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -180,28 +181,33 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
 
     public static async determineBrowserType(): Promise<string | undefined> {
         // There was no browser specified by the user, so we will do some auto-detection to find a browser,
-        // favoring chrome if multiple valid options are installed.
-        const chromeBrowserFinder = new ChromeBrowserFinder(process.env, promises, null);
-        const chromeInstallations = await chromeBrowserFinder.findAll();
-        if (chromeInstallations.length > 0) {
+        // favoring Edge if multiple valid options are installed.
```
allisonchou, line 184: Thought we should re-order browser priority so that we favor Edge over Chrome (since we're Microsoft 😄). Let me know if anyone has other thoughts.

davidwengier (approved):

PR #6271, opened 2023-09-02
Add unit test integration tests + cleanup test running

dibarbet:
Adds integration tests for unit testing and cleans up a bunch of things related to running integration tests.

I recommend reviewing roughly a commit at a time
1.  b637e84d11f2e510ba0ddf5133bfb8674346a628 - cleans up and fixes some of the integration test assets.  For example the slnWithCsproj was actually a broken project.  I moved them all to the integrationTests folder and split up the test asset workspace between LSP and O#.   The only real change here is adding the unit test integration tests.
2. c7ae769646e2b4dd4e1ec997d3cae3ab4f2f44ce - fixes a bunch of different races I encountered when running multiple integration tests at once
    a.  I noticed that sometimes there were two language client instances being created, one by a restart and one by a request attempting to restart the old instance.  I fixed this by changing our code to only ever have a single language client instance that we restart via normal vscode APIs.
    b.  While I was doing a I was getting extremely annoyed by all the irrelevant code we had in our roslynLanguageServer.ts file.  I extracted out a bunch of it into different files.   There is still more that can be done here.
    c.  The debug asset generation would throw an error (and not log anything) if a server restart was occuring.  I modified this to catch the error and actually log it.
    d.  Integration tests were not restoring the project, so the project didn't actually load correctly.  I fixed this to restore the project.
    e.  Some integration tests clean the test workspace after they're done (because they modifiy configs or other things in the project).  However this meant that followup integration tests did not actually have a fully loaded project.  I had to restart the server and wait for the project initialization to complete after restoring to fix this.


.vscode/launch.json
```diff
@@ -15,6 +15,34 @@
             ],
             "preLaunchTask": "buildDev"
         },
+        {
+            "name": "Launch Current File slnWithCsproj Integration Tests",
```
dibarbet, line 19: This was just moved up to the top.

omnisharptest/omnisharpIntegrationTests/testAssets/omnisharpTestAssetWorkspace.ts
```diff
@@ -0,0 +1,30 @@
...
+    singleCsproj,
+    slnWithCsproj,
+    slnFilterWithCsproj,
+    BasicRazorApp2_1,
+    slnWithGenerator,
+};
+
+const workspaceName = vscode.workspace.workspaceFolders![0].uri.fsPath.split(path.sep).pop();
+
+const activeTestAssetWorkspace = new OmnisharpTestAssetWorkspace(testAssetWorkspaces[workspaceName!]);
```
dibarbet, line 28: split into a separate type of O# (for custom O# details, like event names).

omnisharptest/omnisharpIntegrationTests/testAssets/omnisharpTestAssets.ts
```diff
@@ -3,40 +3,17 @@
...
-        const loc = path.join(dir, fileName);
-        await fs.writeTextFile(loc, contents);
-        return vscode.Uri.file(loc);
-    }
-}
+import { ITestAssetWorkspace, TestAssetWorkspace } from '../../../test/integrationTests/testAssets/testAssets';
+import { ActivationResult } from '../integrationHelpers';
 
-export class TestAssetWorkspace {
+export class OmnisharpTestAssetWorkspace extends TestAssetWorkspace {
```
dibarbet, line 14: moved the general functionality to the base and kept this for O# specifics.

src/lsptoolshost/unitTesting.ts
```diff
@@ -6,47 +6,52 @@
...
 
 export function registerUnitTestingCommands(
     context: vscode.ExtensionContext,
     languageServer: RoslynLanguageServer,
     dotnetTestChannel: vscode.OutputChannel
 ) {
     context.subscriptions.push(
-        vscode.commands.registerCommand('dotnet.test.run', async (request) =>
-            runTests(request, languageServer, dotnetTestChannel)
+        vscode.commands.registerCommand(
```
dibarbet, line 17: general fixes in here to allow the tests to see the result of the run tests command.

test/integrationTests/index.ts
```diff
@@ -15,18 +15,27 @@ async function runIntegrationTests() {
...
-            selectProjects: [jestIntegrationTestProjectName],
-            // Since we're running tests in the actual vscode process we have to run them serially.
-            runInBand: true,
-            // Timeout cannot be overriden in the jest config file, so override here.
-            testTimeout: 120000,
-            verbose: true,
-        } as Config.Argv,
-        [jestIntegrationTestProjectName]
-    );
+    const jestConfig = {
```
dibarbet, line 18: changes to add support for running a single file

test/integrationTests/lspInlayHints.integration.test.ts
```diff
@@ -3,20 +3,16 @@
...
 import { InlayHint, InlayHintKind, Position } from 'vscode-languageserver-protocol';
 
-const describeif = (condition: boolean) => (condition ? jestLib.describe : jestLib.describe.skip);
-const skip = !isRazorWorkspace(vscode.workspace) && !isSlnWithGenerator(vscode.workspace);
-
-describeif(skip)(`LSP Inlay Hints ${testAssetWorkspace.description}`, function () {
-    let fileUri: vscode.Uri;
+jestLib.describe(`LSP Inlay Hints ${testAssetWorkspace.description}`, function () {
     jestLib.beforeEach(async function () {
+        await testAssetWorkspace.restoreLspToolsHostAsync();
```
dibarbet, line 15: we weren't restoring before so the project wasn't technically loading correctly.  this fixes that.  At some point we need some tests to verify that projects are all actually loading (with all their references)

test/integrationTests/testAssets/slnWithCsproj/b_SecondInOrder_SlnFile.sln
```diff
@@ -10,8 +10,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "lib", "src\lib\lib.csproj",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "test", "test\test.csproj", "{4679428B-0CA0-4228-B8C0-B676B34A1B30}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "generator", "generator\generator.csproj", "{29F9B261-D48D-4D9B-9005-FB6D3B4F9CC4}"
```
dibarbet, line 13: this project did not actually exist and caused errors.

test/integrationTests/testAssets/testAssets.ts
```diff
@@ -0,0 +1,91 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fs from 'async-file';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import spawnGit from '../../../test/integrationTests/testAssets/spawnGit';
+import { execChildProcess } from '../../../src/common';
+
+export class TestAssetProject {
```
dibarbet, line 12: extracted all this out of the original O# version.

test/integrationTests/lspInlayHints.integration.test.ts
```diff
@@ -50,19 +41,24 @@ describeif(skip)(`LSP Inlay Hints ${testAssetWorkspace.description}`, function (
...
+        }
         const hints: vscode.InlayHint[] = await vscode.commands.executeCommand(
             'vscode.executeInlayHintProvider',
-            fileUri,
+            activeDocument,
             range
         );
 
-        jestLib.expect(hints).toHaveLength(5);
+        jestLib.expect(hints).toHaveLength(6);
```
dibarbet, line 54: now that the project loads we have an extra inlay hint.

davidwengier (approved):
I only looked at the Razor bits, but the rest sounds exciting :)

davidwengier (approved):
I only looked at the Razor bits, but the rest sounds exciting :)

test/integrationTests/integrationHelpers.ts
```diff
@@ -15,19 +17,41 @@ export async function activateCSharpExtension(): Promise<void> {
...
     if (!csharpExtension) {
         throw new Error('Failed to find installation of ms-dotnettools.csharp');
     }
 
+    // Run a restore manually to make sure the project is up to date since we don't have automatic restore.
+    await testAssetWorkspace.restoreLspToolsHostAsync();
+
+    // If the extension is already active, we need to restart it to ensure we start with a clean server state.
+    // For example, a previous test may have changed configs, deleted restored packages or made other changes that would put it in an invalid state.
+    let shouldRestart = false;
```
dibarbet, line 33: @jasonmalinowski could use your thoughts here on how to better handle this.  We have a couple issues here
1.  The server might have activated before we restored the project.  This might be fixed if we auto-restore project when we load it.
2.  A previous test run might have cleaned the workspace (including restored assets), leaving the server missing references.  We do run restore here again, but we currently have no way of waiting for the server to have actually updated (other than explicitly restarting it).

We could fire events every time the server reloads a project?  But it is hard to tell when that is 'done'

jasonmalinowski: Should we do something similar to our asynchronous listeners like we do in Visual Studio for Windows? We could introduce a test-only LSP endpoint that is fired off and completed only once all the work is done. We could either make it a one-off for the project system, or actually wire it into our async listener API so we could use it generally (and pass along the feature name).

dibarbet: Yup I like this idea - basically send the `waitForAllOperations` request to the server and have it return once its done.  I'll take a look at doing this in a followup (this ones already huge)

omnisharptest/omnisharpIntegrationTests/testAssets/omnisharpTestAssets.ts
```diff
@@ -3,40 +3,17 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
```
beccamc, line 3: Shouldn't this file be named `OmnisharpTestAssetWorkspace` and the file above that currently has this name be maybe `ActiveTestAssetWorkspace`

src/main.ts
```diff
@@ -312,18 +323,24 @@ export async function activate(
     if (!useOmnisharpServer) {
         tryGetCSharpDevKitExtensionExports(csharpLogObserver);
 
-        const languageServerExport = new RoslynLanguageServerExport(roslynLanguageServerPromise!);
+        // If we got here, the server should definitely have been created.
+        util.isNotNull(roslynLanguageServerStartedPromise);
```
beccamc, line 327: At this point these are unfulfilled promises, right? When would they be null?

beccamc (approved):

src/main.ts
dibarbet: They can be null if useOmnisharpServer is true - but it should never be null if it was false, hence the assertion that its not null inside this if

dibarbet: Note - this file could probably be refactored so this check didn't need to happen, but that would be a fairly large change and not worth doing in this PR.

PR #6282, opened 2023-09-05
Use NamedPipe for server communication

beccamc:
Current communication between client and server is writing to STDOUT. This makes the connection vulnerable to any code in the process writing to stdout stream and corrupting the data.

This PR uses a named pipe...
- Create a net server that listens to the pipe name
- Start a child process that kicks off the rolsyn language server. The language server is passed the pipe name so it can connect to the same pipe.
- When the connection is complete, pass the resulting MessageReader/MessageWriter to the VS Code language client. 
 
NOTE: Need [this PR](https://github.com/dotnet/roslyn/pull/69816) to go in and then update the server version in this PR.


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -615,7 +631,16 @@ export class RoslynLanguageServer {
             childProcess = cp.spawn(serverPath, args, cpOptions);
         }
 
-        return childProcess;
+        //Log any stderr/stdout messages in our output channel.
+        childProcess.stderr.on('data', (data: { toString: (arg0: any) => any }) =>
```
dibarbet, line 635: At this point, it should basically just be the console logger (before LSP is setup) that we log right?  Did you see anything else in there?

Potentially as well we should prefix these logs with something like [stdout] or [stderr] to indicate where the log is coming from?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -517,7 +528,7 @@ export class RoslynLanguageServer {
         return capabilities;
     }
 
-    private async startServer(logLevel: string | undefined): Promise<cp.ChildProcess> {
+    public async startServer(logLevel: string | undefined): Promise<MessageTransports> {
```
dibarbet, line 531: does this need to be public?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1092,3 +1117,31 @@ function getSessionId(): string {
 
     return sessionId;
 }
+
+async function createClientPipeTransport(
+    pipeName: string,
+    encoding: RAL.MessageBufferEncoding = 'utf-8'
```
dibarbet, line 1123: should make this required

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1092,3 +1117,31 @@ function getSessionId(): string {
...
+): Promise<PipeTransport> {
+    const PIPE_PATH = '\\\\.\\pipe\\';
```
jasonmalinowski, line 1125: What about Mac/Linux?

src/lsptoolshost/roslynLanguageServer.ts
beccamc: The only thing I see is the console logger and the output from Unity.  I figured there wasn't a downside to including it, expect maybe its confusing where all of this is coming from? I think adding a prefix is a good compromise. Are we good with that?

dibarbet: Yeah I agree we should leave it in (with a prefix).

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1092,3 +1118,43 @@ function getSessionId(): string {
...
+
+function getClientPipeName(pipeName: string): string {
+    const isWindows = os.platform() === 'win32';
+
+    if (isWindows) {
+        // Windows pipes are named using the following format
+        return '\\\\.\\pipe\\' + pipeName;
+    } else {
+        // Unix-type pipes are are actually writing to a file
+        return path.join(os.tmpdir(), pipeName + '.sock');
```
dibarbet, line 1154: as in the other PR, the pipename should include the tmpdir so we don't have to re-calculate it on the server.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1092,3 +1118,43 @@ function getSessionId(): string {
...
+        },
+    };
+}
+
+function getClientPipeName(pipeName: string): string {
+    const isWindows = os.platform() === 'win32';
+
+    if (isWindows) {
+        // Windows pipes are named using the following format
+        return '\\\\.\\pipe\\' + pipeName;
```
dibarbet, line 1151: Probably good to make a comment here that the pipeName doesn't include the prefix because node and dotnet have slightly different prefixes when talking about the name.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -615,7 +632,16 @@ export class RoslynLanguageServer {
...
         }
 
-        return childProcess;
+        //Log any stderr/stdout messages in our output channel.
+        childProcess.stderr.on('data', (data: { toString: (arg0: any) => any }) =>
+            _channel.append('[stderr]' + (isString(data) ? data : data.toString(this._encoding)))
+        );
+        childProcess.stdout.on('data', (data: { toString: (arg0: any) => any }) =>
+            _channel.append('[stdout]' + (isString(data) ? data : data.toString(this._encoding)))
+        );
```
jasonmalinowski, line 641:
```suggestion
        childProcess.stderr.on('data', (data: { toString: (arg0: any) => any }) =>
            _channel.append('[stderr] ' + (isString(data) ? data : data.toString(this._encoding)))
        );
        childProcess.stdout.on('data', (data: { toString: (arg0: any) => any }) =>
            _channel.append('[stdout] ' + (isString(data) ? data : data.toString(this._encoding)))
        );
```

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -615,7 +632,16 @@ export class RoslynLanguageServer {
             childProcess = cp.spawn(serverPath, args, cpOptions);
         }
 
-        return childProcess;
+        //Log any stderr/stdout messages in our output channel.
+        childProcess.stderr.on('data', (data: { toString: (arg0: any) => any }) =>
+            _channel.append('[stderr]' + (isString(data) ? data : data.toString(this._encoding)))
```
jasonmalinowski, line 637: What would 'data' be that _isn't_ a string here?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -615,7 +632,16 @@ export class RoslynLanguageServer {
...
+        );
+        childProcess.stdout.on('data', (data: { toString: (arg0: any) => any }) =>
+            _channel.append('[stdout]' + (isString(data) ? data : data.toString(this._encoding)))
+        );
+
+        const protocol = await transport.onConnected();
```
jasonmalinowski, line 643: Should we log something to a trace to say we're waiting for the connection to come in? If anything goes wrong with the named pipe we might otherwise hang here with the process running but no connection.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1092,3 +1118,43 @@ function getSessionId(): string {
 
     return sessionId;
 }
+
+async function createClientPipeTransport(
+    pipeName: string,
+    encoding: RAL.MessageBufferEncoding = 'utf-8'
+): Promise<PipeTransport> {
+    const connected = new Promise<[MessageReader, MessageWriter]>((resolve) => {
+        const server: net.Server = net.createServer((socket: net.Socket) => {
```
jasonmalinowski, line 1127: What is the security on this pipe that's being created? Is it restricted to the user on the machine or is it more open?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1092,3 +1118,43 @@ function getSessionId(): string {
...
+            // Sever has connected
+            server.close();
```
jasonmalinowski, line 1129: If the server never connects (say because something else is broken and we're crashing on startup) are we going to be leaking pipes?

src/lsptoolshost/roslynLanguageServer.ts
beccamc: Updated to create all names on the client, and included this comment... 

```
    // The connection string format for Nodejs and .NET are different.
    const WINDOWS_NODJS_PREFIX = '\\\\.\\pipe\\';
    const WINDOWS_DOTNET_PREFIX = '\\\\.\\';
```

src/lsptoolshost/roslynLanguageServer.ts
beccamc: It's the encoded data buffer!

src/lsptoolshost/roslynLanguageServer.ts
beccamc: Thank you! This has been added.

beccamc:
Closing in favor of https://github.com/dotnet/vscode-csharp/pull/6351

PR #6287, opened 2023-09-05
Validate package.nls.json for \u200b

WardenGnaw:
This PR adds in a check to see if \u200b is included in the loc strings that are used to break hyperlink syntax in descriptions.

Added a new gulp task that will quickly resolve these issues.

Also moved the debugger gulp tasks to tasks\debuggerTasks.ts.


package.nls.cs.json
```diff
@@ -136,9 +136,9 @@
...
-  "generateOptionsSchema.symbolOptions.searchNuGetOrgSymbolServer.description": "Pokud je hodnota true, přidá se do cesty pro hledání symbolů server symbolů pro NuGet.org (https​://symbols.nuget.org​/download/symbols). Pokud tato možnost není zadaná, výchozí hodnota je false.",
-  "generateOptionsSchema.symbolOptions.searchPaths.description": "Pole adres URL serveru symbolů (například: http​://MyExampleSymbolServer) nebo adresářů (například: /build/symbols) k vyhledávání souborů .pdb. Tyto adresáře budou prohledány kromě výchozích umístění – vedle modulu a cesty, kam byl soubor pdb původně přemístěn.",
+  "generateOptionsSchema.symbolOptions.searchMicrosoftSymbolServer.description": "Pokud je hodnota true, přidá se do cesty pro hledání symbolů server symbolů pro produkty Microsoft (https\u200b://msdl.microsoft.com\u200b/download/symbols). Pokud tato možnost není zadaná, výchozí hodnota je false.",
```
jasonmalinowski, line 139: Are these changes just going to get overwritten the next time a localization pass comes along?

dibarbet: @Cosifne is there a way to indicate in the translation that a string should remain unchanged?  
That might not even be enough though, any tool in the localization pipeline could potentially get rid of zero-width spaces.  If these weren't package.json I'd suggest using the string.format version, but not sure how to fix this for package.json strings.

gregg-miskelly: @dibarbet in case you missed this, Andrew already has instructions telling the LOC team not to mess with them. But they are weird, so not surprisingly, sometimes the loc team doesn't follow the instructions.

WardenGnaw: @gregg-miskelly It looks like the loc team kept the \u200b character but something in their pipeline might have escaped it. I think something in their pipeline needs to run `packageNlsJson.replace(/\u200b/gm, '\\u200b');`

gregg-miskelly: What I found earlier that sometimes the `\u200b` was converted to the Unicode character, which maybe isn't ideal because it isn't human readable, but it is okay. But also, sometimes they dropped it all together. If it is hard for the loc team to leave it escaped, I think it is fine to accept the former. But obviously we need something to prevent the later.

dibarbet (approved):

package.nls.cs.json
Cosifne: I think there should a way to make a part of string unchanged.
See https://github.com/microsoft/vscode/issues/125774#issuecomment-1339954388
Can we try to lock the whole link?

Cosifne (approved):

PR #6304, opened 2023-09-06
Reflecting changes in vscode to take ranges in `SerializableSemanticTokensParams` 

maryamariyan:
Reflecting changes in [vscode-csharp](https://github.com/dotnet/vscode-csharp) after making `ProvideSemanticTokensResponse` take ranges as input.

- Modify `SerializableSemanticTokensParams` to take `ranges` rather than `range`
- Modify the data property of the `SemanticTokensResponse` class.
- Changes data to take an array of arrays of number type

Related to dotnet/razor#9092

### Local testing:

- With feature flag disabled (default behavior for VS code), the ranges array contains one element as expected:
image

- With feature flag turned enabled, the ranges array contains sorted disjoined elements as expected:
image

### Observation

- Realized that in vs code we are just using textmate, so the testing involved just assuring the values in ranges are properly mapped over, but the are not really used, in VS code we are just returning dummy empty responses for semantic range requests and just relying on text mate.

- Once we complete issue https://github.com/dotnet/razor/issues/4323 we'd then be matching behaviors between VS and VS code.


src/razor/src/semantic/provideSemanticTokensResponse.ts
```diff
@@ -3,9 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { SemanticTokensResponse } from './semanticTokensResponse';
-
 export class ProvideSemanticTokensResponse {
     // tslint:disable-next-line: variable-name
-    constructor(public Result: SemanticTokensResponse, public HostDocumentSyncVersion: number | null) {}
+    constructor(public Tokens: Array<Array<number>>, public HostDocumentSyncVersion: number | null) {}
```
davidwengier, line 8: The [server version](https://github.com/dotnet/razor/blob/main/src/Razor/src/Microsoft.AspNetCore.Razor.LanguageServer/Semantic/Models/ProvideSemanticTokensResponse.cs) does not have `HostDocumentSyncVersion` as nullable, so could be causing an exception during deserialization.

src/razor/src/semantic/semanticTokensRangeHandler.ts
```diff
@@ -18,8 +17,8 @@ export class SemanticTokensRangeHandler {
         any
     > = new RequestType(SemanticTokensRangeHandler.getSemanticTokensRangeEndpoint);
     private emptySemanticTokensResponse: ProvideSemanticTokensResponse = new ProvideSemanticTokensResponse(
-        new SemanticTokensResponse(new Array<number>(), ''),
-        null
+        new Array<Array<number>>(),
+        1
```
maryamariyan, line 21: Or maybe -1 because this is really not meant to be used.

```suggestion
        -1
```

allisonchou (approved):

src/razor/src/semantic/semanticTokensRangeHandler.ts
```diff
@@ -18,8 +17,8 @@ export class SemanticTokensRangeHandler {
         any
     > = new RequestType(SemanticTokensRangeHandler.getSemanticTokensRangeEndpoint);
     private emptySemanticTokensResponse: ProvideSemanticTokensResponse = new ProvideSemanticTokensResponse(
-        new SemanticTokensResponse(new Array<number>(), ''),
-        null
+        new Array<Array<number>>(),
+        -5
```
maryamariyan, line 21: @davidwengier adding a number like -5 combined with the following commit in dotnet/razor would fix the new bug introduced for VS code that is having all razor components show up as red

https://github.com/dotnet/razor/commit/889f03b9bcf7c05cfc7e9fead16387a6662eff8f

davidwengier: We have access to the right version number, and that would be preferable to a magic number like -5. I'm on my phone so can't get a link, but there is a document synchronizer here too.

The other option is to add a language server feature option on the server side, and stop us bothering to request semantic tokens on vscode at all, and then just delete this whole folder.

src/razor/src/semantic/semanticTokensRangeHandler.ts
```diff
@@ -40,10 +49,33 @@ export class SemanticTokensRangeHandler {
...
+            if (razorDocument === undefined) {
+                return this.emptySemanticTokensResponse;
+            }
+
+            const textDocument = await vscode.workspace.openTextDocument(razorDocumentUri);
+            version = textDocument.version;
+            const synchronized = await this.documentSynchronizer.trySynchronizeProjectedDocument(
+                textDocument,
+                razorDocument.csharpDocument,
+                version,
```
maryamariyan, line 65:
```suggestion
                version, // would have been more correct to have _semanticTokensParams.hostDocumentVersion,
```

For example `formattingParams` has `hostDocumentVersion` property but semantic tokens params does not, so I couldn't find a better version that's accessible here to send down to the synchronizer.

/cc @allisonchou @davidwengier @phil-allen-msft

davidwengier: The version is sent by the language server, just looks like it's not included in the params object in typescript. Compare it to the params object defined in C#, the hostDocumentVersion property is missing.

src/razor/src/semantic/semanticTokensRangeHandler.ts
maryamariyan: Addressed feedback in https://github.com/dotnet/vscode-csharp/pull/6304/commits/f1d84b203a87fdb7f49c0927e5e98e578ab6bc31

maryamariyan:
Noticing the `Test OmniSharpt Test windows` s failing on the CI but locally tests completed successfully for me:

Local output after running `npm run test`:
```
Test Suites: 1 passed, 1 total
Tests:       309 passed, 309 total
Snapshots:   309 passed, 309 total
Time:        14.113 s
```

I will rerun the CI to fix the failure in case it is intermittent.

package.json
```diff
@@ -39,7 +39,7 @@
   "defaults": {
     "roslyn": "4.8.0-3.23455.11",
     "omniSharp": "1.39.7",
-    "razor": "7.0.0-preview.23455.5",
+    "razor": "7.0.0-preview.23456.2",
```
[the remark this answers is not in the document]
maryamariyan: cc @davidwengier 

Would a tool be able to auto-populate the changelog.md diff here?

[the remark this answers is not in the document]
allisonchou: I don't think there's an auto update tool for the changelog unfortunately

src/razor/src/semantic/semanticTokensRangeHandler.ts
maryamariyan: Applied a simpler fix by just including the missing `hostDocumentVersion` property to typescript params

davidwengier (approved):

package.json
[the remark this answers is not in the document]
davidwengier: At least now you're fixing an actual bug, there is something to mention other than just "we changed some types internally" 😁

maryamariyan:
tested further noticed in VS Code the following text would still show up in red

image

This should be fine because before the regression the behavior was the same. 
Seems like VS code does not yet support for providing color for these strings

PR #6310, opened 2023-09-07
Code Actions - Add fix all support

akhera99:
- Creates a new `CodeAction` type, `RoslynFixAllCodeAction` that specifies the scope being used for fix all
- Creates a new request to retrieve the new `CodeAction` type that contains a `WorkspaceEdit` with fix all edits
- Resolves at the middleware layer
- Uses quick picks to determine the selected scope and passes that along
- Converts the LSP `WorkspaceEdit` to a VS Code `WorkspaceEdit` and tries to apply it.

https://github.com/dotnet/vscode-csharp/assets/40616383/25c8384d-de9f-48f2-bda4-aee20c4887dc


ShreyasJejurkar:
Thank you for this, great work! 🎉

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
                 workspace: {
                     configuration: (params) => readConfigurations(params),
                 },
+                resolveCodeAction: async (codeAction, token, next) => {
+                    const lspCodeAction = <CodeAction>codeAction;
+                    const data = lspCodeAction.data;
+                    if (data.FixAllFlavors) {
```
JoeRobich, line 211: I know it comes from Roslyn, but it is amusing that it is called FixAll"Flavors" and not "Scopes" or "Kinds".

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
                 workspace: {
                     configuration: (params) => readConfigurations(params),
                 },
+                resolveCodeAction: async (codeAction, token, next) => {
```
dibarbet, line 208: move this implementation to a separate file (can just export a function and call it from here

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
...
+                    const lspCodeAction = <CodeAction>codeAction;
+                    const data = lspCodeAction.data;
+                    if (data.FixAllFlavors) {
+                        const result = await vscode.window.showQuickPick(data.FixAllFlavors, {
+                            placeHolder: 'Pick a fix all scope',
```
dibarbet, line 213: this should be localized

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
...
+
+                        if (result) {
+                            const fixAllCodeAction: RoslynProtocol.RoslynFixAllCodeAction = {
+                                title: lspCodeAction.title,
+                                edit: lspCodeAction.edit,
+                                data: data,
+                                scope: result,
+                            };
+
+                            const response = await _languageServer.sendRequest(
```
dibarbet, line 224: it could likely take some time to compute this, especially for a solution level fix all in a large solution.  We definitely should wrap this entire operation in a cancellable progress.  See this https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/unitTesting.ts#L84 for an example on how to do it
Some notes
1.  We don't have an explicit progress % for this, so fine to report progress without an increment.
2.  We should allow cancellation up until we call `vscode.workspace.applyEdit` - that part should not be cancellable.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
...
+                    const data = lspCodeAction.data;
+                    if (data.FixAllFlavors) {
+                        const result = await vscode.window.showQuickPick(data.FixAllFlavors, {
+                            placeHolder: 'Pick a fix all scope',
+                        });
+
+                        if (result) {
+                            const fixAllCodeAction: RoslynProtocol.RoslynFixAllCodeAction = {
+                                title: lspCodeAction.title,
+                                edit: lspCodeAction.edit,
```
dibarbet, line 219: why do we need the edit here?

src/lsptoolshost/roslynLanguageServer.ts
dibarbet, line 208: So I like the approach you took here, but it is slightly different from what we talked about before.  I like this better anyway though, but we can simplify a little bit more.

Since you're directly applying the workspace edit (via `vscode.workspace.applyEdit`) and *not* returning a resolved code action to vscode, it would be better to have this implementation just be a command.

Essentially:
1.  On the server side, instead of adding the flavor to the 'data' object, you would return a fix all code action that contains a command (e.g. `roslyn.fixAll`) with arguments containing whatever flavor information you need to call back to the server as you do now.  An example of where we do this is here - https://github.com/dotnet/roslyn/blob/main/src/Features/LanguageServer/Protocol/Handler/CodeLens/CodeLensHandler.cs#L149
2.  Instead of implementing this as part of the `resolveCodeAction` middleware - you would move this implementation to a command handler.  The implementation here basically wouldn't change (minus my other feedback).  For example, where the client implements the command I linked above - https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/unitTesting.ts#L21

Then what happens is that when you select the fix all action in the list - the vscode LSP client will invoke our command handler (based on the command id we returned in the code action)

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
...
+                                const protocolConverter = createConverter(uriConverter, true, true);
+                                const result = await protocolConverter.asWorkspaceEdit(response.edit);
+
+                                if (!(await vscode.workspace.applyEdit(result))) {
+                                    throw new Error(
+                                        'Tried to insert multiple code action edits, but an error occurred.'
+                                    );
+                                }
+                            }
+                            console.log(response);
```
dibarbet, line 242: should remove

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +205,46 @@ export class RoslynLanguageServer {
...
+                            );
+
+                            if (response.edit) {
+                                const uriConverter: URIConverter = (value: string): vscode.Uri =>
+                                    UriConverter.deserialize(value);
+                                const protocolConverter = createConverter(uriConverter, true, true);
+                                const result = await protocolConverter.asWorkspaceEdit(response.edit);
+
+                                if (!(await vscode.workspace.applyEdit(result))) {
+                                    throw new Error(
```
dibarbet, line 237: after you've finished addressing the other comments (e.g. moving to a command), can you make sure that if this fails it shows an error toast?  if it doesn't we might need to explicitly show one.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -202,6 +203,46 @@ export class RoslynLanguageServer {
                 workspace: {
                     configuration: (params) => readConfigurations(params),
                 },
+                // resolveCodeAction: async (codeAction, token, next) => {
```
dibarbet, line 206: should remove

src/lsptoolshost/fixAllCodeAction.ts
```diff
@@ -0,0 +1,67 @@
...
+                        const uriConverter: URIConverter = (value: string): vscode.Uri =>
+                            UriConverter.deserialize(value);
+                        const protocolConverter = createConverter(uriConverter, true, true);
+                        fixAllEdit = await protocolConverter.asWorkspaceEdit(response.edit);
+                    }
+                }
+            }
+        );
+
+        if (!(await vscode.workspace.applyEdit(fixAllEdit))) {
```
dibarbet, line 63: I believe the apply edit should go inside the progress as well.  I think it won't be cancellable because you don't pass in the cancellation token

src/lsptoolshost/fixAllCodeAction.ts
```diff
@@ -0,0 +1,67 @@
...
+            throw new Error('Tried to insert multiple code action edits, but an error occurred.');
```
dibarbet, line 64: test this to make sure the error is visible (hopefully in the C# output window).  If not you may need to manually write it there via passing in the output channel

dibarbet (approved):

PR #6315, opened 2023-09-08
Fix completionComplexEdit: Use fsPath instead of path

### LINKED ISSUE #6011: Override completion broken
No errors in vscode output but autocomplete broken again.

C# exntetion v2.0.320 for vscode 1.80.2. Using dotnet 7.0.306.

Doesnt matter if you press Tab, Click it in the list, or press enter or space, notthing happens:
Code_abCNiakOS2

KCFindstr:
Trying to fix #6011 .

Found that `completionComplexEdit` should find editors according to `fsPath` instead of `path`, similar to other places. `path` might be different on Windows, e.g.

uri.path:
b4f7f0d33c06de5c9a3fa27972f3fc74

editor.document.uri.path:
b8f87cc9ab6d8640baff90288fef5117

After this fix, the autocomplete works properly with mouse clicks / tab autocomplete. However, it still does not respond to enter key (it simply adds a new line instead of inserting the autocompletion).

### DIFF
--- a/src/lsptoolshost/commands.ts
+++ b/src/lsptoolshost/commands.ts
@@ -103,7 +103,7 @@ async function completionComplexEdit(
 ): Promise<void> {
     let success = false;
     const uri = UriConverter.deserialize(uriStr);
-    const editor = vscode.window.visibleTextEditors.find((editor) => editor.document.uri.path === uri.path);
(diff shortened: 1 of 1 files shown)


genlu:
@KCFindstr Thanks a lot for the contribution! From a local repro I was finally able to get, the capitalization of windows drive letter is indeed the cause (although I'm not sure why the uri.path of the document we get from editor would vary)

image

genlu (approved):

PR #6325, opened 2023-09-09
Make complexEdit command more robust

genlu:
... by checking both fsPath and path. Also added logging when there's error.

This is a follow-up on https://github.com/dotnet/vscode-csharp/pull/6315, just in case of unexpected scenarios.

related to:
https://github.com/dotnet/vscode-csharp/issues/6293
https://github.com/dotnet/vscode-csharp/issues/6246

### DIFF
--- a/src/lsptoolshost/commands.ts
+++ b/src/lsptoolshost/commands.ts
@@ -31,7 +31,11 @@ export function registerCommands(
     // so we don't accidentally pass them directly into vscode APIs.
     context.subscriptions.push(vscode.commands.registerCommand('roslyn.client.peekReferences', peekReferencesCallback));
     context.subscriptions.push(
-        vscode.commands.registerCommand('roslyn.client.completionComplexEdit', completionComplexEdit)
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/commands.ts
```diff
@@ -142,7 +149,29 @@ async function completionComplexEdit(
     }
 
     if (!success) {
-        throw new Error('Failed to make a complex text edit for completion.');
+        const componentName = '[roslyn.client.completionComplexEdit]';
+        const errorMessage = 'Failed to make a complex text edit for completion.';
+        outputChannel.show();
```
dibarbet, line 154: One thing to consider here - should we put up a toast as well?  We have utils for showing an error message.  Should only do that though if we want to bring attention to it and don't think it will be popping up all the time.

genlu: good point, we probably can hook up the toast, given this is a failure from user explicitly invoked action. Will add in a follow up PR

PR #6336, opened 2023-09-11
Add integration test for .runsettings in unit tests

dibarbet:
Followup for https://github.com/dotnet/vscode-csharp/pull/6265#discussion_r1321885343

### DIFF
--- a/test/integrationTests/unitTests.integration.test.ts
+++ b/test/integrationTests/unitTests.integration.test.ts
@@ -16,6 +16,9 @@ jestLib.describe(`Unit Testing ${testAssetWorkspace.description}`, function () {
     });
 
     jestLib.beforeEach(async function () {
+        vscode.workspace
+            .getConfiguration()
+            .update('dotnet.unitTests.runSettingsPath', undefined, vscode.ConfigurationTarget.Workspace);
         const fileName = path.join('test', 'UnitTest1.cs');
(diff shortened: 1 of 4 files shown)


test/integrationTests/unitTests.integration.test.ts
```diff
@@ -118,3 +150,15 @@ async function getCodeLensesAsync(): Promise<vscode.CodeLens[]> {
         return a.command!.title.localeCompare(b.command!.command);
     });
 }
+
+async function setConfigurationAndWaitForObserver<T>(configuration: string, value: T): Promise<void> {
+    const changed = new Promise<void>((resolve, _) => {
+        vscode.workspace.onDidChangeConfiguration((e) => {
```
dibarbet, line 156: incredibly annoyingly, `optionProvider.GetLatestOptions()` does not actually read the latest options.  It instead uses whatever the observable last triggered with.  This is extremely racy.

It isn't simple to modify `GetLatestOptions()` because we don't want to be reading the entire option set every single time.  Instead I think I need to refactor options to instead return functions to get the latest value, instead of the entire set of options.

dibarbet: filed https://github.com/dotnet/vscode-csharp/issues/6337 for this

Cosifne (approved):

PR #6340, opened 2023-09-12
Pass ForceNoAlign to ensure msbuild output is parseable by the vscode problem matcher

dibarbet:
When the terminal size is adjusted, msbuild can alter the output making it unparseable by the vscode problem matcher.  This flag makes msbuild not attempt to align the output based on the terminal size.
See https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2022

This flag is also used by devkit for a similar reason, see https://devdiv.visualstudio.com/DevDiv/_git/vs-green?path=/tools/msbuild/Directory.Build.rsp&version=GBmain&line=16&lineEnd=17&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents (internal)

Resolves https://github.com/dotnet/vscode-csharp/issues/6309

Fixed:
image

Before:
image

### DIFF
--- a/omnisharptest/omnisharpFeatureTests/assets.test.ts
+++ b/omnisharptest/omnisharpFeatureTests/assets.test.ts
@@ -78,7 +78,7 @@ suite('Asset generation: csproj', () => {
         // We do not check the watch task since this parameter can break hot reload scenarios.
         tasksJson.tasks
             .filter((task) => task.label !== 'watch')
-            .forEach((task) => task.args!.should.contain('/consoleloggerparameters:NoSummary'));
+            .forEach((task) => task.args!.should.contain('/consoleloggerparameters:NoSummary;ForceNoAlign'));
     });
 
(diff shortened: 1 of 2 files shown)


dibarbet:
@dotnet-policy-service rerun

ShreyasJejurkar:
Any idea why it says cpp? 
image

dibarbet:
> Any idea why it says cpp? image

That is defined here - https://github.com/microsoft/vscode/blob/24739b3e01f8f7d27efa33e7ff9101fb8157032d/src/vs/workbench/contrib/tasks/common/problemMatcher.ts#L1928

I believe because this is used for either cpp or C#.  At some point we may need to change the problem matcher, but its not high priority for me at the moment

dibarbet:
@dotnet-policy-service rerun

PR #6351, opened 2023-09-12
Update language server to use named pipe client

beccamc:
Current communication between client and server is writing to STDOUT. This makes the connection vulnerable to any code in the process writing to stdout stream and corrupting the data.

This PR uses a named pipe...
- Start a child process that kicks off the rolsyn language server. The language server will start the named pipe.
- Listen for the name of the named pipe to arrive over stdin/out. 
- When the name arrives, connect to the pipe using net.createConnection
- When the connection is complete, pass the resulting MessageReader/MessageWriter to the VS Code language client.

NOTE: Need [server PR](https://github.com/dotnet/roslyn/pull/69918) to go in and then update the server version in this PR.

This has been tested on both windows and mac.


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -538,7 +555,40 @@ export class RoslynLanguageServer {
...
+        // and listen for the server to pass back the connection information via stdout.
+        const namedPipeConnectionPromise = new Promise<TransmittedInformation>((resolve) => {
+            childProcess.stdout.on('data', (data: { toString: (arg0: any) => any }) => {
+                const result: string = isString(data) ? data : data.toString(RoslynLanguageServer.encoding);
+                _channel.append('[stdout] ' + result);
+
+                // Use the regular expression to find all JSON lines
+                const jsonLines = result.match(RoslynLanguageServer.namedPipeKeyRegex);
+                if (jsonLines) {
+                    const transmittedPipeNameInfo: TransmittedInformation = JSON.parse(jsonLines[0]);
```
beccamc, line 568: I tried to do something reasonably simple here for a simple communication. Open to feedback or other ideas here.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -538,7 +558,40 @@ export class RoslynLanguageServer {
...
+                    _channel.appendLine('received named pipe information from server');
+                    resolve(transmittedPipeNameInfo);
+                }
+            });
+        });
+
+        // Wait for the server to send back the name of the pipe to connect to.
+        _channel.appendLine('waiting for named pipe information from server...');
+        const pipeConnectionInfo = await namedPipeConnectionPromise;
```
dibarbet, line 580: we should probably add a timeout here and show an error toast if it times out.

src/lsptoolshost/roslynLanguageServer.ts
beccamc: Resolved in Roslyn PR

src/lsptoolshost/roslynLanguageServer.ts
beccamc: Yes that's a good idea. Should probably do it for both receiving the name and making the named pipe connection

beccamc: Updated to throw an error, which will log. The "couldn't create connection" toast will show. 
image

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -538,7 +554,57 @@ export class RoslynLanguageServer {
             childProcess = cp.spawn(serverPath, args, cpOptions);
         }
 
-        return childProcess;
+        // Timeout promise used to time out the connection process if it takes too long.
+        const timeout = new Promise<undefined>((resolve) => {
+            RAL().timer.setTimeout(resolve, 10000);
```
dibarbet, line 559: If we're using the timeout for both the wait and connect, we might want to give a bit bigger of a buffer (this is 10seconds currently right?)

Maybe we start with 30s?

dibarbet (approved):

PR #6369, opened 2023-09-15
Cleanup and fix issues with option provider

dibarbet:
I wanted to add more integration tests that change options, but we had an issue - the `GetLatestOptions` method on the option provider did not actually get the latest options.  Instead it returned whatever the last change event said it was (which in integration tests was extremely racy).

Fixing this lead me down a rabbit whole of options refactoring, but I believe I ended up somewhere clean.
1.  Removed the option class and optionprovider type.
2.  Added option interface implementations backed by getter methods that query the vscode configuration API (see `options.ts`).  These getter methods will *always* return the latest options
3.  Moved a bunch of unit tests to jest (they no longer worked in mocha because there were references to `vscode`)


omnisharptest/omnisharpJestTests/features/reportIssue.test.ts
```diff
@@ -0,0 +1,117 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import reportIssue from '../../../src/shared/reportIssue';
+import { FakeMonoResolver, fakeMonoInfo } from '../../omnisharpUnitTests/fakes/fakeMonoResolver';
+import { FakeDotnetResolver } from '../../omnisharpUnitTests/fakes/fakeDotnetResolver';
+import { DotnetInfo } from '../../../src/shared/utils/dotnetInfo';
+import { jest, describe, test, expect, beforeEach } from '@jest/globals';
+
+describe(`${reportIssue.name}`, () => {
```
dibarbet, line 13: this test was moved and converted to jest.  functionally the same.

omnisharptest/omnisharpJestTests/informationMessageObserver.test.ts
```diff
@@ -0,0 +1,111 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { jest, describe, test, expect, beforeEach } from '@jest/globals';
+import { InformationMessageObserver } from '../../src/observers/informationMessageObserver';
+import { getUnresolvedDependenices, getWorkspaceConfiguration } from '../../test/unitTests/fakes';
+import { Subject, from as observableFrom } from 'rxjs';
+import { timeout } from 'rxjs/operators';
+
+describe('InformationMessageObserver', () => {
```
dibarbet, line 13: this test was moved and converted to jest.  functionally the same.

omnisharptest/omnisharpJestTests/optionStream.test.ts
```diff
@@ -0,0 +1,52 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { jest, describe, test, expect, beforeEach } from '@jest/globals';
+import { Observable } from 'rxjs';
+import createOptionStream from '../../src/shared/observables/createOptionStream';
+import Disposable from '../../src/disposable';
+
+describe('OptionStream', () => {
```
dibarbet, line 12: this test was moved and converted to jest.  functionally the same.

omnisharptest/omnisharpJestTests/options.test.ts
```diff
@@ -0,0 +1,103 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { jest, describe, test, expect, beforeEach } from '@jest/globals';
+import { URI } from 'vscode-uri';
+import * as path from 'path';
+import { commonOptions, omnisharpOptions } from '../../src/shared/options';
+import { getWorkspaceConfiguration } from '../../test/unitTests/fakes';
+
+describe('Options tests', () => {
```
dibarbet, line 13: this test was moved and converted to jest.  functionally the same.

package.json
```diff
@@ -123,7 +123,6 @@
     "@types/del": "3.0.1",
     "@types/fs-extra": "5.0.4",
     "@types/gulp": "4.0.5",
-    "@types/jest": "^24.0.25",
```
dibarbet, line 126: not necessary - we have `@jest/globals` which servers the same purpose.

src/shared/options.ts
```diff
@@ -3,53 +3,103 @@
...
 import * as path from 'path';
 
-export class Options {
-    constructor(
-        public commonOptions: CommonOptions,
-        public omnisharpOptions: OmnisharpServerOptions,
-        public languageServerOptions: LanguageServerOptions,
-        public razorOptions: RazorOptions
-    ) {}
+export interface CommonOptions {
```
dibarbet, line 10: the diff probably won't make much sense side by side - I moved things around.  Essentially we now have 4 interface definitions: `CommonOptions`, `OmnisharpServerOptions`, `LanguageServerOptions`, and `RazorOptions`

Each of those has a class implementation that defines a getter for each option.  The getter reads the option value from vscode.

All of the class implementations are exported (`commonOptions`. `omnisharpOptions`, `languageServerOptions`, `razorOptions`, so to get an option, all a consumer needs to do is:
```
import { commonOptions } from '...options.ts'
const optionValue = commonOptions.dotnetPath
```

which will always return the most up to date value.

beccamc: I assume there are absolutely no perf implications from using a getter to read from the vscode options?

beccamc (approved):
Looks great!

src/shared/options.ts
dibarbet: if the value is needed extremely often in some loop, I'd advise getting it once outside of the loop.  But generally we don't use our options that way, we just access them on startup, or on a specific request which should not be impactful at all.

PR #6387, opened 2023-09-18
Fix a typo in package.nls.json

Cosifne:
We still get many loc PRs like https://github.com/dotnet/vscode-csharp/pull/6383
After changing `comments` to `comment` I see the `lock` block is added to the generated xlf files.
image
So it should address the issue on our side.


gregg-miskelly:
This file is generated. You want to change this code: https://github.com/dotnet/vscode-csharp/blob/53f3a44f7c08a785f3de355a615f257e6d983809/src/tools/generateOptionsSchema.ts#L195

Cosifne:
> This file is generated. You want to change this code:
> 
> https://github.com/dotnet/vscode-csharp/blob/53f3a44f7c08a785f3de355a615f257e6d983809/src/tools/generateOptionsSchema.ts#L195

Ah good catch, I forget this is now generated

gregg-miskelly (approved):

PR #6390, opened 2023-09-18
Add option to disable devkit

dibarbet:


package.json
```diff
@@ -1169,6 +1169,12 @@
             "default": null,
             "description": "Sets a path where MSBuild binary logs are written to when loading projects, to help diagnose loading errors."
           },
+          "dotnet.loadProjectsWithCSharpExtension": {
+            "scope": "machine-overridable",
```
dibarbet, line 1173: Unfortunately this doesn't remove it from user settings.  I saw no way to actually remove it from user settings :(.  Might be a limitation we have to live with.

jasonmalinowski: Hmm, does this work?

> resource - Resource settings, which apply to files and folders, and can be configured in all settings levels, even folder settings.

dibarbet: Resource allows user settings as well.  However it seemed slightly more appropriate
image

package.nls.json
```diff
@@ -5,6 +5,7 @@
   "configuration.dotnet.server.waitForDebugger": "Passes the --debug flag when launching the server to allow a debugger to be attached. (Previously `omnisharp.waitForDebugger`)",
   "configuration.dotnet.server.trace": "Sets the logging level for the language server",
   "configuration.dotnet.server.extensionPaths": "Override for path to language server --extension arguments",
+  "configuration.dotnet.preferCSharpExtension": "Forces projects to load with the C# extension only.  This can be useful when using legacy project types that are not supported by the C# Dev Kit. (Requires window reload)",
```
jasonmalinowski, line 8: Should there be a "the" before "C# Dev Kit" not?

jasonmalinowski (approved):

package.nls.json
mikadumont, line 8: I would take out `the` before C# Dev Kit. Looks good otherwise.

dibarbet, line 8:
```suggestion
  "configuration.dotnet.preferCSharpExtension": "Forces projects to load with the C# extension only.  This can be useful when using legacy project types that are not supported by C# Dev Kit. (Requires window reload)",
```

mikadumont: There shouldn't be a "the" before CDK.

PR #6399, opened 2023-09-19
Show error when process fails to start

dibarbet:


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -532,9 +532,28 @@ export class RoslynLanguageServer {
             childProcess = cp.spawn(serverPath, args, cpOptions);
         }
 
+        // Record the stdout and stderr streams from the server process.
```
dibarbet, line 535: if the process failed to start it would previously silently timeout.  we weren't capturing stderr or erroring on process exit.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -532,9 +532,28 @@ export class RoslynLanguageServer {
...
         // Timeout promise used to time out the connection process if it takes too long.
-        const timeout = new Promise<undefined>((resolve) => {
-            RAL().timer.setTimeout(resolve, 30000);
+        const timeout = new Promise<undefined>((resolve, reject) => {
+            RAL().timer.setTimeout(resolve, languageServerOptions.startTimeout);
+
+            // If the child process exited unexpectedly, reject the promise early.
+            // Error information will be captured from the stdout/stderr streams above.
+            childProcess.on('exit', (code) => {
+                if (code && code !== 0) {
```
dibarbet, line 552: we can't just throw from inside this callback, it won't cause the startServer method to throw since its inside a callback.  Instead we just reject the timeout process early if the process exits with a non-zero exit code.

beccamc (approved):

PR #6405, opened 2023-09-20
Add hiddenWhen to debuggers in package.json

WardenGnaw:
This PR adds the 'hiddenWhen' attribute to debuggers contributed by the C# Extension.

Related: https://github.com/microsoft/vscode/pull/193157

Fixes: 
- https://github.com/microsoft/vscode-dotnettools/issues/14 
- https://github.com/microsoft/vscode-dotnettools/issues/512 
- https://github.com/microsoft/vscode-dotnettools/issues/513

Note: This PR requires the VS Code September 2023 Update

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,7 +108,7 @@
         "webpack-cli": "4.6.0"
       },
       "engines": {
-        "vscode": "^1.73.0"
+        "vscode": "^1.83.0"
       }
     },
     "node_modules/@ampproject/remapping": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,7 +108,7 @@
         "webpack-cli": "4.6.0"
       },
       "engines": {
-        "vscode": "^1.83.0"
+        "vscode": "^1.73.0"
       }
     },
     "node_modules/@ampproject/remapping": {
--- a/package.json
+++ b/package.json
@@ -872,7 +872,7 @@
     }
   ],
   "engines": {
-    "vscode": "^1.73.0"
+    "vscode": "^1.83.0"
   },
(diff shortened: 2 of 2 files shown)


dibarbet:
I assume the plan here is to wait for a 1.83 version of vscode to be released?

timheuer:
> I assume the plan here is to wait for a 1.83 version of vscode to be released?

Do we have a pre-release train this could hop on for validation?

dibarbet:
> > I assume the plan here is to wait for a 1.83 version of vscode to be released?
> 
> Do we have a pre-release train this could hop on for validation?

We don't currently have a planned prerelease, but we can ship a prerelease out of main at any time.  That prerelease would only work on vscode insiders though.

The other, more complicated problem is that our integration tests run on vscode release.  We would also have to switch the integration tests to acquire and run on insiders vscode in order to merge this PR (they're currently all failing on this PR for that reason)

gregg-miskelly:
Do we need to wait? In other words, will older versions of VS Code just ignore the attribute?

WardenGnaw:
These changes work if you use a version of vscode >1.83.0. We do not need the engine upgrade requirement.

package-lock.json
```diff
@@ -108,7 +108,7 @@
         "webpack-cli": "4.6.0"
       },
       "engines": {
-        "vscode": "^1.73.0"
+        "vscode": "^1.83.0"
```
gregg-miskelly, line 111: You want to undo this file

dibarbet (approved):

gregg-miskelly (approved):

PR #6438, opened 2023-09-26
Add support for collecting dumps on crash

dibarbet:
This adds an option to easily capture a crash dump of the language server process.
1.  A vscode option is generally easier than manually trying to launch vscode with a specific set of environment variables
2.  There have been a couple reports where manually setting the env vars isn't easy to do (for example wsl where the code server process is not directly launched by the user) - e.g. https://github.com/microsoft/vscode-dotnettools/issues/522

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -452,6 +452,17 @@ export class RoslynLanguageServer {
         // Save user's DOTNET_ROOT env-var value so server can recover the user setting when needed
         env.DOTNET_ROOT_USER = process.env.DOTNET_ROOT ?? 'EMPTY';
 
+        if (languageServerOptions.crashDumpPath) {
+            // Enable dump collection
+            env.DOTNET_DbgEnableMiniDump = '1';
+            // Collect heap dump
+            env.DOTNET_DbgMiniDumpType = '2';
+            // Collect crashreport.json with additional thread and stack frame information.
+            env.DOTNET_EnableCrashReport = '1';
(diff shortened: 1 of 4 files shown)


arkalyanms (approved):

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -452,6 +452,16 @@ export class RoslynLanguageServer {
...
         env.DOTNET_ROOT_USER = process.env.DOTNET_ROOT ?? 'EMPTY';
 
+        if (languageServerOptions.crashDumpPath) {
+            // Enable dump collection
+            env.DOTNET_DbgEnableMiniDump = '1';
+            // Collect heap dump
+            env.DOTNET_DbgMiniDumpType = '2';
+            // Collect crashreport.json with additional thread and stack frame information.
+            env.DOTNET_EnableCrashReport = '1';
+            env.DOTNET_DbgMiniDumpName = path.join(languageServerOptions.crashDumpPath, '%e.%d.dmp');
```
jasonmalinowski, line 462: You might want to comment what the substitutions do here, especially because I don't see what %d does in the docs. Did you mean %p?

jasonmalinowski (approved):

src/lsptoolshost/roslynLanguageServer.ts
dibarbet: Ah %d is the old one, see https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/botr/xplat-minidump-generation.md#configurationpolicy

I'll add a comment and change to %p

PR #6441, opened 2023-09-26
Fix Razor completion bugs

allisonchou:
These fixes are both pretty hacky, but I believe we lack reasonable alternatives at the moment.

1) Fixes bug where all completion item icons are incorrect (notice `ushort` appears with the snippet icon instead of the keyword icon):
image

This is because we're deserializing the LSP completion item type returned by Roslyn into a VS Code completion item type. Notice that the `CompletionItemKinds` on both types are off by one:
image

2) Fixes https://github.com/dotnet/vscode-csharp/issues/6199

We technically have the same bug in VS, but since we have snippets available there (snippets are currently unavailable for Razor in VS Code), we still get `using` in the completion list (screenshot below is the current experience in VS):
image

However, both VS and VS Code are missing `using` the _keyword_ in the list. The reason we're not getting the keyword is because we're trying to get completions at `__o = [||]`, which is not a valid location for a using statement.

End result after both of these fixes (1 & 2):
image

### DIFF
--- a/src/razor/src/completion/razorCompletionItemProvider.ts
+++ b/src/razor/src/completion/razorCompletionItemProvider.ts
@@ -30,7 +30,8 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
         hostDocumentPosition: vscode.Position,
         projectedPosition: vscode.Position,
         context: vscode.CompletionContext,
-        language: LanguageKind
+        language: LanguageKind,
+        razorDocument: vscode.TextDocument
     ) {
         if (projectedUri) {
             // "@" is not a valid trigger character for C# / HTML and therefore we need to translate
@@ -86,6 +87,11 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
             // In the code behind it's represented as __o = DateTime.
             const completionCharacterOffset = projectedPosition.character - hostDocumentPosition.character;
             for (const completionItem of completionItems) {
+                // vscode.CompletionItemKind is off by one compared to the LSP CompletionItemKind.
+                if (completionItem.kind !== undefined) {
+                    completionItem.kind = completionItem.kind - 1;
+                }
+
                 const doc = completionItem.documentation as vscode.MarkdownString;
                 if (doc && doc.value) {
                     // Without this, the documentation doesn't get rendered in the editor.
@@ -166,6 +172,8 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
                 }
             }
 
+            this.addUsingKeyword(language, razorDocument, hostDocumentPosition, completionItems);
+
             const isIncomplete = completions instanceof Array ? false : completions ? completions.isIncomplete : false;
             return new vscode.CompletionList(completionItems, isIncomplete);
         }
@@ -221,7 +229,8 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
             position,
             projection.position,
             context,
-            projection.languageKind
+            projection.languageKind,
+            document
         );
 
         return completionList;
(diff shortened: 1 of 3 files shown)


src/razor/src/completion/razorCompletionItemProvider.ts
```diff
@@ -278,6 +287,38 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
...
+        completionItems: vscode.CompletionItem[]
+    ) {
+        // This is an ugly hack, but it's needed to get the "using" keyword to show up in the completion list.
+        // The reason it doesn't show up is because the C# generated document puts the position of the cursor
+        // at '__o = [||]', which isn't a valid location for a using statement.
+        if (language == LanguageKind.CSharp) {
+            const line = razorDocument.lineAt(hostDocumentPosition.line);
+            const lineText = line.text.substring(0, hostDocumentPosition.character);
+            if (
+                lineText.endsWith('@') ||
```
allisonchou, line 304: I know this check looks terrible - honestly it makes me cringe too 😅 If anyone has suggestions for an alternative, feel free to suggest

src/razor/src/completion/razorCompletionItemProvider.ts
```diff
@@ -86,6 +87,11 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
             // In the code behind it's represented as __o = DateTime.
             const completionCharacterOffset = projectedPosition.character - hostDocumentPosition.character;
             for (const completionItem of completionItems) {
+                // vscode.CompletionItemKind is off by one compared to the LSP CompletionItemKind.
```
dibarbet, line 90: Wow that is mighty annoying 😆

dibarbet (approved):
*only* signing off on the roslyn-ide changes (l10n)

src/razor/src/completion/razorCompletionItemProvider.ts
```diff
@@ -278,6 +287,38 @@ export class RazorCompletionItemProvider extends RazorLanguageFeatureBase implem
...
+            const line = razorDocument.lineAt(hostDocumentPosition.line);
+            const lineText = line.text.substring(0, hostDocumentPosition.character);
+            if (
+                lineText.endsWith('@') ||
+                lineText.endsWith(
+                    '@u' ||
+                        lineText.endsWith('@us') ||
+                        lineText.endsWith('@usi') ||
+                        lineText.endsWith('@usin') ||
+                        lineText.endsWith('@using')
```
JoeRobich, line 310: You may have come up with the best way. If the user has already finished typing `@using`, does the item still need to be added?

```suggestion
                lineText.endsWith('@') ||
                    lineText.endsWith('@u') ||
                    lineText.endsWith('@us') ||
                    lineText.endsWith('@usi') ||
                    lineText.endsWith('@usin') ||
                    lineText.endsWith('@using')
```

src/razor/src/completion/razorCompletionItemProvider.ts
ryzngard: KermitWorriedGIF

src/razor/src/completion/razorCompletionItemProvider.ts
ryzngard: This works but... ugh... 

What's the benefit of having a using snippet? Can we just make a server fix for Razor to add it's own completion for `@` directives? I think we already do have some... (like `@co`)

ryzngard (approved):
I'm approving, and feel gross with you.

src/razor/src/completion/razorCompletionItemProvider.ts
dibarbet: And https://github.com/dotnet/razor/pull/6012/files#diff-052d5834149367e2bcf4d95d0740ddb4b346116cfcd64e568d0b5eca38f7b8aeL739

genlu:
> We technically have the same bug in VS, but since we have snippets available there

@allisonchou Sounds like this would be an issue if snippet is disabled? Have you considered adding a specific completion provider in razor instead? That should cover both VS and VSCode

src/razor/src/completion/razorCompletionItemProvider.ts
allisonchou: I considered this initially but wanted to match C#'s behavior. They seem to provide the completion item even if the user has finished typing out the whole word:
image

src/razor/src/completion/razorCompletionItemProvider.ts
allisonchou: > What's the benefit of having a using snippet? Can we just make a server fix for Razor to add it's own completion for @ directives? I think we already do have some... 

I'm guessing a using snippet might be beneficial in the case where the user is already in a C# context and would want to insert something like `using () { }`?

I've filed https://github.com/dotnet/razor/issues/9330 to track a (still hacky) server-side fix that we can use in both VS and VS Code. In the interest of getting a temporary fix in I'm planning on merging this version for now, but will follow up soon on the server-side fix.

allisonchou:
> @allisonchou Sounds like this would be an issue if snippet is disabled?

Yeah, it would be an issue in VS if snippets were disabled.

> Have you considered adding a specific completion provider in razor instead? That should cover both VS and VSCode

That's a good idea! I've filed https://github.com/dotnet/razor/issues/9330 and https://github.com/dotnet/razor/issues/9331 as follow ups. I'm going to merge this version for now and then revisit your suggestion soon in the near future in the interest of at least getting some sort of temporary fix in.

PR #6447, opened 2023-09-28
Update filename to use for project info

davidwengier:
Copying Phil in https://github.com/dotnet/roslyn/pull/70156


dibarbet:
Does this require a razor update?  Or can this go in standalone?

davidwengier:
No Razor update needed, but needs to go in at the same time as a Roslyn update that includes Phil's PR

phil-allen-msft (approved):

phil-allen-msft:
Approved, but we should probably wait until the Roslyn side PR comes through so that we write & read the same file.

davidwengier:
Tempted to clean this up and move the logic all to Razor, like it says in the comment

davidwengier:
Actually, even if we consolidate all of the logic into the Razor repo, we'd still need dual insertions to change it because the dependency is backwards here, and Roslyn references Razor. Maybe the real cleanup is to flip that, and have Razor deploy a DLL, and pass it to the language server as an extension.

For now I'll merge the Roslyn PR and update this one with a Roslyn bump.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 ## Latest
+* Update Roslyn version to 4.9.0-1.23502.8 (PR: [#6447](https://github.com/dotnet/vscode-csharp/pull/6447))
+  * Update Razor project configuration file name (PR: [#70156](https://github.com/dotnet/roslyn/pull/70156))
+
+## 2.5.?
```
davidwengier, line 14: @dibarbet not sure what the normal process here is. Will it be 2.5.0? .1?

dibarbet: it is most likely going to be 2.5.30 unless we have to re-spin for some reason (you can check the release branch build number).  You can use that, or ? and we'll update it when we actually release.

davidwengier: I'll leave it then. More likely to notice it needs to be updated if its a question mark

davidwengier:
Finally got the builds to pass, and whilst Windows Sandbox died on me 3 times, I validated on a DevBox and razor seems to continue working fine. My machine is haunted, but that seems unrelated.

dibarbet (approved):

PR #6473, opened 2023-10-03
Initialize Razor even if Razor doc isn't opened yet

allisonchou:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1825836, aka the scenario where:
- There is a Razor doc in the workspace
- No Razor file has been opened yet
- User attempts to open a C# doc referencing Razor components

In the above scenario, we get errors like this in the C# file:
image

We were initially trying to avoid unnecessary work by not initializing Razor until a Razor file is actually opened. However, it creates bugs like above which have been reported multiple times.


davidwengier (approved):
Well I guess this settles the "should Razor be a runtime dependency" argument once and for all 😁

PR #6480, opened 2023-10-03
Track debugging sessions until csdevkit is initialized

isadorasophia:
**Issue**
When using C# dev kit with C# extension, there was a bug that hot reload might not initialize on the first debugging session. This happened due to a race condition that the 'launch' command is fired before the C# dev kit finishes initializing, so no `brokeredServicePipeName` is passed to the debugger.

**Fix**
Starts tracking the debugging session until C# dev kit is initialized. If there is an existing debugging session, this sends a custom `initializeBrokeredServicePipeName` request with the pipe name.

### DIFF
--- a/src/coreclrDebug/provisionalDebugSessionTracker.ts
+++ b/src/coreclrDebug/provisionalDebugSessionTracker.ts
@@ -1,0 +1,91 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+/**
+ * This tracks a debug session until C# dev kit is loaded (if present) and STOPS tracking any existing
+ * debug session after C# dev kit is loaded.
+ * The idea is to avoid a race condition if a debugging session starts before C# dev kit is initialized,
+ * since the brokered service pipe name will be sent with a 'launch' command.
+ * If the extension loads during an existing session, rather than not initializing any C# dev kit services (such as hot reload),
+ * this sends a custom request to the engine with the brokered service pipe name in order to initialize it.
+ */
+export class ProvisionalDebugSessionTracker {
+    private _session: vscode.DebugSession | undefined;
+
+    private _onDidStartDebugSession: vscode.Disposable | undefined;
+    private _onDidTerminateDebugSession: vscode.Disposable | undefined;
+
+    private _brokeredServicePipeName: string | undefined;
+
+    DebuggerSessionTracker() {
+        this._session = undefined;
+    }
+
+    /**
+     * Initializes the debug session handlers.
+     */
+    async initializeDebugSessionHandlers(context: vscode.ExtensionContext): Promise<void> {
+        this._onDidStartDebugSession = vscode.debug.onDidStartDebugSession(this.onDidStartDebugSession.bind(this));
+
+        this._onDidTerminateDebugSession = vscode.debug.onDidTerminateDebugSession(
+            this.onDidTerminateDebugSession.bind(this)
+        );
+
+        context.subscriptions.push(this._onDidStartDebugSession);
+        context.subscriptions.push(this._onDidTerminateDebugSession);
+    }
+
+    /**
+     * Tracks a debug session until it is terminated.
+     * @param session Debug session.
+     */
+    async onDidStartDebugSession(session: vscode.DebugSession): Promise<void> {
+        this._session = session;
+    }
+
+    /**
+     * Notifies that a debug session has been terminated.
+     */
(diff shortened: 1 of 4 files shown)


WardenGnaw (approved):

src/coreclrDebug/provisionalDebugSessionTracker.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * This tracks a debug session until C# dev kit is loaded (if present) and STOPS tracking any existing
+ * debug session after C# dev kit is loaded.
+ * The idea is to avoid a race condition if a debugging session starts before C# dev kit is initialized,
+ * since the brokered service pipe name will be sent with a 'launch' command.
+ * If the extension loads during an existing session, rather than not initializing any C# dev kit services (such as hot reload),
+ * this sends a custom request to the engine with the brokered service pipe name in order to initialize it.
+ */
+export class ProvisionalDebugSessionTracker {
+    private _session: vscode.DebugSession | undefined;
```
gregg-miskelly, line 17: >_session [](http://example.com/codeflow?start=12&length=8)

I believe we can have more than one debug session

src/coreclrDebug/provisionalDebugSessionTracker.ts
```diff
@@ -0,0 +1,91 @@
...
+        context.subscriptions.push(this._onDidStartDebugSession);
+        context.subscriptions.push(this._onDidTerminateDebugSession);
+    }
+
+    /**
+     * Tracks a debug session until it is terminated.
+     * @param session Debug session.
+     */
+    async onDidStartDebugSession(session: vscode.DebugSession): Promise<void> {
+        this._session = session;
```
gregg-miskelly, line 47: >this._session = session; [](http://example.com/codeflow?start=8&length=24)

You should ignore debug sessions that aren't for `coreclr`.

src/coreclrDebug/provisionalDebugSessionTracker.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        context.subscriptions.push(this._onDidStartDebugSession);
+        context.subscriptions.push(this._onDidTerminateDebugSession);
+    }
+
+    /**
+     * Tracks a debug session until it is terminated.
+     * @param session Debug session.
+     */
+    async onDidStartDebugSession(session: vscode.DebugSession): Promise<void> {
```
gregg-miskelly, line 46: >async [](http://example.com/codeflow?start=4&length=5)

Why `async`?

src/coreclrDebug/provisionalDebugSessionTracker.ts
```diff
@@ -0,0 +1,91 @@
...
+     * If there is any active debug session, this notifies the engine that csdevkit was loaded.
+     * @param csDevKitPipeName Brokered service pipe name activated by {@link CSharpDevKitExports}.
+     */
+    async onCsDevKitInitialized(csDevKitPipeName: string): Promise<void> {
+        if (this._session != undefined) {
+            // Debugging session already started, send a custom DAP request to the engine.
+            await this._session.customRequest('initializeBrokeredServicePipeName', csDevKitPipeName);
+        }
+
+        this._brokeredServicePipeName = csDevKitPipeName;
```
gregg-miskelly, line 67: >this [](http://example.com/codeflow?start=8&length=4)

Probably doesn't matter, but should this be done before the `await`?

gregg-miskelly (changes requested):
See comments

src/coreclrDebug/provisionalDebugSessionTracker.ts
isadorasophia: is it okay to clear all sessions at `onDidTerminateDebugSession`?

gregg-miskelly: I am not positive, but the easiest way to find out is to either:

Option 1 - multiple projects
1. Make a workspace with two console apps
2. Change them so that they will never exit (ex: `while (true) Thread.Sleep(100)`)
3. Create a new launch.json file
4. Add a debug configuration for both projects. Something like below.
5. Add a compond configuration: https://code.visualstudio.com/Docs/editor/debugging#_compound-launch-configurations

```
    {
            "name": "C#: project-name-here Debug",
            "type": "dotnet",
            "request": "launch",
            "projectPath": "${workspaceFolder}/project-name-here.csproj"
        }
```

Option 2: launch the same project more than once

Disclaimer: I haven't tried this, but seems like it should work.

1. Create the same project as above,
2. Open solution explorer and right click on the project
3. Invoke the debug launch command
4. Repeat again

isadorasophia: I tested and it seems to work (hot reload works on both apps o:)

gregg-miskelly (approved):

dibarbet (approved):

PR #6481, opened 2023-10-04
Always build release and prerelease VSIXs and allow overriding build …

dibarbet:
…number so that we can ship release or prerelease from any (release/main/patch)

Test builds:
https://dnceng.visualstudio.com/internal/_build/results?buildId=2283050&view=results
https://dnceng.visualstudio.com/internal/_build/results?buildId=2283875&view=results
https://dnceng.visualstudio.com/internal/_build/results?buildId=2283873&view=results

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -1,26 +1,63 @@
 parameters:
-- name: prereleaseFlag
+- name: versionNumberOverride
   type: string
+  default: 'default'
 
-steps:
-- checkout: self
-  clean: true
-  submodules: true
-  fetchTags: false
-  fetchDepth: 0
-
-- template: prereqs.yml
-
-- script: gulp 'vsix:release:package' ${{ parameters.prereleaseFlag }}
+stages:
+- stage: Build
   displayName: 'Build VSIXs'
+  jobs:
+  - job:
+    displayName: 'Build Prerelease VSIXs'
+    steps:
+    - checkout: self
+      clean: true
+      submodules: true
+      fetchTags: false
+      fetchDepth: 0
 
-- task: PublishPipelineArtifact@1
-  # Run the publish step so we have vsix's even if the tests fail.
-  condition: succeededOrFailed()
-  displayName: 'Publish VSIXs'
-  inputs:
-    targetPath: '$(Build.SourcesDirectory)/vsix'
-    artifactName: 'VSIXs - Attempt $(System.JobAttempt)'
+    - template: prereqs.yml
+      parameters:
+        versionNumberOverride: ${{ parameters.versionNumberOverride }}
 
-- script: npm run omnisharptest:artifacts
-  displayName: 'Run artifacts tests'
+    - script: gulp 'vsix:release:package' --prerelease
+      displayName: 'Build VSIXs'
+
+    - task: PublishPipelineArtifact@1
+      # Run the publish step so we have vsix's even if the tests fail.
+      condition: succeededOrFailed()
+      displayName: 'Publish VSIXs'
+      inputs:
+        targetPath: '$(Build.SourcesDirectory)/vsix'
+        artifactName: 'VSIX_Prerelease_$(System.JobAttempt)'
+
+    - script: npm run omnisharptest:artifacts
+      displayName: 'Run artifacts tests'
+
+  - job:
+    displayName: 'Build Release VSIXs'
+    steps:
+    - checkout: self
+      clean: true
+      submodules: true
+      fetchTags: false
+      fetchDepth: 0
+
+    - template: prereqs.yml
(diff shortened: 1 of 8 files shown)


version.json
```diff
@@ -3,7 +3,8 @@
 	"version": "2.6",
 	"publicReleaseRefSpec": [
 		"^refs/heads/release$",
-		"^refs/heads/main$"
+		"^refs/heads/main$",
+		"^refs/heads/patch/.*$"
```
dibarbet, line 7: the patch branches will get 'real' version numbers so we can release out of them.  We also need to setup sync'ing to the internal branch for any patch branch.

dibarbet:
This should not merge until we're ready to do a new release out of the release branch.

azure-pipelines/release.yml
```diff
@@ -11,6 +11,9 @@ parameters:
   - name: test
     type: boolean
     default: true
+  - name: uploadPrerelease
```
dibarbet, line 14: this is now a flag in the pipeline instead of be ing determined by branch name

azure-pipelines/build.yml
```diff
@@ -1,26 +1,63 @@
...
-  displayName: 'Publish VSIXs'
-  inputs:
-    targetPath: '$(Build.SourcesDirectory)/vsix'
-    artifactName: 'VSIXs - Attempt $(System.JobAttempt)'
+    - task: PublishPipelineArtifact@1
+      # Run the publish step so we have vsix's even if the tests fail.
+      condition: succeededOrFailed()
+      displayName: 'Publish VSIXs'
+      inputs:
+        targetPath: '$(Build.SourcesDirectory)/vsix'
```
arkalyanms, line 58: can we make the target paths be different for prerelease and release packet of vsix'es? Might preempt dogfooding and testing confusions.

arkalyanms (approved):

azure-pipelines/build.yml
dibarbet: Is this what you mean?  THey do get uploaded to different paths
image

tasks/offlinePackagingTasks.ts
```diff
@@ -206,13 +206,14 @@ async function acquireNugetPackage(packageName: string, packageVersion: string,
...
 async function doPackageOffline() {
-    // Set the package version using git versioning.
+    // Set the package.json version based on the value in version.json.
     const versionInfo = await nbgv.getVersion();
     console.log(versionInfo.npmPackageVersion);
     await nbgv.setPackageVersion();
 
     let prerelease = false;
     if (argv.prerelease) {
+        console.log('Packaging prerelease version.');
```
jasonmalinowski, line 216: Did you want an 'else' to log the other way as well just so the logs are clear either way?

version.json
jasonmalinowski: Just to confirm my understanding of the docs here, this doesn't necessarily change the algorithm used to choose the version number, but just says that builds out of this branch don't get the commit SHA tagged to them?

azure-pipelines/release.yml
```diff
@@ -55,13 +65,11 @@ stages:
 
               $basePublishArgs = , "publish"
               # Artifacts are published to either pre-release or release based on the build branch, https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions
```
jasonmalinowski, line 67: Stale comment now.

azure-pipelines/release.yml
```diff
@@ -44,8 +47,15 @@ stages:
...
+              $artifactFolderPrefix = "VSIX_Release_"
+              if ("${{ parameters.uploadPrerelease }}" -eq "true") {
+                $artifactFolderPrefix = "VSIX_Prerelease_"
+              }
+
+              Write-Host "Using artifacts from $artifactFolderPrefix"
+
               # Our build pipeline would generated build based on attempt number. Publishing the latest attempt.
-              $allArtifacts = Get-ChildItem -Path "VSIXs - Attempt*" | Sort-Object -Descending
+              $allArtifacts = Get-ChildItem -Path "$artifactFolderPrefix*" | Sort-Object -Descending
```
jasonmalinowski, line 58: Hmm, this is doing a textual sort right, so pray we never have 10 attempts?

jasonmalinowski (approved):

version.json
dibarbet: Thats correct

azure-pipelines/release.yml
dibarbet: Yeah.  That's a pre-existing issue though

azure-pipelines/build.yml
arkalyanms: Yes

PR #6486, opened 2023-10-04
Fix package.json version

akhera99:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "csharp",
   "publisher": "ms-dotnettools",
-  "version": "2.0.0-placeholder",
+  "version": "42.42.42-placeholder",
```
dibarbet, line 4: this ensures we're higher than all the other devkit version checks.

Note that this version is always overwritten by the one from version.props (and git versioning).

dibarbet (approved):

jasonmalinowski (approved):
image

PR #6489, opened 2023-10-05
Add Razor C# semantic tokens support in VS Code

allisonchou:
Transitions C# in Razor to use semantic tokens instead of Textmate.

Fixes https://github.com/dotnet/razor/issues/9336, https://github.com/dotnet/razor/issues/4323, and https://github.com/dotnet/razor/issues/4332

Examples of scenarios this fixes:

- Before:
image
- After:
image

- Before:
image
- After:
image


package.json
```diff
@@ -5175,6 +5175,168 @@
           "markupCommentPunctuation": [
             "punctuation.definition.comment.html",
             "comment.block.html"
+          ],
+          "keyword": [
```
allisonchou, line 5179: This is needed to override the default colors VS Code uses since they can differ compared to what C# uses. For example, `keyword` is colored purple by VS Code, but colored blue by C#.

Thanks @dibarbet for the tip!

src/razor/language-configuration.json
```diff
@@ -20,5 +20,5 @@
 		{ "open": "'", "close": "'" },
 		{ "open": "\"", "close": "\"" },
 		{ "open": "<", "close": ">" }
-	]
+	],
```
ryzngard, line 23:
```suggestion
	]
```

src/razor/src/semantic/semanticTokensRangeHandler.ts
```diff
@@ -16,9 +21,14 @@ export class SemanticTokensRangeHandler {
         ProvideSemanticTokensResponse,
         any
     > = new RequestType(SemanticTokensRangeHandler.getSemanticTokensRangeEndpoint);
-    private emptyTokensInResponse: Array<Array<number>> = new Array<Array<number>>();
+    private emptyTokensResponse: Uint32Array = new Uint32Array();
```
ryzngard, line 24: ❓ what is the benefit of `Uint32Array` over `Array<number>`?

package.json
```diff
@@ -5369,7 +5531,10 @@
       {
         "language": "aspnetcorerazor",
         "scopeName": "text.aspnetcorerazor",
-        "path": "./src/razor/syntaxes/aspnetcorerazor.tmLanguage.json"
+        "path": "./src/razor/syntaxes/aspnetcorerazor.tmLanguage.json",
+        "unbalancedBracketScopes": [
```
allisonchou, line 5535: Needed to address this scenario:
image

Otherwise, VS Code will think there is an unclosed bracket and color it red.

src/razor/src/semantic/semanticTokensRangeHandler.ts
```diff
@@ -4,10 +4,15 @@
...
 import * as vscode from 'vscode';
-import { RequestType } from 'vscode-languageclient';
+import { RequestType, TextDocumentIdentifier } from 'vscode-languageclient';
 import { RazorLanguageServerClient } from '../razorLanguageServerClient';
 import { ProvideSemanticTokensResponse } from './provideSemanticTokensResponse';
 import { SerializableSemanticTokensParams } from './serializableSemanticTokensParams';
+import { RazorDocumentManager } from '../document/razorDocumentManager';
+import { RazorDocumentSynchronizer } from '../document/razorDocumentSynchronizer';
+import { RazorLogger } from '../razorLogger';
+import { provideSemanticTokensRangesCommand as provideSemanticTokensRangesCommand } from '../../../lsptoolshost/razorCommands';
```
ryzngard, line 14: ❓ why is the `as` needed?

src/razor/src/semantic/provideSemanticTokensResponse.ts
```diff
@@ -5,5 +5,5 @@
 
 export class ProvideSemanticTokensResponse {
     // tslint:disable-next-line: variable-name
-    constructor(public Tokens: Array<Array<number>>, public HostDocumentSyncVersion: number) {}
+    constructor(public Tokens: Uint32Array, public HostDocumentSyncVersion: number) {}
```
allisonchou, line 8: This is a `Uint32Array` since that's what `vscode.SemanticTokens.data` uses as their type

ryzngard (approved):

package.json
ryzngard: Interesting, so "keyword.cs" is already blue for VS Code and it just... knows?

src/razor/src/semantic/semanticTokensRangeHandler.ts
allisonchou: Good catch - removed, thanks!

package.json
allisonchou: I'm guessing C# sets the color somewhere, although I'm not sure where. Maybe @dibarbet or @JoeRobich would know?

src/razor/src/semantic/provideSemanticTokensResponse.ts
davidwengier: Isn't this a mismatch to the Razor request type though? @maryamariyan made it return an array of arrays when doing the precise tokens work. I'm surprised this doesn't fail to deserialize on the Razor side

package.json
dibarbet: I'm not sure if this is the only one - you might want to test less than / greather than as well

src/razor/src/semantic/provideSemanticTokensResponse.ts
allisonchou: It looks like the feature flag is off in VS Code? I'm seeing `UsePreciseSemanticTokenRanges` return false

allisonchou: Is the flag meant to be on in VS Code? @maryamariyan

package.json
allisonchou: Ah you're right, and I was missing bit shifting operators as well. Thanks!

src/razor/src/semantic/semanticTokensRangeHandler.ts
allisonchou: I changed the code to use `vscode-languageclient.SemanticTokens` in my latest push - hopefully it's clearer!

package.json
dibarbet: I don't quite remember - it was potentially that the lsp token type for `keyword` mapped to `keyword.control` but `keyword.cs` falls back to `keyword` instead

src/razor/src/semantic/provideSemanticTokensResponse.ts
davidwengier: Looks like the type changed back to a single dimension array in the latest updates on the Razor side. Maybe we just never got around to updating this side of things, since it didn't do anything.

src/lsptoolshost/razorCommands.ts
```diff
@@ -111,6 +112,24 @@ export function registerRazorCommands(context: vscode.ExtensionContext, language
         })
     );
 
+    const semanticTokensRangesRequestType = new RequestType<
+        SerializableSemanticTokensParams,
+        vscode.SemanticTokens,
+        any
+    >('roslyn/semanticTokenRanges');
```
davidwengier, line 119: I think this is the new endpoint for the feature flag, so we probably don't want to use it. We should be able to use the standard `textDocument/semanticTokens/range` method.

package.json
dibarbet: POssibly https://github.com/microsoft/vscode/blob/b96b558f6479557a43de966d7cfdbcdffb82d715/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L545

src/razor/src/semantic/provideSemanticTokensResponse.ts
davidwengier: Okay, latest Razor bump was 5 days ago in https://github.com/dotnet/vscode-csharp/pull/6449 and the new Razor server changes went in 7 days ago, so this change to a single dimensional array makes sense.

src/lsptoolshost/razorCommands.ts
allisonchou: That's fair. It is kind of weird since we're still being sent an array of ranges from the server, so I had to hardcode in `ranges[0]` to use the standard method.

Also, since we're going back to the standard method, I tested out the built-in `'vscode.provideDocumentRangeSemanticTokens` command and was actually able to get it working this time around. 😄 Definitely simplified the code

davidwengier (approved):
Nice work!

PR #6494, opened 2023-10-05
Fire telemetry if Razor buffers get out of sync, and recover from same

davidwengier:
This is the VS Code equivalent of https://github.com/dotnet/razor/pull/9202

Fixes https://github.com/dotnet/vscode-csharp/issues/6488

I'm not sure why this is so easily reproducable for the user reporting the error, but its extremely intermittent for me. I basically hit it twice today, and never again, but I did at least see that we had duplicate HTML content in our generated file when I hit it, so can be sure this is the issue we are running into.

Perhaps a new timing issue with our initialization, now that its not deferred until file open, but with part of Roslyn and the extension still expecting it to be. Hopefully https://github.com/dotnet/razor/issues/9344 will fix those issues.

### DIFF
--- a/src/razor/src/document/razorDocumentManager.ts
+++ b/src/razor/src/document/razorDocumentManager.ts
@@ -245,7 +245,7 @@ export class RazorDocumentManager implements IRazorDocumentManager {
             const csharpProjectedDocument = projectedDocument as CSharpProjectedDocument;
 
             // If the language server is telling us that the previous document was empty, then we should clear
-            // ours out. Hopefully outs would have been empty too, but there are cases where things get out of
+            // ours out. Hopefully ours would have been empty too, but there are cases where things get out of
             // sync
             if (updateBufferRequest.previousWasEmpty && projectedDocument.length !== 0) {
                 this.telemetryReporter.reportBuffersOutOfSync();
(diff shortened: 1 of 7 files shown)


davidwengier:
@allisonchou is this all I need to do for telemetry reporting? It seems too easy :)

src/razor/src/document/razorDocumentManager.ts
```diff
@@ -237,6 +243,15 @@ export class RazorDocumentManager implements IRazorDocumentManager {
             await vscode.workspace.openTextDocument(document.csharpDocument.uri);
 
             const csharpProjectedDocument = projectedDocument as CSharpProjectedDocument;
+
+            // If the language server is telling us that the previous document was empty, then we should clear
+            // ours out. Hopefully outs would have been empty too, but there are cases where things get out of
```
allisonchou, line 248: nit:
```suggestion
            // ours out. Hopefully ours would have been empty too, but there are cases where things get out of
```

src/razor/src/document/razorDocumentManager.ts
```diff
@@ -270,6 +286,15 @@ export class RazorDocumentManager implements IRazorDocumentManager {
             await vscode.workspace.openTextDocument(document.htmlDocument.uri);
 
             const htmlProjectedDocument = projectedDocument as HtmlProjectedDocument;
+
+            // If the language server is telling us that the previous document was empty, then we should clear
+            // ours out. Hopefully outs would have been empty too, but there are cases where things get out of
```
allisonchou, line 291:
```suggestion
            // ours out. Hopefully ours would have been empty too, but there are cases where things get out of
```

allisonchou (approved):
LGTM!

> @allisonchou is this all I need to do for telemetry reporting? It seems too easy :)

I _think_ so? I'm not as familiar with telemetry on the Typescript side of things - @ryzngard would probably know for sure. Does it show up on the GDPR site?

davidwengier:
> Does it show up on the GDPR site?

I haven't actually been able to repro the issue to check, since putting the fix in, but I'll try some more.

I was mainly wondering if it's okay to just be hooking up the TelemetryReporter like this, and calling methods on it at will, or if there needs to be checks of some kind so we only do it in DevKit

ryzngard (approved):
I believe this is correct for telemetry in ts. 

:shipit:

PR #6502, opened 2023-10-06
Fix some Razor colorization issues

allisonchou:
Partially fixes https://github.com/dotnet/razor/issues/9375 (rest of issue in progress)

Examples of scenarios this PR fixes:

Before:
image
After (the parens problem is going to require a more complicated fix that I'm actively working on):
image

Before:
image
After (same comment as above as parens fix):
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -4813,7 +4813,7 @@
         "description": "A Razor component element"
       },
       {
-        "id": "razorComponentAttribute",
+        "id": "RazorComponentAttribute",
         "description": "A Razor component attribute"
       },
       {
@@ -4863,6 +4863,10 @@
       {
         "id": "markupAttribute",
         "description": "The name of a Markup attribute."
+      },
+      {
+        "id": "markupAttributeQuote",
+        "description": "A token that represents an attribute quote in a Markup attribute."
       },
       {
         "id": "markupComment",
@@ -5119,7 +5123,7 @@
           "razorComponentElement": [
             "entity.name.class.element.component"
           ],
-          "razorComponentAttribute": [
+          "RazorComponentAttribute": [
             "entity.name.class.attribute.component"
           ],
           "razorTagHelperElement": [
@@ -5169,12 +5173,18 @@
           "markupAttribute": [
             "entity.other.attribute-name.html"
           ],
+          "markupAttributeQuote": [
+            "punctuation.definition.string.html"
+          ],
           "markupComment": [
             "comment.block.html"
           ],
           "markupCommentPunctuation": [
             "punctuation.definition.comment.html",
             "comment.block.html"
+          ],
+          "markupTagDelimiter": [
+            "punctuation.definition.tag.html"
           ],
           "keyword": [
             "keyword.cs"


package.json
```diff
@@ -4813,7 +4813,7 @@
         "description": "A Razor component element"
       },
       {
-        "id": "razorComponentAttribute",
+        "id": "RazorComponentAttribute",
```
allisonchou, line 4816: This is weird, but somehow the Razor language server defines `RazorComponentAttribute` as uppercase but all other tokens as lowercase: https://github.com/dotnet/razor/blob/main/src/Razor/src/Microsoft.AspNetCore.Razor.LanguageServer/Semantic/Models/RazorSemanticTokensLegend.cs#L31C13-L31C13

We probably want to fix this in the future, although I'm not sure of the implications of making it lowercase/what other things may break

package.json
```diff
@@ -4864,6 +4864,10 @@
         "id": "markupAttribute",
         "description": "The name of a Markup attribute."
       },
+      {
+        "id": "markupAttributeQuote",
```
allisonchou, line 4868: We were missing this semantic token type

package.json
```diff
@@ -5169,13 +5173,19 @@
...
+            "punctuation.definition.string.html"
+          ],
           "markupComment": [
             "comment.block.html"
           ],
           "markupCommentPunctuation": [
             "punctuation.definition.comment.html",
             "comment.block.html"
           ],
+          "markupTagDelimiter": [
```
allisonchou, line 5186: We defined this token earlier in the file but failed to set a scope for it

package.json
```diff
@@ -5119,7 +5123,7 @@
           "razorComponentElement": [
             "entity.name.class.element.component"
           ],
-          "razorComponentAttribute": [
+          "RazorComponentAttribute": [
```
dibarbet, line 5126: same here - this feels really inconsistent.  Is it possible to make these all uniform

package.json
dibarbet: Ah you added this as soon as I hit submit review.  IMHO we should modify the server side.  It would only break people that have added specific theme customization for this token.  But I'm not sure that is super common, and it is easily fixable.

ryzngard: +1 to making a server fix here, unless there's some timeline issues with flowing the bits.

ryzngard (approved):
Signing off but would love to see a server fix of "RazorComponentAttribute" to "razorComponentAttribute"

package.json
allisonchou: So changing the casing on the server side breaks VS:
image

I think this is because we have to make a change to the VS LSP client as well: https://devdiv.visualstudio.com/DevDiv/_git/VSLanguageServerClient?path=/src/product/RemoteLanguage/Impl/Features/SemanticTokens/Tagger/SemanticTokensTaggerBase.cs&version=GBdevelop&line=564&lineEnd=565&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

I'm going to file a follow up issue for this, but since the fix requires multiple repos I think we could still take this as a temporary solution

allisonchou:
> Signing off but would love to see a server fix of "RazorComponentAttribute" to "razorComponentAttribute"

Filed https://github.com/dotnet/razor/issues/9376. The fix is going to involve multi repos so the flow may take a while, so I think we can still take this fix in the meantime

dibarbet (approved):

davidwengier (approved):

PR #6506, opened 2023-10-08
Add setting to control Razor component commit behaviour

davidwengier:
Server side is https://github.com/dotnet/razor/pull/9379

Will bump Razor in this PR before merging


.vscode/launch.json
```diff
@@ -291,8 +291,8 @@
...
+              "NEW_DEPS_VERSION": "7.0.0-preview.23509.2",
```
davidwengier, line 295: @allisonchou is it cool to just update the server? or do I need to update the telemetry too? Before this change, they were on the same version.

allisonchou: I haven't tried updating just the server before so I'm not 100% sure, but I think updating solely the server should be fine as long as there are no telemetry-specific changes. I would verify there are no problems when opening a Razor file just to be sure though

allisonchou (approved):

.vscode/launch.json
davidwengier: Opening razor files works fine, and i don't see the word "telemetry" in any worrying places in the Razor log, so I'm going to assume this all works 😁

PR #6509, opened 2023-10-09
Create Razor integration test system, and formatting test

davidwengier:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1861425

Recommended definitely not looking commit-at-a-time :)

I tried to re-use as much as possible, but with relative file paths and things sometimes specified via configuration, it's not quite ideal. My gut feeling is that the duplication that is here is in files that won't need to be touched, so its an okay balance, if not perfect.


test/razorIntegrationTests/index.ts
```diff
@@ -0,0 +1,49 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as jest from 'jest';
+import { Config } from '@jest/types';
+import * as path from 'path';
+import { jestIntegrationTestProjectName } from './jest.config';
+
+async function runIntegrationTests() {
```
dibarbet, line 11: I would like to re-use the same runner that we use in normal integration tests.  This bit was particularly challenging to get right

It should be possible to export this function from one file and have run just call this.

test/razorIntegrationTests/jestSetup/vsCodeEnvironment.ts
```diff
@@ -0,0 +1,30 @@
...
+        this.global.vscode = vscode;
+    }
+
+    public async teardown() {
+        this.global.vscode = {};
+        return await super.teardown();
+    }
+}
+
+module.exports = VsCodeEnvironment;
```
dibarbet, line 30: same here - you could probably just import the other vscode environment instead of re-defining it.

test/razorIntegrationTests/runIntegrationTests.ts
```diff
@@ -0,0 +1,68 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+import { downloadAndUnzipVSCode, resolveCliArgsFromVSCodeExecutablePath, runTests } from '@vscode/test-electron';
+
+async function main() {
+    try {
```
dibarbet, line 10: same here, I think we shouldn't duplicate this

test/razorIntegrationTests/jestSetup/vsCodeEnvironment.ts
davidwengier: I tried and the relative file in the jest.config works too, so I can remove both files in this folder. Seems like that might be better, in case more mocks are needed in future?

dibarbet: yeah, removing is better if its easy

davidwengier:
Okay, I think I got all of this right now, sharing code etc.

PR #6515, opened 2023-10-10
Fix dotnet path resolution when using snap installed packages

dibarbet:
On systems installing dotnet via snap, the dotnet on the path does not point to the dotnet install location, even via symlinks.  Instead, it is symlinked to the snap executable, e.g. `/usr/bin/snap`.

So calling `which dotnet` and reading symlinks will not give us a valid dotnet installation.

The fix here is to instead parse the output of `dotnet --list-runtimes` which will return the real assembly locations of the dotnet install.  Since the dotnet install has a [known layout](https://github.com/dotnet/designs/blob/main/accepted/2020/install-locations.md#net-core-install-layout), we can easily find the dotnet executable given a runtime install path.

See https://github.com/dotnet/vscode-csharp/issues/6513 for more info.

Resolves https://github.com/dotnet/vscode-csharp/issues/6513

### DIFF
--- a/src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
+++ b/src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
@@ -12,10 +12,9 @@ import { PlatformInformation } from '../shared/platform';
 import { commonOptions } from '../shared/options';
 import { existsSync } from 'fs';
 import { CSharpExtensionId } from '../constants/csharpExtensionId';
-import { promisify } from 'util';
-import { exec } from 'child_process';
 import { getDotnetInfo } from '../shared/utils/getDotnetInfo';
-import { readFile, realpath } from 'fs/promises';
+import { readFile } from 'fs/promises';
+import { RuntimeInfo } from '../shared/utils/dotnetInfo';
 
 export const DotNetRuntimeVersion = '7.0';
 
@@ -62,7 +61,7 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
             dotnetRuntimePath = path.dirname(dotnetInfo.path);
         }
 
-        const dotnetExecutableName = this.platformInfo.isWindows() ? 'dotnet.exe' : 'dotnet';
+        const dotnetExecutableName = this.getDotnetExecutableName();
         const dotnetExecutablePath = path.join(dotnetRuntimePath, dotnetExecutableName);
         if (!existsSync(dotnetExecutablePath)) {
             throw new Error(`Cannot find dotnet path '${dotnetExecutablePath}'`);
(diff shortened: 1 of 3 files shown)


src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -151,39 +150,40 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
+            const runtimeFolderPath = matchingRuntime.Path;
+            const installFolder = path.dirname(path.dirname(runtimeFolderPath));
+            const dotnetExecutablePath = path.join(installFolder, this.getDotnetExecutableName());
+            if (!existsSync(dotnetExecutablePath)) {
+                throw new Error(
+                    `dotnet executable path does not exist: ${dotnetExecutablePath}, dotnet installation may be corrupt.`
+                );
             }
 
             this.channel.appendLine(`Using dotnet configured on PATH`);
```
jasonmalinowski, line 185: Does this message need to change? Or should it now log the path we created?

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -151,39 +150,40 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
-
-            // There could be multiple paths output from where.  Take the first since that is what we used to run dotnet --info.
-            const path = whereOutput.stdout.trim().replace(/\r/gm, '').split('\n')[0];
-            if (!existsSync(path)) {
-                throw new Error(`dotnet path does not exist: ${path}`);
+            // The .NET install layout is a well known structure on all platforms.
+            // See https://github.com/dotnet/designs/blob/main/accepted/2020/install-locations.md#net-core-install-layout
+            //
+            // Therefore we know that the runtime path is always in <install root>/shared/<runtime name>
+            // and the dotnet executable is always at <install root>/dotnet(.exe).
```
jasonmalinowski, line 172: Should we be renaming this function? (findDotnetFromPath)?

jasonmalinowski (approved):

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
dibarbet: the path will be logged in the caller anway, and we're still using the dotnet from the path, so think this is fine.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
dibarbet: I think this is still valid - this is still from the dotnet on path (via executing the `dotnet` commands)

zaballa12:
So what do I have to do to fix this problem?
?

dibarbet:
> So what do I have to do to fix this problem?
> 
> dotnet --list-runtimes Microsoft.AspNetCore.App 7.0.11 [/snap/dotnet-sdk/221/shared/Microsoft.AspNetCore.App] Microsoft.NETCore.App 7.0.11 [/snap/dotnet-sdk/221/shared/Microsoft.NETCore.App]

This change will most likely be in next weeks release which should resolve the issue.  If you want to workaround it for now, I think there are a few options.

If you're only using C# (and not C# devkit), you can workaround this issue by setting `dotnet.dotnetPath` to `/snap/dotnet-sdk/221/` in your vscode configuration.

Alternatively, making sure the `dotnet` executable at `/snap/dotnet-sdk/221/dotnet` is in your path before `/snap/bin/dotnet` should also resolve this.

Another user also mentioned running this - `ln -s /snap/bin/dotnet /usr/bin/dotnet` worked, though I haven't tried it.

PR #6538, opened 2023-10-12
Improve reliability of Razor integration test

davidwengier:
I tried to do fancy things like wait for the Razor server to be activated etc. but it was painful because on my machine when we start waiting, it has already activated so we missed the event. I was going to go further and add a way to query status, but this seems to work just as well for now, and is much easier. It's also how the inlay hints test works, so maybe there is some tribal knowledge that this is the right way to do things.

10 successful runs. That'll do :)

image

### DIFF
--- a/test/razorIntegrationTests/formatting.integration.test.ts
+++ b/test/razorIntegrationTests/formatting.integration.test.ts
@@ -32,8 +32,8 @@ jestLib.describe(`Razor Formatting ${testAssetWorkspace.description}`, function 
         const htmlConfig = vscode.workspace.getConfiguration('html');
         await htmlConfig.update('format.enable', true);
 
-        await integrationHelpers.activateCSharpExtension();
         await integrationHelpers.openFileInWorkspaceAsync(path.join('Pages', 'BadlyFormatted.razor'));
+        await integrationHelpers.activateCSharpExtension();
     });
 
     jestLib.afterAll(async () => {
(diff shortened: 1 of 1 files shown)


test/razorIntegrationTests/formatting.integration.test.ts
```diff
@@ -32,8 +32,8 @@ jestLib.describe(`Razor Formatting ${testAssetWorkspace.description}`, function
         const htmlConfig = vscode.workspace.getConfiguration('html');
         await htmlConfig.update('format.enable', true);
 
-        await integrationHelpers.activateCSharpExtension();
         await integrationHelpers.openFileInWorkspaceAsync(path.join('Pages', 'BadlyFormatted.razor'));
+        await integrationHelpers.activateCSharpExtension();
```
dibarbet, line 36: interesting... not sure why this would help.  Potentially some timing thing where the file needs to be opened for a bit to trigger something on the Razor side?  Weird

dibarbet (approved):
> I tried to do fancy things like wait for the Razor server to be activated etc. but it was painful because on my machine when we start waiting, it has already activated so we missed the event. 

Yeah if you need a specific event the server probably has to be restarted.   The only way around it is if you turn it into a promise in the extension exports itself (like initialization finished).

Not directly related to this, but I'm currently investigating if it is possible to add an LSP API to query async operation listener on the roslyn server side, so the tests can wait for various async operations on the server to complete (instead of events).  For example, processing option changes, processing file watcher changes etc.

PR #6543, opened 2023-10-12
Dedupe Build and Live Diagnostics 

beccamc:
Currently build and live diagnostics are reported independently, causing duplicate errors and errors that don't go away until a re-build. The root cause of this problem is that DevKit extension reports build diagnostics and C# extension reports live diagnostics. 

image

This change creates a separate build diagnostics collection, that filters the diagnostics reported by the build and only shows the ones that are **not** being shown by live. 

1. On build complete, receive build diagnostics from DevKit extension via a brokered service
2. Check if any of the files with build diagnostics have live diagnostics. If they have live diagnostics, display "build-only" diagnostics (e.g. errors we know that will not be caught by live). If they have no diagnostics, shown, they display all of the diagnostics reported by the build. 
3. When a file is opened, check to see if the build reported anything about it. If it did, display the build-only diagnostics. All other diagnostics should be shown by live. 

dedupe

Issue: https://github.com/dotnet/vscode-csharp/issues/5728


src/lsptoolshost/services/descriptors.ts
```diff
@@ -32,4 +32,13 @@ export default class Descriptors {
             protocolMajorVersion: 3,
         }
     );
+
+    static readonly csharpExtensionBuildResultService: ServiceRpcDescriptor = new ServiceJsonRpcDescriptor(
+        ServiceMoniker.create('Microsoft.VisualStudioCode.CSharp.BuildResultService', '0.1'),
```
beccamc, line 37: This name isn't good. Update

beccamc:
If anyone has testing thoughts, please share! I'm sure there are corner cases I am not aware of.

dibarbet:
> If anyone has testing thoughts, please share! I'm sure there are corner cases I am not aware of.

1.  Making sure behavior makes sense with FSA on/off (seems like it does from the description)
2.  Making sure the experience works if you turn FSA on or off (does that require a restart?)
3.  Is it possible to somehow indicate which errors are build only?  Perhaps we could use the diagnostic source?  To potentially alleviate confusion on why some errors go away later.

beccamc:
The source is set to "csharp-build"

image

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -93,6 +94,13 @@ export class RoslynLanguageServer {
     /** The project files previously opened; we hold onto this for the same reason as _solutionFile. */
     private _projectFiles: vscode.Uri[] = new Array<vscode.Uri>();
 
+    /** The diagnostic results from build displayed by VS Code. When live diagnostics are available for a file, these are errors that only build knows about.
+     * When live diagnostics aren't loaded for a file, then these are all of the diagnostics reported by the build.*/
+    private _diagnosticsReportedByBuild: vscode.DiagnosticCollection;
```
dibarbet, line 99: can you extract all the diagnostics stuff out of this file?  Trying to move to a world where `roslynLanguageServer.ts` doesn't really have feature code in it.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -694,6 +705,71 @@ export class RoslynLanguageServer {
...
+        );
+    }
+
+    public clearDiagnostics() {
+        this._diagnosticsReportedByBuild.clear();
+    }
+
+    public async setBuildDiagnostics(buildDiagnostics: Array<[vscode.Uri, vscode.Diagnostic[]]>) {
+        this._allBuildDiagnostics = buildDiagnostics;
+        const buildOnlyIds = await this.getBuildOnlyDiagnosticIds(CancellationToken.None);
```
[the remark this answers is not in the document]
dibarbet: interesting - can the build only diagnostics also change if the project changes?  Because then we'd have to somehow trigger this on project changes.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -694,6 +705,71 @@ export class RoslynLanguageServer {
...
+        const displayedBuildDiagnostics = new Array<[vscode.Uri, vscode.Diagnostic[]]>();
+
+        this._allBuildDiagnostics.forEach((fileDiagnostics) => {
+            const uri = fileDiagnostics[0];
+            const diagnosticList = fileDiagnostics[1];
+
+            // Check if we have live diagnostics shown for this document
+            const liveDiagnostics = vscode.languages.getDiagnostics(uri);
```
dibarbet, line 729: I'm not sure if this check is correct - it could be the case that we have no live diagnostics but we should show only build diagnostics (e.g. the errors were already fixed during the build)

Perhaps it would be better to check the state of the FSA flag - if FSA is workspace, then we know live diagnostics should win.

And of course open document diagnostics should always come from live.

dibarbet: Though this has the risk that build diagnostics don't get shown for a while (if workspace diagnostics are slow to return)

src/lsptoolshost/roslynLanguageServer.ts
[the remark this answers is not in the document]
mavasani: > Do the build only diagnostic ids change after a project is loaded

Generally they should not. Build only compiler diagnostics never change. For analyzers, we determine if a diagnostic is build-only or not based on a special custom tag attached to the analyzer Descriptor. So, the analyzer build-only IDs only change if the underlying AnalyzerReferences change, which is very unlikely to happen too often.

src/lsptoolshost/roslynLanguageServer.ts
mavasani: Hmm the de-dupe logic in the non-LSP implementation is slightly different. Once diagnostics from build are intercepted, we report the ones with build-only diagnostic ID via a build-only error list source. For the rest of them, we throw away all the currently computed/cached live diagnostics (both for open and closed documents) and replace them with the non-build only ones that came from build. These are instantly refreshed on the error list, so your error list content will exactly match the build output window content when a build finishes. Subsequently, we trigger live re-analysis of active and open documents, so in case anything has changed in the snapshot after the build was kicked, we should get refreshed diagnostics for these changed files. Closed files continue having the diagnostics from build, until they get refreshed from FSA analysis (if enabled) OR the document is opened.

The approach chosen here seems to be also fine - though we need to now handle any timing related issues, FSA state, etc. that @dibarbet mentioned

beccamc:
The experience with FSA and very large solutions isn't very good. With this change, FSA solutions don't show build errors that would be duplicates. This means, as David pointed out, you could build the solution but not see errors until FSA finishes. Manish is gathering telemetry on FSA solutions so we can understand how common this is. For now, waiting for feedback and tracking in this issue - https://github.com/dotnet/vscode-csharp/issues/6606

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,146 @@
...
+                uri,
+                BuildDiagnosticsService.filerDiagnosticsFromBuild(diagnosticList, buildOnlyIds, isDocumentOpen),
+            ]);
+        });
+
+        this._diagnosticsReportedByBuild.set(displayedBuildDiagnostics);
+    }
+
+    private compareUri(a: vscode.Uri, b: vscode.Uri): boolean {
+        return a.path.localeCompare(b.path, undefined, { sensitivity: 'accent' }) === 0;
```
beccamc, line 56: @dibarbet Did you mention something about uri comparision with different cases? Sometimes paths are "c:/users/FileName.cs" and "C:/users/FileName.cs".  It's not safe to do an ignore case compare?

src/lsptoolshost/roslynLanguageServer.ts
beccamc: Tracking this here https://github.com/dotnet/vscode-csharp/issues/6606

src/lsptoolshost/buildDiagnosticsService.ts
dibarbet: Its not safe - some file systems are case sensitive

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,146 @@
...
+            const buildDiagnostics = BuildDiagnosticsService.filerDiagnosticsFromBuild(
+                currentFileBuildDiagnostics[1],
+                buildOnlyIds,
+                true
+            );
+            this._diagnosticsReportedByBuild.set(uri, buildDiagnostics);
+        }
+    }
+
+    public static filerDiagnosticsFromBuild(
```
dibarbet, line 75: should be `filter`?

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,146 @@
...
+        diagnosticList: vscode.Diagnostic[],
+        buildOnlyIds: string[],
+        isDocumentOpen: boolean
+    ): vscode.Diagnostic[] {
+        const configuration = vscode.workspace.getConfiguration();
```
dibarbet, line 80: you should be able to use the code from https://github.com/dotnet/vscode-csharp/blob/main/src/shared/options.ts to read these instead of directly looking atthe configuration.

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+            }
+        });
+
+        return buildOnlyDiagnostics;
+    }
+
+    private static isCompilerDiagnostic(d: vscode.Diagnostic): boolean {
+        return d.code ? d.code.toString().startsWith('CS') : false;
```
dibarbet, line 136: does `d.code?.toString()...` work?  If so, just inline these methods.

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private static isAnalyzerDiagnostic(d: vscode.Diagnostic): boolean {
+        return d.code ? !d.code.toString().startsWith('CS') : false;
```
dibarbet, line 140: what is the `!` for here?

src/lsptoolshost/buildDiagnosticsService.ts
beccamc: Yes, thank you!

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,146 @@
...
+
+                    if (
+                        (isAnalyzerDiagnostic && analyzerDiagnosticScope === AnalysisSetting.None) ||
+                        (isCompilerDiagnostic && compilerDiagnosticScope === AnalysisSetting.None)
+                    ) {
+                        // If live diagnostics are completely turned off for this type, then show the build diagnostic
+                        buildOnlyDiagnostics.push(d);
+                    } else if (isDocumentOpen) {
+                        // no-op. The document is open and live diagnostis are on. This diagnostic is already being shown.
+                    } else if (
```
dibarbet, line 118: This flow is a bit hard to follow.  Wondering if we could make it a bit cleaner by doing an upfront grouping of the diagnostics collection into project system diags, build only diags, compiler diags, analyzer diags.

Then iterate each group separately with hopefully simpler logic blocks
e.g.
```
const projectSystemDiagnostics = ...
const buildOnlyDiagnostics = ...
const compilerDiagnostics = ...
const analyzerDiagnostics = ...

buildOnlyDiagnostics.push(projectSystemDiagnostics);
buildOnlyDiagnostics.push(buildOnlyDiagnostics);

if ( compiler scope is none or (scope = open doc and doc not open) {
    buildOnlyDiagnostics.push(compilerDiagnostics);
}

// same for analysis
```

src/lsptoolshost/services/buildResultReporterService.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export class BuildResultDiagnostics implements IBuildResultDiagnostics {
+    constructor(private _languageServerPromise: Promise<RoslynLanguageServer>) {}
+
+    public async buildStarted(): Promise<void> {
+        const langServer = await this._languageServerPromise;
+        langServer._buildDiagnosticService.clearDiagnostics();
+    }
+
+    public async reportBuildResult(buildDiagnostics: Array<[Uri, Diagnostic[]]>): Promise<void> {
```
dibarbet, line 25: I think I'd prefer we have an explicit type here, especially since we're using it across components

src/lsptoolshost/buildDiagnosticsService.ts
beccamc: I think you make a great point! I structured it this way so we don't loop through the diagnostics so many times, but that probably isn't a concern. If a project has millions of diagnostics, possible slowness in the problems window is probably the least of the user's problems.

src/lsptoolshost/buildDiagnosticsService.ts
beccamc: From my understanding, Analyzer diagnostics are anything that doesn't start with 'CS'

dibarbet: 🤦 I thought it was a nullable annotation for some reason.....

src/lsptoolshost/services/buildResultReporterService.ts
dibarbet: Something like Array<DiagnosticInfo> - that way we can access the uri / diagnostics by name (instead of index into the tuple)

src/lsptoolshost/buildDiagnosticsService.ts
beccamc: I didn't use that because the diagnostic can be undefined, in which case I wanted to return false.

beccamc: I could add some tests for undefind diagnostics and some other edge cases.

src/lsptoolshost/buildDiagnosticsService.ts
beccamc: Is it fine to add to [CommonOptions](https://github.com/dotnet/vscode-csharp/blob/5b856d819588b6719886fb9c6d83538c13aede1a/src/shared/options.ts#L10)? Seems a bit odd, but the best out of the categories available. I could also make a new category if that is more appropriate

dibarbet: Ah I didn't realize we didn't have the explicitly there.  I think we should add these to LanguageServerOptions (since O# reads these a little differently).

src/lsptoolshost/buildDiagnosticsService.ts
mavasani: This probably needs a more stronger regex check of 2 letters followed by 4 digits, say regex pattern `[cC][sS][0-9]{4}`. I have seen analyzers that report diagnostics with error codes such as `CSE1000`, which would pass the current check.

src/lsptoolshost/services/buildResultReporterService.ts
beccamc: Yes, I hear you on the tuple issue. What about a dictionary?? Or do you want an object containing a uri and a diagnostic list?

dibarbet: Ah dictionary here is fine as well.

beccamc: Ahh that makes a lot more sense!

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,149 @@
...
+            (compilerDiagnosticScope === AnalysisSetting.OpenFiles && !isDocumentOpen)
+        ) {
+            const compilerDiagnostics = diagnosticList.filter(
+                // Needs to be analyzer diagnostics and not already reported as "build only"
+                (d) => BuildDiagnosticsService.isCompilerDiagnostic(d) && !this.isBuildOnlyDiagnostic(buildOnlyIds, d)
+            );
+            buildDiagnosticsToDisplay.push(...compilerDiagnostics);
+        }
+
+        return buildDiagnosticsToDisplay;
```
dibarbet, line 129: much cleaner 👍

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    private static isBuildOnlyDiagnostic(buildOnlyIds: string[], d: vscode.Diagnostic): boolean {
+        return buildOnlyIds.find((b_id) => b_id === d.code) !== undefined;
+    }
+
+    private static isCompilerDiagnostic(d: vscode.Diagnostic): boolean {
+        // eslint-disable-next-line prettier/prettier
+        const regex = "[cC][sS][0-9]{4}";
```
dibarbet, line 138: what is the prettier warning here?

dibarbet (approved):
I assume this change is safe to merge even if the devkit change is not in?

src/lsptoolshost/buildDiagnosticsService.ts
beccamc: Good catch, it just had a problem with the double quote "

src/lsptoolshost/buildDiagnosticsService.ts
```diff
@@ -0,0 +1,148 @@
...
+            const analyzerDiagnostics = diagnosticList.filter(
+                // Needs to be analyzer diagnostics and not already reported as "build only"
+                (d) => BuildDiagnosticsService.isAnalyzerDiagnostic(d) && !this.isBuildOnlyDiagnostic(buildOnlyIds, d)
+            );
+            buildDiagnosticsToDisplay.push(...analyzerDiagnostics);
+        }
+
+        // Check the compiler diagnostic setting. If the setting is "none" or if the file is closed,
+        // then no live compiler diagnostics are being shown and bulid compiler diagnostics should be added.
+        // If FSA is on, then this is a no-op as FSA will report all compiler diagnostics
```
mavasani, line 117: This approach sounds good to me as a starting point. I think we should dogfood the experience to confirm if the delay between build completion and background analysis reporting live diagnostics for open files (FSA off) or closed files (FSA on) is acceptable for real world solutions with large number of open files across the solution. Especially if there are compiler errors in the files which we are deferring to background analysis to report diagnostics. Otherwise, we may have to introduce some handshake between this diagnostic source and the existing live diagnostic source such that we report all build reported diagnostics for a file here until live background analysis has analyzed that file and reported live diagnostics for it.

PR #6559, opened 2023-10-17
fix(vscode): Add dotnet executable path in list runtimes command

ccastrotrejo:
### Main Code Changes
- Add `dotnetExecutablePath` to the list runtimes command the same way it does to info command

### DIFF
--- a/src/shared/utils/getDotnetInfo.ts
+++ b/src/shared/utils/getDotnetInfo.ts
@@ -70,7 +70,8 @@ async function parseDotnetInfo(dotnetInfo: string, dotnetExecutablePath: string 
         }
 
         const runtimeVersions: { [runtime: string]: RuntimeInfo[] } = {};
-        const listRuntimes = await execChildProcess('dotnet --list-runtimes', process.cwd(), process.env);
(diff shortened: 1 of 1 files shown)


src/shared/utils/getDotnetInfo.ts
```diff
@@ -70,7 +70,8 @@ async function parseDotnetInfo(dotnetInfo: string, dotnetExecutablePath: string
         }
 
         const runtimeVersions: { [runtime: string]: RuntimeInfo[] } = {};
-        const listRuntimes = await execChildProcess('dotnet --list-runtimes', process.cwd(), process.env);
+        const command = dotnetExecutablePath ? `"${dotnetExecutablePath}"` : 'dotnet';
+        const listRuntimes = await execChildProcess(`${command} --list-runtimes`, process.cwd(), process.env);
```
dibarbet, line 74: This is most likely all correct, but give me a bit to pull this down locally and run it through an extra gamut of tests.  Changes here are always a bit dangerous.

ccastrotrejo: Thanks @dibarbet, I totally understand it and really appreciate it.

dibarbet (approved):

PR #6572, opened 2023-10-20
Add nested code actions to csharp extension

akhera99:
- Iterates through nested code actions after a top-level code action selection has been made
- Opens a quick pick for the nested code actions
- Sends either a CodeActionResolveRequest or a CodeActionFixAllResolveRequest based on the selected item
- Applies the edit

https://github.com/dotnet/vscode-csharp/assets/40616383/318a8f5f-40da-47ac-aedb-78b6f970fe71


src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+async function registerNestedResolveCodeAction(
+    languageServer: RoslynLanguageServer,
+    codeActionData: any,
+    outputChannel: vscode.OutputChannel
+) {
+    if (codeActionData) {
+        const data = <LSPAny>codeActionData;
+        const action = data.NestedCodeAction;
+
+        if (action && action.nestedActions && action.nestedActions.length > 0) {
```
JoeRobich, line 35: This could be simplified.
```suggestion
        if (action?.nestedActions?.length > 0) {
```

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+            await vscode.window
+                .showQuickPick(
+                    action.nestedActions?.map((child: { title: string }) => child.title),
+                    {
+                        placeHolder: vscode.l10n.t('Pick a nested action'),
+                        ignoreFocusOut: true,
+                    }
+                )
+                .then(async (selectedValue) => {
```
JoeRobich, line 44: Since you are in an async function there is no need to use `.then(`
```suggestion
            const selectedValue = await vscode.window
                .showQuickPick(
                    action.nestedActions?.map((child: { title: string }) => child.title),
                    {
                        placeHolder: vscode.l10n.t('Pick a nested action'),
                        ignoreFocusOut: true,
                    }
                );
     
```

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+                                        };
+
+                                        response = await languageServer.sendRequest(
+                                            CodeActionResolveRequest.type,
+                                            nestedCodeActionResolve,
+                                            token
+                                        );
+                                    }
+
+                                    if (response.edit) {
```
dibarbet, line 90: Invert this then log and return if this is undefined.
e.g.
```
if (!response.edit) {
    log error to output
    return;
}
```

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+    outputChannel: vscode.OutputChannel
+) {
+    if (codeActionData) {
+        const data = <LSPAny>codeActionData;
+        const action = data.NestedCodeAction;
+
+        if (action && action.nestedActions && action.nestedActions.length > 0) {
+            await vscode.window
+                .showQuickPick(
+                    action.nestedActions?.map((child: { title: string }) => child.title),
```
dibarbet, line 38: you can create a custom object here with the title and a data field containing the actual action.  Then on line 46 you don't have to search for the action again - you already have it in the selected item returned from the quick pick.

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+                        placeHolder: vscode.l10n.t('Pick a nested action'),
+                        ignoreFocusOut: true,
+                    }
+                )
+                .then(async (selectedValue) => {
+                    if (selectedValue) {
+                        const selectedAction = action.nestedActions?.find(
+                            (child: { title: string }) => child.title === selectedValue
+                        );
+                        if (selectedAction) {
```
dibarbet, line 49: instead of this multiple level of nesting, I would use the await as Joey suggested, then invert this and return early if there is no selected action
e.g.
```
if (!selectedAction) {
    return;
}
```

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+            async (request): Promise<void> => registerNestedResolveCodeAction(languageServer, request, outputChannel)
+        )
+    );
+}
+
+async function registerNestedResolveCodeAction(
+    languageServer: RoslynLanguageServer,
+    codeActionData: any,
+    outputChannel: vscode.OutputChannel
+) {
```
dibarbet, line 30:
```suggestion
): Promise<void> {
```

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+                            await vscode.window.withProgress(
+                                {
+                                    location: vscode.ProgressLocation.Notification,
+                                    title: vscode.l10n.t('Nested Code Action'),
+                                    cancellable: true,
+                                },
+                                async (_, token) => {
+                                    let result: string | undefined;
+                                    if (selectedAction.data.FixAllFlavors) {
+                                        result = await vscode.window.showQuickPick(selectedAction.data.FixAllFlavors, {
```
dibarbet, line 59: I think this should be removed out from the progress part

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,122 @@
...
+                        if (selectedAction) {
+                            await vscode.window.withProgress(
+                                {
+                                    location: vscode.ProgressLocation.Notification,
+                                    title: vscode.l10n.t('Nested Code Action'),
+                                    cancellable: true,
+                                },
+                                async (_, token) => {
+                                    let result: string | undefined;
+                                    if (selectedAction.data.FixAllFlavors) {
```
dibarbet, line 58: Instead of having all this branching here based on flavor, could you instead extract the code from https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/fixAllCodeAction.ts#L33 into an exported method that we just call from here if they picked an action with fix all?

And then this file only handles the nested application.

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,131 @@
...
+        const action = data.NestedCodeAction;
+
+        if (action?.nestedActions?.length > 0) {
+            const codeActionTitles: string[] = getCodeActionTitles(action.nestedActions);
+            const selectedValue = await vscode.window.showQuickPick(codeActionTitles, {
+                placeHolder: vscode.l10n.t(action.title),
+                ignoreFocusOut: true,
+            });
+            if (selectedValue) {
+                const selectedAction = retrieveSelectedAction(selectedValue, action.nestedActions);
```
dibarbet, line 42: I think we can make this much easier and entirely delete `retrieveSelectedAction`

Take a look at https://github.com/dotnet/vscode-csharp/blob/6e35fe33f77060f651e81c27450479fc847a0156/src/shared/assets.ts#L64

Basically - where you pass in `codeActionTitles` to `showQuickPick`, you would instead pass in an object that contains a `label` field and a field with the code action object itself.  Then the `selectedAction` returned will have the actual code action object already, so you don't need to search for it again.

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,131 @@
...
+                            token
+                        );
+
+                        if (!response.edit) {
+                            outputChannel.show();
+                            outputChannel.appendLine(`Failed to make an edit for completion.`);
+                            return;
+                        }
+
+                        const uriConverter: URIConverter = (value: string): vscode.Uri =>
```
dibarbet, line 77: Is it possible to use an instance of this instead?
https://github.com/dotnet/vscode-csharp/blob/6e35fe33f77060f651e81c27450479fc847a0156/src/lsptoolshost/uriConverter.ts#L9

src/lsptoolshost/nestedCodeAction.ts
```diff
@@ -0,0 +1,131 @@
...
+
+            // Concatenate multiple strings with " ->"
+            const concatenatedString = codeActionPath.slice(1).join(' -> ');
+            const fixAllString = vscode.l10n.t('Fix All: ');
+            return fixAllFlavors ? [`${fixAllString}${concatenatedString}`] : concatenatedString;
+        }
+    );
+}
+
+function retrieveSelectedAction(selectedValue: string, nestedActions: any): any {
```
dibarbet, line 119: as above, this can be deleted

src/lsptoolshost/nestedCodeAction.ts
akhera99: for what exactly?

dibarbet (approved):

PR #6600, opened 2023-10-26
[Razor] Support platform agnostic language server & telemetry

allisonchou:
Supports platform agnostic language server + telemetry for Razor in the case that the user isn't using an officially supported platform and architecture.

### DIFF
--- a/package.json
+++ b/package.json
@@ -733,6 +733,22 @@
       "integrity": "AB4D44465EA5135912737F8D943DB8DB7ADF1CC213F433188401FEE36B78A319"
     },
     {
+      "id": "Razor",
+      "description": "Razor Language Server (Platform Agnostic)",
+      "url": "https://download.visualstudio.microsoft.com/download/pr/f8b5b74b-3df3-47cc-83b1-cd1d93d1771d/e456a753103b84b87bf5f000499ce3d7/razorlanguageserver-platformagnostic-7.0.0-preview.23513.5.zip",
+      "installPath": ".razor",
+      "platforms": [
+        "neutral"
+      ],
+      "architectures": [
+        "neutral"
+      ],
+      "binaries": [
+        "./rzls"
+      ],
+      "integrity": "0CDC371C58614CBD5BAE83A0AEDF9896115A637E3AF007B7E399F5414CA66451"
+    },
+    {
       "id": "RazorOmnisharp",
       "description": "Razor Language Server for OmniSharp (Windows / x64)",
       "url": "https://download.visualstudio.microsoft.com/download/pr/aee63398-023f-48db-bba2-30162c68f0c4/8d42e62ea4051381c219b3e31bc4eced/razorlanguageserver-win-x64-7.0.0-preview.23363.1.zip",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -732,6 +732,22 @@
       ],
       "integrity": "AB4D44465EA5135912737F8D943DB8DB7ADF1CC213F433188401FEE36B78A319"
     },
+    {
+      "id": "Razor",
+      "description": "Razor Language Server (Platform Agnostic)",
+      "url": "https://download.visualstudio.microsoft.com/download/pr/f8b5b74b-3df3-47cc-83b1-cd1d93d1771d/e456a753103b84b87bf5f000499ce3d7/razorlanguageserver-platformagnostic-7.0.0-preview.23513.5.zip",
```
allisonchou, line 738: This version doesn't exactly match the version above since `23516.2` wasn't published with a platform agnostic language server. This is temporary and should be resolved whenever Razor next updates its versions.

davidwengier: as soon as https://github.com/dotnet/roslyn/pull/70568 merges we can bump razor and rosly, in this PR perhaps?

tasks/offlinePackagingTasks.ts
```diff
@@ -138,18 +138,26 @@ async function acquireRoslyn(
 }
 
 async function installRazor(packageJSON: any, platformInfo: PlatformInformation) {
-    return await installPackageJsonDependency('Razor', packageJSON, platformInfo);
+    if (!(await installPackageJsonDependency('Razor', packageJSON, platformInfo))) {
+        // Try downloading platform neutral package instead
+        const platformNeutral = new PlatformInformation('neutral', 'neutral');
```
dibarbet, line 143: I don't think we should have this fallback.  If the platform doesn't exist that we're packaging the vsix for, that is an error in setup authoring.

Fine with having fallbacks at runtime.  But not in creating the vsix.

dibarbet (changes requested):

tasks/offlinePackagingTasks.ts
allisonchou: That's fair - I updated the code to not fallback in this method

dibarbet (approved):

tasks/offlinePackagingTasks.ts
davidwengier: So when is `platformInfo` undefined here? When someone does a build and pack on a machine architecture we don't officially support? Or are we going to start producing platform agnostic builds of the whole extension?

davidwengier (approved):
I'm approving because the code looks good, but we don't want to merge this, right, because the versions are wrong? Or maybe it doens't matter because not many people are going to actually be trying to pull the platform agnostic builds?

allisonchou:
According to @dibarbet we don't actually publish the platform agnostic extension in the marketplace today, so essentially this PR is for future proofing more than anything. Due to this I think this would be fine to merge and just update the versions in the next Razor update

tasks/offlinePackagingTasks.ts
allisonchou: I think eventually once we do start publishing our language agnostic servers, this method would be changed to return undefined: https://github.com/dotnet/vscode-csharp/blob/1a8a3a9cc742d4bb13e8a791fffd2b7d4395245b/src/shared/platform.ts#L147

I made the condition undefined based on a similar Roslyn condition: https://github.com/dotnet/vscode-csharp/blob/1a8a3a9cc742d4bb13e8a791fffd2b7d4395245b/tasks/offlinePackagingTasks.ts#L122

davidwengier:
Sounds good, thanks for clarifying. Either @ryzngard will be doing an insertion today, or I'll do one on my Monday, so it won't be in this state for long anyway.

PR #6603, opened 2023-10-26
Add support for nuget restore commands

dibarbet:
Server side PR - https://github.com/dotnet/roslyn/pull/70588

restore_projects


src/lsptoolshost/restore.ts
```diff
@@ -0,0 +1,108 @@
...
+            return chooseProjectToRestore(languageServer, restoreChannel);
+        })
+    );
+    context.subscriptions.push(
+        vscode.commands.registerCommand('dotnet.restore.all', async (): Promise<void> => {
+            return restore(languageServer, restoreChannel);
+        })
+    );
+}
+async function chooseProjectToRestore(
```
jasonmalinowski, line 29: The name "choose" made me think this would be returning the project and then calling restore with it or something.

src/lsptoolshost/restore.ts
```diff
@@ -0,0 +1,108 @@
...
+    languageServer: RoslynLanguageServer,
+    restoreChannel: vscode.OutputChannel
+): Promise<void> {
+    const projects = await languageServer.sendRequest0(
+        RestorableProjects.type,
+        new vscode.CancellationTokenSource().token
```
jasonmalinowski, line 35: Is there not a better cancellation token to use for a command, or some UI that can be created for this? I don't think it's important to, as much as I am wondering out loud what the expected pattern is for things like this.

src/main.ts
```diff
@@ -82,6 +82,7 @@ export async function activate(
 
     const csharpChannel = vscode.window.createOutputChannel('C#');
     const dotnetTestChannel = vscode.window.createOutputChannel('.NET Test Log');
+    const dotnetChannel = vscode.window.createOutputChannel('.NET');
```
jasonmalinowski, line 85: I admit I would not have thought the name here would be just '.NET'. '.NET NuGet restore?" Or something similar?

jasonmalinowski (approved):

src/lsptoolshost/restore.ts
dibarbet: We could put a progress bar on this and use cancellation there.  But it didn't seem necessary in this case - the request barely does anything and immediately shows a quick pick when its done.

src/main.ts
dibarbet: I wanted to use the same channel O# did for nuget restore.  It makes sense from a 'shelling out to .net cli' perspective which is what O# uses it as.

src/lsptoolshost/restore.ts
jasonmalinowski: Yeah a progress bar does seem overkill. I wasn't sure if there was a simple pattern for commands with async work that's commonly used.

src/main.ts
jasonmalinowski: @dibarbet: I guess the Omnisharp name doesn't make sense to me either; since I wouldn't expect all dotnet commands to be dumped into the same window!

dibarbet: Well - you'll be happy to learn that there actually only is 1 command in the dotnet channel in O# - restore!  Looks like I can rename it easily enough.

src/lsptoolshost/restore.ts
dibarbet: I don't know of one unfortunately :(

src/main.ts
JoeRobich: I don't know why a new channel was created for it. My feeling is, since it is the C# extension performing the restore, why not just put it in the C# channel?

dibarbet: The C# channel is incredibly verbose if you have logging turned on.  You'll get restore output interleaved with c# server output.  IMHO separate channel is a bit better

PR #6607, opened 2023-10-29
Update razor to 7.0.0-preview.23528.1

ryzngard:


davidwengier (approved):

davidwengier:
I'm surprised and a little scared that this didn't fail the razor integration tests, at least, as I completely forgot that we needed a Roslyn bump too. In theory this is now producing `project.razor.vscode.bin` version 2, but wanting to read version 3. Unless I did something wrong in my PR 🤷‍♂️

ryzngard:
> I'm surprised and a little scared that this didn't fail the razor integration tests, at least, as I completely forgot that we needed a Roslyn bump too. In theory this is now producing `project.razor.vscode.bin` version 2, but wanting to read version 3. Unless I did something wrong in my PR 🤷‍♂️

I thought as long as we went in first it was okay? Let me manually check...

davidwengier:
On my machine, manual testing definitely fails. I expect if (when!) we had semantic tokens integration tests they would have failed.

Now that I think more about it, it's actually not surprising that formatting tests don't fail, as we don't generally delegate to Roslyn to format, we just use the Roslyn public API directly.

davidwengier:
Opened https://github.com/dotnet/vscode-csharp/pull/6608 to fix. Tag helper discovery is broken with just this PR, but that one fixes it.

PR #6608, opened 2023-10-30
Bump Roslyn to 4.9.0-1.23526.14

davidwengier:
Needed to make sure the `project.razor.vscode.bin` file produced matches the one expected by rzls in https://github.com/dotnet/vscode-csharp/pull/6607

Sadly the one integration test we have is resiliant to this scenario 🤦‍♂️
More integration tests coming soon!

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     }
   },
   "defaults": {
-    "roslyn": "4.9.0-1.23525.5",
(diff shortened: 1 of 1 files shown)


ryzngard:
Looks like the packages weren't promoted into the right stream. I pulled and restored locally, should work now (I think)

ryzngard (approved):

PR #6621, opened 2023-11-02
Report jest test results to AzDo

dibarbet:
Looks like

image


test/runIntegrationTests.ts
```diff
@@ -24,19 +24,18 @@ export async function runIntegrationTests(projectName: string) {
         verbose: true,
     } as Config.Argv;
 
-    let filter: string;
     if (process.env.TEST_FILE_FILTER) {
-        // If we have just a file, run that with runTestsByPath.
-        jestConfig.runTestsByPath = true;
+        // If we have just a file, run that with an explicit match.
```
dibarbet, line 28: None of the changes here are explicitly related to the test reporting.  But while trying to get that to work I fixed the issue where the launch.json threw out a ton of errors even when the tests passed.

Now looks like
image

jest.config.ts
```diff
@@ -14,6 +14,23 @@ const config: Config = {
...
+    reporters: [
+        'default',
+        [
+            'jest-junit',
+            {
+                outputDirectory: '<rootDir>/out/',
+                reportTestSuiteErrors: 'true',
+                // Azure DevOps does not display test suites (it ignores them entirely).
+                // So we have to put all the info in the test case name so the UI shows anything relatively useful.
+                // See https://github.com/microsoft/azure-pipelines-tasks/issues/7659
```
dibarbet, line 29: 😢

.vscode/launch.json
```diff
@@ -41,7 +41,8 @@
                 "${workspaceFolder}/**",
                 "!**/node_modules/**"
             ],
-            "preLaunchTask": "buildDev"
+            "preLaunchTask": "buildDev",
+            "internalConsoleOptions": "openOnSessionStart"
```
dibarbet, line 45: Not related to the reporting - just an improvement to open the debug output (where test results are printed) when running the integration tests

dibarbet:
There is an unfortunate issue with jest where if a afterall/beforeall causes a failure, the failure cannot be linked to a specific test.  This appears to be the case for the Razor formatting test.  This looks like
image

jasonmalinowski (approved):

davidwengier:
> There is an unfortunate issue with jest where if a afterall/beforeall causes a failure, the failure cannot be linked to a specific test. This appears to be the case for the Razor formatting test.

Seems like making the display clearly different is a good thing, since if its in afterall/beforeall that means its not my fault, right? 😛

PR #6622, opened 2023-11-02
Add a timeout for downloading razor telemetry

davidwengier:
Fixes https://github.com/dotnet/vscode-csharp/issues/6547

The issue these users are having is to do with their proxy settings in VS Code, and is happening inside where we already have error handling for the download, so it seems like node is just stuck downloading. The only recourse I could work out was a timeout, so at least we're not blocking users _forever_.

### DIFF
--- a/src/packageManager/fileDownloader.ts
+++ b/src/packageManager/fileDownloader.ts
@@ -17,18 +17,20 @@ import { NestedError } from '../nestedError';
 import { parse as parseUrl } from 'url';
 import { getProxyAgent } from './proxy';
 import { NetworkSettingsProvider } from '../networkSettings';
+import { CancellationToken } from 'vscode';
 
 export async function DownloadFile(
     description: string,
     eventStream: EventStream,
     networkSettingsProvider: NetworkSettingsProvider,
     url: string,
-    fallbackUrl?: string
+    fallbackUrl?: string,
+    token?: CancellationToken
 ): Promise<Buffer> {
     eventStream.post(new DownloadStart(description));
 
     try {
-        const buffer = await downloadFile(description, url, eventStream, networkSettingsProvider);
+        const buffer = await downloadFile(description, url, eventStream, networkSettingsProvider, token);
         eventStream.post(new DownloadSuccess(` Done!`));
         return buffer;
     } catch (primaryUrlError) {
@@ -54,7 +56,8 @@ async function downloadFile(
     description: string,
     urlString: string,
     eventStream: EventStream,
-    networkSettingsProvider: NetworkSettingsProvider
+    networkSettingsProvider: NetworkSettingsProvider,
+    token?: CancellationToken
 ): Promise<Buffer> {
     const url = parseUrl(urlString);
     const networkSettings = networkSettingsProvider();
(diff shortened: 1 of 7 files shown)


src/packageManager/fileDownloader.ts
```diff
@@ -66,6 +66,7 @@ async function downloadFile(
         agent: getProxyAgent(url, proxy, strictSSL),
         port: url.port,
         rejectUnauthorized: strictSSL,
+        timeout: 5 * 60 * 1000, // timeout is in milliseconds
```
davidwengier, line 69: I chose a long timeout, because some of us have bad internet connections, I just don't want us to wait literally forever.

src/razor/src/extension.ts
```diff
@@ -99,17 +100,27 @@ export async function activate(
...
         // Set up DevKit environment for telemetry
         if (csharpDevkitExtension) {
             await setupDevKitEnvironment(env, csharpDevkitExtension, logger);
+
+            const telemetryExtensionPath = path.join(
+                util.getExtensionPath(),
+                '.razortelemetry',
+                'Microsoft.VisualStudio.DevKit.Razor.dll'
+            );
+            if (await util.fileExists(telemetryExtensionPath)) {
```
davidwengier, line 113: Moved this all out just because of the `await`

src/razor/src/extension.ts
```diff
@@ -99,17 +100,27 @@ export async function activate(
         // Save user's DOTNET_ROOT env-var value so server can recover the user setting when needed
         env.DOTNET_ROOT_USER = process.env.DOTNET_ROOT ?? 'EMPTY';
 
+        let telemetryExtensionDllPath = '';
         // Set up DevKit environment for telemetry
         if (csharpDevkitExtension) {
             await setupDevKitEnvironment(env, csharpDevkitExtension, logger);
```
maryamariyan, line 106: (Irrelevant to this PR), I would've assumed `setupDevKitEnvironment` would be named `setupDevKitEnvironmentAsync` instead because of the `await`. Do the naming conventions not apply on ts?

From the diff here, I couldn't tell where `csharpDevkitExtension` is defined.

maryamariyan (approved):

src/packageManager/fileDownloader.ts
dibarbet: I'm not sure if even this is long enough.  There might be some extremely slow connections out there and while before they could wait, now they'll never get to a working state.  Should we make this timeout configurable?

dibarbet: I'm almost tempted to turn the download component into a VSCode progress window.  Then the user could cancel it if it was taking too long and not making any progress.

davidwengier: That sounds like a good idea too. Will have a poke around

davidwengier:
Boy this UI stuff is easier in VS Code than in VS 😛

src/razor/razorTelemetryDownloader.ts
```diff
@@ -42,17 +43,27 @@ export class RazorTelemetryDownloader {
...
-                    packagesToInstall,
-                    this.networkSettingsProvider,
-                    this.eventStream,
-                    isValidDownload
-                )
-            ) {
-                this.eventStream.post(new InstallationSuccess());
-                return true;
-            }
+            await vscode.window.withProgress(
```
dibarbet, line 46: 🎉

dibarbet (approved):

davidwengier:
Who wrote that annoying razor formatting test anyway....

PR #6626, opened 2023-11-02
Build VSIXs on multiple platforms to parallelize build

dibarbet:
Building vsixs is the longest part of CI - takes around 20 minutes.

This splits the vsix build into per platform builds.  This as the added benefit of allowing us to do custom per platform steps in the future (e.g. codesign on macos).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -132,6 +132,17 @@
         {
             "type": "node",
             "request": "launch",
+            "name": "Debug gulp task",
+            "preLaunchTask": "build",
+            "program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
+            "args": [
+                "${input:gulpTaskName}"
+            ],
+            "cwd": "${workspaceFolder}"
+        },
+        {
+            "type": "node",
+            "request": "launch",
             "name": "Update package dependencies",
             "preLaunchTask": "build",
             "program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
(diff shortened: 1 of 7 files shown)


tasks/offlinePackagingTasks.ts
```diff
@@ -33,8 +32,14 @@ import path = require('path');
...
+    vsceTarget: string;
+    rid: string;
+    platformInfo: PlatformInformation;
+}
+
 // Mapping of vsce vsix packaging target to the RID used to build the server executable
-export const platformSpecificPackages = [
+export const platformSpecificPackages: VSIXPlatformInfo[] = [
     { vsceTarget: 'win32-x64', rid: 'win-x64', platformInfo: new PlatformInformation('win32', 'x86_64') },
     { vsceTarget: 'win32-ia32', rid: 'win-x86', platformInfo: new PlatformInformation('win32', 'x86') },
```
JoeRobich, line 44: Wonder how much longer we should build a win32 package. See https://code.visualstudio.com/updates/v1_84#_windows-32bit-support-ends

dibarbet: Ah interesting - yeah let me see.

dibarbet:
Time comparison in build vsixes

Old:
image

New:
image

tasks/offlinePackagingTasks.ts
dibarbet: It's under discussion - for now will leave this PR as-is.

.vscode/launch.json
```diff
@@ -129,6 +129,17 @@
             ],
             "preLaunchTask": "buildDev"
         },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Debug gulp task",
```
dibarbet, line 135: just allows you to launch a gulp task under the debugger

arkalyanms (approved):

PR #6627, opened 2023-11-02
Fix C# Debugger telemetry

gregg-miskelly:
The C# extension was still using a `AIF-` Application Insights key, which apparently still at least somewhat works when using the extremely old version of vscode-extension-telemetry that the extension was shipping, but when VS Code tries to send events using this key, it fails due to code [like this](https://github.com/microsoft/vscode-extension-telemetry/blob/cd3176a68113395ceefdef6fe376b7107a37e172/src/node/telemetryReporter.ts#L72).

This PR fixes this by:
* Switching to the same Application Insights key used by the C++, Python and C# Dev Kit extensions
* Updating to the latest version of `extension-telemetry`
* Fixing the breaking changes to the `extension-telemetry` API
* Doing the same thing as C# Dev Kit and ensuring that the telemetry reported is disposed during shutdown.
* The new version of `extension-telemetry` was incompatible with the version of [agent-base](https://www.npmjs.com/package/agent-base) that was being pulled into this extension. 'agent-base' had a bug where it would patch http APIs in a way that didn't work correctly. To fix this, I had to update all of the packages that this extension used that depended on 'agent-base', and deal with the breaking changes in these packages.

## Testing

* [X] Verified using the 'Extension Telemetry' channel that C# extension events are still sent, and still have the extension's version
* [X] Verified that debugger events make it to Kusto (`RawEventsVSCodeExt`)
* [x] Verified that extension events make it to Kusto (`RawEventsVSCodeExt`)
* [X] Verified that http(s) proxy code was at least somewhat functional. I don't have an http proxy to fully test this code, but I at least we correctly parse the proxy URL and return the correct object.


dibarbet (approved):
> The C# extension was still using a AIF- Application Insights key, which apparently still at least somewhat works when using the extremely old version of vscode-extension-telemetry that the extension was shipping

So I've seen telemetry we directly report in the C# extension working (e.g. project load information).  Was it telemetry reported by vscode, using our key that was not working?  Do you have an example of what kind of telemetry that is?

gregg-miskelly:
> So I've seen telemetry we directly report in the C# extension working (e.g. project load information). Was it telemetry reported by vscode, using our key that was not working? Do you have an example of what kind of telemetry that is?

The way debug adapter telemetry works is that a debug adapter sends a ["category": "telemetry" output event](https://github.com/microsoft/debug-adapter-protocol/blob/1701f0c4a191d7efdb01ef7a7780d80ef74d8ba7/debugAdapterProtocol.json#L353C65-L353C74) to VS Code, and VS Code uses the `aiKey` value from package.json to send the telemetry. This broke at some point as VS Code stopped supporting `AIF` keys. I am **guessing** at some point the regular telemetry events would have also broken since these keys are apparently deprecated, but apparently not yet.

WardenGnaw (approved):

dibarbet (approved):

PR #6659, opened 2023-11-13
Remove `https://aka.ms/VSCode-CS-LaunchJson#...`

gregg-miskelly:
OptionsSchema.json referenced `https://aka.ms/VSCode-CS-LaunchJson` with a hash link inside the doc. This isn't flexible enough to redirect to https://code.visualstudio.com/docs/csharp/debugger-settings since the names of the page sections are different. So this updates to use unique aka.ms links for each section.

This also fixes `gulp generateOptionsSchema` which was broken by a previous refactoring of package.json.


dibarbet (approved):
> This also fixes gulp generateOptionsSchema which was broken by a previous refactoring of package.json.

I assume it was the re-ordering of the package.json that caused the issue?  Would it be worth having a unit test that verifies that the task still roughly works so we don't accidentally break it again?

gregg-miskelly:
> > This also fixes gulp generateOptionsSchema which was broken by a previous refactoring of package.json.
> 
> I assume it was the re-ordering of the package.json that caused the issue? Would it be worth having a unit test that verifies that the task still roughly works so we don't accidentally break it again?

It was broken by https://github.com/dotnet/vscode-csharp/pull/6652. I am not sure if it is worth having a test since:
- It would probably never be a product bug if it broke, so we are just changing when it gets fixed
- It has never broken before

If it did become something that broke repeatedly, it would probably be worth it. But at this point I am inclined to think it doesn't matter enough.

WardenGnaw (approved):

PR #6676, opened 2023-11-16
Enable automatic nuget restore on the client

dibarbet:
TODO - Requires server side update as well

See https://github.com/dotnet/roslyn/pull/70851 for details


package.json
```diff
@@ -1742,6 +1742,11 @@
             "default": null,
             "description": "Sets a path where MSBuild binary logs are written to when loading projects, to help diagnose loading errors."
           },
+          "dotnet.projects.enableAutomaticRestore": {
+            "type": "boolean",
+            "default": true,
+            "description": "Sets whether or not automatic nuget restore is enabled if the project system detects assets are missing."
```
jasonmalinowski, line 1748:
```suggestion
            "description": "Enables automatic NuGet restore if the extension detects assets are missing."
```

jasonmalinowski: Also no localization?

src/lsptoolshost/restore.ts
```diff
@@ -22,10 +28,15 @@ export function registerRestoreCommands(
     );
     context.subscriptions.push(
         vscode.commands.registerCommand('dotnet.restore.all', async (): Promise<void> => {
-            return restore(languageServer, restoreChannel);
+            return restore(languageServer, restoreChannel, [], true);
         })
     );
+
+    languageServer.registerOnRequest(UnresolvedProjectDependenciesNotification.type, async (params) => {
```
jasonmalinowski, line 35: Is this a request or a notification? Or I guess the underlying type is a void which means notification?

src/lsptoolshost/restore.ts
```diff
@@ -49,22 +60,25 @@ async function chooseProjectAndRestore(
...
-async function restore(
+export async function restore(
     languageServer: RoslynLanguageServer,
     restoreChannel: vscode.OutputChannel,
-    projectFile?: string
+    projectFiles: string[],
+    showOutput: boolean
 ): Promise<void> {
     if (_restoreInProgress) {
         vscode.window.showErrorMessage(vscode.l10n.t('Restore already in progress'));
```
jasonmalinowski, line 73: Is there a possibility we can now hit this if automatic restore is on? I'm thinking something like the user invokes restore, it starts to restore projects, and tries (but fails) to fully restore one project, call it A. It's still going and marching to restore other projects in the solution, but we saw the restore of A, and then trigger a reload which triggers a check of A's dependencies. We see it's missing, and so we trigger a restore of A _again_, but the restore of the rest of the solution is still going on...

Should we be queueing these instead in some way? That might also make sense if somebody is invoking two project restores one after the other in rapid succession.

dibarbet: This should be safe on the reload side because the restores on the client will block the project load queue until they're done.  So the reload won't be processed until the rest of the projects finish restoring (I actually hit this issue with the first iteration with notifications)

This should only get hit if someone triggers a manual restore while an automatic one is in progress

src/lsptoolshost/restore.ts
dibarbet: Its a request that returns void (but waits until completion) - there are separate APIs for notifications

.roslynDevkit/Microsoft.VisualStudio.Debugger.Contracts.dll
jasonmalinowski, line 1: I'm guessing you didn't mean to check this in? You may want to rebase to also remove from Git history.

src/lsptoolshost/restore.ts
jasonmalinowski: > This should only get hit if someone triggers a manual restore while an automatic one is in progress

Yeah that was the case that I had in mind.

.roslynDevkit/Microsoft.VisualStudio.Debugger.Contracts.dll
dibarbet: fixed!

src/lsptoolshost/restore.ts
dibarbet: Yeah it is possible to hit this with a manual restore.  I think for now though we don't need a queue - it should be a relatively uncommon case.  If its not and ends up being a bad experience for people we can come back and add it.

src/lsptoolshost/roslynProtocol.ts
```diff
@@ -152,17 +152,24 @@ export interface NamedPipeInformation {
...
 
 export interface RestorePartialResult {
     stage: string;
     message: string;
 }
 
+export interface UnresolvedProjectDependenciesParams {
+    /**
+     * The set of projects that have unresolved dependencies and require a restore.
+     */
```
jasonmalinowski, line 169: Might still need a rename or update too since we're also handling other changes.

src/lsptoolshost/roslynProtocol.ts
```diff
@@ -260,3 +267,9 @@ export namespace RestorableProjects {
     export const messageDirection: lsp.MessageDirection = lsp.MessageDirection.clientToServer;
     export const type = new lsp.RequestType0<string[], void>(method);
 }
+
+export namespace ProjectHasUnresolvedDependenciesNotification {
```
jasonmalinowski, line 271:
```suggestion
export namespace ProjectNeedsRestoreRequest {
```

jasonmalinowski: "Request" seems most critical if the API pattern distinguishes request vs. notification to avoid other confusion, but the rest seems good too?

jasonmalinowski (approved):

src/lsptoolshost/roslynProtocol.ts
dibarbet: changed to request.  if I'm changing the underlying method name, it will need a server update.  I'll do all that separately.

PR #6683, opened 2023-11-19
Use a case-insensitive find for files on Mac and Windows

davidwengier:
Fixes https://github.com/dotnet/vscode-csharp/issues/6644

The server side already does [the same thing](https://github.com/dotnet/razor/blob/main/src/Compiler/Microsoft.CodeAnalysis.Razor/src/FilePathComparer.cs#L17-L19) but we have a whole document manager here in TypeScript land, and I guess sometimes CPS uses the path in the sln file as a source of truth.

UPDATE: The razor formatting test failed in a weird spot, so I also changed how it asserts to hopefully make it more reliable.

### DIFF
--- a/src/razor/src/document/razorDocumentManager.ts
+++ b/src/razor/src/document/razorDocumentManager.ts
@@ -18,6 +18,7 @@ import { IRazorDocumentManager } from './IRazorDocumentManager';
 import { RazorDocumentChangeKind } from './razorDocumentChangeKind';
 import { createDocument } from './razorDocumentFactory';
 import { razorInitializeCommand } from '../../../lsptoolshost/razorCommands';
+import { PlatformInformation } from '../../../shared/platform';
 
 export class RazorDocumentManager implements IRazorDocumentManager {
     public roslynActivated = false;
@@ -32,7 +33,8 @@ export class RazorDocumentManager implements IRazorDocumentManager {
     constructor(
         private readonly serverClient: RazorLanguageServerClient,
         private readonly logger: RazorLogger,
-        private readonly telemetryReporter: TelemetryReporter
+        private readonly telemetryReporter: TelemetryReporter,
+        private readonly platformInfo: PlatformInformation
     ) {}
 
     public get onChange() {
@@ -146,7 +148,7 @@ export class RazorDocumentManager implements IRazorDocumentManager {
 
     private _getDocument(uri: vscode.Uri) {
         const path = getUriPath(uri);
-        let document = this.razorDocuments[path];
+        let document = this.findDocument(path);
 
         // This might happen in the case that a file is opened outside the workspace
         if (!document) {
@@ -198,7 +200,7 @@ export class RazorDocumentManager implements IRazorDocumentManager {
 
     private addDocument(uri: vscode.Uri) {
         const path = getUriPath(uri);
-        let document = this.razorDocuments[path];
+        let document = this.findDocument(path);
         if (document) {
             this.logger.logMessage(`Skipping document creation for '${path}' because it already exists.`);
             return document;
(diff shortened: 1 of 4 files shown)


test/razorIntegrationTests/formatting.integration.test.ts
```diff
@@ -59,34 +47,35 @@ describe(`Razor Formatting ${testAssetWorkspace.description}`, function () {
...
-    private string _x = "";
-
-    private void M()
-    {
-        // hi there
-    }
-}
-`)
-        );
+        expect(edits).toHaveLength(13);
```
ryzngard, line 50: I love this change, but is it related to this fix?

src/razor/src/document/razorDocumentManager.ts
```diff
@@ -219,6 +221,26 @@ export class RazorDocumentManager implements IRazorDocumentManager {
         this.notifyDocumentChange(document, RazorDocumentChangeKind.removed);
     }
 
+    private findDocument(path: string) {
+        // This method ultimately gets called from VS Code, using file system paths, or from DevKit, which
+        // can use paths as specified in the sln file. When these don't agree, on case-insensitive operating
+        // systems, we have to be careful to match things up correctly.
+
+        if (this.platformInfo.isLinux()) {
```
ryzngard, line 229: Does Roslyn have to do something similar?

src/razor/src/document/razorDocumentManager.ts
```diff
@@ -219,6 +221,26 @@ export class RazorDocumentManager implements IRazorDocumentManager {
...
+            return this.razorDocuments[path];
+        }
+
+        const lowerCasePath = path.toLowerCase();
+        const key = Object.keys(this.razorDocuments).find(
```
ryzngard, line 234: Can we store the doument path as lowercase to not have to do `toLowerCase` while searching? That seems like it could get expensive fairly quickly

src/razor/src/document/razorDocumentManager.ts
dibarbet, line 229: This will probably work most of the time, but its very possible to have either a case sensitive or insensitive file system on linux (and others).  I'm not sure if there is a way to generally know in node though, other than checking for the existence of the path on disk.

test/razorIntegrationTests/formatting.integration.test.ts
davidwengier: No, was just trying to get my builds green. Sorry :)

src/razor/src/document/razorDocumentManager.ts
davidwengier: Roslyn, like Razor, has basically the same idea: https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/FileSystem/PathUtilities.cs#L675

davidwengier: Neither Roslyn or Razor seem to do anything more sophisticated than this. See https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/FileSystem/PathUtilities.cs#L675

This low down in the code, I don't think file existence is something we can rely on either - we just want to compare two document paths. VS Code will happily keep around a document, and issue LSP requests for it, if it doesn't exist on disk. Likewise Roslyn will happily answer requests about it until it gets a didClose, etc.

src/razor/src/document/razorDocumentManager.ts
davidwengier: Looks like I can switch to a case insensitive comparison, rather than toLowering: https://github.com/dotnet/vscode-csharp/blob/main/src/razor/src/csharp/csharpProjectedDocumentContentProvider.ts#L76

I dare not poke at the bear that is "change the case of everything in the razorDocuments array" 😛

ryzngard (approved):

PR #6711, opened 2023-12-05
Update required VSCode version to account for APIs in vscode-extension-telementry

dibarbet:
Originally discussed in https://github.com/dotnet/vscode-csharp/issues/6707#issuecomment-1839655674

### DIFF
--- a/package.json
+++ b/package.json
@@ -1014,7 +1014,7 @@
     }
   ],
   "engines": {
-    "vscode": "^1.73.0"
+    "vscode": "^1.75.0"
   },
   "activationEvents": [
     "onDebugInitialConfigurations",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1023,18 +1023,7 @@
     "onDebugResolve:clr",
     "onDebugResolve:dotnet",
     "onLanguage:csharp",
-    "onLanguage:aspnetcorerazor",
```
dibarbet, line 1026: updating the version added a warning here that these were not necessary (they are automatically generated now based on the commands we list).

JoeRobich (approved):

package.json
JoeRobich: Why remove the "onLanguage:aspnetcorerazor" one? was it similarly not necessary?

dibarbet: yup!
image

gregg-miskelly (approved):

PR #6746, opened 2023-12-12
Respond to vscode installation layout changes in 1.85.0

dibarbet:


package.json
```diff
@@ -136,7 +136,7 @@
     "@typescript-eslint/eslint-plugin": "^5.61.0",
     "@typescript-eslint/parser": "^5.61.0",
     "@vscode/l10n-dev": "^0.0.30",
-    "@vscode/test-electron": "2.3.4",
+    "@vscode/test-electron": "2.3.8",
```
dibarbet, line 139: this is the fix.  Looks like the zip that the vscode test library downloads changed the location of the vscode executable / .cmd file. 

The old version of the test library then tried to launch the executable from the old path, which no longer existed.

davidwengier (approved):

arkalyanms (approved):

PR #6752, opened 2023-12-13
Support debug apps running under mono runtime using vsdbg.

thaystg:


package.json
```diff
@@ -4875,6 +4877,622 @@
...
+        "languages": [
+          "csharp"
+        ],
+        "variables": {
+          "pickProcess": "csharp.listProcess",
+          "pickRemoteProcess": "csharp.listRemoteProcess",
+          "pickRemoteDockerProcess": "csharp.listRemoteDockerProcess"
+        },
+        "configurationAttributes": {
+          "launch": {
```
WardenGnaw, line 4894: The `launch` and `attach` configurations are generated with the generateOptionsSchema script. You will want to add updating this debugger too.

https://github.com/dotnet/vscode-csharp/blob/9caecf2b984b607f3c9a22b93e50a73dfdf04415/src/tools/generateOptionsSchema.ts#L351-L358

package.json
```diff
@@ -4875,6 +4877,622 @@
...
         ]
+      },
+      {
+        "type": "monovsdbg",
+        "label": ".NET Core using Mono Runtime",
+        "hiddenWhen": "dotnet.debug.serviceBrokerAvailable",
+        "languages": [
+          "csharp"
+        ],
+        "variables": {
```
gregg-miskelly, line 4888: I don't think any of these variables make sense for mono since mono attached by port.

package.json
```diff
@@ -4875,6 +4877,622 @@
...
+          "pickRemoteProcess": "csharp.listRemoteProcess",
+          "pickRemoteDockerProcess": "csharp.listRemoteDockerProcess"
+        },
+        "configurationAttributes": {
+          "launch": {
+            "type": "object",
+            "required": [
+              "program"
+            ],
+            "properties": {
```
gregg-miskelly, line 4899: Do you expect users to ever use `monovsdbg` in a launch.json file? If not, you might not want to document the args.

package.json
gregg-miskelly: NOTE: You also need to update the generator to remove `monoDebuggerOptions` for non-mono configurations.

gregg-miskelly (dismissed):

package.json
gregg-miskelly: Looking at the implementation, if I understand correctly, a `lanunch` request is really an attach, which means I am assuming a user should never be doing this themselves, so it is probably best to not document the arguments.

package.json
gregg-miskelly: Alternatively, you could also undo the Concord-side change to LanchOptions.json, I will leave a comment in the Concord PR with more information.

package.json
```diff
@@ -4875,6 +4877,617 @@
             }
           }
         ]
+      },
+      {
+        "type": "monovsdbg",
+        "label": ".NET Core using Mono Runtime",
+        "hiddenWhen": "dotnet.debug.serviceBrokerAvailable",
```
gregg-miskelly, line 4884: >dotnet.debug.serviceBrokerAvailable [](http://example.com/codeflow?start=23&length=35)

I think you want this to be `true`

src/tools/generateOptionsSchema.ts
```diff
@@ -357,6 +357,9 @@ export function GenerateOptionsSchema() {
     packageJSON.contributes.debuggers[1].configurationAttributes.launch = schemaJSON.definitions.LaunchOptions;
     packageJSON.contributes.debuggers[1].configurationAttributes.attach = schemaJSON.definitions.AttachOptions;
 
+    // Mono
+    packageJSON.contributes.debuggers[2].configurationAttributes.launch = schemaJSON.definitions.LaunchOptions;
```
gregg-miskelly, line 361: I think instead we probably don't want to declare what options `monovsdbg` takes since we don't want people using it directly.

gregg-miskelly (commented):
Extension code changes look good, but I think we probably don't want to document the launch args for `monovsdbg` unless I am wrong and we want to allow folks to directly use it, in which case we probably want to filter out all the arguments that are only applicable for CoreCLR.

src/main.ts
```diff
@@ -425,11 +425,17 @@ function tryGetCSharpDevKitExtensionExports(csharpLogObserver: CsharpLoggerObser
     );
 }
 
+let registered = false;
```
WardenGnaw, line 428: nit: I think we want a comment that this flag is to ensure that `profferBrokeredServices` is only done once.

package.json
WardenGnaw: If users are not going to register anything in a schema, is it good enough to just register the DebugConfigurationProvider and DebugAdapterDescriptorFactory for monosvsdbg?

WardenGnaw (approved):

isadorasophia (approved):

dibarbet (approved):

PR #6770, opened 2023-12-21
Rename razor.trace to razor.server.trace, and use LogLevel values, to match Roslyn (and VS Code conventions)

davidwengier:
Fixes https://github.com/dotnet/razor/issues/9142

Razor logging used to have a setting name that didn't match the `<language>.server.trace` convention, used its own log levels, and the server ignored the setting anyway. This fixes all of that.

Goes with https://github.com/dotnet/razor/pull/9745 and will need a Razor bump before merging


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 ## Latest
 * Update Razor to 7.0.0-preview.23616.1 (PR: [#6760](https://github.com/dotnet/vscode-csharp/pull/6760)
 * Update Roslyn to 4.9.0-3.23611.3 (PR: [#6736](https://github.com/dotnet/vscode-csharp/pull/6736))
+  * Fixed issue where Razor trace level was ignored. New `razor.server.trace` setting created that is not ignored (PR: [dotnet/razor#9745](https://github.com/dotnet/razor/pull/9745))
```
JoeRobich, line 11: Should this be under `* Update Razor to...`?

```suggestion
* Update Razor to 7.0.0-preview.23616.1 (PR: [#6760](https://github.com/dotnet/vscode-csharp/pull/6760)
  * Fixed issue where Razor trace level was ignored. New `razor.server.trace` setting created that is not ignored (PR: [dotnet/razor#9745](https://github.com/dotnet/razor/pull/9745))
* Update Roslyn to 4.9.0-3.23611.3 (PR: [#6736](https://github.com/dotnet/vscode-csharp/pull/6736))
```

package.json
```diff
@@ -1751,21 +1751,20 @@
             "description": "%configuration.razor.languageServer.debug%",
             "order": 90
           },
-          "razor.trace": {
+          "razor.server.trace": {
```
JoeRobich, line 1754: Any benefit in trying to migrate this setting using something like the [migrateOptions.ts](https://github.com/dotnet/vscode-csharp/blob/main/src/shared/migrateOptions.ts)?

davidwengier: I thought about it, but figured it wasnt worth the effort since the Razor server had a bug, and didn't actually honour the old value anyway.

CHANGELOG.md
davidwengier: Whoops, yep!

Needs updating anyway to whatever the new build of Razor is that has the right change. Thanks for the catch though :)

package.json
```diff
@@ -1751,21 +1751,20 @@
...
             "enum": [
-              "Off",
-              "Messages",
-              "Verbose"
-            ],
-            "enumDescriptions": [
-              "%configuration.razor.trace.off%",
-              "%configuration.razor.trace.messages%",
-              "%configuration.razor.trace.verbose%"
+              "Trace",
```
phil-allen-msft, line 1758: Should we still have enum descriptions?

davidwengier: Personally, I don't think add any value.

package.json
```diff
@@ -1751,21 +1751,20 @@
...
+              "Debug",
+              "Information",
+              "Warning",
+              "Error",
+              "Critical",
+              "None"
             ],
-            "description": "%configuration.razor.trace%",
-            "order": 90
```
maryamariyan, line 1768: not sure what this "order" was needed for.

package.json
maryamariyan, line 1754: why is this still called trace when it is actually log level?

why is it not named `razor.server.loglevel`?

package.json
davidwengier: Me neither

package.json
davidwengier: Just following the same convention as Roslyn. I believe its a VS Code convention too.

maryamariyan: that's very confusing

davidwengier:
Ping @dotnet/razor-tooling for reviews

package.json
ryzngard: Agree it's confusing 🤷

ryzngard (approved):

package.json
dibarbet: See https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#logging-support-for-language-server

package.json
dibarbet: this ensures the Razor settings are grouped and ordered together (cc @Cosifne )

package.json
ryzngard: Does that actually change anything that vscode controls? It seems like basically another config to me.

dibarbet: If you specify it exactly as they describe there, this setting will output the full json trace logs (when enabled) to the trace output channel you pass into the vscode LSP client.

Since we added that option though for roslyn, we've modified it a bit, and now we have to manually convert our option values and manually set the trace setting on the LSP client.  We do that here - https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/roslynLanguageServer.ts#L137

davidwengier: Interesting. So Roslyn (and with this PR, Razor) have `<language>.server.trace`, but the doco says `<language>.trace.server`. So, should we be consistent with Roslyn, or with VS Code?

dibarbet: I guess it makes sense to be consistent with Roslyn (for me).  We use `dotnet.server.trace` because
1.  Our configs have a common prefix of `dotnet`
2.  Our configs for server related configurations are `dotnet.server.xyz`
3.  Its tracing configuration, so `dotnet.server.trace`

davidwengier: > Since we added that option though for roslyn, we've changed the set of values you can use, and now we have to manually convert our option values

Yep, this is exactly what this PR is doing too, so I agree it makes sense for the setting name to be consistent if the options are.

PR #6782, opened 2024-01-04
Fix command enablement by using specific activation contexts for O#, Roslyn standalone, and Roslyn devkit

dibarbet:
There were various bugs in the old enablement clauses.  For example
1.  Devkit overrides the useOmnisharp configuration setting.  If useOmnisharp was true and devkit was installed, O# commands were displayed (but would not work since devkit activated, not O#).
2.  Some of the restore commands were only enabled when using Roslyn standalone, but not using O#.

This PR modifies the enablement clauses to look at an activation context.  We explicitly set the activation context after we've run all the other logic to determine which server to actually launch.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1891108 (and other non-filed bugs)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2008,49 +2008,49 @@
         "command": "o.restart",
         "title": "%command.o.restart%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
+        "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "o.pickProjectAndStart",
         "title": "%command.o.pickProjectAndStart%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
+        "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "dotnet.openSolution",
         "title": "%command.dotnet.openSolution%",
         "category": ".NET",
-        "enablement": "!config.dotnet.server.useOmnisharp && dotnet.server.activatedStandalone"
+        "enablement": "dotnet.server.activationContext == 'Roslyn'"
       },
       {
         "command": "o.fixAll.solution",
         "title": "%command.o.fixAll.solution%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
+        "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "o.fixAll.project",
         "title": "%command.o.fixAll.project%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
+        "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "o.fixAll.document",
         "title": "%command.o.fixAll.document%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
+        "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "o.reanalyze.allProjects",
         "title": "%command.o.reanalyze.allProjects%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
+        "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "o.reanalyze.currentProject",
         "title": "%command.o.reanalyze.currentProject%",
         "category": "OmniSharp",
-        "enablement": "config.dotnet.server.useOmnisharp"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -2061,18 +2061,13 @@
...
+        "enablement": "dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp'"
       },
       {
         "command": "dotnet.restore.all",
         "title": "%command.dotnet.restore.all%",
         "category": ".NET",
-        "enablement": "dotnet.server.activatedStandalone"
-      },
-      {
-        "command": "csharp.downloadDebugger",
```
dibarbet, line 2073: as best as I could tell, this command was never registered anywhere.  cc @WardenGnaw

jasonmalinowski:
@dibarbet Does this mean though that commands aren't visible until the extension has launched? For some that's probably not a big deal, but what about the ones that might trigger the server launch in the first place?

JoeRobich (approved):

package.json
WardenGnaw: This was part of the original O# extension where users would download the extension then the dependencies would be downloaded since there was no support for platform specific downloads.

However, I think everything is bundled and platform specific this is just an artifact of the past that can be removed 😀

WardenGnaw (approved):

test/integrationTests/commandEnablement.integration.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        expect(commands).toContain('dotnet.generateAssets');
+        expect(commands).toContain('dotnet.restore.project');
+        expect(commands).toContain('dotnet.restore.all');
+        expect(commands).toContain('dotnet.test.runTestsInContext');
+        expect(commands).toContain('dotnet.test.debugTestsInContext');
+        expect(commands).toContain('csharp.listProcess');
+        expect(commands).toContain('csharp.listRemoteProcess');
+        expect(commands).toContain('csharp.listRemoteDockerProcess');
+        expect(commands).toContain('csharp.attachToProcess');
+        expect(commands).toContain('csharp.reportIssue');
```
jasonmalinowski, line 35: Since we have a very similar list in the omnisharp side of the tests, is it worth just having a shared file that list the commands that are:

1. Expected to be in both.
2. Expected to be in Roslyn only
3. Expected to be in Omnisharp only

That way if nothing else there's not a risk of updating one list but forgetting to update the other.

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -518,14 +518,19 @@ export class RoslynLanguageServer {
                 _channel.appendLine('Activating C# + C# Dev Kit...');
             }
 
+            // Set command enablement to use DevKit commands.
+            vscode.commands.executeCommand('setContext', 'dotnet.server.activationContext', 'RoslynDevKit');
```
jasonmalinowski, line 522: Does the place this is being set in this function matter specifically? i.e. this is after we've activated DevKit I think, but before we've done some other stuff? I think my request here is if order does matter, stick a comment explaining why you chose this spot.

package.json
```diff
@@ -5678,12 +5673,12 @@
       "editor/context": [
         {
           "command": "dotnet.test.runTestsInContext",
-          "when": "editorLangId == csharp && dotnet.server.activatedStandalone",
+          "when": "editorLangId == csharp && dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp'",
```
jasonmalinowski, line 5676:
```suggestion
          "when": "editorLangId == csharp && (dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp')",
```

I'm assuming this behaves like most languages in terms of operator precedence.

package.json
```diff
@@ -5678,12 +5673,12 @@
...
           "group": "2_dotnet@1"
         },
         {
           "command": "dotnet.test.debugTestsInContext",
-          "when": "editorLangId == csharp && dotnet.server.activatedStandalone",
+          "when": "editorLangId == csharp && dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp'",
```
jasonmalinowski, line 5681:
```suggestion
          "when": "editorLangId == csharp && (dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp')",
```

jasonmalinowski (approved):

src/lsptoolshost/roslynLanguageServer.ts
dibarbet: Order isn't super important - we're just setting it as soon as we know devkit is available (this is before its activated)

PR #6785, opened 2024-01-05
Add diagnostics integration tests

dibarbet:


src/lsptoolshost/onAutoInsert.ts
```diff
@@ -40,7 +40,15 @@ export function registerOnAutoInsert(languageServer: RoslynLanguageServer) {
 
             source.cancel();
```
jasonmalinowski, line 41: I don't see source being used anywhere before this point -- can we simplify this?

src/lsptoolshost/onAutoInsert.ts
```diff
@@ -40,7 +40,15 @@ export function registerOnAutoInsert(languageServer: RoslynLanguageServer) {
...
             source = new vscode.CancellationTokenSource();
-            await applyAutoInsertEdit(e, changeTrimmed, languageServer, source.token);
+            try {
+                await applyAutoInsertEdit(e, changeTrimmed, languageServer, source.token);
```
jasonmalinowski, line 44: So what (if anything) is actually going to cancel this token? Because if nothing, then how do we get that CancellationError instead?

src/lsptoolshost/onAutoInsert.ts
```diff
@@ -40,7 +40,15 @@ export function registerOnAutoInsert(languageServer: RoslynLanguageServer) {
...
+            } catch (e) {
```
jasonmalinowski, line 45: We already had an 'e' in the parent scope -- this might not be bad but could get confusing.

src/lsptoolshost/restore.ts
```diff
@@ -41,10 +41,19 @@ async function chooseProjectAndRestore(
...
+    let projects: string[];
+    try {
+        projects = await languageServer.sendRequest0(
+            RestorableProjects.type,
+            new vscode.CancellationTokenSource().token
+        );
+    } catch (e) {
+        if (e instanceof vscode.CancellationError) {
+            return;
+        }
```
jasonmalinowski, line 53: Ditto to earlier: how do we get this if we aren't passing in a cancellable token?

jasonmalinowski (approved):

src/lsptoolshost/onAutoInsert.ts
dibarbet: The vscode jsonrpc can return a cancellation error if the server shuts down for example.  So it doesn't have to be this token cancelling.

arkalyanms (approved):

PR #6806, opened 2024-01-12
Copy the BuildHost deps.json to the right place

jasonmalinowski:
This avoids having to merge https://github.com/dotnet/roslyn/pull/71597 and then do a server update, to reduce risk.


arkalyanms (approved):

jasonmalinowski:
This didn't work: it's copying for the wrong framework since the contents aren't quite right.

PR #6814, opened 2024-01-17
Improve completionComplexEdit command

etvorun:
We need to open a document if it is not opened yet when handling complex edit like adding a new event handler to MyPage.xaml.cs after selecting "add new event handler" in MyPage.xaml completion popup.

### DIFF
--- a/src/lsptoolshost/commands.ts
+++ b/src/lsptoolshost/commands.ts
@@ -103,73 +103,70 @@ async function completionComplexEdit(
     newOffset: number,
     outputChannel: vscode.OutputChannel
 ): Promise<void> {
-    let success = false;
-    const uri = UriConverter.deserialize(uriStr);
-    const editor = vscode.window.visibleTextEditors.find(
-        (editor) => editor.document.uri.path === uri.path || editor.document.uri.fsPath === uri.fsPath
+    const componentName = '[roslyn.client.completionComplexEdit]';
+
+    // Find TextDocument, opening if needed.
+    const document = await vscode.workspace.openTextDocument(uriStr);
+    if (document === undefined) {
+        outputAndThrow(outputChannel, `${componentName} Can't open document with path: '${uriStr}'`);
+    }
+
+    // Use editor if we need to deal with selection or snippets.
+    let editor: vscode.TextEditor | undefined = undefined;
+    if (isSnippetString || newOffset >= 0) {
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/commands.ts
```diff
@@ -103,73 +103,70 @@ async function completionComplexEdit(
...
+    }
 
-            if (success) {
-                break;
-            }
+    // Use editor if we need to deal with selection or snippets.
+    let editor: vscode.TextEditor | undefined = undefined;
+    if (isSnippetString || newOffset >= 0) {
+        editor = await vscode.window.showTextDocument(document);
+        if (editor == undefined) {
```
dibarbet, line 118:
```suggestion
        if (editor === undefined) {
```

src/lsptoolshost/commands.ts
```diff
@@ -103,73 +103,70 @@ async function completionComplexEdit(
...
-        outputChannel.show();
-        outputChannel.appendLine(`${componentName} ${errorMessage}`);
-
-        if (editor === undefined) {
-            outputChannel.appendLine(
-                `${componentName} Can't find visible document with uri.fsPath: '${uri.fsPath}' and uri.path: '${uri.path}'`
-            );
-
-            outputChannel.appendLine(`${componentName} URIs of all visible documents:`);
-            for (const visibleEditor of vscode.window.visibleTextEditors) {
```
dibarbet, line 160: @genlu do you think we still need to keep this logging or can it be removed?

etvorun: I found logging useless. I can remove it you also see no value here.

genlu: Those were added when we tried to figure out the uri issue that broke complex edit. If we haven't seen any such issues recently, then I'm fine removing them

etvorun: Logging removed

dibarbet: we should keep at least the logging we had when we were not able to apply the edits.  throwing an error doesn't always bubble up to a visible location.

My original comment was asking Gen about the logging of the visible editor stuff that was also removed here.  It sounds like this part is OK to remove.

Switching it back to how you originally had it and it should be good to go in.

dibarbet (changes requested):

dibarbet (approved):

PR #6819, opened 2024-01-19
Updates the CONTRIBUTING.md file

maryamariyan:
These changes should make the CONTRIBUTING.md file easier to follow for new contributors. 

- Makes instructions more clear and actionable
- Improves headers structure and adds table of contents
- Slightly updates the instructions for building, running, and testing the repository

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,51 +1,112 @@
-## Local development
+## Table of Contents
 
-### Pre-requisites
+- [Setting Up Local Development Environment](#setting-up-local-development-environment)
+  - [Prerequisites for Development](#prerequisites-for-development)
+  - [Building, Running, and Testing the Repository](#building-running-and-testing-the-repository)
+  - [Setting Up Local Language Servers](#setting-up-local-language-servers)
+    - [Roslyn](#roslyn)
+    - [Razor](#razor)
+  - [Debugging Local Language Servers](#debugging-local-language-servers)
+    - [Roslyn](#roslyn-1)
+    - [Razor](#razor-1)
+  - [Creating VSIX Packages for the Extension](#creating-vsix-packages-for-the-extension)
+- [Updating the `Roslyn` Language Server Version](#updating-the-roslyn-language-server-version)
 
-First install:
+## Setting Up Local Development Environment
+
+Setting up your local development environment for the vscode-csharp repository involves several steps. This guide will walk you through the process.
+
+### Prerequisites for Development
+
+Before you start, make sure you have the following software installed on your machine:
 
 * Node.js ([v18.17.0 LTS](https://nodejs.org/en/blog/release/v18.17.0) is recommended).
 * Npm (The version shipped with node is fine)
-* .NET 7.0 SDK (dotnet should be on your path)
+* .NET 8.0 SDK (dotnet should be on your path)
 
-### Build and run the extension
+Once you have these installed, you can proceed to clone the vscode-csharp repository:
 
-To run and develop the extension do the following:
+```bash
+git clone https://github.com/your-username/vscode-csharp.git
+cd vscode-csharp
+```
 
-* Run `npm i`
-* Run `npm i -g gulp`
-* Run `gulp installDependencies` (this will download the various dependencies as specified by the version in the [package.json](package.json))
-* Open in Visual Studio Code (`code .`)
-* _Optional:_ run `npm run watch`, make code changes
-* Press <kbd>F5</kbd> to debug
+Next, install the necessary npm packages:
 
-To **test** do the following: `npm run test` or <kbd>F5</kbd> in VS Code with the "Launch Tests" debug configuration.
+```bash
+npm install
+```
 
-### Using a locally developed Roslyn server
+Now, you're ready to start building, running, and testing the repository.
 
-https://github.com/dotnet/roslyn contains the server implementation.  Follow the instructions there to build the repo as normal.  Once built, the server DLL will be located in the build output directory, typically
+### Building, Running, and Testing the Repository
 
-`$roslynRepoRoot/artifacts/bin/Microsoft.CodeAnalysis.LanguageServer/Debug/net7.0/Microsoft.CodeAnalysis.LanguageServer.dll`
+Follow these steps to build, run, and test the repository:
 
-depending on which configuration is built.  Then, launch the extension here and change the VSCode setting `dotnet.server.path` to point to the Roslyn dll path you built above and restart the language server.
+#### Building
 
-If you need to debug the server, you can set the VSCode setting `dotnet.server.waitForDebugger` to true.  This will trigger a `Debugger.Launch()` on the server side as it starts.
+1. Run `npm i` - This command installs the project dependencies.
+2. Run `npm i -g gulp` - This command installs Gulp globally.
+3. Run `gulp installDependencies` - This command downloads the various dependencies as specified by the version in the [package.json](package.json) file.
+4. Run `code .` - This command opens the project in Visual Studio Code.
+5. Run `npm run watch` (Optional) - This command watches for code changes.
 
-### Using a locally developed Razor server
+#### Running
 
-https://github.com/dotnet/razor contains the server implementation.  Follow the instructions there to build the repo as normal.  Once built, the server will be located in the build output directory, typically
+After completing the build steps:
 
-`$razorRepoRoot/artifacts/bin/rzls/Debug/net7.0`
+1. Press <kbd>Ctrl+Shift+D</kbd> to open the Run view in VS Code and ensure `Launch Extension` is selected.
+2. Start debugging by pressing <kbd>F5</kbd>.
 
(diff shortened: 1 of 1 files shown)


maryamariyan:
@jordi1215 please take a look at the new instructions and let me know if this works for you

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
+
+1. Press <kbd>Ctrl+Shift+D</kbd> to open the Run view in VS Code and ensure `Launch Extension` is selected.
+2. Start debugging by pressing <kbd>F5</kbd>.
+
+#### Testing
+
+To run tests:
+
+1. Execute `npm run test` or press <kbd>F5</kbd> in VS Code with the "Launch Tests" debug configuration selected.
+2. For integration tests, select `Razor Integration Tests` and press <kbd>F5</kbd> to start debugging.
```
dibarbet, line 67: For integration tests it might be better to mention the two 'current file' integration tests (for roslyn and for razor). those will allow you to actually debug the test, the 'Razor integration tests' configuration does not.

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
+
+Now, you're ready to start building, running, and testing the repository.
+
+### Building, Running, and Testing the Repository
+
+Follow these steps to build, run, and test the repository:
+
+#### Building
+
+1. Run `npm i` - This command installs the project dependencies.
```
dibarbet, line 49: npm i is the same as npm install, which you mentioned above.  doesn't need to be run twice

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
+2. Run `npm i -g gulp` - This command installs Gulp globally.
+3. Run `gulp installDependencies` - This command downloads the various dependencies as specified by the version in the [package.json](package.json) file.
+4. Run `code .` - This command opens the project in Visual Studio Code.
+5. Run `npm run watch` (Optional) - This command watches for code changes.
```
dibarbet, line 53: npm run watch is only applicable if you're actually running something, so maybe move to the running section.

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
+### Setting Up Local Language Servers
+
+This section shows how to set up local Razor or Roslyn language servers for debugging with the VSCode C# extension.
+
+#### Roslyn
+
+1. Clone the [Roslyn repository](https://github.com/dotnet/roslyn). This repository contains the Roslyn server implementation.
+2. Follow the build instructions provided in the repository.
+
+The server DLL is typically at `$roslynRepoRoot/artifacts/bin/Microsoft.CodeAnalysis.LanguageServer/Debug/net8.0/Microsoft.CodeAnalysis.LanguageServer.dll`, but this may vary based on the built configuration.
```
dibarbet, line 78: currently we're still producing a net7.0 dll.  
```suggestion
The server DLL is typically at `$roslynRepoRoot/artifacts/bin/Microsoft.CodeAnalysis.LanguageServer/Debug/net7.0/Microsoft.CodeAnalysis.LanguageServer.dll`, but this may vary based on the built configuration.
```

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
+Setting up your local development environment for the vscode-csharp repository involves several steps. This guide will walk you through the process.
+
+### Prerequisites for Development
+
+Before you start, make sure you have the following software installed on your machine:
 
 * Node.js ([v18.17.0 LTS](https://nodejs.org/en/blog/release/v18.17.0) is recommended).
 * Npm (The version shipped with node is fine)
-* .NET 7.0 SDK (dotnet should be on your path)
+* .NET 8.0 SDK (dotnet should be on your path)
```
dibarbet, line 26: Technically .net8 should work, but IMHO we should stay with recommending .net7 for now since
1.  CI uses .net 7
2.  the extension uses a .net7 runtime still.

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
 
-depending on which configuration is built.  Then, launch the extension here and change the VSCode setting `razor.languageServer.directory` to point to the Razor executable path you built above and reload the window.
+After completing the configuration, follow these steps:
 
-If you need to debug the server, you can set the VSCode setting `razor.languageServer.debug` to true.  This will trigger a `Debugger.Launch()` on the server side as it starts. You can also set `razor.server.trace` to `Debug` to get more log messages in the output window
+1. Ensure the language server is fully built in Debug mode.
+2. Open the `.sln` solution file for the language server in a Visual Studio instance. Keep this instance open for use in a later step.
+3. Press `Ctrl+Shift+P` and select `Reload Window`. This ensures the changes from the configuration step are applied.
+4. After reloading, a window will pop up prompting you to select or open a Visual Studio instance. Click `Cancel` at this stage. This is because we want to debug the child VS Code instance that will open after the next step, not the current instance.
```
dibarbet, line 136: Can you clarify this?  At least for roslyn, if you set `waitForDebugger` you should only get 1 window prompt and you should attach to the first one.

dibarbet (commented):
Some nice improvements!  A few suggestions as well

CONTRIBUTING.md
maryamariyan:
```suggestion
* .NET 7.0 SDK (dotnet should be on your path)
```

CONTRIBUTING.md
davidwengier: > Can you clarify this? At least for roslyn, if you set `waitForDebugger` you should only get 1 window prompt and you should attach to the first one.

I think this is because the advice is to set up the local language servers in the main VS Code settings. I would advise against that, and only configure local language servers in the workspaces that you intend to debug, that way your VS Code installation is unaffected.

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
-To **test** do the following: `npm run test` or <kbd>F5</kbd> in VS Code with the "Launch Tests" debug configuration.
+1. Search for `dotnet server`.
+2. Set `dotnet.server.path` to the path of your Roslyn DLL.
+3. Enable `dotnet.server.waitForDebugger`.
 
-### Using a locally developed Roslyn server
+#### Razor
 
-https://github.com/dotnet/roslyn contains the server implementation.  Follow the instructions there to build the repo as normal.  Once built, the server DLL will be located in the build output directory, typically
+1. Navigate to `C:\Users\YourUsername\AppData\Roaming\Code\User\settings.json` and add the following lines:
```
davidwengier, line 110: Aside from the fact that these are Windows specific paths, I think it's better to tell people to create a settings.json in the folders they want to use for debugging VS Code, or use the "Workspace" tab in the settings UI when changing settings

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
 
 * Node.js ([v18.17.0 LTS](https://nodejs.org/en/blog/release/v18.17.0) is recommended).
 * Npm (The version shipped with node is fine)
 * .NET 7.0 SDK (dotnet should be on your path)
 
-### Build and run the extension
+Once you have these installed, you can proceed to clone the vscode-csharp repository:
+
+```bash
+git clone https://github.com/your-username/vscode-csharp.git
```
davidwengier, line 31: You can't clone from `your-username` without first forking the repo, so worth mentioning to do that from the GitHub UI first.

CONTRIBUTING.md
```diff
@@ -1,51 +1,147 @@
...
+#### Razor
+
+1. Clone the [Razor repository](https://github.com/dotnet/razor). This repository contains the Razor server implementation.
+2. Follow the build instructions provided in the repository.
+
+The server DLL is typically at `$razorRepoRoot/artifacts/bin/rzls/Debug/net8.0`.
+
+### Configuring Local Language Servers
+
+This section provides instructions on how to debug locally built Roslyn and Razor language servers. You can do this by either directly editing the `settings.json` file or through the VSCode settings interface.
```
davidwengier, line 89: The other option would be to mention that these instructions are for Windows only, both because of the paths but also the use of Visual Studio and the debugger dialog. I don't honestly know what happens on Mac or Linux when setting the `waitForDebugger` setting to true 😁

CONTRIBUTING.md
maryamariyan: I just updated the PR to used the "workspace" tab :)

CONTRIBUTING.md
maryamariyan: yeah I noticed this is all windows only

CONTRIBUTING.md
maryamariyan: I did a redo of this section, using workspace approach instead

CONTRIBUTING.md
maryamariyan: okay I just updated the windows specific paths

CONTRIBUTING.md
maryamariyan: updated the windows specific paths the newer commit

PR #6844, opened 2024-02-01
Allow rzls to launch using standard dotnet (Redo)

maryamariyan:
- Updates where to pick up dotnet.exe.
- Contributes to making rzls move away from being self-contained.
- Unifies how roslyn and razor servers acquire dotnet path/env variables.
- Picks up leftover changes from https://github.com/dotnet/vscode-csharp/pull/5855 into main

### DIFF
--- a/src/razor/src/razorLanguageServerOptionsResolver.ts
+++ b/src/razor/src/razorLanguageServerOptionsResolver.ts
@@ -7,6 +7,7 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import * as vscodeAdapter from './vscodeAdapter';
+import * as vscode from 'vscode';
 import { RazorLanguageServerOptions } from './razorLanguageServerOptions';
 import { RazorLogger } from './razorLogger';
 import { LogLevel } from './logLevel';
@@ -44,7 +45,9 @@ function findLanguageServerExecutable(withinDir: string) {
     }
 
     if (!fs.existsSync(fullPath)) {
-        throw new Error(`Could not find Razor Language Server executable within directory '${withinDir}'`);
+        throw new Error(
+            vscode.l10n.t("Could not find Razor Language Server executable within directory '{0}'", withinDir)
+        );
     }
 
     return fullPath;
--- a/src/razor/src/razorLanguageServerOptionsResolver.ts
+++ b/src/razor/src/razorLanguageServerOptionsResolver.ts
@@ -7,7 +7,6 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import * as vscodeAdapter from './vscodeAdapter';
-import * as vscode from 'vscode';
 import { RazorLanguageServerOptions } from './razorLanguageServerOptions';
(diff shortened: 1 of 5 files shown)


src/razor/src/razorLanguageServerOptionsResolver.ts
```diff
@@ -35,23 +34,17 @@ export function resolveRazorLanguageServerOptions(
...
-            throw new Error(
-                vscode.l10n.t("Could not find Razor Language Server executable within directory '{0}'", withinDir)
-            );
-        }
+        fullPath = path.join(withinDir, 'rzls.dll');
+    }
 
-        fullPath = dllPath;
+    if (!fs.existsSync(fullPath)) {
+        throw new Error(`Could not find Razor Language Server executable within directory '${withinDir}'`);
```
phil-allen-msft, line 47: The pattern in our code is to translate the errors (l10n.t); is there a reason to remove the translation of this error?

maryamariyan: good call out, I'll bring it back

src/razor/src/razorLanguageServerOptionsResolver.ts
```diff
@@ -35,23 +35,19 @@ export function resolveRazorLanguageServerOptions(
...
-                vscode.l10n.t("Could not find Razor Language Server executable within directory '{0}'", withinDir)
-            );
-        }
+        fullPath = path.join(withinDir, 'rzls.dll');
+    }
 
-        fullPath = dllPath;
+    if (!fs.existsSync(fullPath)) {
+        throw new Error(
+            vscode.l10n.t("Could not find Razor Language Server executable within directory '{0}'", withinDir)
```
davidwengier, line 49: I know the old code didn't do this either, but while you're here anyway, it occurs to me it would be much more helpful if this logged `fullPath`, so the user would also know which file it was looking for. I've hit this message before and not been sure if it was the DLL or EXE that was an issue.

davidwengier (approved):

src/razor/src/razorLanguageServerOptionsResolver.ts
```diff
@@ -35,23 +35,19 @@ export function resolveRazorLanguageServerOptions(
...
     } else {
-        // Exe doesn't exist.
-        const dllPath = path.join(withinDir, 'rzls.dll');
-
-        if (!fs.existsSync(dllPath)) {
-            throw new Error(
-                vscode.l10n.t("Could not find Razor Language Server executable within directory '{0}'", withinDir)
-            );
-        }
+        fullPath = path.join(withinDir, 'rzls.dll');
```
dibarbet, line 44: Just a note here - it is possible to publish a non-self contained executable if you wanted to
https://learn.microsoft.com/en-us/dotnet/core/deploying/#publish-framework-dependent

The benefit of the executable is that your process name is the executable name.  Whereas if you launch with `dotnet rzls.dll` your executable name will be 'dotnet'

dibarbet (approved):
Seems reasonable to me!

PR #6867, opened 2024-02-07
adding lower case razorComponentAttribute to package.json

jordi1215:
This pull request (PR) will create a transition state to fix the capitalization of a `RazorSemanticTokensLegend` named `RazorComponentAttribute` in the razor repo. (See [Razor repo PR](https://github.com/dotnet/razor/pull/9909)). The changes made in this PR will ensure that the re-named semantic token can be recognized on VS code as well.

### Summary of the changes
- replaced two entries in the `package.json` file to help recognize the lower-case version of `razorComponentAttribute`
### Testing
- tested with the updated razor repo and the coloring of razor component attribute works (see pic below)
razorComponentAttribute


davidwengier:
I don't think we need to leave the old names in here. The version of Razor being used is part of the same `package.json` file, so we can make both changes together. In other words, we can change the classifications to lowercase in the same PR that we insert a build of Razor that uses the lowercase names.

jordi1215:
> I don't think we need to leave the old names in here. The version of Razor being used is part of the same `package.json` file, so we can make both changes together. In other words, we can change the classifications to lowercase in the same PR that we insert a build of Razor that uses the lowercase names.

That sounds right! Let me make the appropriate changes then.

davidwengier:
Needs to be merged along with an insertion to build 7.0.0-preview.24168.2 or later

.vscode/launch.json
```diff
@@ -169,8 +169,8 @@
...
+              "NEW_DEPS_VERSION": "7.0.0-preview.24168.4",
```
davidwengier, line 173: This version doens't match what is in `package.json`. Though I guess it doesn't matter since package.json is internaly consistent 🤷‍♂️

davidwengier (commented):
We need to bump the telemetry packages too, as they depend on types in the language server package, so will presumably fail to load.

Also this will conflict with https://github.com/dotnet/vscode-csharp/pull/6983 :D

.vscode/launch.json
ryzngard: *shrug * idk what happened. Will look

ryzngard:
> We need to bump the telemetry packages too, as they depend on types in the language server package, so will presumably fail to load.
> 
> Also this will conflict with #6983 :D

Good call. I'll handle that tomorrow and see if I can help @jordi1215 with conflicts

.vscode/launch.json
ryzngard: as it turns out, if you have a space in the url the whole thing fails to update

davidwengier (approved):

PR #6879, opened 2024-02-12
Fix `Attribute` snippet

josebaliberal:
This PR aims to correct the `attribute` snippet. Currently the attribute type is defined with a syntax error on the second line like ```sealed class ${Class1}Attribute : System.Attribute``` 
where 'Class1.cs' is the file name.
It also makes it so it changes the attribute prefix on the class definition and the ctor at the same time


dibarbet (approved):
Thanks for the fix!

PR #6887, opened 2024-02-14
Add grammar support for new directives: rendermode, preservewhitespace and typeparam

jordi1215:
This pull request fixes dotnet/razor#9882 in the razor repository. It updates the TextMate grammar to include code coloring for three new Razor directives: rendermode, preservewhitespace, and typeparam. You can find more information about these directives in the [razor documentation](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-8.0#preservewhitespace).

### tmLanguage.yml

This file defines the TextMate grammar. I added support for the three new Razor directives, following the same format as the existing directives.

 ### tmLanguage.json

This file is automatically generated to reflect the changes in tmLanguage.yml. You can update it by running the razorTests or the `npm run compile:razorTextMate` command.

### razorTests

I added three new entries in grammarTests.test.ts and created three new test files for the new directives. These tests follow the same format as the existing tests for other Razor directives. 

### Before
before_vs_code

### After
after_vs_code_2

### DIFF
--- a/src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml
+++ b/src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml
@@ -441,7 +441,7 @@ repository:
     captures:
       1: { patterns: [ include: '#transition' ] }
       2: { name: 'keyword.control.razor.directive.preservewhitespace'}
-      3: { patterns: [ include: 'source.cs#type' ] }
+      3: { patterns: [ include: 'source.cs#boolean-literal' ] }
 
   #>>>>> @typeparam <<<<<
 
--- a/src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml
+++ b/src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml
@@ -281,6 +281,9 @@ repository:
       - include: '#section-directive'
       - include: '#layout-directive'
       - include: '#using-directive'
+      - include: '#rendermode-directive'
+      - include: '#preservewhitespace-directive'
+      - include: '#typeparam-directive'
 
   #>>>>> @code and @functions <<<<<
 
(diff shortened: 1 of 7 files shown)


src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml
```diff
@@ -420,6 +423,36 @@ repository:
...
+
+  #>>>>> @preservewhitespace <<<<<
+
+  preservewhitespace-directive:
+    name: 'meta.directive'
+    match: '(@)(preservewhitespace)\s+([^$]+)?'
+    captures:
+      1: { patterns: [ include: '#transition' ] }
+      2: { name: 'keyword.control.razor.directive.preservewhitespace'}
+      3: { patterns: [ include: 'source.cs#type' ] }
```
davidwengier, line 444: `preserveWhitespace` only allows `true` or `false` after it, not a C# type. I think the regex can probably just be updated to `(true|false)` instead of `([^$]+)`, but I don't know what classification name to give. Might need to look in the C# grammar and see what it uses

test/razorTests/Microsoft.AspNetCore.Razor.VSCode.Grammar.Test/tests/typeparamDirective.ts
```diff
@@ -0,0 +1,33 @@
...
+        it('No type', async () => {
+            await assertMatchesSnapshot('@typeparam');
+        });
+
+        it('No type spaced', async () => {
+            await assertMatchesSnapshot('@typeparam              ');
+        });
+
+        it('Incomplete type, generic', async () => {
+            await assertMatchesSnapshot('@typeparam SomeViewBase<string');
```
davidwengier, line 22: Great job, thinking to test generics with this! Maybe for completeness, a test that includes the closing `>` would be good?

davidwengier (commented):
> addresses an https://github.com/dotnet/razor/issues/9882 in the razor repository

GitHub Pro-tip, if you include the word "fixes" or "closes" immediately before an issue number, then GitHub will automatically close the issue when the PR merges.

src/razor/syntaxes/aspnetcorerazor.tmLanguage.yml
jordi1215: Updating the regex might not be what we want. since TextMate will color the whole line once a match is found. I kept the same regex so that the `@preserveWhitespace` can be colored as soon as the user type a space bar after the keyword, but I updated the last matching pattern for the line to be a `boolean-literal`, so it will change colors if it sees a `true` or a `false`. For visualization please see the updated picture in the PR description. Thanks!

davidwengier: Even better! Great work.

davidwengier (approved):

test/razorTests/Microsoft.AspNetCore.Razor.VSCode.Grammar.Test/tests/typeparamDirective.ts
jordi1215: Thanks for the tip! Although I can't take the credit for that since I merely modeled the test cases after the other test cases in the test suite.

Issue #6893, opened 2024-02-16
Incorrect language strings for Razor Language Server Trace configuration


rsandbach:
When #6770 was merged, the translation files may have been overlooked. The `Shared` output in vscode is reporting:

```
0000-00-00 00:00:00.000 [warning] [csharp]: Couldn't find message for key configuration.razor.server.trace.
```

_package.json_ was updated with the new resource string:
https://github.com/dotnet/vscode-csharp/blob/bb2134804be15ae64b715cc7a29b06eed82cdbc5/package.json#L1767

_package.nls.json_ still has the old key:
https://github.com/dotnet/vscode-csharp/blob/bb2134804be15ae64b715cc7a29b06eed82cdbc5/package.nls.json#L75

[These](https://github.com/search?q=repo%3Adotnet%2Fvscode-csharp+configuration.razor.trace&type=code) translation files still have the wrong key. I can throw together a pull request if desired, but am not sure if anything __else__ needs to be updated.

dibarbet:
Resolved by https://github.com/dotnet/vscode-csharp/pull/6894, thanks!

PR #6894, opened 2024-02-16
Update message keys for configuration.razor.trace

rsandbach:
PR for #6893


rsandbach:
@dotnet-policy-service agree

davidwengier:
@Cosifne Was I wrong to think these should have automatically been updated by the localization process, as would happen with `.xlf` files in VS land? Is there some other part of the process I missed to clean up localization?

Cosifne:
@davidwengier  the key here is auto-generated by vscode-l10n, not sure why it's out-off sync originally, let me check

rsandbach:
> if possible could you check if the config is still grouped with other razor configs?

I think this is what you are asking. There are other Razor settings in the LSP Server group, but they are not all together as one. 

image

Razor > Server is not grouped with Razor > Language Server

image

When editing the settings JSON it also fails to translate the descriptions of the settings:
image

Out of curiosity, is there a difference between Razor > Server and Razor > Language Server?

davidwengier:
> Out of curiosity, is there a difference between Razor > Server and Razor > Language Server?

Naming is hard :)

davidwengier (approved):
Thanks for the PR! I took the liberty of pushing a change to clean up the unused strings, for each trace value. Thanks also for noticing the issue with the code action handler!

PR #6933, opened 2024-02-26
Undo temporary 'using' keyword hack

alexgav:
Removing temporary hack to add "using" keyword to completion since we have a better change to add it in both VS and VSCode (see https://github.com/dotnet/razor/pull/9982). 

For the most part it's an undo of prior PR https://github.com/dotnet/vscode-csharp/pull/6441, but keeping one minor change from there that adjust completion item kind which is off by one with VS, this one

https://github.com/dotnet/vscode-csharp/blob/7e16947e68cb7a70177d3afba1d4e5116574259b/src/razor/src/completion/razorCompletionItemProvider.ts#L91C1-L93C18

which seems like a good change to keep.


davidwengier (approved):
👍

We should do the corresponding Razor insertion in this branch before merging.

davidwengier:
Inserted a new Razor build into this PR, which should also fix https://github.com/dotnet/razor/pull/10014 / https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1977817

PR #6939, opened 2024-03-02
Add option for formatting code block open braces

davidwengier:
VS Code option for https://github.com/dotnet/razor/pull/10018


package.json
```diff
@@ -1978,6 +1978,12 @@
             "default": true,
             "description": "Enable/disable default Razor formatter."
           },
+          "razor.format.codeBlockBraceOnNextLine": {
+            "type": "boolean",
+            "scope": "window",
+            "default": false,
+            "description": "Forces the open brace after an @code or @functions directive to be on the following line."
```
Cosifne, line 1985: Do you expect this to be localized?

Cosifne (approved):
Fine with the change to be Eng only right now. Seems other options are also not localized. So it might be good to do that in another PR.

ryzngard (approved):

davidwengier:
> Fine with the change to be Eng only right now. Seems other options are also not localized. So it might be good to do that in another PR.

Thanks for the heads up, I just copied the existing pattern and didn't really think about it. Logged https://github.com/dotnet/razor/issues/10025 to fix this in another PR.

PR #7004, opened 2024-03-26
Insertion PR for roslyn

maryamariyan:
Needed for https://github.com/dotnet/vscode-csharp/pull/6960#issuecomment-2019103906

cc @ryzngard @davidwengier @dibarbet

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     }
   },
   "defaults": {
-    "roslyn": "4.10.0-3.24168.7",
+    "roslyn": "4.10.0-3.24168.8",
     "omniSharp": "1.39.11",
     "razor": "7.0.0-preview.24168.4",
     "razorOmnisharp": "7.0.0-preview.23363.1",


davidwengier:
Did we really only miss it by one build? That seems terribly unlucky 😁

My only real comment is that we should probably just cherry-pick this commit into the other PR, so it all goes at once. No need to do two PRs I don't think?

ryzngard:
> Did we really only miss it by one build? That seems terribly unlucky 😁
> 
> My only real comment is that we should probably just cherry-pick this commit into the other PR, so it all goes at once. No need to do two PRs I don't think?

Yes we did 😢

dibarbet:
> My only real comment is that we should probably just cherry-pick this commit into the other PR, so it all goes at once. No need to do two PRs I don't think?

Agree with this - fine to cherry-pick this change into your previous PR that adds the flag.  Also make sure to run `gulp updateRoslynVersion` locally to update the new package versions in the nuget feed

maryamariyan:
sounds good, I'll close this and cherry pick on the other pr

PR #7007, opened 2024-03-29
Add github actions for issue health

arkalyanms:


.github/workflows/close-stale-needs-more-info.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        
+      - name: Close Stale Issues with Needs More Info Label and Comment
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const labelName = 'needs-more-info';
```
dibarbet, line 25: we have two needs more info labels.  Fine with using this one, but we should transition issues under the old one to it and delete the old one 
image

.github/workflows/stale.yml
```diff
@@ -0,0 +1,76 @@
...
+          "Debugger": "@wardengnaw",
+          "Debugger-ExpressionEvaluation": "@wardengnaw",
+          "Debugger-FixedPendingInsertion": "@wardengnaw",
+          "Debugger-Install": "@wardengnaw",
+          "Debugger-Poachable": "@wardengnaw",
+          "Documentation": "@mikadumont",
+          "Project System": "@jasonmalinowski",
+          "Question": "@mikadumont",
+          "Razor": "@phil-allen-msft",
+          "Unity": "@jbevain"
```
dibarbet, line 26: At some point it might be useful to create (or re-use) a few different github groups here.  For example we could have a debugger set, razor set, unity set, etc.  Then we can live update the groups without needing to modify this.
Totally fine with this as v1 though.

dibarbet (commented):
otherwise seems reasonable!

.github/workflows/close-stale-needs-more-info.yml
arkalyanms: All our new labels seem to be of the space format. Do we want to keep that and delete the hyphenated version? The hyphenated has 0 issues under it anyway

dibarbet: Fine with either.  easiest is probably just to use the one with spaces since we on't have to move labels

PR #7010, opened 2024-03-29
Test more versions of .NET using docker containers

dibarbet:
Adds CI runs for net6/7/8 on Linux.

Windows is possible in the future, but we likely need to create custom containers for it - the dotnet ones are based on windows server nano which doesn't seem to include powershell, a hard requirement for windows Azdo containers.  Encountered this issue without it - https://github.com/microsoft/azure-pipelines-tasks/issues/11448

### DIFF
--- a/test/integrationTests/testAssets/slnWithCsproj/test/test.csproj
+++ b/test/integrationTests/testAssets/slnWithCsproj/test/test.csproj
@@ -1,11 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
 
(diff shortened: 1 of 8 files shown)


test/integrationTests/testAssets/slnWithCsproj/test/test.csproj
```diff
@@ -1,11 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
ryzngard, line 4: nit: Should this be in a props file somewhere just to keep in sync? Now that the synchronization between them actually matters

dibarbet: good call, I can do that.

ryzngard (approved):
As much as I can read yml and understand it, this looks correct to me

PR #7011, opened 2024-03-29
Update roslyn

dibarbet:
Updates Roslyn to https://dnceng.visualstudio.com/internal/_build/results?buildId=2418271&view=results

Also adds support for the obsolete classification in the default themes:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 ## Latest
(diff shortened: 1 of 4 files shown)


arkalyanms (approved):

dibarbet:
> Just updated to latest bits, but this doesn't work for me. Is there a setting one should use?

@lonix1 if you are not using the vs2019 dark or light themes that we ship with the extension, you may have to add the following to your settings.json:
```
"editor.semanticTokenColorCustomizations": {
    "rules": {
        "*.deprecated": {
            "strikethrough": true
        }
    },
}
```

Potentially we can consider adding a `configurationDefaults` to our package.json so it would automatically apply to any theme being used.

PR #7037, opened 2024-04-10
Gracefully handle failures activating the intellicode extension

dibarbet:
Partially resolves https://github.com/dotnet/vscode-csharp/issues/7031

A recent issue is causing the intellicode extension to fail to activate (see MicrosoftDocs/intellicode/issues/552).  This is a problem for us because if intellicode activation throws, we will fail to start the C# language server and all C# language features are lost.  

However, it's totally possible for the C# server to function normally without intellicode, it'll just be missing the intellicode provided completion features.  To that end, this PR modifies the activation code to gracefully handle failures in intellicode activation by showing an info box and then continuing on without it.

The info box (with optional button to hide it)
intellicode_failure


Cosifne (approved):

l10n/bundle.l10n.json
```diff
@@ -135,6 +135,9 @@
   "Choose": "Choose",
   "Choose and set default": "Choose and set default",
   "Do not load any": "Do not load any",
+  "Intellicode completions will not be provided, {0} failed to activate.": "Intellicode completions will not be provided, {0} failed to activate.",
```
vivlimmsft, line 138:
```suggestion
  "IntelliCode features will not be available, {0} failed to activate.": "IntelliCode features will not be available, {0} failed to activate.",
```
Slightly tweaking this to both match the `IntelliCode` casing and be less specific about what features won't be available.

Theoretically we should be able to still provide starred completions even with the current problem, as those are two parallel features with no shared dependencies, but our activation code currently explodes if either fails.

vivlimmsft (approved):
LGTM. This change makes a lot of sense to me given the _many_ ways this cross-extension dependency can break, and continuing to provide C# functionality in those cases seems like a major improvement to me.

333fred (approved):

PR #7039, opened 2024-04-11
Package xaml content

dibarbet:
Packages components that are required to run inside the Roslyn language server to support experimental xaml tools support for C# Dev Kit.  These are only used if c# devkit is enabled.


package.json
```diff
@@ -41,7 +41,8 @@
     "omniSharp": "1.39.11",
     "razor": "7.0.0-preview.24178.4",
     "razorOmnisharp": "7.0.0-preview.23363.1",
-    "razorTelemetry": "7.0.0-preview.24178.4"
+    "razorTelemetry": "7.0.0-preview.24178.4",
+    "xamlDesignTools": "17.11.34810.23"
```
dibarbet, line 45: TODO - naming?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -819,6 +819,14 @@ export class RoslynLanguageServer {
         args.push('--devKitDependencyPath', devKitDepsPath);
 
         args.push('--sessionId', getSessionId());
+
+        // Also include the Xaml Dev Kit extensions
+        const xamlBasePath = path.join(clientRoot, '..', '.xamlDesignTools', 'lib', 'netstandard2.0');
```
dibarbet, line 824: TODO - Allow local overrides for these extension paths (to allow locally built binaries to replace the ones shipped in the extension).

dibarbet: TODO - Unify path constants between package download and reading them in the extension.

BretJohnson: @dibarbet - can you say how you're thinking local overrides for the extension paths, for locally built binary binaries, would work - especially if you have any existing conventions there

dibarbet: @BretJohnson I don't yet - the only overrides we have are for the roslyn and razor serverdlls directly (e.g. `dotnet.server.path`).  But that currently doesn't allow you to override the roslynDevkit path for instnace.
Some options I'm thinking of
1.  Look too see if the same dll name was specified in the extension paths setting, and using that path instead if set.
2.  Have some kind of `dotnet.server.componentOverrides` setting, perhaps like
```
dotnet.server.componentOverrides: {
    "roslynDevkit": <path to roslyndevkit drop>
    "xamlDesignTools": <path to xaml design tools drop>
}
```

It would be definitely configured in a vscode setting (similar to dotnet.server.path) though.

dibarbet: this is what I ended up doing
```
"dotnet.server.componentPaths": {
    "roslynDevKit": "",
    "xamlDesignTools": "C:\\Users\\dabarbet\\Downloads\\.xamlDesignTools"
},
```

image

package.json
```diff
@@ -1666,6 +1682,12 @@
             "default": null,
             "description": "%configuration.dotnet.server.crashDumpPath%"
           },
+          "dotnet.server.enableXamlIntellisense": {
```
etvorun, line 1685: How about enableXamlTools

package.nls.json
```diff
@@ -26,11 +26,15 @@
...
   "configuration.dotnet.dotnetPath": "Specifies the path to a dotnet installation directory to use instead of the default system one. This only influences the dotnet installation to use for hosting the language server itself. Example: \"/home/username/mycustomdotnetdirectory\".",
   "configuration.dotnet.server.path": "Specifies the absolute path to the server (LSP or O#) executable. When left empty the version pinned to the C# Extension is used. (Previously `omnisharp.path`)",
+  "configuration.dotnet.server.componentPaths": "Allows overriding the folder path for built in components of the language server (for example, override the .roslynDevKit path in the extension directory to use locally built components)",
+  "configuration.dotnet.server.componentPaths.roslynDevKit": "Overrides the folder path for the .roslynDevKit component of the language server",
+  "configuration.dotnet.server.componentPaths.xamlDesignTools": "Overrides the folder path for the .xamlDesignTools component of the language server",
```
etvorun, line 31: xamlTools instead of xamlDesignTools

etvorun (approved):

package.json
BretJohnson: I think we ended up with `xamlTools`

src/lsptoolshost/roslynLanguageServer.ts
BretJohnson: Probably this should get renamed to `xamlTools` too. (And thanks for sharing this, so I can copy some of the code.)

package.json
BretJohnson: I think we ended up with `dotnet.enableXamlTools` and I'm adding a matching `dotnet.enableMobileHotReload`.

package.json
```diff
@@ -41,7 +41,8 @@
     "omniSharp": "1.39.11",
     "razor": "7.0.0-preview.24178.4",
     "razorOmnisharp": "7.0.0-preview.23363.1",
-    "razorTelemetry": "7.0.0-preview.24178.4"
+    "razorTelemetry": "7.0.0-preview.24178.4",
+    "xamlTools": "17.11.34815.8"
```
etvorun, line 45: If not too late, can you please change it to 17.11.34816.20. It includes first working end-to-end XAML Hot Reload.

akhera99 (approved):

PR #7051, opened 2024-04-15
Textmate: fix Razor expression classification inside of HTML tags

jordi1215:
- added textmate grammar to both razor and html (test) files so that we can correctly label razor components that are within html tags. We will need to update the html grammar used in razor as well
- updated tests to reflect the desire behavior, note that we changed some of the classification from existing tests as they were not correctly labeling razor grammar in html tags either

### Before 
before_vs_code

### After
after_vs_code

### DIFF
--- a/src/razor/syntaxes/aspnetcorerazor.tmLanguage.json
+++ b/src/razor/syntaxes/aspnetcorerazor.tmLanguage.json
@@ -64,6 +64,16 @@
         }
       ]
     },
+    "expression": {
+      "patterns": [
+        {
+          "include": "#explicit-razor-expression"
+        },
+        {
+          "include": "#implicit-expression"
+        }
+      ]
+    },
     "escaped-transition": {
       "name": "constant.character.escape.razor.transition",
       "match": "@@"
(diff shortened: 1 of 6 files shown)


src/razor/syntaxes/aspnetcorerazor.tmLanguage.json
```diff
@@ -64,6 +64,16 @@
         }
       ]
     },
+    "expression": {
```
davidwengier, line 67: Nit: Perhaps a more specific name would be good here, just in case. Even "implicit-or-explicit-expression" would be good, and less overloaded than just "expression"

davidwengier (approved):
As discussed, I think this is good as a workaround for VS, but would prefer a better solution where semantic classification is considered.

src/razor/syntaxes/aspnetcorerazor.tmLanguage.json
jordi1215: That makes sense! I will push out another commit for that

PR #7055, opened 2024-04-17
Textmate: fix Razor expression classification inside of HTML tags using injection

jordi1215:
This fixes some problem created in dotnet/vscode-csharp#7051. In this new approach, we use the "injection" feature of the textmate grammar so that we don't have to change the grammar of html.


davidwengier (approved):
Amazing!

GlacierFox:
Noticed a changed when writing code in Blazor and just popping in to say awesome job, works great 👍

davidwengier:
Thanks for confirming, and the kind words @GlacierFox 

FYI @jordi1215 😄

PR #7061, opened 2024-04-20
Merge main into prerelease

dibarbet:


arkalyanms (approved):

azure-pipelines/loc.yml
```diff
@@ -32,10 +32,10 @@ variables:
 - group: OneLocBuildVariables
 
 extends:
-  template: v1/1ES.Unofficial.PipelineTemplate.yml@1ESPipelineTemplates
```
arkalyanms, line 35: @Cosifne or @dibarbet just this would be get into release branch.

dibarbet: @arkalyanms the pipeline only ever runs against the main branch, so it seems fine to let it flow in our normal release schedule

PR #7074, opened 2024-04-27
Merge main into prerelease

dibarbet:


dibarbet:
override merging to unblock snap.  simple merge with no changes that aren't already validated in main.

PR #7082, opened 2024-04-30
List solution filter files (.slnf) in the 'Open Solution' command.

JoeRobich:
- Fixes sorting of solutions in the workspace root path.

### DIFF
--- a/src/lsptoolshost/commands.ts
+++ b/src/lsptoolshost/commands.ts
@@ -178,7 +178,7 @@ async function openSolution(languageServer: RoslynLanguageServer): Promise<vscod
         return undefined;
     }
 
-    const solutionFiles = await vscode.workspace.findFiles('**/*.sln');
+    const solutionFiles = await vscode.workspace.findFiles('**/*.{sln,slnf}');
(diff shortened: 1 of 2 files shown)


src/lsptoolshost/commands.ts
```diff
@@ -178,7 +178,7 @@ async function openSolution(languageServer: RoslynLanguageServer): Promise<vscod
         return undefined;
     }
 
-    const solutionFiles = await vscode.workspace.findFiles('**/*.sln');
+    const solutionFiles = await vscode.workspace.findFiles('**/*.{sln,slnf}');
```
dibarbet, line 181: Does the server side actually work for this?  Specifically I don't know if the msbuild API we're calling parses the sln file
https://github.com/dotnet/roslyn/blob/main/src/Workspaces/Core/MSBuild.BuildHost/BuildHost.cs#L159

JoeRobich: This does work. Verified with local testing and reviewing the MSBuild API itself. Looks like we could simplify some MSBuildWorkspace handling of Solution Filters by using this same api.

dibarbet (approved):

PR #7112, opened 2024-05-08
Merge feature/xamlTools into main

dibarbet:


src/shared/options.ts
```diff
@@ -398,6 +400,12 @@ class LanguageServerOptionsImpl implements LanguageServerOptions {
     public get compilerDiagnosticScope() {
         return readOption<string>('dotnet.backgroundAnalysis.compilerDiagnosticsScope', 'openFiles');
     }
+    public get componentPaths() {
+        return readOption<{ [key: string]: string }>('dotnet.server.componentPaths', {});
+    }
+    public get enableXamlToolsPreview() {
+        return readOption<boolean>('dotnet.enableXamlToolsPreview', false);
```
mgoertz-msft, line 407: Isn't this supposed to be true now?

arkalyanms (approved):

mgoertz-msft (approved):
Besides the default setting for the xamlTools feature flag everything looks good to me.

src/shared/options.ts
dibarbet: ah, its true in the package.json, but I'll update it here too.

PR #7137, opened 2024-05-17
Bump xamlTools to latest signed release (17.11.34917.22)

mgoertz-msft:
I verified the following:
- onTypeFormat no longer causes exceptions when typing in XAML
- Format Document/Selection works as expected
- onAutoInsert no longer requires a space after the tag name to work
- completion item documentation tooltips are back


dibarbet (commented):
Change seems fine, but needs to be retargeted to prerelease

PR #7138, opened 2024-05-17
Remove xamlToolsPreview switch

mgoertz-msft:


package.json
```diff
@@ -1719,12 +1719,6 @@
             "default": null,
             "description": "%configuration.dotnet.server.crashDumpPath%"
           },
-          "dotnet.enableXamlToolsPreview": {
```
dibarbet, line 1722: Lets keep the option, but change the default

dibarbet (changes requested):

dibarbet:
Also, this needs to target the prerelease branch if its going into release next week.

package.json
mgoertz-msft: The default is already `true`. The idea (and plan) is to remove the option altogether for release.

mgoertz-msft:
Ugh, I guess it isn't as simple as changing the target branch to prerelease

PR #7220, opened 2024-06-10
Adding changes to support use monovsdbg to debug wasm apps.

thaystg:
Adding changes to support debug wasm apps using monovsdbg.

It's a draft becauls I'm still waiting for VSWebAssemblyBridge publication to consume it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1412,6 +1412,11 @@
             "description": "%generateOptionsSchema.expressionEvaluationOptions.showRawValues.description%",
             "default": false
           },
+          "csharp.mono.debug.useVSDbg": {
+            "type": "boolean",
+            "description": "%generateOptionsSchema.useVSDbg.description%",
+            "default": false
+          },
           "dotnet.unitTestDebuggingOptions": {
             "type": "object",
             "description": "%configuration.dotnet.unitTestDebuggingOptions%",
@@ -4817,6 +4822,11 @@
               "env": {
                 "type": "object",
                 "description": "Environment variables passed to dotnet. Only valid for hosted apps."
+              },
+              "useVSDbg": {
+                "type": "boolean",
+                "default": false,
+                "description": "%generateOptionsSchema.useVSDbg.description%"
               },
               "dotNetConfig": {
                 "description": "Options passed to the underlying .NET debugger. For more info, see https://github.com/dotnet/vscode-csharp/blob/main/debugger.md.",
--- a/package.nls.json
+++ b/package.nls.json
@@ -435,5 +435,6 @@
     "comment": [
       "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
     ]
-  }
+  },
+  "generateOptionsSchema.useVSDbg.description": "Enable new .NET 8+ Mono Debugger (preview)"
 }
--- a/src/omnisharp/protocol.ts
+++ b/src/omnisharp/protocol.ts
@@ -343,6 +343,7 @@ export interface MSBuildProject {
     IsWebProject: boolean;
     IsBlazorWebAssemblyStandalone: boolean;
     IsBlazorWebAssemblyHosted: boolean;
+    IsWebAssemblyProject: boolean;
 }
 
 export interface TargetFramework {
--- a/src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
+++ b/src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
@@ -10,10 +10,13 @@ import { fileURLToPath } from 'url';
 import * as vscode from 'vscode';
 import { ChromeBrowserFinder, EdgeBrowserFinder } from '@vscode/js-debug-browsers';
 import { RazorLogger } from '../razorLogger';
-import { JS_DEBUG_NAME, SERVER_APP_NAME } from './constants';
+import { ONLY_JS_DEBUG_NAME, MANAGED_DEBUG_NAME, JS_DEBUG_NAME, SERVER_APP_NAME } from './constants';
 import { onDidTerminateDebugSession } from './terminateDebugHandler';
 import showInformationMessage from '../../../shared/observers/utils/showInformationMessage';
 import showErrorMessage from '../../../observers/utils/showErrorMessage';
+import path = require('path');
+import * as cp from 'child_process';
+import { getExtensionPath } from '../../../common';
 
 export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
     private static readonly autoDetectUserNotice: string = vscode.l10n.t(
@@ -21,6 +24,8 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
     );
     private static readonly edgeBrowserType: string = 'msedge';
     private static readonly chromeBrowserType: string = 'chrome';
+    private static readonly pidsByUrl = new Map<string, number | undefined>();
+    private static readonly vsWebAssemblyBridgeOutputChannel = vscode.window.createOutputChannel('VsWebAssemblyBridge');
 
     constructor(private readonly logger: RazorLogger, private readonly vscodeType: typeof vscode) {}
 
@@ -97,6 +102,7 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
             launchBrowser: {
                 enabled: false,
             },
+            cascadeTerminateToConfigurations: [ONLY_JS_DEBUG_NAME, MANAGED_DEBUG_NAME, JS_DEBUG_NAME],
             ...configuration.dotNetConfig,
         };
 
@@ -125,6 +131,13 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
         inspectUri: string,
         url: string
     ) {
+        const wasmConfig = vscode.workspace.getConfiguration('csharp');
+        const useVSDbg = configuration.useVSDbg || wasmConfig.get<boolean>('mono.debug.useVSDbg') == true;
+        let portBrowserDebug = -1;
+        if (useVSDbg) {
+            [inspectUri, portBrowserDebug] = await this.attachToAppOnBrowser(folder, configuration);
+        }
+
         const configBrowser = configuration.browser;
         const browserType =
             configBrowser === 'edge'
@@ -137,7 +150,7 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
         }
 
         const browser = {
-            name: JS_DEBUG_NAME,
+            name: useVSDbg ? ONLY_JS_DEBUG_NAME : JS_DEBUG_NAME,
             type: browserType,
             request: 'launch',
             timeout: configuration.timeout || 30000,
@@ -149,8 +162,12 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
             ...configuration.browserConfig,
             // When the browser debugging session is stopped, propogate
             // this and terminate the debugging session of the Blazor dev server.
-            cascadeTerminateToConfigurations: [SERVER_APP_NAME],
+            cascadeTerminateToConfigurations: [SERVER_APP_NAME, MANAGED_DEBUG_NAME],
         };
+
+        if (useVSDbg) {
+            browser.port = portBrowserDebug;
+        }
 
         try {
             /**
@@ -177,6 +194,150 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
                 }
             });
         }
+    }
+
+    private async attachToAppOnBrowser(
+        folder: vscode.WorkspaceFolder | undefined,
+        configuration: vscode.DebugConfiguration
+    ): Promise<[string, number]> {
+        const [inspectUriRet, portICorDebug, portBrowserDebug] =
+            await BlazorDebugConfigurationProvider.launchVsWebAssemblyBridge(configuration.url);
+
+        const cwd = configuration.cwd || '${workspaceFolder}';
+        const args = configuration.hosted ? [] : ['run'];
+        const app = {
+            name: MANAGED_DEBUG_NAME,
+            type: 'monovsdbg',
+            request: 'launch',
+            //program: 'C:\\diag\\icordebug_wasm_vscode\\test\\obj\\Debug\\net9.0\\test.dll',
+            args,
+            cwd,
+            cascadeTerminateToConfigurations: [ONLY_JS_DEBUG_NAME, SERVER_APP_NAME, JS_DEBUG_NAME],
+            ...configuration.dotNetConfig,
+        };
+
+        app.monoDebuggerOptions = {
+            ip: '127.0.0.1',
+            port: portICorDebug,
+            platform: 'browser',
+            isServer: true,
+            //assetsPath: 'C:\\diag\\icordebug_wasm_vscode\\test\\bin\\Debug\\net9.0\\',
+        };
+
+        try {
+            await this.vscodeType.debug.startDebugging(folder, app);
+            const terminate = this.vscodeType.debug.onDidTerminateDebugSession(async (event) => {
+                if (process.platform !== 'win32') {
+                    const blazorDevServer = 'blazor-devserver\\.dll';
+                    const dir = folder && folder.uri && folder.uri.fsPath;
+                    const regexEscapedDir = dir!.toLowerCase()!.replace(/\//g, '\\/');
+                    const launchedApp = configuration.hosted
+                        ? app.program
+                        : `${regexEscapedDir}.*${blazorDevServer}|${blazorDevServer}.*${regexEscapedDir}`;
+                    await onDidTerminateDebugSession(event, this.logger, launchedApp);
+                    terminate.dispose();
+                }
+                this.vscodeType.debug.stopDebugging();
+            });
+        } catch (error) {
+            this.logger.logError('[DEBUGGER] Error when launching application: ', error as Error);
+        }
+        return [inspectUriRet, portBrowserDebug];
+    }
+
+    private static async launchVsWebAssemblyBridge(urlStr: string): Promise<[string, number, number]> {
+        const dotnetPath = process.platform === 'win32' ? 'dotnet.exe' : 'dotnet';
+        const vsWebAssemblyBridge = path.join(
+            getExtensionPath(),
+            '.vswebassemblybridge',
+            'Microsoft.Diagnostics.BrowserDebugHost.dll'
+        );
+        const devToolsUrl = `http://localhost:0`; // Browser debugging port address
+        const spawnedProxyArgs = [
+            `${vsWebAssemblyBridge}`,
+            '--DevToolsUrl',
+            `${devToolsUrl}`,
+            '--UseVsDbg',
+            'true',
+            '--iCorDebugPort',
+            '-1',
+            '--OwnerPid',
+            `${process.pid}`,
+        ];
+        const spawnedProxy = cp.spawn(dotnetPath, spawnedProxyArgs);
+
+        try {
+            let chunksProcessed = 0;
+            let proxyICorDebugPort = -1;
+            let proxyBrowserPort = -1;
+            for await (const output of spawnedProxy.stdout) {
+                // If we haven't found the URL in the first ten chunks processed
+                // then bail out.
+                if (chunksProcessed++ > 10) {
+                    return ['', -1, -1];
+                }
+                BlazorDebugConfigurationProvider.vsWebAssemblyBridgeOutputChannel.appendLine(output);
+                // The debug proxy server outputs the port it is listening on in the
+                // standard output of the launched application. We need to pass this URL
+                // back to the debugger so we extract the URL from stdout using a regex.
+                // The debug proxy will not exit until killed via the `killDebugProxy`
+                // method so parsing stdout is necessary to extract the URL.
+                const matchExprProxyUrl = 'Now listening on: (?<url>.*)';
+                const matchExprICorDebugPort = 'Listening iCorDebug on: (?<port>.*)';
+                const matchExprBrowserPort = 'Waiting for browser on: (?<port>.*)';
+                const foundProxyUrl = `${output}`.match(matchExprProxyUrl);
+                const foundICorDebugPort = `${output}`.match(matchExprICorDebugPort);
+                const foundBrowserPort = `${output}`.match(matchExprBrowserPort);
+                const proxyUrlString = foundProxyUrl?.groups?.url;
+                if (foundICorDebugPort?.groups?.port != undefined) {
+                    proxyICorDebugPort = Number(foundICorDebugPort?.groups?.port);
+                }
+                if (foundBrowserPort?.groups?.port != undefined) {
+                    proxyBrowserPort = Number(foundBrowserPort?.groups?.port);
+                }
+                if (proxyUrlString) {
+                    BlazorDebugConfigurationProvider.vsWebAssemblyBridgeOutputChannel.appendLine(
+                        `Debugging proxy is running at: ${proxyUrlString}`
+                    );
+                    const oldPid = BlazorDebugConfigurationProvider.pidsByUrl.get(urlStr);
+                    if (oldPid != undefined) {
+                        process.kill(oldPid);
+                    }
+                    BlazorDebugConfigurationProvider.pidsByUrl.set(urlStr, spawnedProxy.pid);
+                    const url = new URL(proxyUrlString);
+                    return [
+                        `${url.protocol.replace(`http`, `ws`)}//${url.host}{browserInspectUriPath}`,
+                        proxyICorDebugPort,
+                        proxyBrowserPort,
+                    ];
+                }
+            }
+
+            for await (const error of spawnedProxy.stderr) {
+                BlazorDebugConfigurationProvider.vsWebAssemblyBridgeOutputChannel.appendLine(`ERROR: ${error}`);
+            }
+        } catch (error: any) {
+            if (spawnedProxy.pid) {
+                BlazorDebugConfigurationProvider.vsWebAssemblyBridgeOutputChannel.appendLine(
+                    `Error occured while spawning debug proxy. Terminating debug proxy server.`
+                );
+                process.kill(spawnedProxy.pid);
+            }
+            throw error;
+        }
+        return ['', -1, -1];
+    }
+
+    public static async tryToUseVSDbgForMono(urlStr: string): Promise<[string, number, number]> {
+        const wasmConfig = vscode.workspace.getConfiguration('csharp');
+        const useVSDbg = wasmConfig.get<boolean>('mono.debug.useVSDbg') == true;
+        if (useVSDbg) {
+            const [inspectUri, portICorDebug, portBrowserDebug] =
+                await BlazorDebugConfigurationProvider.launchVsWebAssemblyBridge(urlStr);
+
+            return [inspectUri, portICorDebug, portBrowserDebug];
+        }
+        return ['', -1, -1];
     }
 
     public static async determineBrowserType(): Promise<string | undefined> {
--- a/src/shared/assets.ts
+++ b/src/shared/assets.ts
@@ -344,6 +344,18 @@ export class AssetGenerator {
 
         return result;
     }
+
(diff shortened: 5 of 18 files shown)


package.json
```diff
@@ -4818,6 +4823,11 @@
                 "type": "object",
                 "description": "Environment variables passed to dotnet. Only valid for hosted apps."
               },
+              "useVSDbg": {
+                "type": "boolean",
+                "default": false,
+                "description": "%generateOptionsSchema.useVSDbg.description%"
```
gregg-miskelly: >generateOptionsSchema [](http://example.com/codeflow?start=33&length=21)

Do you want this to be a launch configuration attribute, or just a preference? If you want it to be a launch configuration attribute, then you need modify the [launch-option generator](https://github.com/dotnet/vscode-csharp/blob/main/src/tools/generateOptionsSchema.ts) to add this, as I am almost certain when you re-run it, this change will get blown away. Also, I am guessing maybe you want a user-preference for this also, and that would happen automatically if you did that. #Closed

package.nls.json
```diff
@@ -435,5 +435,6 @@
     "comment": [
       "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
     ]
-  }
+  },
+  "generateOptionsSchema.useVSDbg.description": "Enable new .NET 8+ Mono Debugger (preview)"
```
gregg-miskelly: >Mono [](http://example.com/codeflow?start=68&length=4)

In VS Code, will this option apply to anything besides WASM? (since the MAUI scenarios are already on vsdbg right? Or is that still optional?) #Closed

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -125,6 +131,13 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
         inspectUri: string,
         url: string
     ) {
+        const wasmConfig = vscode.workspace.getConfiguration('csharp');
+        const useVSDbg = configuration.useVSDbg || wasmConfig.get<boolean>('mono.debug.useVSDbg') == true;
```
gregg-miskelly: >wasmConfig.get<boolean>('mono.debug.useVSDbg') == true [](http://example.com/codeflow?start=51&length=54)

Do we need to do **something** to better handle pre-.NET 9 projects? (Assuming the new architecture doesn't handle them)

Two possible versions of something that occur to me:
* Move the user preference to C# Dev Kit, and only apply it there since the C# extension doesn't normally have a lot of information on the target framework. We could leave the debug configuration option here, but change the description.
* Change this configuration provider so it does check the target framework #Closed

src/shared/utils.ts
```diff
@@ -43,6 +43,14 @@ export function isWebProject(projectPath: string): boolean {
     return projectFileText.toLowerCase().indexOf('sdk="microsoft.net.sdk.web"') >= 0;
 }
 
+export function isWebAssemblyProject(projectPath: string): boolean {
+    const projectFileText = fs.readFileSync(projectPath, 'utf8');
```
gregg-miskelly, line 47: >const projectFileText = fs.readFileSync(projectPath, 'utf8'); [](http://example.com/codeflow?start=4&length=61)

Instead of reading the file a second time, I would suggest changing `isWebProject` to `getProjectFlags` and returning both.

src/shared/assets.ts
```diff
@@ -344,6 +344,18 @@ export class AssetGenerator {
 
         return result;
     }
+
+    public getAssetsPathAndProgram(): [string, string] {
```
gregg-miskelly, line 348: >() [](http://example.com/codeflow?start=34&length=2)

Shouldn't this require the path to the project instead of looping over all project?

thaystg: We need to loop all the projects, usually we are launching the server project but we need the assetsPath of the client project which is the webassembly one.

gregg-miskelly: Can we do something to avoid this? At least having an option, since a given workspace could have multiple projects and I am assuming we don't want to pick the wrong one.

package.nls.json
thaystg: For MAUI the default is enabled and the option is in their extension. I would like to unify it and use only this option in the future. But it's okay, maybe this is very complex, I will change mono to wasm here.

package.json
thaystg: I will remove it and keep only the useVSDbg option in the preferences.

image

src/shared/assets.ts
thaystg: Done, we will continue passing the assets path but if for some reason the MVID of a loaded module is different from the one loaded from assets path we will load the metadata information from debugger protocol and ignore assets path.

src/shared/utils.ts
thaystg: done

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: Checking it in two different places, one if has c# dev kit installed using the QueryExecutionService, and the other one is basically reading the project file and trying to detect if c# dev kit is not installed. 
If for some reason I can't get the framework it will not use vsdbg for wasm.

thaystg:
@gregg-miskelly can you please review again?
I think I addressed all your comments.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -179,6 +262,220 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                    const url = new URL(proxyUrlString);
+                    return [
+                        `${url.protocol.replace(`http`, `ws`)}//${url.host}{browserInspectUriPath}`,
+                        proxyICorDebugPort,
+                        proxyBrowserPort,
+                    ];
+                }
+            }
+
+            for await (const error of spawnedProxy.stderr) {
```
gregg-miskelly, line 376: I am assuming that once we return the port, this code will not run anymore? Should we instead structure this function in such a way that we continue to monitor stderr + process exit until we are ready for the debug session to end?

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -179,6 +262,220 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                                activeConfig.BuildProperties[0].Value == 'net7.0' ||
+                                activeConfig.BuildProperties[0].Value == 'net8.0'
+                            ) {
+                                isNet9OrNewer = false;
+                            }
+                        });
+                    });
+                    return isNet9OrNewer;
+                }
+            }
```
gregg-miskelly, line 438: Do you need some sort of `else throw devkit-not-ready` here in case you can't get the proxy yet, but dev kit is installed?

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -179,6 +262,220 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+        } catch (err) {
+            throw new Error('Exception while talking to proxy: ' + err);
+        } finally {
+            proxy?.dispose();
+        }
+        return isNet9OrNewer;
+    }
+    private static async useVSDbg(): Promise<boolean> {
+        const wasmConfig = vscode.workspace.getConfiguration('csharp');
+        const useVSDbg = wasmConfig.get<boolean>('wasm.debug.useVSDbg') == true;
```
gregg-miskelly, line 448: Early out if `useVSDbg` is false?

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -179,6 +262,220 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+        const proxy = await getCSharpDevKit()?.exports.serviceBroker.getProxy<IQueryExecutionService>(
+            Descriptors.projectQueryExecutionService
+        );
+        try {
+            if (proxy) {
+                const result = await proxy.ExecuteQueryAsync(queryString);
+                const queryResult = JSON.parse(result) as QueryResult<Project>;
+                if (queryResult && queryResult.Items) {
+                    isNet9OrNewer = true;
+                    queryResult.Items.forEach((project) => {
```
gregg-miskelly, line 424: Can't we check the project we are trying to launch instead of every project?

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -179,6 +262,220 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                        project.ActiveConfigurations?.forEach((activeConfig) => {
+                            if (
+                                activeConfig.BuildProperties[0].Value == 'net5.0' ||
```
gregg-miskelly, line 427: Instead of a list of configurations which aren't .NET 9, shouldn't we use a regex to test if this is a `net<ver>` configuration (with anything other then a 'net<ver>' configuration saying "no"), and extract the version number, then use semver (or some simple hand rolled code to decide if that is '9.0' or newer.

src/shared/assets.ts
```diff
@@ -344,6 +344,33 @@ export class AssetGenerator {
...
+        return [assetsPath, this.executableProjects[0].outputPath];
+    }
+
+    public isDotNet9OrNewer(): boolean {
+        let ret = false;
+        this.executableProjects.forEach((project) => {
+            if (project.isWebAssemblyProject) {
+                const projectFileText = fs.readFileSync(project.projectPath, 'utf8');
+                ret =
+                    projectFileText.toLowerCase().indexOf('net8.0') < 0 &&
```
gregg-miskelly, line 366: I feel like we should be using an XML parser, or at least a regex to try and find `<TargetFramework>` or `<TargetFramworks>` here rather than just searching blindly in the entire document for these strings.

gregg-miskelly, line 366: Similar comment here about not just looking for the names of various frameworks that we don't support but instead looking for ones we do support.

src/shared/assets.ts
```diff
@@ -344,6 +344,33 @@ export class AssetGenerator {
...
+                assetsPath += ';';
+            }
+        });
+        assetsPath = assetsPath.slice(0, -1);
+        return [assetsPath, this.executableProjects[0].outputPath];
+    }
+
+    public isDotNet9OrNewer(): boolean {
+        let ret = false;
+        this.executableProjects.forEach((project) => {
```
gregg-miskelly, line 362: Same comment here about taking a path to a project instead of looking at all of them in the entire workspace.

src/shared/utils.ts
```diff
@@ -35,12 +35,15 @@ export function findNetStandardTargetFramework(tfmShortNames: string[]): string
...
 
-export function isWebProject(projectPath: string): boolean {
+export function isWebProject(projectPath: string): [boolean, boolean] {
     const projectFileText = fs.readFileSync(projectPath, 'utf8');
 
     // Assume that this is an MSBuild project. In that case, look for the 'Sdk="Microsoft.NET.Sdk.Web"' attribute.
     // TODO: Have OmniSharp provide the list of SDKs used by a project and check that list instead.
-    return projectFileText.toLowerCase().indexOf('sdk="microsoft.net.sdk.web"') >= 0;
+    return [
+        projectFileText.toLowerCase().indexOf('sdk="microsoft.net.sdk.web"') >= 0,
```
gregg-miskelly: nit: Save `projectFileText.toLowerCase()` into a variable so it only needs to be done once. #Closed

gregg-miskelly (changes requested):

src/shared/assets.ts
thaystg: I was trying to avoid need to change it everytime that we ship a new product? like when we ship .net 10 we will need to add it here?

thaystg: Ah okay, ignore me, a regex will avoid it. > 9 should be valid.

gregg-miskelly: I wonder if it would be worth just asking the language service team to return this, as they already have it.

gregg-miskelly: Roslyn code for this: https://github.com/search?q=repo%3Adotnet%2Froslyn+targetsDotnetCore&type=code

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -149,9 +214,27 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
             // this and terminate the debugging session of the Blazor dev server.
-            cascadeTerminateToConfigurations: [SERVER_APP_NAME],
+            cascadeTerminateToConfigurations: [SERVER_APP_NAME, MANAGED_DEBUG_NAME],
         };
 
+        if (useVSDbg) {
+            browser.port = portBrowserDebug;
+        }
+
+        const monovsdbgSession = debugSessionTracker.getDebugSessionByName('monovsdbg');
```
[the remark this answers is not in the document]
thaystg: Is this what you mean?
image

Here is the name of the debug session that the user will see:
https://github.com/dotnet/vscode-csharp/blob/e2802c1c99127104f8509af022a146d5e56c4605/src/razor/src/blazorDebug/constants.ts#L9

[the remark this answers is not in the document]
gregg-miskelly: Yup, that sounds reasonable. Am I miss understanding something that `getDebugSessionByName` is looking for a session with the name `monovsdbg`? If so, where does that come from?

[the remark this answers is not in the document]
thaystg: Ah, it's looking for a debug session with this type, the name of the function is very bad, I will fix it.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: Done.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: Done.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: done.

src/shared/utils.ts
thaystg: done.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: If the newer devkit is installed and we try to call tryToUseVSDbgForMono this function will not exist in C# extension yet and this is addressed on devkit side. if the function exists the vswebassembly will also exists.

Am I understanding the question wrongly?

gregg-miskelly: What I mean is -- this code will do:

```
if (proxy) {
    ....
}
// no else here, so the code will fall through and return `false`
```

I think instead the code should be doing something like:
```
if (!proxy) {
   throw new Error("Unable to obtain required service from C# Dev Kit.");
}
```

To avoid possible races if we try to start debugging and C# Dev Kit hasn't yet initalized.

thaystg: Thanks for the explanation.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
gregg-miskelly: Ideally yes: At least in general, I don't think it is an uncommon pattern for folks to only bother upgrading only projects in active development, and some folks like to make rather large solutions (and probably workspaces in VS Code, though I have less data there). So, I wouldn't be surprised to see a workspace that has lots of different projects with different versions of .NET. The tooling also encourages that pattern since one needs to start hand-modifying project files to come up with good options to keep all projects on the same target framework.

thaystg: Okay! When we are launching an app without having the assets generated, using the C# devkit I really have the project path that is being launched and I can only look at this project. But when we don't have C# devkit or we have the launch.json generated I don't have this information and then I will loop in all the webassembly projects, does it make sense for you?

thaystg:
@gregg-miskelly I think I'm done again. If you can review again it will be great.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -179,6 +263,228 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+        const proxy = await getCSharpDevKit()?.exports.serviceBroker.getProxy<IQueryExecutionService>(
+            Descriptors.projectQueryExecutionService
+        );
+        if (!proxy) {
+            throw new Error('Unable to obtain required service from C# Dev Kit.');
+        }
+        try {
+            const result = await proxy.ExecuteQueryAsync(queryString);
+            const queryResult = JSON.parse(result) as QueryResult<Project>;
+            const pattern = /net(\d+\.\d+)/;
```
gregg-miskelly: Should this be `/^net(\d+\.\d+)$/`? Or maybe `\b` instead of `$` if you want to support one of the `-` platform things. #Closed

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -125,6 +184,13 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
         inspectUri: string,
         url: string
     ) {
+        const originalInspectUri = inspectUri;
+        const useVSDbg = await BlazorDebugConfigurationProvider.useVSDbg('');
```
gregg-miskelly: Should this try to workout the path to the project using `configuration.cwd` if it is set like [here](https://github.com/thaystg/vscode-csharp/blob/4e7091703de13a06ec5abb2eb61fcafcbe24fea2/src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts#L102)? #Closed

gregg-miskelly (approved):
Otherwise LGTM

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: done.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: done.

gregg-miskelly (approved):

WardenGnaw:
Did you want to update a case for the updatePackageDependencies script?

https://github.com/dotnet/vscode-csharp/blob/602cd792949d79d86333fa5cadad0aa1f3b8eb08/src/tools/updatePackageDependencies.ts#L276

thaystg:
> Did you want to update a case for the updatePackageDependencies script?
> 
> https://github.com/dotnet/vscode-csharp/blob/602cd792949d79d86333fa5cadad0aa1f3b8eb08/src/tools/updatePackageDependencies.ts#L276

I'm using npm run updatePackageDependencies to update the webassembly package but I didn't need to change this file.

JoeRobich:
@thaystg Is this still a change we want to consider?

thaystg:
Correct, we will work on it yet, but I will mark it as a draft for now.

thaystg:
Can I get a review again here?

package.json
```diff
@@ -707,6 +707,23 @@
       "binaries": [
         "./rzls"
       ]
+    },
+    {
+      "id": "VSWebAssemblyBridge",
```
gregg-miskelly, line 702: >id [](http://example.com/codeflow?start=7&length=2)

Should this be added [here](https://github.com/dotnet/vscode-csharp/blob/4cb5e00211cfabee4b3693f8ba6371038b070e5e/src/tools/updatePackageDependencies.ts#L273) so you can use the update dependencies script?

src/shared/assets.ts
```diff
@@ -345,6 +345,49 @@ export class AssetGenerator {
...
+    public getAssetsPathAndProgram(): [string, string] {
+        let assetsPath = ``;
+        this.executableProjects.forEach((project) => {
+            if (project.isWebAssemblyProject) {
+                assetsPath += path.join(path.dirname(project.outputPath), path.sep);
+                assetsPath += ';';
+            }
+        });
+        assetsPath = assetsPath.slice(0, -1);
+        return [assetsPath, this.executableProjects[0].outputPath];
```
gregg-miskelly, line 361: >this.executableProjects[0].outputPath [](http://example.com/codeflow?start=28&length=37)

I am not quite following what this is doing -- we are returning the assetsPath of all the web assembly projects, but then the output path of whatever random project happens to be the first executable project in the solution.

src/shared/assets.ts
```diff
@@ -345,6 +345,49 @@ export class AssetGenerator {
...
+    }
+
+    public isDotNet9OrNewer(): boolean {
```
gregg-miskelly, line 375: >isDotNet9OrNewer [](http://example.com/codeflow?start=11&length=16)

Shouldn't we combine this together with `getAssetsPathAndProgram` so we only return assets for wasm projects that target .NET 9? #Closed

src/shared/utils.ts
```diff
@@ -35,12 +35,15 @@ export function findNetStandardTargetFramework(tfmShortNames: string[]): string
     return tfmShortNames.find((tf) => tf.startsWith('netstandard'));
 }
 
-export function isWebProject(projectPath: string): boolean {
-    const projectFileText = fs.readFileSync(projectPath, 'utf8');
+export function isWebProject(projectPath: string): [boolean, boolean] {
```
gregg-miskelly, line 66: We now have two ways to detect a Blazor project. Should a project that is a Blazor project always have both? Or is one of them just going to return true for a subset of blazor projects? #Closed

src/shared/assets.ts
thaystg: The program name is basically not used, is a requirement for concord, but we don't use it for wasm, that is why I'm getting the first one.
The assets path is used to load the metadata, this one is important.

macel94:
can we get some news here?
i am one of the 2 that actually use blazor wasm in wsl2 and would like to be able to debug :D

src/shared/utils.ts
thaystg: Correct, not all the Blazor projects will have both.

src/shared/assets.ts
thaystg: done

package.json
thaystg: done

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -184,6 +273,228 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+            ip: '127.0.0.1',
+            port: portICorDebug,
+            platform: 'browser',
+            isServer: true,
+        };
+
+        try {
+            await this.vscodeType.debug.startDebugging(folder, app);
+            const terminate = this.vscodeType.debug.onDidTerminateDebugSession(async (event) => {
+                if (process.platform !== 'win32') {
```
gregg-miskelly, line 358: >if ( [](http://example.com/codeflow?start=16&length=4)

Do you need a check that the terminating session is the one you are interested in?

src/omnisharp/protocol.ts
```diff
@@ -343,6 +343,7 @@ export interface MSBuildProject {
     IsWebProject: boolean;
     IsBlazorWebAssemblyStandalone: boolean;
     IsBlazorWebAssemblyHosted: boolean;
+    IsWebAssemblyProject: boolean;
```
gregg-miskelly, line 346: I don't this you want this modification -- I think this would imply that OmniSharp is returning this new property, but I am assuming you didn't add it? #Closed

gregg-miskelly: Oh, I see

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -184,6 +273,228 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+                this.vscodeType.debug.stopDebugging();
+            });
+        } catch (error) {
+            this.logger.logError('[DEBUGGER] Error when launching application: ', error as Error);
+        }
+        return [inspectUriRet, portBrowserDebug];
+    }
+
+    private static async launchVsWebAssemblyBridge(urlStr: string): Promise<[string, number, number]> {
+        const dotnetPath = process.platform === 'win32' ? 'dotnet.exe' : 'dotnet';
```
gregg-miskelly: >const dotnetPath = process.platform === 'win32' ? 'dotnet.exe' : 'dotnet'; [](http://example.com/codeflow?start=8&length=74)

This will use the version of `dotnet` which is globally installed. Is that the one you want? Or do you want the version that comes with the C# extension?

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -184,6 +273,228 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+        ];
+        const cpOptions: cp.SpawnOptionsWithoutStdio = {
+            detached: true,
+            windowsHide: true,
+        };
+        let chunksProcessed = 0;
+        let proxyICorDebugPort = -1;
+        let proxyBrowserPort = -1;
+        let newUri = '';
+        const spawnedProxy = cp.spawn(dotnetPath, spawnedProxyArgs, cpOptions);
```
gregg-miskelly, line 446: >spawnedProxy [](http://example.com/codeflow?start=14&length=12)

Do you need to handle the process exited before it says it is ready?

src/shared/assets.ts
gregg-miskelly: Should we just use a sentinal program name? Ex: "Blazor WASM"?

package.json
gregg-miskelly: Did you push this fix? I don't see it

thaystg: I forgot to push, it's there now.

src/shared/assets.ts
thaystg: We can't it's required that the program exists in the disk.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: I fixed it, now it's using the one that comes with the C# extension.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: done.

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
thaystg: done,

thaystg:
@gregg-miskelly can I get another review here?
Thanks a lot!

src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
```diff
@@ -184,6 +321,364 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
...
+            },
+            versionSpecRequirement: 'greater_than_or_equal',
+            rejectPreviews: true,
+        };
+
+        try {
+            let acquireResult = await vscode.commands.executeCommand<IDotnetAcquireResult | undefined>(
+                'dotnet.findPath',
+                findPathRequest
+            );
```
gregg-miskelly, line 397: I am not following what this code is doing -- don't you need to check what version and which runtime this is? #Closed

gregg-miskelly: Oh, I see

PR #7236, opened 2024-06-14
Download razor telemetry nupkg and include in vsix

ryzngard:
Updates to https://dev.azure.com/dnceng/internal/_build/results?buildId=2475112&view=results to have correct nuget packaging.


tasks/offlinePackagingTasks.ts
```diff
@@ -175,6 +182,9 @@ gulp.task(
     // Run the fetch of all packages, and then also installDependencies after
     gulp.series(async () => {
         await updateNugetPackageVersion(allNugetPackages.razor);
+
+        // Also pull in the Roslyn DevKit dependencies nuget package.
```
davidwengier, line 186: Nit
```suggestion
        // Also pull in the Razor DevKit dependencies nuget package.
```

davidwengier (approved):
It's not clear to me why we need two different packages, or to put things in two different folders, it seems like we're probably just making our install size bigger by duplicating DLLs. But thats not really a reason to block this PR so have a green tick :)

ryzngard:
> It's not clear to me why we need two different packages, or to put things in two different folders, it seems like we're probably just making our install size bigger by duplicating DLLs. But thats not really a reason to block this PR so have a green tick :)

I don't have a strong opinion here. the duplicate dll is just the telemetry dll, and technically we may be on a separate version than Roslyn. If we decide to require the same version then we could probably reduce size slightly

davidwengier:
I was thinking duplicated dlls between our two nupkgs, like shared code or workspaces.DLL maybe. If there isn't any, then great, ignore me :)

src/lsptoolshost/builtInComponents.ts
```diff
@@ -27,9 +27,14 @@ export const componentInfo: { [key: string]: ComponentInfo } = {
...
     },
+    razorDevKit: {
+        defaultFolderName: '.razorDevKit',
+        optionName: 'razorDevKit',
+        componentDllPaths: ['Microsoft.VisualStudio.DevKit.Razor.dll'],
+    },
 };
 
-export function getComponentPaths(componentName: string, options: LanguageServerOptions): string[] {
+export function getComponentPaths(componentName: string, options: LanguageServerOptions | undefined): string[] {
```
dibarbet, line 37: instead of having this undefined you could consider adding the razorDevkit folder as a component path option.  Then you would be able to override it for local development.

dibarbet (commented):
> or to put things in two different folders

For us it's a nice way to ensure they don't even accidentally load in non-devkit scenarios if they're in a different folder.

src/lsptoolshost/builtInComponents.ts
ryzngard: I think the only issue I have with that is the structure `LanguageServerOptions` has a lot of concepts that don't apply to razor. Could we move the options to a lighter interface?

dibarbet: I guess technically the other options aren't relevant to you, but you can just import `languageServerOptions` and pass it to the function (without caring what else is on it).  Actually, frankly `builtInComponents.ts` should probably just call `languageServerOptions.componentPaths` and not even have options be a parameter.

dibarbet: We could move it to `CommonOptions` as well, but not sure if its necessary to do that

ryzngard: > Actually, frankly builtInComponents.ts should probably just call languageServerOptions.componentPaths and not even have options be a parameter

I'll look into that. Seems cleaner

ryzngard: Actually let me see about that in a follow up. Would like to get this in for now since I'll likely have another razor change coming in that will require dual insertion

Issue #6675, opened 2023-11-16
Report generic project load or reload progress


dibarbet:
Today, there is no real indication of how much progress the server is making when loading projects.  The most information is in the C# logs where we indicate when we've loaded a project.

 
We should add a mechanism on the server side that sends progress reports to the client on project loading.
The client can then use these reports to display a progress bar indicating what is happening and how far along we are in project load.

We can use this to also move automatic restore entirely server side (and report using the same progress mechanism).

PR #7254, opened 2024-06-21
Add workspace status item to the Language Status bar.

JoeRobich:
This makes a little progress on #6675. It shows a work indicator for the initial workspace load. It will also show the open solution file name and provides a link for users to open a different solution. This does not move NuGet restore to the server nor does it show progress when reloading projects.

https://github.com/user-attachments/assets/314e4967-065a-4568-8b29-24004967a777


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -262,6 +267,16 @@ export class RoslynLanguageServer {
         await this._languageClient.restart();
     }
 
+    public workspaceDisplayName(): string {
+        if (this._solutionFile !== undefined) {
```
dibarbet, line 271: with devkit, you may have a solution open, but this will not be defined.  If we want to have consistent behavior in both, it would probably be better to get the item to display from the server.  As the server will know the solution name in either case.

JoeRobich: We hide this status item when DevKit is enabled.

src/lsptoolshost/languageStatusBar.ts
```diff
@@ -0,0 +1,42 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { RoslynLanguageServer } from './roslynLanguageServer';
+import { RoslynLanguageServerEvents } from './languageServerEvents';
+import { languageServerOptions } from '../shared/options';
+import { ServerState } from './serverStateChange';
+import { getCSharpDevKit } from '../utils/getCSharpDevKit';
+
+export function registerLanguageStatusItems(
```
dibarbet, line 13: Not relevant for this PR, but something to check when we do TFMs - can two different extensions register language status items for the same language?  Devkit does have a language status item (launch.json stuff).

dibarbet (approved):
Totally good with this as the first part!

PR #7300, opened 2024-07-02
Use rzls.dll on macOS

davidwengier:
https://github.com/dotnet/vscode-csharp/commit/c2666b4af8f1f56bb7eccc773cc923f17a1ef1e6 but for Razor

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2111269

My hypothesis is that now that we use nupkgs, our publishing happens on windows so can't be signed in a mac compatible way.

### DIFF
--- a/src/razor/src/razorLanguageServerOptionsResolver.ts
+++ b/src/razor/src/razorLanguageServerOptionsResolver.ts
@@ -37,12 +37,29 @@ export function resolveRazorLanguageServerOptions(
 }
 
 function findLanguageServerExecutable(withinDir: string) {
-    // Prefer using executable over fallback to dll.
-    const fileName = isWindows() ? 'rzls.exe' : 'rzls';
-    let fullPath = path.join(withinDir, fileName);
-    if (!fs.existsSync(fullPath)) {
-        fullPath = path.join(withinDir, 'rzls.dll');
+    // On Windows we use the executable, which is "rzls.exe".
+    // On macOS we use the dll, which is "rzls.dll".
+    // On everything else we use the executable, which is "rzls".
+
+    const fileName = 'rzls';
+    let extension = '';
+
+    if (isWindows()) {
+        extension = '.exe';
     }
+
+    if (isMacOS()) {
+        // Use the DLL on MacOS to work around signing issue tracked by https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1767519/
+        extension = '.dll';
+    }
+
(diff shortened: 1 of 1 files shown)


src/razor/src/razorLanguageServerOptionsResolver.ts
```diff
@@ -37,13 +37,30 @@ export function resolveRazorLanguageServerOptions(
...
+        extension = '.dll';
     }
 
+    let pathWithExtension = `${fileName}${extension}`;
+    if (!fs.existsSync(pathWithExtension)) {
+        // We might be running a platform neutral vsix which has no executable, instead we run the dll directly.
+        pathWithExtension = `${fileName}.dll`;
+    }
+
+    let fullPath = path.join(withinDir, pathWithExtension);
```
davidwengier, line 62:
```suggestion
    const fullPath = path.join(withinDir, pathWithExtension);
```

src/razor/src/razorLanguageServerOptionsResolver.ts
```diff
@@ -56,3 +73,7 @@ function findLanguageServerExecutable(withinDir: string) {
 function isWindows() {
     return !!os.platform().match(/^win/);
 }
+
+function isMacOS() {
+    return os.platform() === 'darwin';
+}
```
davidwengier, line 79:
```suggestion
}

```

ryzngard (approved):

dibarbet:
> My hypothesis is that now that we use nupkgs, our publishing happens on windows so can't be signed in a mac compatible way.

Yeah if all your build and packaging happens on Windows mac executables won't be `codesign`d

ryzngard:
> > My hypothesis is that now that we use nupkgs, our publishing happens on windows so can't be signed in a mac compatible way.
> 
> Yeah if all your build and packaging happens on Windows mac executables won't be `codesign`d

Is there a way we can build the nugets on macos? It seems like we were building on macos before so we have it as part of our pool. Maybe it won't be too hard to build mac on mac?

dibarbet:
> Is there a way we can build the nugets on macos? It seems like we were building on macos before so we have it as part of our pool. Maybe it won't be too hard to build mac on mac?

I assume its possible.  Probably need to split your signed build into multiple parts (a windows step and mac step) and only build mac binaries on mac.

We haven't done it for Roslyn because its quite a bit of work to split our the signed build.

PR #7321, opened 2024-07-09
Add Project Context status item

JoeRobich:
Adds a status item to display the project context of the active C# file.

For files in the open workspace:
image

For files in the misc workspace:
image

When running with DevKit:
image

### DIFF
--- a/src/lsptoolshost/languageStatusBar.ts
+++ b/src/lsptoolshost/languageStatusBar.ts
@@ -19,6 +19,7 @@ export function registerLanguageStatusItems(
     if (!getCSharpDevKit()) {
         WorkspaceStatus.createStatusItem(context, languageServerEvents);
     }
+    ProjectContextStatus.createStatusItem(context, languageServer, languageServerEvents);
 }
 
 class WorkspaceStatus {
@@ -38,5 +39,45 @@ class WorkspaceStatus {
             item.text = e.workspaceLabel;
             item.busy = e.state === ServerState.ProjectInitializationStarted;
         });
+    }
+}
+
+class ProjectContextStatus {
+    static createStatusItem(
+        context: vscode.ExtensionContext,
+        languageServer: RoslynLanguageServer,
+        languageServerEvents: RoslynLanguageServerEvents
+    ) {
+        const projectContextService = languageServer._projectContextService;
+
+        const item = vscode.languages.createLanguageStatusItem(
+            'csharp.projectContextStatus',
+            languageServerOptions.documentSelector
+        );
+        item.name = vscode.l10n.t('C# Project Context Status');
+        item.detail = vscode.l10n.t('Project Context');
+        context.subscriptions.push(item);
+
+        updateItem(vscode.window.activeTextEditor);
+        context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(updateItem));
+
+        languageServerEvents.onServerStateChange((e) => {
+            if (e.state === ServerState.ProjectInitializationComplete) {
+                projectContextService.clear();
+                updateItem(vscode.window.activeTextEditor);
+            }
+        });
+
+        async function updateItem(e: vscode.TextEditor | undefined) {
+            if (e?.document.languageId !== 'csharp') {
+                item.text = '';
+                return;
+            }
+
+            const projectContext = await projectContextService.getCurrentProjectContext(e.document.uri);
+            if (projectContext) {
+                item.text = projectContext._vs_label;
+            }
+        }
     }
 }
--- a/src/lsptoolshost/languageStatusBar.ts
+++ b/src/lsptoolshost/languageStatusBar.ts
@@ -19,7 +19,7 @@ export function registerLanguageStatusItems(
     if (!getCSharpDevKit()) {
         WorkspaceStatus.createStatusItem(context, languageServerEvents);
     }
-    ProjectContextStatus.createStatusItem(context, languageServer, languageServerEvents);
+    ProjectContextStatus.createStatusItem(context, languageServer);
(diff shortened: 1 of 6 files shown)


src/lsptoolshost/languageStatusBar.ts
```diff
@@ -19,6 +19,7 @@ export function registerLanguageStatusItems(
     if (!getCSharpDevKit()) {
         WorkspaceStatus.createStatusItem(context, languageServerEvents);
     }
+    ProjectContextStatus.createStatusItem(context, languageServer, languageServerEvents);
```
dibarbet, line 22: Check that this works with devkit - not sure what happens when there are two extensions both contributing items

src/lsptoolshost/languageStatusBar.ts
```diff
@@ -40,3 +41,43 @@ class WorkspaceStatus {
...
+        languageServerEvents: RoslynLanguageServerEvents
+    ) {
+        const projectContextService = languageServer._projectContextService;
+
+        const item = vscode.languages.createLanguageStatusItem(
+            'csharp.projectContextStatus',
+            languageServerOptions.documentSelector
+        );
+        item.name = vscode.l10n.t('C# Project Context Status');
+        item.detail = vscode.l10n.t('Project Context');
```
dibarbet, line 58: Instead of 'Project Context' in the right hand part of the status bar item, maybe 'Active File Context'?

src/lsptoolshost/services/projectContextService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    async getCurrentProjectContext(uri: string | vscode.Uri): Promise<VSProjectContext | undefined> {
+        const projectContexts = await this.getProjectContexts(uri);
+        return projectContexts?._vs_projectContexts[projectContexts._vs_defaultIndex];
+    }
+
+    async getProjectContexts(uri: string | vscode.Uri): Promise<VSProjectContextList | undefined> {
+        const uriString = uri instanceof vscode.Uri ? UriConverter.serialize(uri) : uri;
+
+        if (!(uriString in this._projectContexts)) {
```
dibarbet, line 30: A little concerned about this caching.  It kind of seems like we should just query the server every time the active text editor changes, regardless of if we had contexts before.  It isn't a hugely expensive request.

For example, if we removed/added a TFM, this cache doesn't seem like it would get refreshed.

I believe VS already even sends the request on typing.

src/lsptoolshost/languageStatusBar.ts
```diff
@@ -40,3 +41,43 @@ class WorkspaceStatus {
...
+        context.subscriptions.push(item);
+
+        updateItem(vscode.window.activeTextEditor);
+        context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(updateItem));
```
dibarbet, line 62: I presume we can only have one active text editor right?  If so we should like have a cancellation source here where the prior request gets cancelled every time this changes.

e.g. similar to https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/onAutoInsert.ts#L66

src/lsptoolshost/languageStatusBar.ts
JoeRobich: image

src/lsptoolshost/languageStatusBar.ts
JoeRobich: I'll change to that. @mikadumont is there a name you prefer here?

src/lsptoolshost/services/projectContextService.ts
JoeRobich: Removing caching from this PR

src/lsptoolshost/languageStatusBar.ts
JoeRobich: Updated.

src/lsptoolshost/services/projectContextService.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        try {
+            const contextList = await this.getProjectContexts(uri, this._source.token);
+            if (!contextList) {
+                return;
+            }
+
+            const context = contextList._vs_projectContexts[contextList._vs_defaultIndex];
+            this._contextChangeEmitter.fire({ uri, context });
+        } catch {
```
dibarbet, line 60: I wonder if we can differentiate between a cancellation exception and something else.  Would be nice to log if something actually failed.

src/lsptoolshost/services/projectContextService.ts
```diff
@@ -0,0 +1,78 @@
...
+            // This request was cancelled
+        }
+    }
+
+    private async getProjectContexts(
+        uri: vscode.Uri,
+        token: vscode.CancellationToken
+    ): Promise<VSProjectContextList | undefined> {
```
dibarbet, line 68: DOes this ever actually return undefined?  I think sendRequest returns the <T> or throws, and the <T> should just be `VSProjectContextList`

dibarbet (approved):
A couple small comments, but nothing blocking.  Would be great to get this in for next weeks release!

src/lsptoolshost/services/projectContextService.ts
JoeRobich: Moved the try/catch here so we now return `undefined` when cancelled.

src/lsptoolshost/services/projectContextService.ts
JoeRobich: Updated to only swallow `CancellationError` and re-throw others.

dibarbet (approved):

PR #7328, opened 2024-07-10
Add ability to select a document's project context

JoeRobich:
Adds the ability to select a document's project context. Uses the key added in https://github.com/dotnet/roslyn/pull/81940 to ensure all documents that have the same set of contexts use the same active context.

- Updates Roslyn to bring in new refresh handler and project context list key
- Adds select project context command
- Adds the select command to the Project Context status item
- Updates middleware to send selected context with server requests
- Sends a provider refresh notification when selected context is changed

https://github.com/user-attachments/assets/01ab29f2-b89d-49c7-9c3f-2858fa64eeb6

<details>
<summary>See all Roslyn changes...</summary>

[View Complete Diff of Changes](https://github.com/dotnet/roslyn/compare/19f995da648f0afdaffe499200e9c50dc0568eb2...79ce4ab3329da76c6793f1bf453efb7793834685?w=1)
* Add a notification handler for project context changed. (PR: [#81942](https://github.com/dotnet/roslyn/pull/81942))
* Ensure IDE does not use .editorconfig for source generated files (PR: [#81911](https://github.com/dotnet/roslyn/pull/81911))
* Send a key across identifying contents of the project context list (PR: [#81940](https://github.com/dotnet/roslyn/pull/81940))
* Include document path in InvalidOperationException when document not found (PR: [#81969](https://github.com/dotnet/roslyn/pull/81969))
* Include Enc diagnostics in Razor requests (PR: [#81941](https://github.com/dotnet/roslyn/pull/81941))
* Fix completions after attribute list in lambdas (PR: [#81961](https://github.com/dotnet/roslyn/pull/81961))
</details>


l10n/bundle.l10n.json
```diff
@@ -161,8 +161,12 @@
   "Fix All: ": "Fix All: ",
   "C# Workspace Status": "C# Workspace Status",
   "Open solution": "Open solution",
+  "Select context": "Select context",
+  "C# Project Context Status": "C# Project Context Status",
+  "Active File Context": "Active File Context",
```
dibarbet, line 166: Wondering - could we merge this into the select button?  
e.g.
```ProjectA (net6.0)            Change Active File Context```

or maybe we could even put the document name in it?
```ProjectA (net6.0)            Change Program.cs Context```

Comment (JoeRobich) [reply l10n/bundle.l10n.json:166]: Sure, although the select command is only visible when there are multiple project contexts. Would it be jarring for the text to move from the item details to the command and back again as you navigate?

Comment (dibarbet) [reply l10n/bundle.l10n.json:166]: ah hmm.  I wonder if we should just always show the button even if there's only one.  Not sure

File: src/lsptoolshost/projectContext/projectContextService.ts [hunk 2]
```diff
@@ -64,6 +67,55 @@ export class ProjectContextService {
...
+        if (!this._projectContextMap.has(key)) {
+            this._projectContextMap.set(key, contextList._vs_projectContexts[contextList._vs_defaultIndex]);
+        }
+
+        return this._projectContextMap.get(key);
+    }
+
+    private getContextKey(contextList: VSProjectContextList): string {
+        return contextList._vs_projectContexts
+            .map((context) => context._vs_label)
```
dibarbet: should this be using the server side key?

CHANGELOG.md
```diff
@@ -4,23 +4,30 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 # 2.112.x
+* Add ability to select a doecument's project context (PR: [#7328](https://github.com/dotnet/vscode-csharp/pull/7328))
```
JoeRobich:
```suggestion
* Add ability to select a document's project context (PR: [#7328](https://github.com/dotnet/vscode-csharp/pull/7328))
```

CHANGELOG.md
```diff
@@ -4,23 +4,30 @@
...
-  * Rename component tags and type references when a Razor file is renamed (PR: [#12561](https://github.com/dotnet/razor/pull/12561))                                                                                                                    
-  * Handle conflict markers (PR: [#12642](https://github.com/dotnet/razor/pull/12642))                                                                                                                                                                   
-  * Fix cross project span and edit mapping (PR: [#12614](https://github.com/dotnet/razor/pull/12614))                                                                                                                                                   
-* Update Roslyn to 5.4.0-2.26060.1 (PR: [#8877](https://github.com/dotnet/vscode-csharp/pull/8877))  
+  * Filter our html diagnostics when a tag helper attribute spans multiple lines (PR: [#12654](https://github.com/dotnet/razor/pull/12654))
```
JoeRobich, line 15: Removed some trailing whitespace from Razor entries.

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -64,6 +67,55 @@ export class ProjectContextService {
...
+    ): Promise<VSProjectContext>;
+    public async getDocumentContext(
+        uri: string | vscode.Uri,
+        contextList?: VSProjectContextList | undefined
+    ): Promise<VSProjectContext | undefined> {
+        // To find the current context for the specified document we need to know the list
+        // of contexts that it is a part of.
+        contextList ??= await this.getProjectContexts(uri, CancellationToken.None);
+        if (contextList === undefined) {
+            return undefined;
```
dibarbet: consider trace logging something here

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -64,6 +67,55 @@ export class ProjectContextService {
...
+        uri: string | vscode.Uri,
+        contextList?: VSProjectContextList | undefined
+    ): Promise<VSProjectContext>;
+    public async getDocumentContext(
+        uri: string | vscode.Uri,
+        contextList?: VSProjectContextList | undefined
+    ): Promise<VSProjectContext | undefined> {
+        // To find the current context for the specified document we need to know the list
+        // of contexts that it is a part of.
+        contextList ??= await this.getProjectContexts(uri, CancellationToken.None);
```
dibarbet: my understanding might be wrong here, but it looks like `tryAddProjectContext` queries `getDocumentContext` without a contextList, meaning that it is going to make a project context request to the server for every request?

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -508,6 +529,23 @@ export class RoslynLanguageServer {
         }
     }
 
+    public async refreshFeatureProviders(): Promise<void> {
+        const uri = getActiveDocumentUri();
+        const request = uri !== undefined ? { uri: UriConverter.serialize(uri) } : undefined;
```
dibarbet: Now that I look at this a second time - I think it may not be correct for the refresh features notification to take in a URI.  Typically on the server side, the refresh queue looks at the URI to avoid sending a refresh if the document is open (as the client typically knows about changes to open documents). 

So if we pass the URI here to the server, I think the queue will skip refreshing?
https://github.com/dotnet/roslyn/pull/81942/changes#diff-6635807165b7ab31eb31a3143e04cea70b5a7b5980dca99e461a632ee89b1ccfL111

JoeRobich: Oh interesting! I would prefer not to send one either =)

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: Sigh, yes. I had previously pushed an update to use it but I guess that was from a different machine and I didn't pull before updating to use the provider refresh. Updated!

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: reworked that code out.

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: Added another cache so I can track document uri -> context key which makes it possible to lookup active context.

src/lsptoolshost/server/roslynLanguageServer.ts
JoeRobich: works better not sending one

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -17,19 +17,24 @@ import { LanguageServerEvents, ServerState } from '../server/languageServerEvent
...
+
 // We want to verify the project context is in a stable state before warning the user about miscellaneous files.
 const VerificationDelay = 5 * 1000;
 
 let _verifyTimeout: NodeJS.Timeout | undefined;
 let _documentUriToVerify: vscode.Uri | undefined;
 
 export class ProjectContextService {
+    private readonly _projectContextMap: Map<ContextKey, VSProjectContext> = new Map();
+    private readonly _contextKeyMap: Map<UriString, ContextKey> = new Map();
```
dibarbet: I think it'd be nice to have some doc comments on what the two maps are for and how they are supposed to be updated

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -90,38 +128,69 @@ export class ProjectContextService {
...
-            return await this._languageServer.sendRequest(
+            const contextList = await this._languageServer.sendRequest(
                 VSGetProjectContextsRequest.type,
                 { _vs_textDocument: textDocument },
                 token
             );
+
+            // Update our caches so that we can quickly lookup the active context later.
+            // No need to track if there is only one context because the default context
+            // will always be active.
```
dibarbet: is the idea here that the lookup for the context on the client side will fail, but that is fine because the server will automatically choose only the single context?

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -90,38 +128,69 @@ export class ProjectContextService {
...
+                if (!this._projectContextMap.has(contextList._vs_key)) {
+                    this._projectContextMap.set(
+                        contextList._vs_key,
+                        contextList._vs_projectContexts[contextList._vs_defaultIndex]
+                    );
+                }
+            } else {
+                // If memory becomes a concern we could ref count how many documents are
+                // using each context key and remove them here as well.
+                this._contextKeyMap.delete(uriString);
```
dibarbet: We should definitely hook up something on document close to delete items out of this map.

For `_projectContextMap`, is it possible to clear that set?  We don't really know which contexts sets are active right?  Like for example what happens if you change the TFMs for a project over and over - will we keep accumulating new items in this map?

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -90,38 +128,69 @@ export class ProjectContextService {
...
         // If we do not recieve a refresh even within the timout period, send a verified event.
         _verifyTimeout = setTimeout(() => {
-            this._contextChangeEmitter.fire({ languageId, uri, context, isVerified: true });
+            this._contextChangeEmitter.fire({ document, context, isVerified: true, hasAdditionalContexts });
         }, VerificationDelay);
     }
 
-    private async getProjectContexts(
-        uri: vscode.Uri,
+    public async getProjectContexts(
```
dibarbet: nit - rename to queryServerProjectContexts to distinguish better from  `getDocumentContext`

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: Yes, by not providing a project context the server will use the default one as it does today.

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: It will accumulate new contexts over time but only for new context keys. Assuming the user isn't constantly changing TFMs on their projects this shouldn't be a problem as the keys will not be changing. Would be simple enough to ref count and remove entries when they reach 0.

JoeRobich: Added ref counting and cache cleanup.

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: Doc'd!

src/lsptoolshost/projectContext/projectContextService.ts
JoeRobich: I updated this comment to say that more explicitly

l10n/bundle.l10n.json
JoeRobich: I think we should get it into preview in its current form and work out user experiences issues from there.

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -130,4 +195,46 @@ export class ProjectContextService {
...
+
+    private updateCaches(uriString: UriString, contextList: VSProjectContextList) {
+        const oldContextKey = this._uriToContextKeyMap.get(uriString);
+        const newContextKey = contextList._vs_key;
+
+        if (oldContextKey !== undefined) {
+            if (oldContextKey === newContextKey) {
+                // We have already seen this context key and it hasn't changed, nothing to do.
+                return;
+            }
```
dibarbet: suggestion - move this earlier outside of the inner `if (oldContextKey !== undefined)`.  Think it would make it easier to reason about the method flow.  `newContextKey` should never be undefined so should be fine.

Then this inner block just ends up being only about decrementing the old key when the key changes.

src/lsptoolshost/projectContext/projectContextService.ts
```diff
@@ -130,4 +195,46 @@ export class ProjectContextService {
...
+                this._contextKeyToRefCountMap.set(oldContextKey, oldRefCount - 1);
+            }
+        }
+
+        // Update our caches so that we can quickly lookup the active context later.
+        // No need to track when there is only one context because the server will use
+        // the default context automatically.
+        if (contextList._vs_projectContexts.length > 1) {
+            // Increment the ref count for the new context key.
+            const newRefCount = (this._contextKeyToRefCountMap.get(newContextKey) || 0) + 1;
```
dibarbet: took me a bit to convince myself that always incrementing here is fine - might be good to add a comment stating that we only get here if the key for an existing tracked URI is changing, or if we're adding a key for a new URI (both of which need to increment the ref count).

dibarbet (approved):

hamish-smith:
@JoeRobich Thank you! Video on this PR has shown me how to find a feature that will make working on multi-targeted libraries a lot easier. Thank you for this!

PR #7333, opened 2024-07-11
Enable Project Context status item for Razor files

JoeRobich:
Requests the virtual C# file path and uses that to get project context information.

image

When used with DevKit:
image

### DIFF
--- a/src/lsptoolshost/services/projectContextService.ts
+++ b/src/lsptoolshost/services/projectContextService.ts
@@ -10,6 +10,9 @@ import { TextDocumentIdentifier } from 'vscode-languageserver-protocol';
 import { UriConverter } from '../uriConverter';
 import { LanguageServerEvents } from '../languageServerEvents';
 import { ServerState } from '../serverStateChange';
+import { DynamicFileInfoHandler } from '../../razor/src/dynamicFile/dynamicFileInfoHandler';
+import { ProvideDynamicFileResponse } from '../../razor/src/dynamicFile/provideDynamicFileResponse';
+import { ProvideDynamicFileParams } from '../../razor/src/dynamicFile/provideDynamicFileParams';
 
 export interface ProjectContextChangeEvent {
     uri: vscode.Uri;
@@ -39,11 +42,22 @@ export class ProjectContextService {
 
     public async refresh() {
         const textEditor = vscode.window.activeTextEditor;
-        if (textEditor?.document?.languageId !== 'csharp') {
+        const languageId = textEditor?.document?.languageId;
+        if (languageId !== 'csharp' && languageId !== 'aspnetcorerazor') {
             return;
         }
 
-        const uri = textEditor.document.uri;
+        let uri = textEditor!.document.uri;
+
+        // If the active document is a Razor file, we need to map it back to a C# file.
+        if (languageId === 'aspnetcorerazor') {
+            const virtualUri = await this.getVirtualCSharpUri(uri);
+            if (!virtualUri) {
+                return;
+            }
+
+            uri = virtualUri;
+        }
 
         // If we have an open request, cancel it.
         this._source.cancel();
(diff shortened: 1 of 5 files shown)


src/lsptoolshost/services/projectContextService.ts
```diff
@@ -58,6 +72,20 @@ export class ProjectContextService {
         this._contextChangeEmitter.fire({ uri, context });
     }
 
+    private async getVirtualCSharpUri(uri: vscode.Uri): Promise<vscode.Uri | undefined> {
+        const response = await vscode.commands.executeCommand<ProvideDynamicFileResponse>(
```
ryzngard, line 76: @davidwengier I'm not sure if there will be any problem using this endpoint or if there's something else that should be used to find the dynamic file URI

davidwengier: I don't think it would cause any issues. In theory it could trigger Razor activation, but if this only gets called when a Razor document is open in the editor, that either has happened, or is about to anyway.

src/lsptoolshost/services/projectContextService.ts
```diff
@@ -58,6 +72,20 @@ export class ProjectContextService {
...
+            DynamicFileInfoHandler.provideDynamicFileInfoCommand,
+            new ProvideDynamicFileParams([uri.fsPath])
+        );
+
+        const responseUri = response.generatedFiles[0];
+        if (!responseUri) {
```
davidwengier, line 82: Does this need an explicit check for `null` too?
https://github.com/dotnet/vscode-csharp/blob/d7439d3fa6a13800dc12f3ce719c1010c2cb38d3/src/razor/src/dynamicFile/dynamicFileInfoHandler.ts#L47

davidwengier (approved):

src/lsptoolshost/services/projectContextService.ts
ryzngard: I think it would get added to the document list with a wrong uri and either be fixed or show up in misc files

src/lsptoolshost/services/projectContextService.ts
JoeRobich: The false-y check should cover null, undefined, and i think '' as well.

src/razor/src/document/razorDocumentSynchronizer.ts
```diff
@@ -119,6 +119,8 @@ export class RazorDocumentSynchronizer {
     }
 
     private removeSynchronization(context: SynchronizationContext) {
+        context.dispose();
```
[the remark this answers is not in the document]
davidwengier: Not sure I see why the token would get cancelled, but clearing the rejections like this doesn't seem like it would be an issue.

src/razor/src/document/razorDocumentSynchronizer.ts
```diff
@@ -119,6 +119,8 @@ export class RazorDocumentSynchronizer {
...
+
         const documentKey = getUriPath(context.projectedDocument.uri);
         const synchronizations = this.synchronizations[documentKey];
         clearTimeout(context.timeoutId);
```
davidwengier, line 126: Nit: Consider moving this inside `context.Dispose()` now that it exists?

src/razor/src/document/razorDocumentSynchronizer.ts
```diff
@@ -167,6 +169,11 @@ export class RazorDocumentSynchronizer {
                     reject(reason);
                 }
             },
+            dispose: () => {
+                while (rejectionsForCancel.length > 0) {
+                    rejectionsForCancel.pop();
+                }
```
davidwengier, line 175: Nit: I think `rejectionsForCancel.length = 0` works too

davidwengier (approved):

dibarbet (approved):

src/razor/src/document/razorDocumentSynchronizer.ts
JoeRobich: Since integration test have been flaky for me, I take your suggestions as a follow up and get this merged for snap.

PR #7352, opened 2024-07-15
Update the workspace status bar when the server is stopped.

JoeRobich:
- Sets the workspace status item to the Warning state to display an indicator that there has been an issue.
- Provides easy access to restart the LSP server.

https://github.com/user-attachments/assets/892186ae-0af3-4854-b7e5-2d1e6d371d9e


src/lsptoolshost/services/projectContextService.ts
```diff
@@ -22,13 +22,18 @@ export interface ProjectContextChangeEvent {
 export class ProjectContextService {
     private readonly _contextChangeEmitter = new vscode.EventEmitter<ProjectContextChangeEvent>();
     private _source = new vscode.CancellationTokenSource();
+    private readonly _emptyProjectContext: VSProjectContext = {
+        _vs_id: '',
+        _vs_kind: '',
+        _vs_label: '',
```
dibarbet, line 28: are the empty strings needed for anything?  Might prefer to send undefined if not

dibarbet (approved):

src/lsptoolshost/services/projectContextService.ts
JoeRobich: I do use it to blank out the Project Context item text. Going to leave as is for now and we can revisit if necessary.

PR #7359, opened 2024-07-16
Fix gotodef integration test failures throwing cancellation

dibarbet:
Promise was being returned immediately, so the catch block never got hit when the request was cancelled.


JoeRobich (approved):

dibarbet:
a few other misc failures, but none of the goto def test for now

Issue #6349, opened 2023-09-12
[SUGGESTION] Disable or Ability to Disable Autocomplete for XML Documentation Comments


arunchndr:
From vscode-dotnettools created by [Sour-Codes](https://github.com/Sour-Codes): microsoft/vscode-dotnettools#475  

### Describe the feature you'd like

The swap from Omnisharp has really put a damper on my productivity.  I know it takes a while to iron things out to be smooth, but this feels like we're walking backwards...

As you type `///`, LSP is quick to transform it into...

```
///	<summary>
///	
///	</summary>
```

...which is great and all, but subsequent line breaks after generates...

```
///	<summary>
///	
///	</summary>
///	
```

or when I type in anything that Intellisense can determine is XML documentation comments for autocompletion, it fails horribly.

For example, if I start typing a type reference `<see cref="System.Object" />`, the context menu will show `<see cref="System.Object" />` for autocompletion, but will only output `<s` and then I am stuck typing the rest of it out anyway.  In fact, any of the autocompletion for XML documentation comments will have a similar effect, unfortunately.

To be honest, I've got so used to this not being automated going from Visual Studio, which handles this effectively, to Omnisharp, where it didn't automate this at all.    These automation attempts with errors makes me wish that `///` didn't autocomplete without my choice.

(I force `<GenerateDocumentationFile>` to `true` on all of my projects and a client's project has over 1,200 warnings of missing XML documentation comments.  My plan was to tackle them this weekend, but not with this `///` bug...)

PR #7381, opened 2024-07-22
Adds an option to disable auto insert

akhera99:
Roslyn side fix: https://github.com/dotnet/roslyn/pull/74485
To fix: #6349


package.json
```diff
@@ -732,6 +732,11 @@
         "title": "Text Editor",
         "order": 1,
         "properties": {
+          "dotnet.autoInsert.enableAutoInsert": {
```
JoeRobich, line 735: Will auto insert be used for more than doc comments? Perhaps the settings should be `dotnet.autoInsert.insertDocComments` imagining that certain auto inserts might be disabled without disabling all auto inserts.

JoeRobich: If we do just want one big switch, then I would suggest `dotnet.autoInsert.enabled`

Cosifne, line 735: This needs to be covered by test.
https://github.com/dotnet/vscode-csharp/blob/main/test/lsptoolshost/unitTests/configurationMiddleware.test.ts

dibarbet (commented):
looks fine - but we should include the server side update in this PR as well

dibarbet (approved):

PR #7427, opened 2024-08-09
Fix Attach to .NET Process before Roslyn LSP is ready

WardenGnaw:
This PR passes the roslynServerStartedProcess to the coreClrDebug activate call as the Roslyn Server registers the 'coreclr' DebugConfigurationProvider that will handle the process listing for the user to select the process ID.

If the configuration provider is not registered but the adapter factory is, it will send it to the debugger without a processName or processId and will fail with that error.

Fixes: https://github.com/microsoft/vscode-dotnettools/issues/1350

### DIFF
--- a/l10n/bundle.l10n.json
+++ b/l10n/bundle.l10n.json
@@ -191,6 +191,7 @@
   "Unexpected message received from debugger.": "Unexpected message received from debugger.",
   "[ERROR]: C# Extension failed to install the debugger package.": "[ERROR]: C# Extension failed to install the debugger package.",
   "Could not find a process id to attach.": "Could not find a process id to attach.",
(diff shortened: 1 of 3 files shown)


JoeRobich:
Would this also be an issue for O#? Maybe we could pass in `roslynLanguageServerStartedPromise || omnisharpLangServicePromise` and have it work the same for both.

WardenGnaw:
> Would this also be an issue for O#? Maybe we could pass in `roslynLanguageServerStartedPromise || omnisharpLangServicePromise` and have it work the same for both.

I wasn't able to hit it with O#, but I'll pass it in since it registers the configuration provider similar to Roslyn.

JoeRobich (approved):

gregg-miskelly (approved):

PR #7438, opened 2024-08-12
Fix handling razor files with non-Ascii characters

ryzngard:
Fixes https://github.com/microsoft/vscode-dotnettools/issues/1096

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -66,6 +66,8 @@ import { getComponentPaths } from './builtInComponents';
 import { OnAutoInsertFeature } from './onAutoInsertFeature';
 import { registerLanguageStatusItems } from './languageStatusBar';
 import { ProjectContextService } from './services/projectContextService';
+import { ProvideDynamicFileResponse } from '../razor/src/dynamicFile/provideDynamicFileResponse';
+import { ProvideDynamicFileParams } from '../razor/src/dynamicFile/provideDynamicFileParams';
 
 let _channel: vscode.OutputChannel;
 let _traceChannel: vscode.OutputChannel;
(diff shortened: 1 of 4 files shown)


ryzngard:
@dibarbet I don't know if the current behavior is intentional for `GetAbsoluteUri` but I didn't want to venture to change that because I'd undoubtedly break a lot of things. Is it worthy of a bug or just something that should be filed?

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1131,3 +1141,20 @@ function getSessionId(): string {
...
 }
+
+// matches https://github.com/dotnet/roslyn/blob/6696095f5aa51290707b5d8518b861ee0bab7884/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs#L21
+interface ProvideDynamicFileInfoParams {
+    razorFiles: string[];
+}
+
+// https://github.com/dotnet/roslyn/blob/21e07ecbcfa306608dc17418b5893ba50bb2fea9/src/LanguageServer/Protocol/Extensions/ProtocolConversions.cs#L199
+// CreateAbsoluteUri has different behavior if a string is Ascii or not. For Ascii, it provides an absolute path such as 'C:\project\page.razor,
+// but for non-Ascii characters it prepends `file://` and escapes the uri. This is the inverse of that by assuming anything starting with `file://` can
```
dibarbet, line 1152: this is weird - createabsoluteuri should always return a URI object.  When the URI object is serialized, it should always be serialized with the URI scheme (file://, or whatever else) on it.  If you're not seeing the scheme it likely means something is wrong with the serialization.

I suspect that the issue is that the type does not specify the DocumentUriConverter as the JsonConverter for it - https://github.com/dotnet/roslyn/blob/6696095f5aa51290707b5d8518b861ee0bab7884/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs#L23

Not blocking here, but definitely the server side should be fixed to serialize URIs correctly, otherwise you're going to get more URI issues here.

dibarbet (approved):

src/lsptoolshost/roslynLanguageServer.ts
ryzngard: I'll see if I can fix that on the server side. Was unaware there was a specific converter needed

dibarbet: I'm not certain the best way to specify a converter for an array.  Potentially the easiest solution is to switch from URI to `TextDocumentIdentifier` which already has the necessary converter.

E.g. https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Protocol/TextDocumentIdentifier.cs#L21

davidwengier: If we're updating the server anyway, I'd just make the property a `Uri` instead of `Uri[]`. It never has more than one element. I think it was a bit of ambitious design

ryzngard:
Closing to fix on server side

PR #7440, opened 2024-08-12
Include tooltip to Razor provisional completion

### LINKED ISSUE #7250: Provisional Completion in Razor should show full tooltip info
1. Open Visual Studio Code
2. Open a Razor file
3. Start completing `@DateTime.$$`

RESULT:
You can arrow up/down through the list, but the items don't have any tooltip to show their meaning:
image

EXPECT:
Tooltips with more data:
image
.... like in C# files.

jordi1215:
This resolves #7250. The cause was that when we added a provisional dot to the generated virtual csharp document, the change didn't get reflected on disk before the Roslyn server took the completion request.

Changes made in completionHandler.ts:
1. Made changes so that both provisional and non-provisional code completion uses the same `provideCompletionsCommand` (roslyn.provideCompletions).
2. On top of adding a temporary provisional dot "." after each provisional completion request, we are adding a "resolve" provisional dot for each resolve completion request. The dot is essential for the resolve request to be correct and therefor for the tooltip to show up.
3. Use vscode.workspace.onDidChangeTextDocument to ensure that the virtual documents get updated before sending the request to the roslyn server for both completion and resolve completion requests.

#### Before
image

#### What could have gone wrong if it wasn't for David's comment

https://github.com/user-attachments/assets/8147402c-a93e-4b59-aaff-fc682073bb8e

#### After
image

https://github.com/user-attachments/assets/534ca754-9cbe-4a03-9ba6-3b08c39416b2

### DIFF
--- a/src/razor/src/completion/completionHandler.ts
+++ b/src/razor/src/completion/completionHandler.ts
@@ -167,7 +167,8 @@ export class CompletionHandler {
             ) {
                 return delegatedCompletionItemResolveParams.completionItem;
             } else {
-                // will add a provisional dot to the C# document if a C# completion was triggered
+                // will add a provisional dot to the C# document if a C# provisional completion triggered
+                // this resolve completion request
                 if (virtualCsharpDocument.addResolveProvisionalDotAt()) {
                     const absoluteIndex = virtualCsharpDocument.getResolveProvisionalEditIndex();
                     if (absoluteIndex !== undefined) {
@@ -192,8 +193,7 @@ export class CompletionHandler {
             this.logger.logWarning(`${CompletionHandler.completionResolveEndpoint} failed with ${error}`);
         } finally {
             // remove the provisional dot after the resolve has completed and if it was added
-            const clearEditIndex = true;
-            if (virtualCsharpDocument.removeResolveProvisionalDot(clearEditIndex)) {
+            if (virtualCsharpDocument.removeResolveProvisionalDot()) {
                 const absoluteIndex = virtualCsharpDocument.getResolveProvisionalEditIndex();
                 const removeDot = true;
                 if (absoluteIndex !== undefined) {
@@ -219,6 +219,11 @@ export class CompletionHandler {
         // Convert projected position to absolute index for provisional dot
         const absoluteIndex = CompletionHandler.getIndexOfPosition(virtualDocument, projectedPosition);
         try {
+            // currently, we are temporarily adding a '.' to the C# document to ensure correct completions are provided
+            // for each roslyn.resolveCompletion request and we remember the location from the last provisional completion request.
+            // Therefore we need to remove the resolve provisional dot position
+            // at the start of every completion request in case a '.' gets added when it shouldn't be.
+            virtualDocument.clearResolveProvisionalEditIndex();
             if (provisionalTextEdit) {
                 // provisional C# completion
                 // add '.' to projected C# document to ensure correct completions are provided
--- a/src/razor/src/completion/completionHandler.ts
+++ b/src/razor/src/completion/completionHandler.ts
@@ -190,10 +190,17 @@ export class CompletionHandler {
         try {
             if (provisionalTextEdit) {
                 // provisional C# completion
-                return this.provideCSharpProvisionalCompletions(triggerCharacter, virtualDocument, projectedPosition);
+                // add '.' to projected C# document to ensure correct completions are provided
+                // This is because when a user types '.' after an object, it is initially in
+                // html document and not generated C# document.
+                const absoluteIndex = CompletionHandler.getIndexOfPosition(virtualDocument, projectedPosition);
+                if (absoluteIndex === -1) {
+                    return CompletionHandler.emptyCompletionList;
+                }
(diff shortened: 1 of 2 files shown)


davidwengier:
> The documentation gets synced afterwards 

This is only if the completion is actually taken, right? Do things still work correctly otherwise? eg, type `DateTime.`, press ESC to dismiss completion, and then scroll the text into view and out of view, or hover over something etc. I would expect not cleaning up the `.` would cause disco colors etc.

jordi1215:
> > The documentation gets synced afterwards
> 
> This is only if the completion is actually taken, right? Do things still work correctly otherwise? eg, type `DateTime.`, press ESC to dismiss completion, and then scroll the text into view and out of view, or hover over something etc. I would expect not cleaning up the `.` would cause disco colors etc.
> 
> TL;DR, If the Razor document doesn't change after typing the dot, presumably the generated document never gets re-synced



> > The documentation gets synced afterwards
> 
> This is only if the completion is actually taken, right? Do things still work correctly otherwise? eg, type `DateTime.`, press ESC to dismiss completion, and then scroll the text into view and out of view, or hover over something etc. I would expect not cleaning up the `.` would cause disco colors etc.
> 
> TL;DR, If the Razor document doesn't change after typing the dot, presumably the generated document never gets re-synced

Thanks for the heads up! I did manage to get it into a "bad" state. Assuming `var name="Jordi"`, `@name.` will trigger completion. If nothing gets accepted then C# will think it is an incomplete statement and give the red squiggles whereas HTML will interpret it as `Jordi.`. 

I think I will try add and remove the provisional dot for both completion and resolveCompletion command. 

Thanks so much!

src/razor/src/completion/completionHandler.ts
```diff
@@ -217,54 +259,39 @@ export class CompletionHandler {
...
-            if (virtualDocument.removeProvisionalDot()) {
-                await this.ensureProjectedCSharpDocumentUpdated(virtualDocument.uri);
-            }
-        }
+    private async delay(ms: number) {
+        return new Promise((resolve) => setTimeout(resolve, ms));
     }
 
-    private async ensureProjectedCSharpDocumentUpdated(virtualDocumentUri: vscode.Uri) {
+    private async ensureProvisionalDotUpdatedInCSharpDocument(
```
ryzngard, line 276: I wonder if it would be better to hook to `vscode.workspace.onDidChangeTextDocument` to wait for the buffer to be updated. Assuming we sync all changes before artifically adding the `.`

src/razor/src/csharp/csharpProjectedDocument.ts
```diff
@@ -80,6 +82,7 @@ export class CSharpProjectedDocument implements IProjectedDocument {
         if (this.provisionalEditAt && this.preProvisionalContent) {
             // Undo provisional edit if one was applied.
             this.setContent(this.preProvisionalContent);
+            this.resolveProvisionalEditAt = this.provisionalEditAt;
```
davidwengier, line 85: Perhaps this field should be cleared in `addProvisionalDotAt`, just to ensure the previous resolve index doesn't stick around?

src/razor/src/csharp/csharpProjectedDocument.ts
```diff
@@ -88,6 +91,49 @@ export class CSharpProjectedDocument implements IProjectedDocument {
         return false;
     }
 
+    // add resolve provisional dot if a provisional completion request was made
+    // A resolve provisional dot is the same as a provisional dot, but it remembers the
+    // last provisional dot inserted location and is used for the roslyn.resolveCompletion API
+    public addResolveProvisionalDotAt() {
```
davidwengier, line 97: Nit: I think `ensureResolveProvisionalDot` would be a better name for this. It doesn't always add, and on first reading of the handler that seemed like a potential bug, but it was just the name confusing me. Also drop the `At` since it doesn't take the index, it already knows.

davidwengier (commented):
I agree with Andrew, it would be much better if we can hook the didChange rather than use a delay. The rest makes sense, though I must admit I find this whole area confusing and hard to reason about. Could just be my TypeScript skills :)

src/razor/src/completion/completionHandler.ts
jordi1215: This makes so much sense! The reflected update is on the new commit. Thanks!

src/razor/src/csharp/csharpProjectedDocument.ts
jordi1215: The problem here is right now the resolve completion request doesn't know whether the completion request was provisional or not. So one bad state could be: 
-> (Provisional completion) user triggers provisional completion
-> (resolve completion) provisional dot are added and removed as the user toggles completion list 
-> (regular completion) as the user continues to type, completion list shows up 
-> (resolve completion) provisional dot are added and removed still since the field is not cleared

src/razor/src/csharp/csharpProjectedDocument.ts
jordi1215: That makes sense! Thank you!

davidwengier (approved):

ryzngard (approved):

PR #7442, opened 2024-08-13
Fix handling files with non-Ascii characters by sending a TextDocumentIdentifier for dynamic file requests/responses

ryzngard:
VS Code side of https://github.com/dotnet/roslyn/pull/74727

### DIFF
--- a/src/lsptoolshost/uriConverter.ts
+++ b/src/lsptoolshost/uriConverter.ts
@@ -24,6 +24,6 @@ export class UriConverter {
     }
 
     public static deserialize(value: string): vscode.Uri {
-        return vscode.Uri.parse(value);
+        return vscode.Uri.parse(value, true);
     }
 }
--- a/src/razor/src/dynamicFile/removeDynamicFileParams.ts
+++ b/src/razor/src/dynamicFile/removeDynamicFileParams.ts
@@ -3,8 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { DocumentUri } from 'vscode-languageclient/node';
+import { TextDocumentIdentifier } from 'vscode-languageclient/node';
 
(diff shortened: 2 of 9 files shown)


src/lsptoolshost/uriConverter.ts
```diff
@@ -24,6 +24,6 @@ export class UriConverter {
     }
 
     public static deserialize(value: string): vscode.Uri {
-        return vscode.Uri.parse(value);
+        return vscode.Uri.parse(value, true);
```
dibarbet, line 27: what does true do here?

ryzngard: it's the `strict` parameter. It causes it to throw if the Uri doesn't have a schema

dibarbet (approved):
pending roslyn update

src/razor/src/dynamicFile/removeDynamicFileParams.ts
```diff
@@ -3,8 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { DocumentUri } from 'vscode-languageclient/node';
+import { TextDocumentIdentifier } from 'vscode-languageclient/node';
 
+// matches https://github.com/dotnet/roslyn/blob/9e91ca6590450e66e0041ee3135bbf044ac0687a/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/HostWorkspace/RazorDynamicFileInfoProvider.cs#L28
 export class RemoveDynamicFileParams {
-    constructor(public readonly razorFiles: DocumentUri[]) {}
+    constructor(public readonly csharpDocument: TextDocumentIdentifier) {}
```
ryzngard, line 10: so ironically I named this wrong in the original PR 😅. I'll get that fixed later

davidwengier (approved):

PR #7490, opened 2024-08-29
Add signing support to VSIX

dibarbet:
Verified with vsce-sign and binlogs
image

https://dnceng.visualstudio.com/internal/_build/results?buildId=2531117&view=results


.vscodeignore
```diff
@@ -14,7 +14,7 @@
 .vscode-test/**
 coverage/**
 out/**
-server/**
+msbuild/**
```
JoeRobich, line 17: Does this mean we will be packaging the ServerDownload.csproj into the extension?

JoeRobich: Never mind, I see now that it is nested.

JoeRobich (approved):

dibarbet:
Thanks for the review!  Since this affects publishing (and we always run the publish pipeline on main changes), I'm going to hold off on merging for now until we've done our publishing for this week.

dibarbet:
Merging this now that the release stuff is all merged.

PR #7506, opened 2024-09-04
Fix up Node version

JoeRobich:


dibarbet (approved):
I'm not sure how I managed to only change one of the numbers haha

PR #7529, opened 2024-09-07
Move OmniSharp test code under the /test folder

JoeRobich:
Rearranges the `/test` folder to mirror the `/src` folder by having `omnisharp`, `lsptoolshost`, and `razor` subfolders. I moved fakes and the jestSetup code files into `/test` as they were used by all the different tests.


wallaby.js
```diff
@@ -6,10 +6,10 @@ module.exports = function(_wallaby) {
     files: [
       "src/**/*.ts*",
       "src/**/*.css",
-      "omnisharptest/omnisharpUnitTests/**/*.ts*",
-      "!omnisharptest/omnisharpUnitTests/**/*.test.ts*"
+      "test/omnisharp/omnisharpUnitTests/**/*.ts*",
```
JoeRobich, line 9: After searching for what wallaby.js is, I feel confident we can just delete this file.

dibarbet (approved):
Looks like we're seeing the same number of tests being run in CI, so think this is good.  I didn't look at every file though

PR #7536, opened 2024-09-09
Ensure the fix all resolve request reads an existing field

dibarbet:
The title property never existed on the data object we get from the server.  A server side change made it required for all code action requests (per LSP spec), so started throwing when we passed null.

This changes the client side to pass a value that is always provided in the data - https://github.com/dotnet/roslyn/blob/bd5c00e5e09de8564093f42d87fe49d4971f2e84/src/LanguageServer/Protocol/Handler/CodeActions/CodeActionResolveData.cs#L26

Resolves https://github.com/dotnet/vscode-csharp/issues/7535

### DIFF
--- a/src/lsptoolshost/fixAllCodeAction.ts
+++ b/src/lsptoolshost/fixAllCodeAction.ts
@@ -24,10 +24,14 @@ export function registerCodeActionFixAllCommands(
 }
 
 export async function getFixAllResponse(
-    data: LSPAny,
+    data: RoslynProtocol.CodeActionResolveData,
     languageServer: RoslynLanguageServer,
     outputChannel: vscode.OutputChannel
 ) {
+    if (!data.FixAllFlavors) {
+        throw new Error(`FixAllFlavors is missing from data ${JSON.stringify(data)}`);
+    }
+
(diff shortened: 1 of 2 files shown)


src/lsptoolshost/fixAllCodeAction.ts
```diff
@@ -24,10 +24,14 @@ export function registerCodeActionFixAllCommands(
 }
 
 export async function getFixAllResponse(
-    data: LSPAny,
+    data: RoslynProtocol.CodeActionResolveData,
     languageServer: RoslynLanguageServer,
     outputChannel: vscode.OutputChannel
 ) {
+    if (!data.FixAllFlavors) {
```
arkalyanms, line 31: nit: check the uniqueID resolves too?

arkalyanms (approved):

src/lsptoolshost/fixAllCodeAction.ts
dibarbet: fixallflavors is a nullable property on the data field, but uniqueId is not, we shouldn't need to verify it again

PR #7578, opened 2024-09-18
Register Copilot `relatedFilesProvider` for C#

genlu:
`registerRelatedFilesProvider` is a new method on the copilot extension. `RelatedDocumentsHandler` is implemented [here](https://github.com/CyrusNajmabadi/roslyn/blob/main/src/LanguageServer/Protocol/Handler/RelatedDocuments/RelatedDocumentsHandler.cs) in Roslyn language server.

### DIFF
--- a/src/lsptoolshost/copilot.ts
+++ b/src/lsptoolshost/copilot.ts
@@ -1,0 +1,89 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { CSharpExtensionId } from '../constants/csharpExtensionId';
+import { CopilotRelatedDocumentsReport, CopilotRelatedDocumentsRequest } from './roslynProtocol';
+import { RoslynLanguageServer } from './roslynLanguageServer';
+import { UriConverter } from './uriConverter';
+import { TextDocumentIdentifier } from 'vscode-languageserver-protocol';
+
+export async function registerCopilotExtension(languageServer: RoslynLanguageServer, channel: vscode.OutputChannel) {
+    const ext = vscode.extensions.getExtension('github.copilot');
+    if (!ext) {
+        channel.appendLine('GitHub Copilot extension not installed. Skipping call to `registerRelatedFilesProvider`');
+        return;
+    }
+    await ext.activate();
+    const relatedAPI = ext.exports as
+        | {
+              registerRelatedFilesProvider(
+                  providerId: { extensionId: string; languageId: string },
+                  callback: (
+                      uri: vscode.Uri
+                  ) => Promise<{ entries: vscode.Uri[]; traits?: { name: string; value: string }[] }>
+              ): void;
+          }
+        | undefined;
+    if (!relatedAPI) {
+        channel.appendLine(
+            'Incompatible GitHub Copilot extension installed. Skipping call to `registerRelatedFilesProvider`'
+        );
+        return;
+    }
+    channel.appendLine('registerRelatedFilesProvider succeeded.');
+
+    const id = {
+        extensionId: CSharpExtensionId,
+        languageId: 'csharp',
+    };
+
+    relatedAPI.registerRelatedFilesProvider(id, async (uri) => {
+        const writeOutput = (output: CopilotRelatedDocumentsReport[], builder: vscode.Uri[] | null) => {
+            if (output) {
+                for (const report of output) {
+                    if (report._vs_file_paths) {
+                        for (const filePath of report._vs_file_paths) {
+                            channel.appendLine('found related file: ' + filePath);
+                            builder?.push(vscode.Uri.file(filePath));
+                        }
+                    }
+                }
+            }
+        };
+
+        const relatedFiles: vscode.Uri[] = [];
+        const uriString = UriConverter.serialize(uri);
+        const textDocument = TextDocumentIdentifier.create(uriString);
+        const responsePromise = languageServer.sendRequestWithProgress(
+            CopilotRelatedDocumentsRequest.type,
+            {
+                _vs_textDocument: textDocument,
+                position: {
+                    line: 0,
+                    character: 0,
+                },
+            },
+            async (p) => {
+                writeOutput(p, relatedFiles);
+            }
+        );
+
+        await responsePromise.then(
+            (result) => {
+                writeOutput(result, null);
+                return;
+            },
+            (err) => {
+                channel.appendLine(err);
+                return;
+            }
+        );
+
+        return {
+            entries: relatedFiles,
+        };
+    });
+}
--- a/src/lsptoolshost/copilot.ts
+++ b/src/lsptoolshost/copilot.ts
@@ -10,30 +10,33 @@ import { RoslynLanguageServer } from './roslynLanguageServer';
 import { UriConverter } from './uriConverter';
 import { TextDocumentIdentifier } from 'vscode-languageserver-protocol';
 
-export async function registerCopilotExtension(languageServer: RoslynLanguageServer, channel: vscode.OutputChannel) {
+interface CopilotRelatedFilesProviderRegistration {
+    registerRelatedFilesProvider(
+        providerId: { extensionId: string; languageId: string },
+        callback: (uri: vscode.Uri) => Promise<{ entries: vscode.Uri[]; traits?: { name: string; value: string }[] }>
+    ): void;
+}
+
+export async function registerCopilotExtensionAsync(
+    languageServer: RoslynLanguageServer,
+    tracingChannel: vscode.OutputChannel
+) {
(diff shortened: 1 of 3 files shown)


src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export async function registerCopilotExtension(languageServer: RoslynLanguageServer, channel: vscode.OutputChannel) {
+    const ext = vscode.extensions.getExtension('github.copilot');
+    if (!ext) {
+        channel.appendLine('GitHub Copilot extension not installed. Skipping call to `registerRelatedFilesProvider`');
+        return;
+    }
+    await ext.activate();
+    const relatedAPI = ext.exports as
+        | {
```
dibarbet, line 21: Can you create a separate type definition for this?  It's a little hard to read inline

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+                  providerId: { extensionId: string; languageId: string },
+                  callback: (
+                      uri: vscode.Uri
+                  ) => Promise<{ entries: vscode.Uri[]; traits?: { name: string; value: string }[] }>
+              ): void;
+          }
+        | undefined;
+    if (!relatedAPI) {
+        channel.appendLine(
+            'Incompatible GitHub Copilot extension installed. Skipping call to `registerRelatedFilesProvider`'
```
dibarbet, line 32: It is possible to show a notification asking the user to upgrade to a new version if you think it would be valuable.  But I'd only do that if the extension is already installed

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+import { CSharpExtensionId } from '../constants/csharpExtensionId';
+import { CopilotRelatedDocumentsReport, CopilotRelatedDocumentsRequest } from './roslynProtocol';
+import { RoslynLanguageServer } from './roslynLanguageServer';
+import { UriConverter } from './uriConverter';
+import { TextDocumentIdentifier } from 'vscode-languageserver-protocol';
+
+export async function registerCopilotExtension(languageServer: RoslynLanguageServer, channel: vscode.OutputChannel) {
+    const ext = vscode.extensions.getExtension('github.copilot');
+    if (!ext) {
+        channel.appendLine('GitHub Copilot extension not installed. Skipping call to `registerRelatedFilesProvider`');
```
dibarbet, line 16: These are technically user visible strings.  Consider making it more evident want this actually means in terms of feature set

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+              ): void;
+          }
+        | undefined;
+    if (!relatedAPI) {
+        channel.appendLine(
+            'Incompatible GitHub Copilot extension installed. Skipping call to `registerRelatedFilesProvider`'
+        );
+        return;
+    }
+    channel.appendLine('registerRelatedFilesProvider succeeded.');
```
dibarbet, line 36: similar to above, would suggest re-wording to something like 'Registered copilot extension' or something similar

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+        languageId: 'csharp',
+    };
+
+    relatedAPI.registerRelatedFilesProvider(id, async (uri) => {
+        const writeOutput = (output: CopilotRelatedDocumentsReport[], builder: vscode.Uri[] | null) => {
+            if (output) {
+                for (const report of output) {
+                    if (report._vs_file_paths) {
+                        for (const filePath of report._vs_file_paths) {
+                            channel.appendLine('found related file: ' + filePath);
```
dibarbet, line 49: This should probably only be logged if the log verbosity is set to trace.  I think we do it elsewhere.

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+    channel.appendLine('registerRelatedFilesProvider succeeded.');
+
+    const id = {
+        extensionId: CSharpExtensionId,
+        languageId: 'csharp',
+    };
+
+    relatedAPI.registerRelatedFilesProvider(id, async (uri) => {
+        const writeOutput = (output: CopilotRelatedDocumentsReport[], builder: vscode.Uri[] | null) => {
+            if (output) {
```
dibarbet, line 45: If the output is actually nullable, it probably should be `output?: CopilotRelatedDocumentsReport[]`.  Otherwise this check should be removed.

The client response type is defined as non-null in `roslynProtocol`, but the server defines the response as nullable -https://github.com/dotnet/roslyn/pull/74918/files#diff-0b28133c769dc0e92c52213acbee40648526328ab9cec6c7c0e2a9b5a43fa76bR70

It doesn't look like the server ever returns null though - so potentially the server annotation needs to be updated?

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+    channel.appendLine('registerRelatedFilesProvider succeeded.');
+
+    const id = {
+        extensionId: CSharpExtensionId,
+        languageId: 'csharp',
+    };
+
+    relatedAPI.registerRelatedFilesProvider(id, async (uri) => {
+        const writeOutput = (output: CopilotRelatedDocumentsReport[], builder: vscode.Uri[] | null) => {
```
dibarbet, line 44: does `builder?: vscode.Uri[]` work?

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,89 @@
...
+                    line: 0,
+                    character: 0,
+                },
+            },
+            async (p) => {
+                writeOutput(p, relatedFiles);
+            }
+        );
+
+        await responsePromise.then(
```
dibarbet, line 74: I suspect the other places using `languageServer.sendRequestWithProgress` need to be updated - but I think it should be fine to just await it directly (in a try catch if you want to handle errors).

src/lsptoolshost/roslynProtocol.ts
```diff
@@ -221,6 +221,16 @@ export interface ProjectNeedsRestoreName {
     projectFilePaths: string[];
 }
 
+export interface CopilotRelatedDocumentsParams extends lsp.WorkDoneProgressParams, lsp.PartialResultParams {
+    _vs_textDocument: lsp.TextDocumentIdentifier;
+    position: lsp.Position;
+}
+
+export interface CopilotRelatedDocumentsReport {
+    _vs_resultId: string | null;
```
dibarbet, line 230: I noticed we have a resultId in the server response, but its not used, and it is not present in the client request either.  That seems like a bug - otherwise we're going to be recomputing this information on every single request.

src/lsptoolshost/copilot.ts
genlu: hmm, maybe I should move this to the trace log. This is intended to provide additional context implvitly (i.e. no user facing option to turn it on and off)

dibarbet: Also fine with this being written when trace logging is enabled.  I would still write to the C# output window though

e.g. https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/roslynLanguageServer.ts#L614

src/lsptoolshost/roslynProtocol.ts
genlu: created a PR to remove resultID as discussed offline:
https://github.com/dotnet/roslyn/pull/75176

genlu:
@dibarbet Could you please take another look? (also I think this doesn't require https://github.com/dotnet/roslyn/pull/75176 to work, since the protocol is still compatible with change in this PR, pls correct me if I was wrong)

src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -1036,6 +1037,7 @@ export async function activateRoslynLanguageServer(
     );
 
     registerLanguageStatusItems(context, languageServer, languageServerEvents);
+    await registerCopilotExtensionAsync(languageServer, _traceChannel);
```
dibarbet, line 1040: I'd recommend not using the trace channel - that is generally just for LSP trace logs.  What it should do is write to the normal output channel only when trace logging is enabled

src/lsptoolshost/copilot.ts
```diff
@@ -0,0 +1,82 @@
...
+                position: {
+                    line: 0,
+                    character: 0,
+                },
+            },
+            async (r) => buildResult(r, relatedFiles)
+        );
+
+        try {
+            await responsePromise;
```
dibarbet, line 74: I'd just inline this

dibarbet (commented):
> @dibarbet Could you please take another look? (also I think this doesn't require https://github.com/dotnet/roslyn/pull/75176 to work, since the protocol is still compatible with change in this PR, pls correct me if I was wrong)

I believe so, since it was optional.

src/lsptoolshost/roslynLanguageServer.ts
genlu: Done!

dibarbet (approved):

PR #7581, opened 2024-09-19
Add a TextDocumentContentProvider for source-generated files

dibarbet:
This allows the Roslyn LSP to return document references with a roslyn-source-generated URI, and we'll turn around and ask the LSP for those files when they're opened.

For now we don't worry about refreshing files if they've changed, just to keep this simple.


src/lsptoolshost/sourceGeneratedFilesContentProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function registerSourceGeneratedFilesContentProvider(
+    context: vscode.ExtensionContext,
+    languageServer: RoslynLanguageServer
+) {
+    context.subscriptions.push(
+        vscode.workspace.registerTextDocumentContentProvider(
+            'roslyn-source-generated',
+            new (class implements vscode.TextDocumentContentProvider {
+                async provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): Promise<string> {
```
dibarbet, line 20: we definitely need to add refresh support in.  There is some interesting behavior where VSCode doesn't actually always call didClose on the document even if the only active view closes (it calls sometimes much later).  So the content gets reused even if the view is closed and another go to def request is made.

I could implement a command to refresh, or refresh on textview changing, but IMHO we should just implement server side refresh instead.

JoeRobich (approved):

JoeRobich:
We should definitely have a test for this. I think it is fine if it comes in a follow up.

PR #7599, opened 2024-09-24
Enable no floating promises lint rule

dibarbet:
Enables the no floating promises eslint rule.

Main refactoring was updating and more consistently handling how we show toasts.  Please take a careful look and make sure that I did not break any places where we were *explicitly* not awaiting the promise.


src/razor/src/document/razorDocumentManager.ts
```diff
@@ -179,8 +179,8 @@ export class RazorDocumentManager implements IRazorDocumentManager {
             this.razorDocumentGenerationInitialized = true;
             const pipeName = generateUuid();
 
-            vscode.commands.executeCommand(razorInitializeCommand, pipeName);
-            this.serverClient.connectNamedPipe(pipeName);
+            await vscode.commands.executeCommand(razorInitializeCommand, pipeName);
```
ryzngard, line 182: Making a note I need to come back and look at this area. I _think_ this is fine at first glance but need to sit down and look more

src/shared/configurationProvider.ts
```diff
@@ -226,58 +233,69 @@ export class BaseVsDbgConfigurationProvider implements vscode.DebugConfiguration
...
                 errorCode === CertToolStatusCodes.CertificateNotTrusted ||
                 errorCode === CertToolStatusCodes.ErrorNoValidCertificateFound
             ) {
-                const labelYes = vscode.l10n.t('Yes');
+                const labelYes: ActionOption = {
+                    title: vscode.l10n.t('Yes'),
+                    action: async () => {
+                        const returnData = await createSelfSignedCert(dotnetPath);
+                        if (returnData.error === null) {
+                            //if the prcess returns 0, returnData.error is null, otherwise the return code can be acessed in returnData.error.code
```
ryzngard, line 248:
```suggestion
                            // if the prpcess returns 0, returnData.error is null, otherwise the return code can be acessed in returnData.error.code
```

ryzngard (approved):
Overall looks good. Made a note of a place that would be a potential cause of bugs but I think all it will do is slightly slow down activation?

gregg-miskelly (approved):
I can't claim to *know* VS Code well enough to say if we should be awaiting vs. fire-and-forgetting in all those cases, but it *looks* reasonable

ryzngard (approved):
Signing off on razor code

src/shared/observers/utils/showMessage.ts
```diff
@@ -0,0 +1,108 @@
...
+    delegateFunc: showMessageFunc,
+    message: string,
+    options: MessageOptions | undefined,
+    ...items: (CommandOption | ActionOption | string)[]
+) {
+    const messageOptions = options === undefined ? {} : options;
+    delegateFunc(message, messageOptions, ...items).then(
+        async (selectedItem) => {
+            if (selectedItem) {
+                const item = items.find((i) => i);
```
JoeRobich, line 84: Should this just be using `selectedItem`?

src/shared/observers/utils/showMessage.ts
```diff
@@ -0,0 +1,108 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { MessageOptions, vscode } from '../../../vscodeAdapter';
+
+/**
+ * Show an error message toast.
+ * This function returns immediately and does not wait for the user to select an option or dismiss the message.
+ */
+export function showErrorMessage(
```
JoeRobich, line 12: Love the consolidation of messaging helpers

src/shared/observers/utils/showMessage.ts
dibarbet: good catch, legacy of multiple re-writes of this method

test/omnisharp/omnisharpUnitTests/logging/errorMessageObserver.test.ts
```diff
@@ -16,7 +15,7 @@ describe('ErrorMessageObserver', () => {
     let errorMessage: string | undefined;
     const observer = new ErrorMessageObserver(vscode);
 
-    vscode.window.showErrorMessage = async (message: string, ..._: string[]) => {
+    vscode.window.showErrorMessage = async <T extends MessageItem>(message: string, ..._items: T[]) => {
         errorMessage = message;
         return Promise.resolve<string>('Done');
```
JoeRobich, line 20: This seems suspect now that we've changed a lot of places to use ActionItems and such.

dibarbet: this is mocking the VSCode API call, which uses `MessageItem` objects.  The actual `errorMessageObserver.ts` will call our helper with `ActionOption` (or w.e other one) which then calls the VSCode API (which we mock here).

JoeRobich (approved):

test/omnisharp/omnisharpIntegrationTests/integrationHelpers.ts
```diff
@@ -17,16 +17,6 @@ export interface ActivationResult {
 }
 
 export async function activateCSharpExtension(): Promise<ActivationResult> {
-    const configuration = vscode.workspace.getConfiguration();
-    await configuration.update(
-        'omnisharp.enableLspDriver',
-        process.env.OMNISHARP_DRIVER === 'lsp' ? true : false,
```
dibarbet, line 23: this env var is never set.  And this configuration update was throwing (and failing tests now that we awaited it).

PR #7605, opened 2024-09-26
update to new registerRelatedFilesProvider API

genlu:


src/lsptoolshost/copilot.ts
```diff
@@ -35,58 +38,60 @@ export async function registerCopilotExtensionAsync(
...
-    await ext.activate();
-    const relatedAPI = ext.exports as CopilotRelatedFilesProviderRegistration | undefined;
-    if (!relatedAPI) {
-        if (isTraceLogLevel) {
-            channel.appendLine(
-                'Incompatible GitHub Copilot extension installed. Skip registeration of C# related files provider.'
-            );
+    ext.activate().then(() => {
+        const relatedAPI = ext.exports as CopilotRelatedFilesProviderRegistration | undefined;
+        if (!relatedAPI) {
```
dibarbet, line 43: Seems like this should work, but I'm not super confident on how it works in JS.  Can you verify that nothing throws if you have an older version of copilot with the previous version of the API?

src/lsptoolshost/copilot.ts
```diff
@@ -35,58 +38,60 @@ export async function registerCopilotExtensionAsync(
...
+            return;
         }
-        return;
-    }
 
-    if (isTraceLogLevel) {
-        channel.appendLine('registeration of C# related files provider for GitHub Copilot extension succeeded.');
-    }
+        if (isTraceLogLevel) {
+            channel.appendLine('registeration of C# related files provider for GitHub Copilot extension succeeded.');
```
dibarbet, line 53:
```suggestion
            channel.appendLine('registration of C# related files provider for GitHub Copilot extension succeeded.');
```

dibarbet (approved):
lgtm assuming this gracesfully handles the previous version of the API

src/lsptoolshost/copilot.ts
genlu: Hmm, it seems to work with old API as well,. My TS knowledge is poor so this is unexpected to me, I thought `relatedAPI ` for copilot extension with old API would be undefined. What I saw under debugger is the registration just work and callback when called got an undefined cancellation token.

PR #7622, opened 2024-10-02
Update registerRelatedFilesProvider API

genlu:
Which was changed recently in an internal repo ([this](https://github.com/github/copilot-client/pull/6001) and [this](https://github.com/github/copilot-client/pull/5950/))

FYI @sandersn

### DIFF
--- a/src/lsptoolshost/copilot.ts
+++ b/src/lsptoolshost/copilot.ts
@@ -16,6 +16,7 @@ interface CopilotRelatedFilesProviderRegistration {
         providerId: { extensionId: string; languageId: string },
         callback: (
             uri: vscode.Uri,
+            context: { flags: Record<string, unknown> },
             cancellationToken?: vscode.CancellationToken
         ) => Promise<{ entries: vscode.Uri[]; traits?: { name: string; value: string }[] }>
     ): vscode.Disposable;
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/copilot.ts
```diff
@@ -55,7 +63,7 @@ export function registerCopilotExtension(languageServer: RoslynLanguageServer, c
             languageId: 'csharp',
         };
 
-        relatedAPI.registerRelatedFilesProvider(id, async (uri, token) => {
+        relatedAPI.registerRelatedFilesProvider(id, async (uri, _, token) => {
```
dibarbet, line 66: Just want to confirm that we don't have issues with either
1.  An old copilot extension calling into this.  I would guess that we would end up discarding the cancellation token as it would be in the second parameter, and token would be undefined.
2.  A new copilot extension calling into an old C# version - where it would pass 3 parameters where we only have two.

If the mismatch is being resolved elsewhere that is fine, but want to make sure.

genlu: I tested both scenarios, they would at least return related files w/o error/crashing. Given that copilot releases at a fast cadence (once every couple of days) and the experiment would last until next C# extension stable release (in the experiment we would control which versions of C# extension to be part of it), I'm not very concerned about mismatch.

dibarbet (approved):

PR #7106, opened 2024-05-07
Add temporary option to allow suppression of recoverable LSP error toasts

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/6977

Add temporary option to suppress LSP recoverable errors until we can resolve the architectural issues causing the failed requests.
Related to https://github.com/dotnet/vscode-csharp/issues/6973, https://github.com/microsoft/vscode-dotnettools/issues/722 and more

Implemented as suggested in https://github.com/microsoft/vscode-languageserver-node/issues/1449#issuecomment-2074436676

Additional notes
1.  VSCode's notification settings intentionally do not allow errors to be suppressed.
2.  Ideally we could look at VSCode's do not disturb settings or notification settings for the extension instead of our own option, but the API for that does not exist.

PR #7624, opened 2024-10-03
Disable all LSP Server toasts

ryzngard:
[DRAFT]

Should we include an option to enable? I believe this should be disabled by default.
Should we filter based on lsp method?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 # Latest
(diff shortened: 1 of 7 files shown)


davidwengier (approved):
I think we should have an option, like Roslyn: https://github.com/dotnet/vscode-csharp/pull/7106

(though I wouldn't call it 'temporary' 😛)

ryzngard:
> I think we should have an option, like Roslyn: #7106
> 
> (though I wouldn't call it 'temporary' 😛)

Honestly think once we get cohosting in a reasonable spot we should enable by default. I guess it's not temporary in that people could turn it off, but temporary that it's off by default

davidwengier:
You're right, and actually once we get cohosting we'd be using the Roslyn option I linked to anyway, so this is definitely temporary.

davidwengier (approved):

PR #7628, opened 2024-10-04
Warn in the statusbar when the active file is not part of the open workspace.

JoeRobich:
To help make it more clear when the active file is not part of the open workspace, we will show a warning in the C# language status bar. This change requires a server side change to send the workspace kind as part of the project context. See https://github.com/dotnet/roslyn/pull/75384

https://github.com/user-attachments/assets/9f6a3513-bed5-4d1c-a227-cb78f90ca8b7


dibarbet (approved):
lgtm, pending server update

PR #7645, opened 2024-10-09
Make sure named pipe awaits server start

ryzngard:
Fixes https://github.com/microsoft/vscode-dotnettools/issues/1470

In cases where the workspace does not contain a razor or cshtml file startup is delayed for razor.
This helps make sure we don't initialize when not needed. That also means that if a user opens a solution that contains a reference
to a project outside of the workspace then startup will be delayed. This is generally fine because it will be
initialized when needed, but there was an issue where the named pipe setup was trying to wait the startup task
that had not been created yet. An await on undefined immediately returns and the project system will never get
information about the project through the named pipe.

### DIFF
--- a/src/razor/src/razorLanguageServerClient.ts
+++ b/src/razor/src/razorLanguageServerClient.ts
@@ -223,7 +223,7 @@ export class RazorLanguageServerClient implements vscode.Disposable {
     }
 
     public async connectNamedPipe(pipeName: string): Promise<void> {
-        await this.startHandle;
+        await this.start();
 
         // Params must match https://github.com/dotnet/razor/blob/92005deac54f3e9d1a4d1d8f04389379cccfa354/src/Razor/src/Microsoft.CodeAnalysis.Razor.Workspaces/Protocol/RazorNamedPipeConnectParams.cs#L9
         await this.sendNotification('razor/namedPipeConnect', { pipeName: pipeName });


src/razor/src/razorLanguageServerClient.ts
```diff
@@ -223,7 +223,7 @@ export class RazorLanguageServerClient implements vscode.Disposable {
     }
 
     public async connectNamedPipe(pipeName: string): Promise<void> {
-        await this.startHandle;
+        await this.start();
```
davidwengier, line 226: I'm not sure I understand this. You said that previous it could wait on `startHandle` before it had been created, due to delayed initialization, which makes sense, but this makes it look like this will actually start things up instead. Doesn't that break the delayed initialization?

This change makes sense if we know that `connectNamedPipe` is only called after something else has started the Razor bits, but if that were true then I don't see how the old code would have been problematic.

Or does `start()` have a wait inside it that waits for a file to appear in the workspace? That would make sense, but if so, perhaps a comment here mentioning that, and why its okay to call start here, would be good to add.

ryzngard: From my understanding this happens anytime roslyn would ask for dynamic file information from razor: https://github.com/dotnet/vscode-csharp/blob/ec1d55dc1b5cf06cf0500d28d4c15565c523b261/src/razor/src/document/razorDocumentManager.ts#L173 

At that point we should be starting up right? Otherwise dynamic file info will never be provided. If that's not the case then I could add this to check if razor is started, add a callback for after it started

davidwengier: Okay, I see it now. There is a gap where a there could be no Razor files in the VS Code workspace, but they do exist in a Roslyn project. Odd, but fair enough. I had a quick look at it seems fine if this happens to be where the language server starts.

davidwengier (approved):

PR #7652, opened 2024-10-10
Add  miscellaneous file open notifications 

JoeRobich:
Gives the user more warning that they are editing a file outside their workspace. 

image

The same notification text is now used in the language status bar when a miscellaneous file is open.

image

Features:
- Allows the user to easily silence the notification at the workspace level.
- A setting was added to silence the notification at the user or workspace level.
- Only shown once a session for a particular document.


dibarbet:
> Should we show the filename in the notification? When the notification text is different multiple notifications can stack up.

Can we potentially throttle it somehow?  For example if a notification is already showing, we update the notification to be 'Multiple open files..." so that further notifications don't stack up?

> What should the message say?
I'm fine iwth the text shown in the picture, seems clear enough

> Should we notify about the same file more than once a session?
I would say no

> Do we need the granularity of disabling at the workspace level?
I would lean towards yes - there might be workspaces that have lots of misc files and others not.  But not 100% sure.

> Should we store the "Do not show ever" setting in the users settings so that it can roam with them?
I think that would be a good idea actually - yeah.

src/lsptoolshost/miscellaneousFileNotifier.ts
```diff
@@ -0,0 +1,58 @@
...
+const NotifiedDocuments = new Set<string>();
+
+export function registerMiscellaneousFileNotifier(
+    context: vscode.ExtensionContext,
+    languageServer: RoslynLanguageServer
+) {
+    context.workspaceState.update(SuppressMiscellaneousFilesToastsOption, undefined);
+
+    languageServer._projectContextService.onActiveFileContextChanged((e) => {
+        const hash = createHash(e.uri.toString(/*skipEncoding:*/ true));
```
dibarbet, line 23: maybe move this down below the other checks, since it could be more expensive?

src/lsptoolshost/miscellaneousFileNotifier.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        if (context.workspaceState.get<boolean>(SuppressMiscellaneousFilesToastsOption, false)) {
+            return;
+        }
+
+        NotifiedDocuments.add(hash);
+
+        const message = vscode.l10n.t(
+            'The active document is not part of the open workspace. Not all language features will be available.'
```
dibarbet, line 43: I was wondering if we wanted to add like a learn more docs link with how they might tell why the file is in misc.  For example, making sure the file is part of a project, making sure the language server has a project loaded, making sure the project is part of the loaded sln, etc.

dibarbet (approved):

src/lsptoolshost/miscellaneousFileNotifier.ts
JoeRobich: I think that is a good idea and we can do it in a follow up. 

@webreidi I see VS has [a page on Misc Files](https://learn.microsoft.com/en-us/visualstudio/ide/reference/miscellaneous-files?view=vs-2022). Is that a good one to link to for now? Should we have a C# centric page created?

JoeRobich: Created draft PR https://github.com/dotnet/vscode-csharp/pull/7654 to explore this further.

webreidi: I think the VS Link will only cause confusion. But we can add to the [FAQ](https://code.visualstudio.com/docs/csharp/cs-dev-kit-faq) the specifics of this.

PR #7654, opened 2024-10-11
Add 'Learn more' to miscellaneous files warning.

JoeRobich:
As [suggested](https://github.com/dotnet/vscode-csharp/pull/7652#discussion_r1796219025) at #7652, a "Learn more" link would be used for educating users about miscellaneous files. 

This PR adds a command for opening a VS doc page for Miscellaneous Files. We may want a C# extension centric page for our users instead.

image

image


src/lsptoolshost/miscellaneousFileNotifier.ts
arkalyanms, line 1: @webreidi would you vet the learn link below? It currently takes you to the VS page. If we want a custom page that has potential ways to fix like "go restore your project", we should introduce a new page. For that, we may also need an understanding of common reasons why a file might go into miscellaneous.

arkalyanms (approved):

PR #7663, opened 2024-10-15
Lookup Razor file context using its document uri.

JoeRobich:
Previously we had been looking up Razor document context from its mapped C# file. This is unnecessary as we can look it up directly from the razor file uri.

### DIFF
--- a/src/lsptoolshost/miscellaneousFileNotifier.ts
+++ b/src/lsptoolshost/miscellaneousFileNotifier.ts
@@ -18,8 +18,12 @@ export function registerMiscellaneousFileNotifier(
     languageServer: RoslynLanguageServer
 ) {
     languageServer._projectContextService.onActiveFileContextChanged((e) => {
-        // Only warn for miscellaneous files when the workspace is fully initialized.
-        if (!e.context._vs_is_miscellaneous || languageServer.state !== ServerState.ProjectInitializationComplete) {
+        // Only warn for C# miscellaneous files when the workspace is fully initialized.
+        if (
+            e.languageId !== 'csharp' ||
+            !e.context._vs_is_miscellaneous ||
+            languageServer.state !== ServerState.ProjectInitializationComplete
+        ) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


davidwengier:
> This is unnecessary

Only if the .razor and .cshtml files are in the Roslyn workspace presumably, which is not 100% of projects that have Razor files in them. Not sure if you care though :)

A possible alternative would be to send the request to the Razor server, but I'm not sure what you do with the project context response, and it will not match Roslyn project Ids etc.

JoeRobich:
> I'm not sure what you do with the project context response, and it will not match Roslyn project Ids etc.

We display the context name in the language status bar as the active file context. Longer term we would want to look at using the returned project contexts in a picker to choose the active file context. In that future, TextDocumentIdentifiers in requests to Roslyn LSP will be updated with the selected active project context.

davidwengier:
> That would mean that this warning is still potentially correct.

"this warning"? I have no context on what this change is fixing, so I can't really say if this is better or worse than before 😛

>  TextDocumentIdentifiers in requests to Roslyn LSP will be updated with the selected active project context

Until cohosting is a thing, there presumably won't be any requests to the Roslyn LSP for Razor documents. Right now its a little odd not asking Razor for project contexts when it is the server that will be sent subsequent requests, but if those requests don't have project contexts then it doesn't matter which contexts the requests don't have 🤷‍♂️

TL;DR: I have no idea if this is a good change or a bad change, but the stakes seem pretty low at the moment so I wouldn't block anything on my comments.

dibarbet (approved):

davidwengier:
I looked at the bug this is fixing, and it seems to me the most appropriate thing here would be to just remove "aspnetcorerazor" from the language checks. The fact is that not all language features _are_ available in Razor files, and since Roslyn isn't the one providing them anyway, it seems odd that it would show any kind of warning at all.

JoeRobich:
> it seems odd that it would show any kind of warning at all

Updated to not show warning toasts for Razor files.

davidwengier (approved):

PR #7684, opened 2024-10-23
Use the .NET runtime extension to find an appropriate .NET install

dibarbet:
C# extension side of https://github.com/microsoft/vscode-dotnettools/issues/1243

Switches from our custom dotnet resolution to call the .net install tool API.  Also ensures that a high enough version of the install tool exists.


src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -5,72 +5,61 @@
...
-        if (!dotnetRuntimePath) {
-            const dotnetPath = await this.findDotnetFromPath();
-            if (dotnetPath) {
-                return {
-                    version: '' /* We don't need to know the version - we've already verified its high enough */,
-                    path: dotnetPath,
-                    env: this.getEnvironmentVariables(dotnetPath),
-                };
+        let dotnetExecutablePath: string;
+        if (commonOptions.dotnetPath) {
```
dibarbet, line 33: this should be deprecated in favor of the options on the .net install tool side.  However I plan to tackle that in a separate PR as we'll need to separate out the O# usage of this option.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -5,72 +5,61 @@
...
+            const findPathRequest: IDotnetFindPathContext = {
+                acquireContext: {
+                    version: DotNetRuntimeVersion,
+                    requestingExtensionId: CSharpExtensionId,
+                    architecture: extensionArchitecture,
+                    mode: 'runtime',
+                },
+                versionSpecRequirement: 'greater_than_or_equal',
+            };
+            const result = await vscode.commands.executeCommand<string>('dotnet.findPath', findPathRequest);
```
dibarbet, line 52: fyi @nagilson - https://github.com/dotnet/vscode-dotnet-runtime/blob/a2e6a5c385190fe8e48e51dca17897b3a209d4f1/vscode-dotnet-runtime-extension/src/test/functional/DotnetCoreAcquisitionExtension.test.ts#L235 appears to indicate it returns a `IDotnetAcquireResult`, but it appears to only ever return a string

src/main.ts
```diff
@@ -83,6 +84,35 @@ export async function activate(
...
+    const dotnetRuntimeExtensionVersion = dotnetRuntimeExtension?.packageJSON.version;
+    if (lt(dotnetRuntimeExtensionVersion, requiredDotnetRuntimeExtensionVersion)) {
+        const button = vscode.l10n.t('Update and reload');
+        const prompt = vscode.l10n.t(
+            'The {0} extension requires version {1} or greater of the .NET Install Tool ({2}) extension.  Please update to continue',
+            context.extension.packageJSON.displayName,
+            requiredDotnetRuntimeExtensionVersion,
+            dotnetRuntimeExtensionId
+        );
+        await vscode.window.showErrorMessage(prompt, button).then(async (selection) => {
```
JoeRobich, line 100: nit: No need to use `.then()`

```suggestion
        const selection = await vscode.window.showErrorMessage(prompt, button);
```

src/main.ts
```diff
@@ -83,6 +84,35 @@ export async function activate(
         requiredPackageIds.push('OmniSharp');
     }
 
+    const dotnetRuntimeExtensionId = 'ms-dotnettools.vscode-dotnet-runtime';
+    const requiredDotnetRuntimeExtensionVersion = '2.2.1';
```
JoeRobich, line 88: Too bad this can't be expressed in the package.json

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -5,72 +5,61 @@
...
-        if (!dotnetRuntimePath) {
-            const dotnetInfo = await this.acquireDotNetProcessDependencies(serverPath);
-            dotnetRuntimePath = path.dirname(dotnetInfo.path);
-        }
 
-        const dotnetExecutableName = this.getDotnetExecutableName();
-        const dotnetExecutablePath = path.join(dotnetRuntimePath, dotnetExecutableName);
-        if (!existsSync(dotnetExecutablePath)) {
-            throw new Error(`Cannot find dotnet path '${dotnetExecutablePath}'`);
+            this.channel.appendLine(`Acquiring .NET runtime version ${DotNetRuntimeVersion}`);
```
JoeRobich, line 41: This might give the impression we are always pulling down a runtime. Perhaps
```suggestion
            this.channel.appendLine(`Locating .NET runtime version ${DotNetRuntimeVersion}`);
```

src/main.ts
dibarbet: Yeah, unfortunately no versions in extension dependencies :(

JoeRobich (approved):

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
nagilson: It can return an `IDotnetAcquireResult` if the `existingDotnetPath` is set. It should always return an `IDotnetAcquireResult`. This was a mistake on my part, thank you for bringing this to my attention. I will make a fix.

dibarbet:
converting to draft - there is one slight modification to the API coming in version `2.2.2` to consistently return a `IDotnetAcquireResult` instead of just a string.  Will update this PR once it is available.

dibarbet:
2.2.3 is now available which also includes support for requiring at least a patch version.  this is ready for review

JoeRobich (approved):

src/main.ts
```diff
@@ -82,6 +83,34 @@ export async function activate(
...
 
+    const dotnetRuntimeExtensionId = 'ms-dotnettools.vscode-dotnet-runtime';
+    const requiredDotnetRuntimeExtensionVersion = '2.2.3';
+
+    const dotnetRuntimeExtension = vscode.extensions.getExtension(dotnetRuntimeExtensionId);
+    const dotnetRuntimeExtensionVersion = dotnetRuntimeExtension?.packageJSON.version;
+    if (lt(dotnetRuntimeExtensionVersion, requiredDotnetRuntimeExtensionVersion)) {
+        const button = vscode.l10n.t('Update and reload');
+        const prompt = vscode.l10n.t(
+            'The {0} extension requires version {1} or greater of the .NET Install Tool ({2}) extension.  Please update to continue',
```
nagilson, line 94:
```suggestion
            'The {0} extension requires version {1} or greater of the .NET Install Tool ({2}) extension. Please update to continue',
```
Same nit here.

src/main.ts
```diff
@@ -82,6 +83,34 @@ export async function activate(
         requiredPackageIds.push('OmniSharp');
     }
 
+    const dotnetRuntimeExtensionId = 'ms-dotnettools.vscode-dotnet-runtime';
+    const requiredDotnetRuntimeExtensionVersion = '2.2.3';
+
+    const dotnetRuntimeExtension = vscode.extensions.getExtension(dotnetRuntimeExtensionId);
```
nagilson, line 89: Note: Our extension requires vscode 1.81.1 or higher. Looks like you already are at ^1.93.0 though so its good.

src/main.ts
nagilson, line 94: Some people who install C# have no idea what .NET is, so we might expect some amount of confusion from users who don't understand why they would need a '.NET Install Tool' -- it may be helpful to include more context, but people don't tend to read long error messages, so this is likely better.

src/main.ts
```diff
@@ -82,6 +83,34 @@ export async function activate(
...
+            dotnetRuntimeExtensionId
+        );
+        const selection = await vscode.window.showErrorMessage(prompt, button);
+        if (selection === button) {
+            await vscode.commands.executeCommand('workbench.extensions.installExtension', dotnetRuntimeExtensionId);
+            await vscode.commands.executeCommand('workbench.action.reloadWindow');
+        } else {
+            throw new Error(
+                vscode.l10n.t(
+                    'Version {0} of the .NET Install Tool ({1}) was not found, will not activate.',
```
nagilson, line 106:
```suggestion
                    'Version {0} of the .NET Install Tool ({1}) was not found, C# will not activate.',
```
nit

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
nagilson: I thought about whether we should keep the existing logic as a backup, but if this has about a mo in pre-release I'm not so concerned.

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -39,38 +36,47 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
     private hostInfo: HostExecutableInformation | undefined;
 
     async getHostExecutableInfo(): Promise<HostExecutableInformation> {
-        let dotnetRuntimePath = commonOptions.dotnetPath;
```
nagilson, line 42: A breaking change document for dotnetPath should be created, IMO. I would also make sure this is remarked in the release notes, and that our existingPath setting is mentioned.
 https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.vscode-dotnet-runtime&ssr=false#overview:~:text=I%20already%20have%20a%20.NET%20Runtime%20or%20SDK%20installed%2C%20and%20I%20want%20to%20use%20it

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
```diff
@@ -119,106 +121,29 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
...
-            const dotnetInfo = await getDotnetInfo([]);
-
-            const extensionArchitecture = await this.getArchitectureFromTargetPlatform();
-            const dotnetArchitecture = dotnetInfo.Architecture;
-
-            // If the extension arhcitecture is defined, we check that it matches the dotnet architecture.
-            // If its undefined we likely have a platform neutral server and assume it can run on any architecture.
-            if (extensionArchitecture && extensionArchitecture !== dotnetArchitecture) {
-                throw new Error(
-                    `The architecture of the .NET runtime (${dotnetArchitecture}) does not match the architecture of the extension (${extensionArchitecture}).`
```
nagilson, line 167: While this will show up in our logging extensively, it's a bit unfortunate that there is no message for the user we would give here now. Not much I can think of for now. Maybe we will make an API to get the error message if it returned undefined (cant change IDotnetAcquireContext to have a string? anymore and error instead of undefined at this point)

nagilson (approved):
Looks great, thank you :) Left some thoughts.

src/main.ts
dibarbet: Yeah - I think this is already fairly long and am hesitant to increase it.  cc @webreidi for thoughts.

l10n/bundle.l10n.json
```diff
@@ -1,6 +1,9 @@
 {
   ".NET Test Log": ".NET Test Log",
   ".NET NuGet Restore": ".NET NuGet Restore",
+  "Update and reload": "Update and reload",
+  "The {0} extension requires version {1} or greater of the .NET Install Tool ({2}) extension.  Please update to continue": "The {0} extension requires version {1} or greater of the .NET Install Tool ({2}) extension.  Please update to continue",
```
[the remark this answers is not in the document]
dibarbet: Habit - I'm a two spaces after periods person ;)

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
dibarbet: I haven't removed this in this PR - plan to in a followup.  What might be an even better option is to migrate existing users of `dotnet.dotnetPath` to the new setting.  We already do that for migrating O# options

src/lsptoolshost/dotnetRuntimeExtensionResolver.ts
dibarbet: Note - this error only showed up in the logs before as well.  It is caught above, and we downloaded the runtime.  If you're including the error in your logs, it doesn't feel that different from before.

dibarbet:
merging - validation looked fine (encountered only known issues that behave the same way in the current version)

PR #7689, opened 2024-10-24
Delay the Misc Files warning toast.

JoeRobich:
Give the project system time to update with any newly added or renamed files. Then, refresh the active document's project context before displaying a warning toast.


src/lsptoolshost/miscellaneousFileNotifier.ts
```diff
@@ -13,11 +13,35 @@ import { languageServerOptions } from '../shared/options';
...
+            // If we have changed active document then do not notify about the previous one.
+            clearTimeout(_notifyTimeout);
+            _notifyTimeout = undefined;
+        }
+
+        if (_documentUriToNotify) {
+            if (e.uri.toString() === _documentUriToNotify.toString()) {
+                // We have rerequested project contexts for the active document
+                // and we can now notify if the document isn't part of the workspace.
+                isNotifyPass = true;
```
dibarbet, line 39: do we need to add to NotifiedDocuments? Looks like we're no longer adding to it anywhere

dibarbet: nvm I see the update

dibarbet (approved):

PR #7717, opened 2024-10-31
Use insiders for code action fix

dibarbet:


JoeRobich:
Would there be value in testing against release and insiders going forward?

dibarbet:
Definitely would like to test insiders - but not on PRs.  There's enough flakiness already that we need to get out of the tests.

dibarbet:
superceded by https://github.com/dotnet/vscode-csharp/pull/7784

PR #7731, opened 2024-11-01
Disable server restart command with devkit

dibarbet:
This gets people into trouble since restarting the server means project information is lost when using devkit.  See https://github.com/dotnet/vscode-csharp/issues/5882

So disable this command when devkit is enabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -1898,7 +1898,7 @@
         "command": "dotnet.restartServer",
         "title": "%command.dotnet.restartServer%",
         "category": ".NET",
-        "enablement": "dotnet.server.activationContext != 'OmniSharp'"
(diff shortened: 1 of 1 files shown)


JoeRobich (approved):

package.json
```diff
@@ -1898,7 +1898,7 @@
         "command": "dotnet.restartServer",
         "title": "%command.dotnet.restartServer%",
         "category": ".NET",
-        "enablement": "dotnet.server.activationContext != 'OmniSharp'"
+        "enablement": "dotnet.server.activationContext == 'Roslyn'"
```
dibarbet, line 1901: 'Roslyn' means standalone C#

PR #7809, opened 2024-11-20
Exclude active document from C# completion's related files

genlu:


CyrusNajmabadi (approved):

src/lsptoolshost/copilot.ts
```diff
@@ -51,12 +51,22 @@ export function registerCopilotExtension(languageServer: RoslynLanguageServer, c
...
+            ) => {
                 if (reports) {
                     for (const report of reports) {
                         if (report._vs_file_paths) {
                             for (const filePath of report._vs_file_paths) {
-                                builder?.push(vscode.Uri.file(filePath));
+                                // The Roslyn related document service would return the active document as related file to itself
+                                // if the code contains reference to the types defined in the same document. Skip it so the active file
+                                // won't be used as additonal context.
+                                const relatedUri = vscode.Uri.file(filePath);
```
dibarbet, line 66: fine with this - but wouldn't it be better to filter on the server (it should know the request document)?  Then each client doesn't have to do this

dibarbet (approved):

src/lsptoolshost/copilot.ts
genlu: Yea, I thought about that, but ended up taking this approach because conceptually the server returning active document itself if referenced makes sense and is consistent with other similar Roslyn APIs (like getting all linked document from an ID), different client might want to do something different with it

PR #7825, opened 2024-11-25
Delete dotnet.dotnetPath option and migrate to replacement

dibarbet:
This removes the `dotnet.dotnetPath` option and adds support for migrating it to the replacement setting (defined by the .NET install tool).

Resolves C# side of https://github.com/microsoft/vscode-dotnettools/issues/1243

Prior to this change, both `dotnet.dotnetPath` and the .NET Install Tool's `dotnetAcquisitionExtension.existingDotnetPath` setting both could influence the runtime that the server used.  It is unnecessary to support both of these options at this point, and only causes needless confusion.  

Here we delete `dotnet.dotnetPath` and migrate existing values to `dotnetAcquisitionExtension.existingDotnetPath`.  This should be unnoticeable in terms of behavior.  The migration is slightly more complicated than other settings as `dotnet.dotnetPath` required a folder containing the `dotnet` executable whereas the .NET install tool setting requires the actual executable path.

OmniSharp as not yet fully migrated over to the .NET install tool for .NET discovery, so O# users of `dotnet.dotnetPath` will be migrated to `omnisharp.dotnetPath`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,6 +7,9 @@
     // Set the context to the workspace folder to allow us to copy files from it.
     "context": ".."
   },
+  "mounts": [
+    "type=volume,target=${containerWorkspaceFolder}/node_modules"
+  ],
   "customizations": {
     "vscode": {
       "settings": {
@@ -37,7 +40,8 @@
         "powershell.integratedConsole.showOnStartup": false,
         "powershell.startAutomatically": false,
         // ms-azure-devops.azure-pipelines settings
-        "azure-pipelines.customSchemaFile": ".vscode/dnceng-schema.json"
+        "azure-pipelines.customSchemaFile": ".vscode/dnceng-schema.json",
+        "jest.jestCommandLine": "./node_modules/.bin/jest"
       },
       "extensions": [
         "ms-dotnettools.csharp",
@@ -52,5 +56,5 @@
       ]
     }
   },
-  "postCreateCommand": "npm ci && npx gulp installDependencies"
+  "postCreateCommand": "npm i && npx gulp installDependencies && npm i -g gulp"
 }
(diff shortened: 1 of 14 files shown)


.devcontainer/devcontainer.json
```diff
@@ -52,5 +56,5 @@
       ]
     }
   },
-  "postCreateCommand": "npm ci && npx gulp installDependencies"
+  "postCreateCommand": "npm i && npx gulp installDependencies && npm i -g gulp"
```
JoeRobich, line 59: Why change from `ci` to `i`? With `ci` it already has the dependency information in the lockfile.

src/utils/dotnetDevCertsHttps.ts
```diff
@@ -5,23 +5,18 @@
...
 
 // Will run `dotnet dev-certs https --trust` to prompt the user to create a trusted self signed certificates. Retruns true if sucessfull.
-export async function createSelfSignedCert(dotnetPath: string): Promise<ExecReturnData> {
-    return await execChildProcess(
-        `${dotnetPath ? dotnetPath : 'dotnet'} dev-certs https --trust`,
-        process.cwd(),
-        process.env
-    );
+export async function createSelfSignedCert(): Promise<ExecReturnData> {
+    const dotnetPath = (commonOptions.useOmnisharpServer ? omnisharpOptions.dotnetPath : undefined) ?? 'dotnet';
```
JoeRobich, line 18: Maybe extract this into a function where it can be written in a more readable form.

JoeRobich (approved):

.devcontainer/devcontainer.json
dibarbet: I was having an issue with jest not getting installed correctly - but I don't think this was the cause.  I will revert

PR #7826, opened 2024-11-26
Don't open closed documents

ryzngard:
This updates the dynamic document system to keep track of edits for closed documents and report those to Roslyn. This helps us avoid creating buffers and opening documents to send content across, which should help avoid any issues with keeping buffers open as well as no longer block the Roslyn LSP on background document changes.

VS Code side of https://github.com/dotnet/roslyn/pull/76050


src/razor/src/csharp/csharpProjectedDocument.ts
```diff
@@ -150,4 +168,18 @@ export class CSharpProjectedDocument implements IProjectedDocument {
...
         this.content = content;
     }
+
+    private updateContent(edits: ServerTextChange[]) {
+        if (edits.length === 0) {
+            return;
+        }
+
+        let content = this.content;
+        for (const edit of edits.reverse()) {
```
davidwengier, line 178: I'm surprised this works, given elsewhere the edits are concated together. Wouldn't we need to insert subsequent edits at position 0?

Though that could be at odds with the Roslyn algorithm that processes the same edits?

ryzngard: that's a good point. I don't think I've actually hit a scenario where they got concat together in testing so probably missed this. Idk why reverse was used initially. Let me look...

davidwengier: Reverse makes sense for a single batch of ordered changes, it means you don't need to track forward when applying edits earlier in the file, and adjust positions later in the file. But the batches would need to be applied in order.

ryzngard: I think I can just update to applying stored edits first instead of doing a concat

ryzngard: This should be better now. I updated to keep track of `Updates` as individual units sent from Razor. How they get applied  remains consistent and the edits being returned on ordered to be sent back to Roslyn

src/razor/src/dynamicFile/dynamicFileInfoHandler.ts
```diff
@@ -33,39 +37,57 @@ export class DynamicFileInfoHandler {
...
-            this.documentManager.roslynActivated = true;
+            const virtualCsharpUri = UriConverter.serialize(razorDocument.csharpDocument.uri);
 
-            // Normally we start receiving dynamic info after Razor is initialized, but if the user had a .razor file open
-            // when they started VS Code, the order is the other way around. This no-ops if Razor is already initialized.
-            await this.documentManager.ensureRazorInitialized();
+            if (this.documentManager.isRazorDocumentOpenInCSharpWorkspace(vscodeUri)) {
+                // Open documents have didOpen/didChange to update the csharp buffer. Razor
+                // does not send edits and instead lets vscode handle them.
+                return new ProvideDynamicFileResponse({ uri: virtualCsharpUri }, null);
```
davidwengier, line 78: Should this reset any stored edits, so that when a document is closed, things have a clean slate?

ie, when a document is opened, LSP will send a didOpen with the full contents, then any future edits, and then when its closed, are we at risk of sending some edits that were made before the open?

src/razor/src/dynamicFile/dynamicFileInfoHandler.ts
```diff
@@ -33,39 +37,57 @@ export class DynamicFileInfoHandler {
...
+            } else {
+                // Closed documents provide edits since the last time they were requested since
+                // there is no open buffer in vscode corresponding to the csharp content.
+                const csharpDocument = razorDocument.csharpDocument as CSharpProjectedDocument;
+                const edits = csharpDocument.getAndApplyEdits();
+
+                return new ProvideDynamicFileResponse({ uri: virtualCsharpUri }, edits ?? null);
```
davidwengier, line 85: Given this can send null, and open docs always do, and thinking about the above comment, I wonder if these two branches can be combined. ie, on open, clear the edits, then provide the dynamic file response with `edits ?? null`.

davidwengier (approved):
I think this makes sense. I suspect there are still plenty of weird races around multiple updates and their edits being applied successfully here and in Roslyn, but it would be silly to block on it I think.

dibarbet (approved):

PR #7843, opened 2024-12-03
Restore WASM debug hook

ryzngard:

PR #7844, opened 2024-12-04
Restore WASM debug hook (#7843)

ryzngard:


src/razor/src/extension.ts
```diff
@@ -53,6 +53,7 @@ import { RazorFormatNewFileHandler } from './formatNewFile/razorFormatNewFileHan
 import { InlayHintHandler } from './inlayHint/inlayHintHandler';
 import { InlayHintResolveHandler } from './inlayHint/inlayHintResolveHandler';
 import { getComponentPaths } from '../../lsptoolshost/builtInComponents';
+import { BlazorDebugConfigurationProvider } from './blazorDebug/blazorDebugConfigurationProvider';
```
phil-allen-msft, line 56: This looks like a copy of line 9.

phil-allen-msft: @dibarbet, perhaps we don't need this particular cherry-pick.

dibarbet: @ryzngard @phil-allen-msft I'm not seeing the fix in the prerelease branch, so as far as I know we need it
https://github.com/dotnet/vscode-csharp/blob/prerelease/src/razor/src/extension.ts

ryzngard: I'll find out why the cherry pick didn't work

dibarbet (approved):

PR #7875, opened 2024-12-12
Don't try to send dynamicInfoChanged if the server isn't running

ryzngard:
razor/dynamicFileInfoChanged can be attempted to be sent if there's a time when the Razor server is running but the Roslyn server isn't. This could happen during shutdown or if some other error occurred. Occasionally seeing this in tests and this change should help the reliability of those tests.

### DIFF
--- a/src/lsptoolshost/roslynLanguageServer.ts
+++ b/src/lsptoolshost/roslynLanguageServer.ts
@@ -797,7 +797,7 @@ export class RoslynLanguageServer {
                 if (this.isRunning()) {
                     this.sendNotification<RazorDynamicFileChangedParams>('razor/dynamicFileInfoChanged', notification)
                 } else {
-                    _channel.debug('Tried to send razor/dynamicFileInfoChanged while server is not running');
+                    _channel.warn('Tried to send razor/dynamicFileInfoChanged while server is not running');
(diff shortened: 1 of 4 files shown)


src/lsptoolshost/roslynLanguageServer.ts
```diff
@@ -793,7 +793,13 @@ export class RoslynLanguageServer {
         vscode.commands.registerCommand(
             DynamicFileInfoHandler.dynamicFileUpdatedCommand,
             async (notification: RazorDynamicFileChangedParams) =>
-                this.sendNotification<RazorDynamicFileChangedParams>('razor/dynamicFileInfoChanged', notification)
+            {
+                if (this.isRunning()) {
+                    this.sendNotification<RazorDynamicFileChangedParams>('razor/dynamicFileInfoChanged', notification)
+                } else {
+                    _channel.debug('Tried to send razor/dynamicFileInfoChanged while server is not running');
```
davidwengier, line 800: If this is rare and only during shutdown, would it be better to log it as a more important message, so that if it appears in a user report of general instability, we'll know something is up with server lifetimes?

davidwengier (approved):

PR #7876, opened 2024-12-13
JSON does not allow trailing commas

iratebadger:
This json is causing errors in other extensions for some reason. Lets use valid json. Filtering the comments is easy, fixing trailing commas is hard.

### DIFF
--- a/themes/vs2019_dark.json
+++ b/themes/vs2019_dark.json
@@ -493,7 +493,7 @@
                 "variable.other.enummember"
             ],
             "settings": {
-                "foreground": "#51B6C4",
+                "foreground": "#51B6C4"
             }
         },
         {
@@ -641,7 +641,7 @@
             "name": "Constant",
             "scope": "variable.other.constant",
             "settings": {
-                "foreground": "#D4D4D4",
+                "foreground": "#D4D4D4"
             }
         },
         {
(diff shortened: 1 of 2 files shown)


iratebadger:
@dotnet-policy-service agree

jasonmalinowski:
@iratebadger: thanks for the pull request; the build failures look like something has just broken in our build system in the last 24 hours or so and we'll need to chase that down.

JoeRobich:
@iratebadger Can you please link to or share examples of the problems this is causing? Not that I am necessarily against this change, but VSCode itself uses both [comments](https://github.com/microsoft/vscode/blob/31092ac1a0f9e68026cd0796f184fd3dfe978069/extensions/theme-defaults/themes/dark_plus.json#L13) and [trailing commas](https://github.com/microsoft/vscode/blob/31092ac1a0f9e68026cd0796f184fd3dfe978069/extensions/theme-defaults/themes/dark_plus.json#L203) in their themes.

jasonmalinowski:
@iratebadger Do you have an example we could use here?

JoeRobich:
@iratebadger Thanks for opening this PR. If you can provide an example of the problem that is being caused by the trailing commas and comments, then we can review the necessity of this change. Closing this for now.

Issue #7165, opened 2024-05-25
Launching a project should use a terminal window for process output


davidfowl:
When you launch a console or .NET project in Visual Studio with F5 there's an external console window that pops up with the output (this can be configured to use the in IDE terminal window as well).

VS 

https://github.com/dotnet/vscode-csharp/assets/95136/3a32dbbf-4f8a-41b1-bd8c-6c0550e5ecf0

In VS code the focus is on the debug output and the console output and debug spew is interleaved making it extremely hard to look at the output:

VS Code

https://github.com/dotnet/vscode-csharp/assets/95136/7c0fdfbe-c2a5-460a-a9b7-2beabcd8d502

Instead, we should show the console output from launches in a terminal window (similar to how to build is done)

WardenGnaw:
For Console Projects, the default will be the `Terminal` output tab or if a user has configured a `csharp.debug.console` setting, it will use that instead.

For ASP.NET projects, due to how VS Code detects when to auto launch the web browser, it is currently required to print the output into the `Debug Console`. 

See https://code.visualstudio.com/docs/editor/debugging#_automatically-open-a-uri-when-debugging-a-server-program

One thought is if we can also just log the ASP.NET URI into the debug console for the `serverReadyAction` but default all stdout/stderr to the Terminal Window and use that as the default

cc: @gregg-miskelly

davidfowl:
I think we should aim to make the experience similar to how the console apps work and use a terminal for output. If using the terminal and launching via the browser is impossible then we need to ask vscode for a new feature.

The console is in important for logs and the debug window is fully of other debug messages that make it hard to visually parse the output.

Anecdotally we taught a C# beginner course and used dotnet run because the debug window is visually noisy.

gregg-miskelly:
I think it would be possible to support using the internal/external terminal, though not particularly cheaply. We still need to fire an event to VS Code to let them know that the web browser is ready, which means that the debugger needs to redirect stdout to itself, but what we could do is:
* Have the root terminal process be vsdbg.exe, or, if we want, some other launcher exe
* Have this exe turn around and write any output to the terminal
* If the output line matches the 'server ready' regex, have it also echo that one line back to VS Code as an output event

davidfowl:
Seems like we have this today when you run without the debugger:


https://github.com/dotnet/vscode-csharp/assets/95136/16514cc2-38c9-4bed-91f1-75cd218fa107

gregg-miskelly:
[...]

PR #7878, opened 2024-12-14
respect csharp.debug.console in all cases

zacuke:
Please see issue https://github.com/microsoft/vscode-dotnettools/issues/1473 for more info.

The [documentation](https://code.visualstudio.com/docs/csharp/debugger-settings#_console-terminal-window) doesn't say that `"csharp.debug.console": "externalTerminal"` is only valid for Console Apps.

I've created this PR to fix the issue of `"csharp.debug.console": "externalTerminal"` being silently ignored in many cases when combined with `"type":"dotnet"`.

This was previously working as expected with  `"type": "coreclr"`.

### DIFF
--- a/src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
+++ b/src/razor/src/blazorDebug/blazorDebugConfigurationProvider.ts
@@ -80,6 +80,7 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
         const program = configuration.hosted ? configuration.program : 'dotnet';
         const cwd = configuration.cwd || '${workspaceFolder}';
         const args = configuration.hosted ? [] : ['run'];
+        const console = configuration.console;
 
         const app = {
             name: SERVER_APP_NAME,
@@ -96,6 +97,7 @@ export class BlazorDebugConfigurationProvider implements vscode.DebugConfigurati
             launchBrowser: {
                 enabled: false,
             },
+            console: console,
             ...configuration.dotNetConfig,
         };
 
--- a/src/shared/assets.ts
+++ b/src/shared/assets.ts
@@ -173,7 +173,7 @@ export class AssetGenerator {
                     }
 
                     // Override console option with user option.
-                    if (forDotnetConfiguration && programLaunchType === ProgramLaunchType.Console && key == 'console') {
+                    if (forDotnetConfiguration && key == 'console') {
                         const consoleOption: string | undefined = AssetGenerator.getConsoleDebugOption();
                         if (consoleOption) {
                             configuration.console = consoleOption;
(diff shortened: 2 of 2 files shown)


JoeRobich:
@zacuke Just to let you know, we are working through some issues with our CI. Once we have it straightened out, I will rerun the checks for you.

zacuke:
After review I do see the documentation says the setting is for _console projects_, but I think it's important to call out that all dotnet projects have a console.

I think people will stumble into this issue because csharp.debug.console is mysteriously ignored. Their _app's console output_ continues to show up in the built-in debug console. Besides the output, the debug console has [known limitations](https://stackoverflow.com/questions/55045070/pass-sigint-while-pressing-ctrlc-on-using-debugger-in-vscode) including [ReadLine](https://stackoverflow.com/questions/41195432/debug-console-window-cannot-accept-console-readline-input-during-debugging). 

Even with web apps, there are cases during debugging we might want to handle ctrl-c or take user input as part of debugging exercise. We might also have web apps that need to handle termination a specific way, or ask for input if a parameter is not passed in on command line. With the debug console it's difficult to debug these edge cases.

I also want to reiterate that opening an external console worked before with type=coreclr regardless of project type (and when using traditional Visual Studio).

I appreciate the consideration of merging this pull request. If nothing else I had fun digging through the extension's code, and I like the direction Microsoft is going with enhancing the C# experience inside vscode.

gregg-miskelly:
Hi @zacuke, 

Thanks for taking the time to investigate this and open a PR! However, I am going to have to decline this because, as you probably are now aware, it is very much intentional that "csharp.debug.console" is only for console apps. There is a feature request issue to support console for web projects (#7165) a fix for this is much more involved that just what you have here because what you have here would break automatic web browser launch.

Note that the `ReadLine` limitation no longer applies (see [documentation](https://code.visualstudio.com/docs/csharp/debugger-settings#_inputting-text-into-the-target-process-when-using-internalconsole)). Ctrl-C still does, so if that is an important feature for you, please open an issue.

In terms of better documenting "csharp.debug.console", that does seem like something worth improving. I opened https://github.com/microsoft/vscode-docs/pull/7871 to try and do that. Let me know if there is anything else you feel like is missing for the documentation.

In VS Code itself, there was already a message to this effect:
image

But again, let me know if there is something you think is worth improving there.

zacuke:
My changes don't seem to break automatic web browser launch in my testing, as the message is duplicated in both Debugger console and External terminal. That seems like an improvement over silently ignoring the setting.

If there's still a known regression, please elaborate and I can try to update my PR changes with further work to satisfy all requirements. If it's just the browser not opening, that doesn't seem to be an issue assuming the content is duplicated on both external window and debug console (which it seems to). Is there some specific condition where the message isn't duplicated thus leading to a browser not opening? Maybe something in vscode changed and that reasoning can be re-evaluated to allow this PR, or something like it, to be merged in.
 
I opened a ticket for Ctrl+C and to try to clarify my story https://github.com/dotnet/vscode-csharp/issues/7882

PR #7907, opened 2025-01-07
Update tsconfig and remove use of `require()`

JoeRobich:


src/lsptoolshost/onAutoInsertFeature.ts
```diff
@@ -22,7 +22,7 @@ import {
 } from 'vscode-languageserver-protocol';
 
 import * as RoslynProtocol from './roslynProtocol';
-import * as UUID from 'vscode-languageclient/lib/common/utils/uuid';
+import { generateUuid } from 'vscode-languageclient/lib/common/utils/uuid';
```
JoeRobich, line 25: webpack didn't love that this path didn't have a `default` export. So updated the import to just pull in the used method.

tsconfig.json
```diff
@@ -1,14 +1,13 @@
 {
     "compilerOptions": {
-        "target": "es6",
-        "module": "commonjs",
+        "target": "ES2022",
+        "module": "Preserve",
```
dibarbet, line 4: should these be upper case or lower case or does it not matter?
https://www.typescriptlang.org/tsconfig/#module

dibarbet (approved):

src/razor/src/blazorDebug/terminateDebugHandler.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as psList from 'ps-list';
+import psList from 'ps-list';
```
JoeRobich, line 6: TS then wanted other packages which didn't have a default export to be imported differently.

tsconfig.json
JoeRobich: I used the TS autocomplete suggested casing.

PR #7921, opened 2025-01-11
Implement context provider

genlu:


src/lsptoolshost/copilot.ts
```diff
@@ -26,74 +27,129 @@ interface CopilotRelatedFilesProviderRegistration {
...
+    getContextProviderAPI(version: string): Promise<ContextProviderApiV1 | undefined>;
 }
 
-export function registerCopilotExtension(languageServer: RoslynLanguageServer, channel: vscode.LogOutputChannel) {
+async function getCopilotAPIsAsync(): Promise<CopilotAPIs | undefined> {
     const ext = vscode.extensions.getExtension('github.copilot');
     if (!ext) {
-        channel.debug('GitHub Copilot extension not installed. Skip registeration of C# related files provider.');
-        return;
+        return undefined;
```
dibarbet, line 36: I would keep the log messages if possible (trace/debug only though)

src/lsptoolshost/copilot.ts
```diff
@@ -26,74 +27,129 @@ interface CopilotRelatedFilesProviderRegistration {
...
-            );
-            return;
-        }
 
-        channel.debug('registration of C# related files provider for GitHub Copilot extension succeeded.');
+    if (!ext.isActive) {
+        try {
+            return await ext.activate();
+        } catch {
+            return undefined;
```
dibarbet, line 43: Is the catch necessary here? There is already a `.catch` in the caller which also logs the error.  
If you're keeping this catch here, it definitely needs to log the error.

src/lsptoolshost/copilot.ts
```diff
@@ -26,74 +27,129 @@ interface CopilotRelatedFilesProviderRegistration {
...
+        }
+    } else {
```
dibarbet, line 45: no need for the if else here.  activate should no-op if already active

src/lsptoolshost/copilot.ts
```diff
@@ -26,74 +27,129 @@ interface CopilotRelatedFilesProviderRegistration {
...
+        channel.debug('Failed to get context provider API from GitHub Copilot extension.');
+        return;
+    }
+
+    contextAPI.registerContextProvider<CodeSnippet>({
+        id: 'csharpContextProvider',
+        selector: [{ language: 'csharp' }],
+        resolver: {
+            resolve: async (request, token) => {
+                return [{ uri: 'testUri', value: 'testValue', additionalUris: [] }];
```
dibarbet, line 132: assuming this is the part that will eventually call out to the server

src/lsptoolshost/copilot.ts
```diff
@@ -26,74 +27,129 @@ interface CopilotRelatedFilesProviderRegistration {
...
 
-        channel.debug('registration of C# related files provider for GitHub Copilot extension succeeded.');
+    if (!ext.isActive) {
+        try {
+            return await ext.activate();
+        } catch {
+            return undefined;
+        }
+    } else {
+        return ext.exports as CopilotAPIs | undefined;
```
dibarbet, line 46: guessing this fails if the interface definition doesn't match?  e.g. the version of the extension doesn't have a `getContextProviderAPI` export?

genlu:
Closed in favor of an alternative approach

PR #7929, opened 2025-01-15
TextMate bug: recognize codeblocks with curly braces in switch statements

jordi1215:
This fixes https://github.com/dotnet/razor/issues/11249.

The problem here is that curly braces are not required within the `case` clause. I added one line in the TextMate syntax to account for this potential usage.

For this to work on visual studio, I will have to manually copy the TextMate syntax to our razor repo, which is tracked here https://github.com/dotnet/razor/pull/11397.

Before
image

After
image


davidwengier (approved):
Lovely

ryzngard (approved):

PR #7961, opened 2025-02-04
Add profiling pipeline

dibarbet:
Passing run which produced mibc
https://dnceng.visualstudio.com/internal/_build/results?buildId=2634936&view=results (internal)

### DIFF
--- a/src/lsptoolshost/profiling.ts
+++ b/src/lsptoolshost/profiling.ts
@@ -1,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { EOL } from 'os';
+import { LogOutputChannel } from 'vscode';
+
+export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.ProcessEnv {
+    let profilingEnvVars = {};
+    if (process.env.ROSLYN_DOTNET_EventPipeOutputPath) {
+        profilingEnvVars = {
+            DOTNET_EnableEventPipe: 1,
+            DOTNET_EventPipeConfig: 'Microsoft-Windows-DotNETRuntime:0x1F000080018:5',
+            DOTNET_EventPipeOutputPath: process.env.ROSLYN_DOTNET_EventPipeOutputPath,
+            DOTNET_ReadyToRun: 0,
+            DOTNET_TieredCompilation: 1,
+            DOTNET_TC_CallCounting: 0,
+            DOTNET_TC_QuickJitForLoops: 1,
+            DOTNET_JitCollect64BitCounts: 1,
+        };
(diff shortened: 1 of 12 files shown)


src/lsptoolshost/profiling.ts
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { EOL } from 'os';
+import { LogOutputChannel } from 'vscode';
+
+export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.ProcessEnv {
```
dibarbet, line 9: see https://devdiv.visualstudio.com/DevDiv/_git/vs-green?path=/src/profiling.ts (internal)

tasks/profilingTasks.ts
```diff
@@ -0,0 +1,112 @@
...
+
+createProfilingTasks();
+
+function createProfilingTasks() {
+    const profilingOutputFolder = path.join(outPath, 'profiling');
+    gulp.task(`profiling:csharp:${basicSlnTestProject}`, async () => {
+        // Ensure the profiling output folder exists, otherwise the outputs will not get written.
+        fs.mkdirSync(path.join(outPath, 'profiling'), { recursive: true });
+
+        await runIntegrationTest(
```
dibarbet, line 21: For now we just run the C# integration tests.  The devkit ones seem to be slightly less stable.  We can consider expanding this in the future.

Most of the data should be the same regardless.

tasks/profilingTasks.ts
```diff
@@ -0,0 +1,112 @@
...
+
+async function mergeTraces(profilingOutputFolder: string) {
+    const files = fs.readdirSync(profilingOutputFolder);
+    const nettraceFiles = files.filter((f) => f.endsWith('.nettrace'));
+    if (nettraceFiles.length === 0) {
+        throw new Error('No .nettrace files found in the profiling output folder.');
+    }
+
+    // Function to spawn a tool
+    function spawnTool(command: string, args: string[], warnOnError = false) {
```
dibarbet, line 54: this is the same as https://devdiv.visualstudio.com/DevDiv/_git/vs-green?path=/azure-pipelines/profiling/mibc.yml&version=GBmain&_a=contents&line=95&lineStyle=plain&lineEnd=96&lineStartColumn=1&lineEndColumn=1 (internal) but converted from powershell

tasks/testHelpers.ts
```diff
@@ -0,0 +1,141 @@
...
+
+import fs from 'fs';
+import * as path from 'path';
+import { rootPath, outPath } from './projectPaths';
+import { prepareVSCodeAndExecuteTests } from '../test/vscodeLauncher';
+
+export const basicSlnTestProject = 'slnWithCsproj';
+export const integrationTestProjects = [basicSlnTestProject];
+
+export async function runDevKitIntegrationTests(
```
dibarbet, line 14: everything here is just moved basically unchanged.

JoeRobich (approved):

PR #7965, opened 2025-02-06
Sync whitespace options even when detectIndentation is on

dibarbet:
Resolves https://github.com/dotnet/vscode-csharp/issues/7963
Resolves https://github.com/dotnet/vscode-csharp/issues/6244
Resolves https://github.com/microsoft/vscode-dotnettools/issues/1651

Previously, we would completely ignore the values of tabSize/indentSize/insertSpaces when `editor.detectIndentation` was enabled (which it is by default).  Instead, a hardcoded value would be sent to the server.

Normally this isn't super noticeable - features like formatting pass the settings specific to that document (based on the detection) when invoking the language server.  However some features like code actions do not specify these options, and so the code would only ever be generated with the hardcoded defaults.

This PR updates it so that we now use the user's settings instead of a hardcoded default even when `editor.detectIndentation` is enabled.  (Note - .editorconfig values still override the user setting as normal).

There is still a caveat here however - we are not using the real detected indentation.  So it is still possible that we generate code that does not match the rest of the document's formatting (because the detected indentation does not match the user's setting). 
 At some point I need to investigate if it is possible to sync the detected per document formatting settings over to the server (so code actions use the detected indentation).

spaces_with_detect_indentation

### DIFF
--- a/src/lsptoolshost/universalEditorConfigProvider.ts
+++ b/src/lsptoolshost/universalEditorConfigProvider.ts
@@ -29,11 +29,11 @@ export function readEquivalentVsCodeConfiguration(serverSideOptionName: string):
 }
 
 function readTabSize(configuration: WorkspaceConfiguration): string {
-    return readValueIfDetectIndentationIsOff(configuration, 'editor.tabSize', '4');
+    return readVsCodeConfigurations<string>(configuration, 'editor.tabSize');
 }
 
 function readIndentSize(configuration: WorkspaceConfiguration): string {
-    const indentSize = readValueIfDetectIndentationIsOff<string>(configuration, 'editor.indentSize', '4');
+    const indentSize = readVsCodeConfigurations<string>(configuration, 'editor.indentSize');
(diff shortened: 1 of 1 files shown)


JoeRobich:
Do changes to `editor.detectIndentation` trigger an option sync? I imagine that formatting options may change when it changes.

dibarbet:
> Do changes to editor.detectIndentation trigger an option sync? I imagine that formatting options may change when it changes.

We don't sync this option specifically, we do sync the other whitespace settings though.  I don't think we need to know about detectIndentation on the server side

> Formatting settings are really at a document level. If there is no source of truth like an .editorconfig, then each document could have different detected indentation.

agreed - and there might be a way in LSP to get the per document settings (you can pass resources to the client when you request settings).  But we'd have to come up with some mechanism on the server side to store that information per document as well.

An alternative for document-only code actions is that the LSP client adds the options to the request (similar to how the formatting commands work).  Of course that wouldn't work for workspace edits either.

JoeRobich (approved):

PR #7984, opened 2025-02-10
Manage information diagnostic severity on the client-side.

JoeRobich:
- A server-side change is being made to report diagnostics with their reported severity (https://github.com/dotnet/roslyn/pull/77145).
- This adds client-side middleware to allow users who prefer to not see information diagnostics in their editor have those diagnostics be reported as hint.
- Middleware required updated vscode-languageclient packages which caused many imported to be updated due changes in what types are exported from which packages.

Resolves https://github.com/dotnet/vscode-csharp/issues/6723


src/lsptoolshost/autoInsert/onAutoInsertFeature.ts
```diff
@@ -56,7 +56,7 @@ export class OnAutoInsertFeature implements DynamicFeature<RoslynProtocol.OnAuto
             this._registrations.delete(id);
         }
     }
-    dispose(): void {
+    clear(): void {
```
JoeRobich, line 59: API change

src/lsptoolshost/diagnostics/fixAllCodeAction.ts
```diff
@@ -58,7 +58,7 @@ export async function getFixAllResponse(
 
                 if (response.edit) {
                     const uriConverter: URIConverter = (value: string): vscode.Uri => UriConverter.deserialize(value);
-                    const protocolConverter = createConverter(uriConverter, true, true);
+                    const protocolConverter = createConverter(uriConverter, true, true, true);
```
JoeRobich, line 61: createConverter added support for Theme Icons

src/lsptoolshost/server/roslynLanguageClient.ts
```diff
@@ -3,13 +3,9 @@
...
  *--------------------------------------------------------------------------------------------*/
 
-import {
-    CancellationToken,
-    LanguageClient,
-    LanguageClientOptions,
-    MessageSignature,
-    ServerOptions,
-} from 'vscode-languageclient/node';
+import { CancellationToken, MessageSignature } from 'vscode-jsonrpc';
```
JoeRobich, line 6: Lots of places where types are only being exported from their implementation package.

src/lsptoolshost/utils/combineDocumentSelectors.ts
```diff
@@ -4,7 +4,10 @@
...
 
 import * as vscode from 'vscode';
+import { DocumentSelector } from 'vscode-languageserver-protocol/lib/common/protocol';
 
-export function combineDocumentSelectors(...selectors: vscode.DocumentSelector[]): vscode.DocumentSelector {
-    return selectors.reduce<(string | vscode.DocumentFilter)[]>((acc, selector) => acc.concat(selector), []);
+export function combineDocumentSelectors(
+    ...selectors: (DocumentSelector | vscode.DocumentSelector)[]
+): vscode.DocumentSelector {
+    return selectors.reduce<(string | vscode.DocumentFilter)[]>((acc, selector) => acc.concat(<any>selector), []);
```
JoeRobich, line 12: DocumentSelector types do not quite match up anymore. I expect this will be resolved in a future "@types/vscode" package update.

src/shared/options.ts
```diff
@@ -511,4 +515,5 @@ export const LanguageServerOptionsThatTriggerReload: ReadonlyArray<keyof Languag
     'componentPaths',
     'enableXamlTools',
     'useServerGC',
+    'reportInformationAsHint',
```
JoeRobich, line 518: Changing 'reportInformationAsHint' will prompt for the user to reload the workspace

tsconfig.json
```diff
@@ -7,6 +7,14 @@
         "lib": [
             "ES2021"
         ],
+        "paths": {
```
JoeRobich, line 10: Exports have changed in these packages. I think if we were able to use a different module resolution behavior they would not be necessary. Added alias paths here and in webpack config.

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -284,6 +282,8 @@ export class RoslynLanguageServer {
                 protocol2Code: UriConverter.deserialize,
             },
             middleware: {
+                provideDiagnostics,
+                provideWorkspaceDiagnostics,
```
davidwengier, line 286: Can/should this be put in the razorLanguageServer too? Presumably we'll now be relaying the information diagnostics back from Roslyn un-changed, so may as well honour the same setting?

JoeRobich: Seems reasonable. =)

package.json
```diff
@@ -839,6 +841,12 @@
             "description": "%configuration.dotnet.backgroundAnalysis.compilerDiagnosticsScope%",
             "order": 30
           },
+          "dotnet.diagnostics.reportInformationAsHint": {
+            "type": "boolean",
+            "default": false,
```
dibarbet, line 846: we should default this to on (report has hint), otherwise we're going to get a ton of complaints when we ship this.

src/lsptoolshost/extensions/roslynLanguageServerExportChannel.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 import * as vscode from 'vscode';
-import { RequestType } from 'vscode-languageclient/node';
+import { RequestType } from 'vscode-jsonrpc';
```
dibarbet, line 6: What happened here - did types get moved around?  Would have thought we would be using the language client API

src/lsptoolshost/utils/combineDocumentSelectors.ts
dibarbet: Shouldn't we be using the vscode document selector types here?  And not the LSP versions?

dibarbet (changes requested):
fine with the majority of the changes - but definitely think we should avoid changing the default for now here.

package.json
```diff
@@ -116,10 +116,11 @@
     "tmp": "0.0.33",
     "uuid": "^9.0.0",
     "vscode-html-languageservice": "^5.0.1",
-    "vscode-jsonrpc": "8.2.0-next.0",
-    "vscode-languageclient": "8.2.0-next.1",
-    "vscode-languageserver-protocol": "3.17.4-next.1",
-    "vscode-languageserver-textdocument": "^1.0.5",
+    "vscode-jsonrpc": "9.0.0-next.7",
```
dibarbet, line 119: I think the new server package version is available now, we should probably include it in this PR

package.json
JoeRobich: Should we prompt users on first launch to make a choice? I think if we ship it on then no one will ever find it. Users who expect info diagnostics will continue to think our extension is broken and may never check GH.

package.json
JoeRobich: I also need to update the diagnostic integration tests for this change.

src/lsptoolshost/utils/combineDocumentSelectors.ts
JoeRobich: we weren't previously but that doesn't mean that we can't. Will look at it.

package.json
JoeRobich: Changed the default to match the current behavior. We can look at prompting the user in the future.

dibarbet (approved):

dibarbet (approved):

src/lsptoolshost/utils/combineDocumentSelectors.ts
JoeRobich: Long story short, we need to use both types of DocumentSelector. ClientLanguageOptions requires the LSP type and various calls require the vscode type. When vscode moves to the newer vscode-languageserver package, I expect these to become equivalent again.

mikes-gh:
@JoeRobich Thank you for giving us the choice.  Been waiting a long time to get my infos back.
I can confirm my infos are now listed in the problems page and I can navigate to them from there.

PR #8000, opened 2025-02-21
Activation perf

dibarbet:
Was looking through startup perf at the request of devkit to see if there were things we could improve.  The first commit is just misc cleanup and improvements - removes a few warnings that get reported in the extension host on startup.

In the second commit I switched to esbuild for bundling instead of webpack.  I had noticed that the time to load extension code (~300ms) was much higher than our actual activation implementation (~30ms).  Wanted to experiment to see if esbuild bundling reduced the output JS code size.

Unfortunately, both code size and time to load the code was essentially unchanged.  However - there is an improvement in build times that I think makes this worth taking (and is expected, esbuild is way faster than webpack).  In the linux leg in particular I'm seeing a roughly 3 min reduction in time
before:
image
after:
image

There are likely other things to improve there as well (e.g. removing eslint from the vsix build step).

At the same time, attempted to cleanup some of the scripts to make things a bit easier to understand.


package.json
```diff
@@ -1412,7 +1408,7 @@
         "order": 9,
         "properties": {
           "dotnet.preferCSharpExtension": {
-            "scope": "resource",
+            "scope": "machine-overridable",
```
JoeRobich, line 1411: I wonder if this should be "window".

scope ↓ / target → | user | remote | workspace | folder
-- | -- | -- | -- | --
application | ✔️ | ✖️ | ✖️ | ✖️
machine | ✖️ | ✔️ | ✖️ | ✖️
window | ✔️ | ✔️ | ✔️ | ✖️
resource | ✔️ | ✔️ | ✔️ | ✔️

dibarbet: You're probably right - since it isn't possible to have a different value for this for different folders (theres only one extension active at a time).

JoeRobich (approved):

JoeRobich (approved):

ryzngard (approved):

PR #8002, opened 2025-02-21
Copy CPU profiles generated in tests to logs folder for upload

dibarbet:


azure-pipelines/test.yml
```diff
@@ -33,6 +33,13 @@ steps:
     DISPLAY: :99.0
     CODE_VERSION: ${{ parameters.testVSCodeVersion }}
 
+- powershell: |
+    $tempPath = [System.IO.Path]::GetTempPath()
+    echo "Temp Path: $tempPath"
+    Get-ChildItem -Path $tempPath -Filter *.cpuprofile | Copy-Item -Destination "$(Build.SourcesDirectory)/out/logs"
```
jasonmalinowski, line 39: Was the regular copy files Azure DevOps task unavailable because of the fact it's temp?

dibarbet: Yeah for two reasons
1.  The pipeline variable for temp path is the pipeline agent temp path, not the machine temp path.
2.  I can't hard code it either as the temp path (and temp path env vars) differ.

Easiest solution was to retrieve it via the actual C# api.

JoeRobich (approved):

PR #8020, opened 2025-02-25
Always set fuse

ryzngard:
Fixes [AB#2397532](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2397532/?view=edit)

It was fuse. `@layout` is missing mappings and for some reason we will happily send diagnostics with -1 as the value

### DIFF
--- a/src/razor/src/razorLanguageServerClient.ts
+++ b/src/razor/src/razorLanguageServerClient.ts
@@ -268,10 +268,8 @@ export class RazorLanguageServerClient implements vscode.Disposable {
             args.push('--SingleServerCompletionSupport');
             args.push('true');
 
-            if (options.forceRuntimeCodeGeneration) {
-                args.push('--ForceRuntimeCodeGeneration');
-                args.push('true');
-            }
+            args.push('--ForceRuntimeCodeGeneration');
+            args.push(options.forceRuntimeCodeGeneration ? 'true' : 'false');
 
             if (options.useNewFormattingEngine) {
                 args.push('--UseNewFormattingEngine');


davidwengier (approved):
Interesting these don't break VS. Or do we not have the same test coverage?

Should we keep the bug open, or log a new one in the Razor repo, to track the real fixes?

ryzngard:
> Interesting these don't break VS. Or do we not have the same test coverage?
> 
> Should we keep the bug open, or log a new one in the Razor repo, to track the real fixes?

I'll open a new one. Doesn't break in VS because the code path depends on `DelegateToCSharpOnDiagnosticPublish` being set to true.

src/razor/src/razorLanguageServerClient.ts
```diff
@@ -268,10 +268,8 @@ export class RazorLanguageServerClient implements vscode.Disposable {
             args.push('--SingleServerCompletionSupport');
             args.push('true');
 
-            if (options.forceRuntimeCodeGeneration) {
-                args.push('--ForceRuntimeCodeGeneration');
-                args.push('true');
-            }
+            args.push('--ForceRuntimeCodeGeneration');
```
dibarbet, line 271: two questions
1.  Does this need to be backported to prerelease?
2.  If we are backporting to prerelease, are we OK with the release going out next week with fuse on?

ryzngard: 1. yes 
2. no

Assuming by doing 1 we prevent fuse from default on in vs code. Will see what the best fix is moving forward. We could probably get closer to VS behavior with https://github.com/dotnet/razor/issues/9150 and avoid part of this problem (although the missing source mapping should still be added, which will be a separate issue)

dibarbet: My apologies, I didn't read the change closely enough.  I see now its switching back to the option.  LGTM

dibarbet (approved):

src/razor/src/razorLanguageServerClient.ts
dibarbet: Also realized we don't have the backport bot here yet - PR for that here - https://github.com/dotnet/vscode-csharp/pull/8021

PR #8027, opened 2025-02-27
Bump Razor to 9.0.0-preview.25125.9

davidwengier:
[View Complete Diff of Changes](https://github.com/dotnet/razor/compare/82bb05d0f43e11b7068addd9c0762b3372580c1f...0184787f9c22a83848ebd40d9a3bcf5f92943ce6?w=1)

@phil-allen-msft @ryzngard this build doesn't pass symbol check, but I'm working on the assumption that the diagnostics fix is more important, and that the C# extension build pipeline won't complain, but please confirm whether that's the way to go, or if we should wait for the symbol issue to be solved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 # 2.68.x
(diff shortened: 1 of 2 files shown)


dibarbet (approved):

phil-allen-msft (approved):

davidwengier:
I will take that as a yes :)

thanks for all your work @dibarbet !

PR #8028, opened 2025-02-27
Only publish cpuprofile logs for integration tests.

JoeRobich:
Since unit tests do not start a VSCode instance there is no log folder to copy into. This was causing CI to always fail.

So, I had to do this a wonky way. Apparently my new `isIntegration` test.yml template parameter couldn't be used for template conditions because it is being sourced from the `isIntegration` variable in the test-matrix.yml. It would be evaluated to `$(isIntegration)` when used in `${{ if }}` statements. It was usable from within scripts as that replacement seems to be made at runtime. So I use a script to set a variable that can then be used from conditions in order to conditionally publish integration test logs.

### DIFF
--- a/azure-pipelines/test.yml
+++ b/azure-pipelines/test.yml
@@ -46,25 +46,24 @@ steps:
     mergeTestResults: true
     testRunTitle: $(System.StageDisplayName) $(Agent.JobName) (Attempt $(System.JobAttempt))
 
-- ${{ if eq(parameters.isIntegration, 'true') }}:
-  - powershell: |
-      $tempPath = [System.IO.Path]::GetTempPath()
-      echo "Temp Path: $tempPath"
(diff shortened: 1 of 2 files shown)


davidwengier (approved):

azure-pipelines/test.yml
```diff
@@ -51,16 +46,28 @@ steps:
     mergeTestResults: true
     testRunTitle: $(System.StageDisplayName) $(Agent.JobName) (Attempt $(System.JobAttempt))
 
+- powershell: |
```
JoeRobich, line 49: The template parameter `isIntegration` is sourced back to a variable in test-matrix.yml. This makes it unusable for `${{ if }}` statements. So instead of excluding steps based on a template condition, I will set a variable which can be used to conditionally run the integration log publishing steps.

PR #8060, opened 2025-03-13
Set the BuildNumber in the release pipeline

JoeRobich:
Sets the BuildNumber in the release pipeline to easily identify what build was being published.

- Set the BuildNumber in the form `# (Test)? Publish $(branch) $(version) #`
  - ex. `# Publish prerelease 2.69.22 #`

[Test Run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2662783&view=results)

### DIFF
--- a/azure-pipelines/release.yml
+++ b/azure-pipelines/release.yml
@@ -113,6 +113,35 @@ extends:
                       Write-Host "##[command]vsce $publishArgs"
                       vsce @publishArgs
                     }
+
+    - stage: 'SetBuildNumber'
+      displayName: 'Set BuildNumbder'
+      jobs:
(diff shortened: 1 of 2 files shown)


ryzngard (approved):

azure-pipelines/release.yml
```diff
@@ -51,6 +51,27 @@ extends:
           runOnce:
             deploy:
               steps:
+              - pwsh: |
```
JoeRobich, line 54: Downside is that this doesn't run until pretty late in the run but the resource variables aren't available until runtime and this is the earliest we can set it.

arunchndr (approved):

PR #8096, opened 2025-03-25
Add .razorExtension

ryzngard:
Depends on the following PRs: 

https://github.com/dotnet/razor/pull/11510
https://github.com/dotnet/roslyn/pull/77715


.vscodeignore
```diff
@@ -7,6 +7,7 @@
 !.razor/**
 !.razorDevKit/**
 !.razoromnisharp/**
+!.razorExtension/** */
```
davidwengier, line 10: Just a guess, but looks odd
```suggestion
!.razorExtension/**
```

src/razor/src/dynamicFile/dynamicFileInfoHandler.ts
```diff
@@ -99,22 +92,25 @@ export class DynamicFileInfoHandler {
...
                 // Open documents have didOpen/didChange to update the csharp buffer. Razor
                 // does not send edits and instead lets vscode handle them.
-                return new ProvideDynamicFileResponse(
-                    { uri: virtualCsharpUri },
-                    null,
-                    csharpDocument.checksum,
-                    csharpDocument.checksumAlgorithm,
-                    csharpDocument.encodingCodePage
-                );
+                // return new ProvideDynamicFileResponse(
```
davidwengier, line 95: Intentional?

ryzngard: Nope. I need to double check this is correct. prior to this I think it was unnecessary to send this and null should be fine. Left a commented out to remind myself then promptly forgot

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -615,6 +615,13 @@ export class RoslynLanguageServer {
             path.join(razorPath, 'Targets', 'Microsoft.NET.Sdk.Razor.DesignTime.targets')
         );
 
+        const razorExtensionPath =
```
dibarbet, line 622: I think we should probably use this support instead of a new specific option for the razorExtensionPath
https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/extensions/builtInComponents.ts#L16

This would move the option into `componentPaths` instead, but the API handles that for you.

dibarbet (changes requested):

src/lsptoolshost/server/roslynLanguageServer.ts
ryzngard: oh good catch, I completely forgot that

dibarbet (approved):

PR #8119, opened 2025-03-31
Revert xamlTools for this week's prerelease update

BretJohnson:
Revert xamlTools to what it was 2 weeks ago, to avoid the problem in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2429317, with references to the wrong version of Microsoft.VisualStudio.Threading.

A proper fix will go to main, but for this week's prerelease update we're just reverting to the xamlTools version currently shipping in prerelease. There aren't any important fixes getting missed, so revert isn't a problem.


BretJohnson:
Closing, to retarget to prerelease

PR #8132, opened 2025-04-03
Acquire aspnetcore runtime when using devkit to avoid double acquisition

dibarbet:
C# Dev Kit requires the aspnetcore runtime, and requests it from the acquisition tool.  If C# requests the regular runtime, the .NET install tool cannot de-duplicate the requests and only download the normal runtime once.  Instead, if using devkit we can just require the aspnetcore runtime.  There should be no change in behavior as devkit already requires it.

Before, you get a download for both the runtime, and the aspnetcore runtime (which includes the normal runtime):
image

After, you just get a single download for the aspnetcore runtime
image

### DIFF
--- a/src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionApi.ts
+++ b/src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionApi.ts
@@ -24,7 +24,7 @@ export interface IDotnetFindPathContext {
 /**
  * https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts
  */
-interface IDotnetAcquireContext {
+export interface IDotnetAcquireContext {
     version: string;
     requestingExtensionId?: string;
     errorConfiguration?: AcquireErrorConfiguration;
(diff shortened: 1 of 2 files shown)


src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionApi.ts
```diff
@@ -24,7 +24,7 @@ export interface IDotnetFindPathContext {
 /**
  * https://github.com/dotnet/vscode-dotnet-runtime/blob/main/vscode-dotnet-runtime-library/src/IDotnetAcquireContext.ts
  */
-interface IDotnetAcquireContext {
+export interface IDotnetAcquireContext {
```
jasonmalinowski, line 27: Is there no npm package for this?

JoeRobich (approved):

src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionApi.ts
dibarbet: I don't believe so - @nagilson there is no NPM package for the .net install tool types right?

nagilson: Unfortunately, no, and this makes me very sad
(we should probably do that)

PR #8135, opened 2025-04-04
Don't show duplicate toast if C# Dev Kit fails to activate

dibarbet:
came up in another discussion - currently the vscode lsp client displays a toast if server startup fails.  In server startup we're awaiting devkit activation - so if devkit activation fails we get two overlapping toasts.

This change moves us to await devkit activation outside of the server startup path.  If devkit fails, we rely on them to display an error.

### DIFF
--- a/src/lsptoolshost/server/roslynLanguageServer.ts
+++ b/src/lsptoolshost/server/roslynLanguageServer.ts
@@ -255,6 +255,11 @@ export class RoslynLanguageServer {
         channel: vscode.LogOutputChannel,
         traceChannel: vscode.OutputChannel
     ): Promise<RoslynLanguageServer> {
+        const devKit = getCSharpDevKit();
+        if (devKit) {
+            await devKit.activate();
+        }
+
         const serverOptions: ServerOptions = async () => {
             return await this.startServer(
                 platformInfo,


arunchndr (approved):

ryzngard (approved):

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -255,6 +255,11 @@ export class RoslynLanguageServer {
         channel: vscode.LogOutputChannel,
         traceChannel: vscode.OutputChannel
     ): Promise<RoslynLanguageServer> {
+        const devKit = getCSharpDevKit();
+        if (devKit) {
+            await devKit.activate();
```
AArnott, line 260: Allowing the exception to propagate will still produce an error toast in VS Code. But I don't know... maybe it'll be smaller?

AArnott (approved):

src/lsptoolshost/server/roslynLanguageServer.ts
dibarbet: I don't believe this will show a toast - this is in a promise that isn't awaited in activation and not part of a callback from vscode .  We write an error, but I haven't managed to get a toast to display in any of the editors - I only get a devkit toast.

Issue #8149, opened 2025-04-08
Command '.NET: Generate Assets for Build and Debug' is missing


Oblomoff:
Type: <b>Bug</b>

Command '.NET: Generate Assets for Build and Debug' is missing from Command Panel (F1).
      
## Environment information ##

**VSCode version**: 1.99.0
**C# Extension**: 2.72.27
**Using OmniSharp**: false

<details><summary>Dotnet Information</summary>
.NET SDK:
 Version:           9.0.201
 Commit:            071aaccdc2
 Workload version:  9.0.200-manifests.a3a1a094
 MSBuild version:   17.13.13+1c2026462

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26100
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.201\

.NET workloads installed:
 [aspire]
   Installation Source: VS 17.13.35919.96
   Manifest Version:    8.2.2/8.0.100
   Manifest Path:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.aspire\8.2.2\WorkloadManifest.json
   Install Type:              Msi

Configured to use loose manifests when installing new manifests.

Host:
  Version:      9.0.3
  Architecture: x64
  Commit:       831d23e561

.NET SDKs installed:
  8.0.303 [C:\Program Files\dotnet\sdk]
  9.0.201 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 9.0.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.36 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
[...]

PR #8169, opened 2025-04-11
Support showing Generate Build and Debug assets for CDK

WardenGnaw:
This PR re-enables the .NET: Generate Build and Debug Assets when C# Dev Kit is installed. However, it will prompt users first to utilize the new dynamic configurations, but if users really want to use the older configuration style, they can cancel the modal dialog and it will work similar to how if they just have C# extension installed.

GenerateAssetsCDK

Related: #8149


package.json
```diff
@@ -1828,7 +1828,7 @@
         "command": "dotnet.generateAssets",
         "title": "%command.dotnet.generateAssets.currentProject%",
         "category": ".NET",
-        "enablement": "dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp'"
+        "enablement": "dotnet.server.activationContext == 'RoslynDevKit' || dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp'"
```
dibarbet, line 1831: Should be fine to delete the enablement clause entirely - Roslyn/RoslynDevKit and OmniSharp are the only options

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -51,9 +51,54 @@ export function registerDebugger(
...
+            await generateAssets(workspaceInformationProvider, selectedIndex);
+        })
     );
 }
+
+async function promptForDevKitDebugConfigurations(): Promise<boolean> {
+    if (getCSharpDevKit()) {
+        let result: boolean | undefined = undefined;
+
+        while (result === undefined) {
```
dibarbet, line 70: as far as I know the while loop isn't needed - `await vscode.window.showInformationMessage` will block until either an option is selected or the dialog is dismissed.

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -51,9 +51,54 @@ export function registerDebugger(
...
+
+            if (dialogResult === labelYes) {
+                await vscode.commands.executeCommand('workbench.action.debug.selectandstart', 'dotnet');
+                result = false;
+            } else if (dialogResult === labelMoreInfo) {
+                const launchjsonDescriptionURL = 'https://aka.ms/VSCode-CS-DynamicDebugConfig';
+                await vscode.env.openExternal(vscode.Uri.parse(launchjsonDescriptionURL));
+            } else if (dialogResult === undefined) {
+                // User cancelled dialog and wishes to continue generating assets.
+                result = true;
```
gregg-miskelly, line 96: Rather than doing this with a cancel button, can we add a `No` button that returns 'true' and either get rid of the 'Cancel' button, of if that isn't possible, have the 'Cancel' button return false? I feel like a button called 'Cancel' should cancel the command.

package.json
gregg-miskelly: We still want an enablement command, but maybe there is a better one -- we only want the command to be enabled when a language server is running. See [this commit](https://github.com/dotnet/vscode-csharp/commit/357cefbdb9921831b0a8c93bcc583829d80a7fcc).

dibarbet: Ah I missed that.  Likely we should create a different 'server activated context' (would help with the perf work I was doing before), but I won't block on that.

src/lsptoolshost/debugger/debugger.ts
WardenGnaw: `await vscode.window.showInformationMessage` will block but the loop is specifically for if the user selected `More Information` and the dialog will just show up again.

src/lsptoolshost/debugger/debugger.ts
WardenGnaw: Yeah, I'm not seeing a way ti hide "cancel" unless its non-modal but then we also lose the description.

gregg-miskelly (approved):

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -51,9 +51,59 @@ export function registerDebugger(
...
+
+            if (dialogResult === labelYes) {
+                await vscode.commands.executeCommand('workbench.action.debug.selectandstart', 'dotnet');
+                result = false;
+            } else if (dialogResult === labelNo) {
+                // User cancelled dialog and wishes to continue generating assets.
+                result = true;
+            } else if (dialogResult === labelMoreInfo) {
+                const launchjsonDescriptionURL = 'https://aka.ms/VSCode-CS-DynamicDebugConfig';
+                await vscode.env.openExternal(vscode.Uri.parse(launchjsonDescriptionURL));
```
JoeRobich, line 98: Should this branch set the `result` variable?

JoeRobich: Nevermind. I see the loop now.

dibarbet (approved):

PR #8187, opened 2025-04-14
Support new context resolve handler 

akhera99:
Same as https://github.com/dotnet/vscode-csharp/pull/8126 but with the roslyn version updated.


akhera99:
Could just update Gen's PR. Closing out

PR #8189, opened 2025-04-16
Add cohost option to Razor

davidwengier:
Goes with [roslyn] and [razor] to enable the start of Razor cohosting in VS Code.

Mainly just adding a new option, and using it to turn off a bunch of things, but also a slight tweak to configuration names which I'll call out.

### DIFF
--- a/src/lsptoolshost/options/optionNameConverter.ts
+++ b/src/lsptoolshost/options/optionNameConverter.ts
@@ -3,16 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import assert from 'node:assert';
-
 export function convertServerOptionNameToClientConfigurationName(section: string): string | null {
     // Server name would be in format {languageName}|{grouping}.{name} or
-    // {grouping}.{name} if this option can be applied to multiple languages.
+    // {grouping}.{name} if this option can be applied to multiple languages or
+    // just {name} if this option is not in a group.
     const languageNameIndex = section.indexOf('|');
     if (languageNameIndex == -1 || section.substring(0, languageNameIndex) == 'csharp') {
         // 1. locate the last dot to find the {name} part.
         const lastDotIndex = section.lastIndexOf('.');
-        assert(lastDotIndex !== -1, `There is no . in ${section}.`);
         const optionName = section.substring(lastDotIndex + 1);
 
         // 2. Get {grouping} part.
@@ -25,7 +23,7 @@ export function convertServerOptionNameToClientConfigurationName(section: string
         // Example:
         // Grouping: implement_type
         // Name: dotnet_insertion_behavior
-        // Expect result is: dotnet.implmentType.insertionBehavior
+        // Expect result is: dotnet.implementType.insertionBehavior
         const prefixes = ['dotnet', 'csharp'];
         const optionNamePrefix = getPrefix(optionName, prefixes);
 
@@ -34,9 +32,11 @@ export function convertServerOptionNameToClientConfigurationName(section: string
         // Finally, convert everything to camel case and put them together.
         const camelCaseGroupName = convertToCamelCase(optionGroupName, '_');
         const camelCaseFeatureName = convertToCamelCase(featureName, '_');
-        return optionNamePrefix == ''
-            ? camelCaseGroupName.concat('.', camelCaseFeatureName)
-            : convertToCamelCase(optionNamePrefix, '_').concat('.', camelCaseGroupName, '.', camelCaseFeatureName);
+        const camelCasePrefixName = convertToCamelCase(optionNamePrefix, '_');
+
+        // Concatenate the three parts together, with dots as separators, but only if they are not empty.
+        const parts = [camelCasePrefixName, camelCaseGroupName, camelCaseFeatureName].filter((part) => part !== '');
+        return parts.join('.');
     }
 
     return null;
--- a/package.json
+++ b/package.json
@@ -1536,6 +1536,11 @@
             "type": "boolean",
             "default": true,
             "description": "%configuration.razor.languageServer.useNewFormattingEngine%"
+          },
+          "razor.languageServer.cohostingEnabled": {
+            "type": "boolean",
+            "default": false,
+            "description": "%configuration.razor.languageServer.cohostingEnabled%"
           }
         }
       },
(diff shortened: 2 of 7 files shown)


src/lsptoolshost/options/optionNameConverter.ts
```diff
@@ -3,16 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import assert from 'node:assert';
-
 export function convertServerOptionNameToClientConfigurationName(section: string): string | null {
```
davidwengier, line 6: Changes in this method are to allow the [`WorkspaceConfigurationOptionsStorage.SourceGeneratorExecution`](https://github.com/dotnet/roslyn/pull/78167/files#diff-ed915d03999e9221ff42b872c1c81d6dadb4cb6465b38fcb290939c403305653R63) option to be synced, and for Razor to check [its own options](https://github.com/dotnet/razor/pull/11742/files#diff-72e220682be1a76fdf9e2a9ee992d75e5f9cd5737f82f3e1c56f8fdc585b2234R55-R57)

src/shared/options.ts
```diff
@@ -430,6 +431,9 @@ class RazorOptionsImpl implements RazorOptions {
     public get razorServerPath() {
         return readOption<string>('razor.languageServer.directory', '');
     }
+    public get cohostingEnabled() {
```
dibarbet, line 434: I assume this option requires a reload?  If so we should add it to the needs reload options

src/lsptoolshost/options/optionNameConverter.ts
```diff
@@ -3,16 +3,14 @@
...
     // Server name would be in format {languageName}|{grouping}.{name} or
-    // {grouping}.{name} if this option can be applied to multiple languages.
+    // {grouping}.{name} if this option can be applied to multiple languages or
+    // just {name} if this option is not in a group.
```
dibarbet, line 9: which options don't have a group?  I might be misremembering but they should all be `csharp` / `dotnet` / `razor`?

dibarbet (commented):
A couple things we need to think about eventually
1.  Do we want to group Razor server logs wit hthe C# logs?  Or should we split them out instead?
2.  Razor report issue - does the command need to be updated to pull different logs / describe the cohosting option value?
3.   For VSCode, the experience is using the same language server instance as C# right?  Just with language specific handlers and dynamic registration of those features?
4.   There's a lot of random feature related code in the client side, e.g. https://github.com/dotnet/vscode-csharp/blob/main/src/razor/src/completion/completionHandler.ts - how much of this applies to cohosting?

src/shared/options.ts
davidwengier: I think all Razor options currently need a reload :)

davidwengier:
> A couple things we need to think about eventually

Very good questions! This is definitely just a start on cohosting, and https://github.com/dotnet/razor/pull/11742 is only adding semantic tokens support specifically because it allows us to punt on some of these things and do them in separate PRs.
 
> 1. Do we want to group Razor server logs wit hthe C# logs?  Or should we split them out instead?

Logs from Razor currently come from [here](https://github.com/dotnet/razor/pull/11742/files#diff-21aece6ecf41ca0eb514d9afac87294819df187e10f9b04f9d58d0fc577b9e8cR52) and it would be pretty trivial to use a custom message and put them back in the Razor output window. 

> 2. Razor report issue - does the command need to be updated to pull different logs / describe the cohosting option value?

Probably.

> 3. For VSCode, the experience is using the same language server instance as C# right?  Just with language specific handlers and dynamic registration of those features?

Thats the end goal, yes.

> 4. There's a lot of random feature related code in the client side, e.g. [`main`/src/razor/src/completion/completionHandler.ts](https://github.com/dotnet/vscode-csharp/blob/main/src/razor/src/completion/completionHandler.ts) - how much of this applies to cohosting?

The bits that apply are the ones that maintain Html virtual documents, and forward requests on to them. That's why I started with semantic tokens, because it doesn't. Those endpoints will have to move though, to the roslyn server, and some of our services will either need to be able to take both roslyn and razor servers as parameters, or more likely we can use a common interface. There is more work on the Razor side before that is needed though.

src/lsptoolshost/options/optionNameConverter.ts
davidwengier: This one!
https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Features/Options/WorkspaceConfigurationOptionsStorage.cs#L19

dibarbet: Hmm - that one shouldn't really be exposed to VSCode - setting to balanced will break a lot of things.  We haven't implemented the hooks to run sg on build or save or anything like that.

davidwengier: I added it because we need to manually set it to automatic in order to dogfood cohosting. @chsienki is working on making that unnecessary in the long term.

Though it sounds like you're saying the default would be automatic? That's a surprise based on the setting default, but if the code path in vs code is different enough to make the setting unnecessary, that's fine by me!

dibarbet: See https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Microsoft.CodeAnalysis.LanguageServer/Program.cs#L106

davidwengier: Oh excellent! Thank you

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -617,15 +617,23 @@ export class RoslynLanguageServer {
...
                 : razorOptions.razorServerPath;
 
-        args.push('--razorSourceGenerator', path.join(razorPath, 'Microsoft.CodeAnalysis.Razor.Compiler.dll'));
+        let razorComponentPath = '';
+        getComponentPaths('razorExtension', languageServerOptions).forEach((extPath) => {
+            additionalExtensionPaths.push(extPath);
+            razorComponentPath = path.dirname(extPath);
+        });
+
+        // If cohosting is enabled we get the source generator from the razor component path
```
ryzngard, line 626: Shouldn't they be the same?

ryzngard (approved):

src/lsptoolshost/server/roslynLanguageServer.ts
davidwengier: In a real build, yes they'd be the same, but for local development not necessarily. If you're only working on cohosting, then there is no need to specify a local path for rzls at all, since it won't run. Plus this makes more sense in future when the .razor folder will presumably go away entirely.

dibarbet (approved):

PR #8210, opened 2025-04-24
Support Html requests in cohosting

davidwengier:
Part of https://github.com/dotnet/razor/issues/11759

This is the guts of how html requests work in cohosting. Relatively simple, most of the work is now on the server. Also adds support for document color and color presentation as the first two test candidates.

### DIFF
--- a/src/lsptoolshost/razor/htmlDocumentManager.ts
+++ b/src/lsptoolshost/razor/htmlDocumentManager.ts
@@ -1,0 +1,118 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { RazorLogger } from '../../razor/src/razorLogger';
+import { PlatformInformation } from '../../shared/platform';
+import { getUriPath } from '../../razor/src/uriPaths';
+import { virtualHtmlSuffix } from '../../razor/src/razorConventions';
+import { HtmlDocumentContentProvider } from './htmlDocumentContentProvider';
+import { HtmlDocument } from './htmlDocument';
+
+export class HtmlDocumentManager {
+    private readonly htmlDocuments: { [hostDocumentPath: string]: HtmlDocument } = {};
+    private readonly contentProvider: HtmlDocumentContentProvider;
+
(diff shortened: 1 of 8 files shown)


src/lsptoolshost/razor/htmlDocumentManager.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        this.logger.logVerbose(`New content for '${uri}', updating '${document.path}'.`);
+
+        document.setContent(text);
+
+        this.contentProvider.fireDidChange(document.uri);
+    }
+
+    private async closeDocument(uri: vscode.Uri) {
+        const document = await this.getDocument(uri);
```
ryzngard, line 54: Should this be `this.findDocument` so it doesn't add one if it didn't exist?

src/lsptoolshost/razor/razorEndpoints.ts
```diff
@@ -5,16 +5,76 @@
...
-    razorLogger: RazorLogger
+    razorLogger: RazorLogger,
+    platformInfo: PlatformInformation
 ) {
     const logNotificationType = new NotificationType<LogMessageParams>('razor/log');
     languageServer.registerOnNotificationWithParams(logNotificationType, (params) =>
         razorLogger.log(params.message, params.type)
     );
+
+    if (!razorOptions.cohostingEnabled) {
```
ryzngard, line 39: Do we need to remove registration of dynamicFile endpoints if cohosting is on?

ryzngard (approved):

src/lsptoolshost/razor/htmlDocumentManager.ts
davidwengier: Good catch, thanks

src/lsptoolshost/razor/razorEndpoints.ts
davidwengier: Yeah, I think so. I think we never call them (though tbh I've put almost no effort into initialization yet)

ryzngard: can be another pr for sure

src/lsptoolshost/razor/htmlDocumentManager.ts
davidwengier: Oh my god, thank you again. This comment totally found a bug, I was listening for the wrong document being closed, and if I didn't incorrectly call `getDocument` the whole thing would have stopped working. Fixed to listen to the Razor document now.

PR #8250, opened 2025-05-06
Add FAR across razor and cs files

ryzngard:
Renames tests to reference instead of definition, but I kept the definition test there since it's sort of under the umbrella of that and naming is hard.

Adds a new test that uses a code behind to find references that exist in both cs and razor.

Found an issue with why some tests are hard to write. I added documentation of my findings but haven't found a good solution, just a workaround to avoid getting into that state. We'll see how bad that gets because it may mean that tests will race against eachother. It also may mean my fix isn't really a fix at all and we'll need something better...


test/lsptoolshost/integrationTests/integrationHelpers.ts
```diff
@@ -13,7 +13,19 @@ import testAssetWorkspace from './testAssets/testAssetWorkspace';
...
+    // properly with CSharp. That means allowing a restart of the CSharp server
+    // will improperly break that state. This will be fixed in cohosting and
+    // could be potentially fixed now but I timeboxed to unblock adding more tests.
+    // Without this the symptoms in a test will be that all razor files are considered
+    // in the misc workspace.
+    await activateCSharpExtension(false);
+}
+
+export async function activateCSharpExtension(allowRestart?: boolean): Promise<void> {
+    allowRestart = allowRestart ?? true;
```
dibarbet, line 27: use default parameters perhaps?

test/lsptoolshost/integrationTests/integrationHelpers.ts
```diff
@@ -50,7 +62,7 @@ export async function activateCSharpExtension(): Promise<void> {
     console.log('ms-dotnettools.csharp activated');
     console.log(`Extension Log Directory: ${csharpExtension.exports.logDirectory}`);
 
-    if (shouldRestart) {
+    if (shouldRestart && allowRestart) {
```
dibarbet, line 65: this has a good chance of being flaky.  Without restarting after a restore we have no consistent project load state in the test.  And other tests may have changed various settings that impact this running test.

C# devkit has a similar issue (where we cannot restart) - but it is resolved by running each suite (file) in a separate test run *and* devkit project load ensures that all restores and everything is complete.  We could consider running the razor tests only in the devkit mode.

ryzngard: Yea I had a feeling that was the answer. I have to head out but I'll reach out tomorrow and get something more solid here. It's possible we can gracefully handle a restart in razor with some extra work.

davidwengier: I had similar thoughts. I thought the restart was there because we might have had to download the .NET runtime or something, so could see that not working in CI for sure.

I'm curious what state do we need between tests in Razor, that this causes problems?

ryzngard: Talked offline, we can use DevKit tests to avoid restart

alexgav (approved):
:shipit:

test/razor/razorIntegrationTests/reference.integration.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    test('Go To Definition', async () => {
+        if (!integrationHelpers.isRazorWorkspace(vscode.workspace)) {
+            return;
+        }
+
+        const position = new vscode.Position(6, 41);
+
+        await integrationHelpers.waitForExpectedResult<vscode.Location[]>(
+            async () =>
+                await vscode.commands.executeCommand(
```
dibarbet, line 38: is this done in waitForExpectedResults in order to retry?  Generally awaiting the command should return the results

dibarbet (approved):

test/razor/razorIntegrationTests/reference.integration.test.ts
ryzngard: Yea, it's just for the retry logic. Since this now is actually using project information instead of misc project it can happen before syncing csharp if the test runs too fast.

PR #8251, opened 2025-05-06
Fix custom Razor commands for cohosting

davidwengier:
Fixes https://github.com/dotnet/razor/issues/11785

Makes the "Report Razor issue" command work with cohosting, and hides the "Show Razor C#" command (not possible in cohosting) and "Show Razor Html" command (I'm lazy and I don't think it has any value).

The "Report issue" command still gets the Html content, for whatever its worth.


JoeRobich (approved):

ryzngard (approved):

dibarbet:
> "Show Razor C#" command (not possible in cohosting) 

Is it possible to go to the source generated razor file using the normal source generator opening support we have?

davidwengier:
Yeah, I'm sure that would work too. When I say "not possible", I really mean "not possible now". If we reuse the existing generated document provider, we'd probably need to go to the server to find out the filename/documentId, but if we're going to the server we could also just get the contents of the file :)

PR #8281, opened 2025-05-12
Upgrade the LSP client library for better logging support 

jasonmalinowski:
10.0.0-next.15 brings full support for LogOutputChannels which cleans up the logging of severities and timestamps. Otherwise we end up with confusing duplicate severities being listed with duplicate timesamps.


src/lsptoolshost/activate.ts
```diff
@@ -51,8 +51,7 @@ export async function activateRoslynLanguageServer(
     // Create a channel for outputting general logs from the language server.
     _channel = outputChannel;
     // Create a separate channel for outputting trace logs - these are incredibly verbose and make other logs very difficult to see.
-    // The trace channel verbosity is controlled by the _channel verbosity.
-    _traceChannel = vscode.window.createOutputChannel(vscode.l10n.t('C# LSP Trace Logs'));
+    _traceChannel = vscode.window.createOutputChannel(vscode.l10n.t('C# LSP Trace Logs'), { log: true });
```
dibarbet, line 54: A slightly interesting question.  Now that both of these are logger output windows, do we now have a separate log level configuration for the trace window?  

Previously I manually tied it to the C# output window log level (by detecting changes to the level, and updating the LSP clients trace level), but maybe we shouldn't do that any more?

If so, we should remove the code that updates the trace log level based on the C# output window level, and update Support.md to mention how to get the trace lgos.

.vscode/tasks.json
```diff
@@ -15,7 +15,8 @@
             "group": {
                 "kind": "build",
                 "isDefault": true
-            }
+            },
+            "problemMatcher": "$tsc-watch",
```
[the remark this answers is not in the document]
jasonmalinowski: > What was mandating this?

Me wanting errors to appear when I hit F5 and it failed to build.

> And is tsc-watch correct even if we're not doing a live watch?

Ah good call. Will fix.

src/lsptoolshost/activate.ts
jasonmalinowski: They both have a setting now yes. I was still noticing that it seemed I had to set the setting for the base one to also get the other one. Where is that code?

dibarbet: should be https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/server/roslynLanguageServer.ts#L184

jasonmalinowski: Fixed.

dibarbet (approved):

PR #8311, opened 2025-05-20
Porting more completion tests and adding text change methods

alexgav:
Porting most completion test from VS (all applicable) and adding text change methods that ignore cancellation exceptions.

### DIFF
--- a/test/razor/razorIntegrationTests/completion.integration.test.ts
+++ b/test/razor/razorIntegrationTests/completion.integration.test.ts
@@ -44,6 +44,8 @@ integrationHelpers.describeIfDevKit(`Razor Hover ${testAssetWorkspace.descriptio
             vscode.CompletionItemKind.TypeParameter,
             'text'
         );
+
+        await clearLine(4);
     });
 
     test('Div Tag', async () => {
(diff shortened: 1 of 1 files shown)


test/razor/razorIntegrationTests/completion.integration.test.ts
```diff
@@ -65,8 +67,128 @@ integrationHelpers.describeIfDevKit(`Razor Hover ${testAssetWorkspace.descriptio
...
+        if (!activeEditor) {
+            throw new Error('No active editor');
+        }
+
+        try {
+            await activeEditor.edit((builder) => {
+                builder.insert(position, text);
+            });
+        } catch (error) {
+            if (error instanceof vscode.CancellationError) {
```
dibarbet, line 162: It seems a bit suspicious to be catching the cancellation error here specifically.  It looks like the test itself is attempting to insert the text, but the insert gets cancelled.  Wouldn't that fail the test because the text did not successfully get applied?

I wouldn't expect the actual insertion to get cancelled in general, unless someone else is changes the file underneath.  I would generally expect other requests (e.g. a classification request) to get cancelled if you change the file - but I don't think that would cause the edit application itself to throw a cancellation?

alexgav: Though looking at the failures from the test run, I am catching them in the wrong place. At least some of the time synchronization appears to happen after the insertion/deletion (e.g. outside of insertText). I guess I could wrap test method itself including and after the modification operations.

dibarbet: > At least some of the time synchronization appears to happen after the insertion/deletion (e.g. outside of insertText). I guess I could wrap test method itself including and after the modification operations.

So if the synchronizer is changing the document underneath you may have to catch the exception here, think I get that part now (since the edit may fail).

If you're seeing them thrown outside of the insert text part, you probably want to catch them in the actual synchronization code path then?  Unless some caller of it needs to observe the cancellation.

alexgav: it doesn't throw explicitly, at least not that I can easily see - I'll debug through that some more

https://github.com/dotnet/vscode-csharp/blob/3c12e2709a71838f2e2f0b62b0a08655d514c261/src/razor/src/document/razorDocumentSynchronizer.ts#L143

dibarbet: Ah it is throwing a promise rejection - https://github.com/dotnet/vscode-csharp/blob/3c12e2709a71838f2e2f0b62b0a08655d514c261/src/razor/src/document/razorDocumentSynchronizer.ts#L155 which I believe if unhandled will cause jest to fail.

Theoretically the promise rejection should be handled in here - https://github.com/dotnet/vscode-csharp/blob/3c12e2709a71838f2e2f0b62b0a08655d514c261/src/razor/src/document/razorDocumentSynchronizer.ts#L64
and in https://github.com/dotnet/vscode-csharp/blob/3c12e2709a71838f2e2f0b62b0a08655d514c261/src/razor/src/document/razorDocumentSynchronizer.ts#L82

and I wouldn't expect it to bubble up.  But perhaps there are promises in that set that are not caught somehow?  Probably needs to be debugged to figure out what rejection is not being caught.  It does seem suspicious that a promise with 1 resolve/reject could have reject called multiple times (for each item in the rejectionsForCancel).

alexgav: Debugging it still, unfortunately the issue is pretty intermittent on my machine (though CI machines have no problem hitting it :) ).

dibarbet: fwiw the code in there seems quite complex and a bit suspect in general.  I wonder if it would be worth rewriting it instead of spending a bunch of time debugging.  I'm no TS expert, but the multiple events rejecting the same promise, possibly entirely outside of the await context seems very prone to random issues.

alexgav: After my last commit hover and rename tests are failing (though not on my box of course). I'll convert this PR to draft and mark with WIP while I figure out what's going on.

Issue #8314, opened 2025-05-21
[BUG]: Adding new file inside directory causes file to be treated as a project and bin + obj folders to be generated causing build to fail


StevanFreeborn:
Type: <b>Bug</b>

## Issue Description ##

If you open a project or solution in Visual Studio Code while using version `2.78.15` of the C# extension and you attempt to create a file within a folder the new file is treated as a project and bin + obj folders are created. These resulting artifacts then prevent further builds of the project or solution.

There seems to be a related issue logged here on the C# Dev Kit repo. => [https://github.com/microsoft/vscode-dotnettools/issues/2024](https://github.com/microsoft/vscode-dotnettools/issues/2024)

## Steps to Reproduce ##

- Ensure you have version `2.78.15` of the C# extension installed in Visual Studio Code.
- Create a new directory and go to the directory

```pwsh
mkdir example
cd example
```

- Create a new console application in the root of the directory

```pwsh
dotnet new console -o .
```

- Open the directory in Visual Studio Code

```pwsh
code .
```

- Create a new directory in the root of your workspace called `new_folder`
- Create a new file in the `new_folder` directory named `Test.cs`

You should find that after creating the file a `bin` and `obj` folder is generated in the `new_folder` directory.
You should find that attempting to build the project generates the following similar errors:

```txt
  Determining projects to restore...
  All projects are up-to-date for restore.
C:\Users\sfree\software_projects\c_sharp_extension_issue\obj\Debug\net9.0\.NETCoreApp,Version=v9.0.AssemblyAttributes.cs(4,12): error CS0579: Duplicate 'global::System.Runtime.Versioning.TargetFrameworkAttribute' attribute [C:\Users\sfree\software_projects\c_sharp_extension_issue\c_sharp_extension_issue.csproj]
C:\Users\sfree\software_projects\c_sharp_extension_issue\obj\Debug\net9.0\c_sharp_extension_issue.AssemblyInfo.cs(13,12): error CS0579: Duplicate 'System.Reflection.AssemblyCompanyAttribute' attribute [C:\Users\sfree\software_projects\c_sharp_extension_issue\c_sharp_extension_issue.csproj]
[...]

PR #8315, opened 2025-05-21
fix: set default value for enableFileBasedPrograms to false

StevanFreeborn:
This pull request includes a single change to the `package.json` file, updating the default value for the `dotnet.projects.enableFileBasedPrograms` configuration setting from `true` to `false`.

This is meant to address issue #8314

### DIFF
--- a/package.json
+++ b/package.json
@@ -1509,7 +1509,7 @@
           },
           "dotnet.projects.enableFileBasedPrograms": {
             "type": "boolean",
-            "default": true,
+            "default": false,
             "description": "%configuration.dotnet.projects.enableFileBasedPrograms%",
             "tags": [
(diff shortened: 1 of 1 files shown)


JoeRobich:
@StevanFreeborn Thanks for opening a PR. We have already taken a fix for this issue in #8298 and are going through validation to release it.

StevanFreeborn:
> @StevanFreeborn Thanks for opening a PR. We have already taken a fix for this issue in #8298 and are going through validation to release it.

No problem at all! Thanks for looking at it. I always feel like if I open an issue, I should at least attempt trying to do something about it.

Love all the work that has been and is being done to make writing C# in VS Code a great experience.

PR #8352, opened 2025-06-06
Remove mention of non-existent "Dotnet > Server: Trace" setting

mriehm:
The "Dotnet > Server: Trace" setting was removed in #7688, so it should no longer be mentioned in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,19 +30,7 @@ For more information you can:
 If you don’t want to take advantage of the great Language Server features, you can revert back to using OmniSharp by going to the Extension settings and setting `dotnet.server.useOmnisharp` to true. Next, uninstall or disable C# Dev Kit. Finally, restart VS Code for this to take effect.
 
 ## Found a Bug?
-To file a new issue:
-
-1. Open the Command Palette with Ctrl+Shift+P (Cmd+Shift+P on macOS), type "Open Settings (UI)", and press Enter.
-2. Search for "trace" in the search bar. Under "Dotnet > Server: Trace" select "Trace" from the drop down. This will provide more output information.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,19 +30,7 @@ For more information you can:
 If you don’t want to take advantage of the great Language Server features, you can revert back to using OmniSharp by going to the Extension settings and setting `dotnet.server.useOmnisharp` to true. Next, uninstall or disable C# Dev Kit. Finally, restart VS Code for this to take effect.
 
 ## Found a Bug?
-To file a new issue:
-
-1. Open the Command Palette with Ctrl+Shift+P (Cmd+Shift+P on macOS), type "Open Settings (UI)", and press Enter.
-2. Search for "trace" in the search bar. Under "Dotnet > Server: Trace" select "Trace" from the drop down. This will provide more output information.
```
jasonmalinowski, line 36: Although I absolutely agree this is out of date, should we instead update this to say how to set the severity now? I also wonder if @dibarbet we should have some sort of "enable logging and restart" command which generally could replace steps 1 through 5.

dibarbet (approved):

README.md
dibarbet: > Although I absolutely agree this is out of date, should we instead update this to say how to set the severity now? I also wonder if @dibarbet we should have some sort of "enable logging and restart" command which generally could replace steps 1 through 5.

Those are detailed in support.md.  You also don't need to always reload the window any more (we live update the level).  I like having the readme point to the support doc so we don't have to keep two places up to date.

dibarbet:
thanks for the contribution @mriehm!

PR #8371, opened 2025-06-18
Add command to easily collect trace of LSP process

dibarbet:
It is often useful to get traces of the language server process.  Rather than have users figure out how to get dotnet-trace, find the LSP server processId, figure out which trace args to use, we can provide a command to take a trace automatically.

dotnet_trace_gif

### DIFF
--- a/l10n/bundle.l10n.json
+++ b/l10n/bundle.l10n.json
@@ -186,6 +186,49 @@
   "Sending request": "Sending request",
   "C# Project Context Status": "C# Project Context Status",
   "Active File Context": "Active File Context",
+  "Initializing dotnet-trace.../dotnet-trace is a command name and should not be localized": {
+    "message": "Initializing dotnet-trace...",
+    "comment": [
+      "dotnet-trace is a command name and should not be localized"
+    ]
+  },
+  "Failed to execute dotnet-trace command: {0}/dotnet-trace is a command name and should not be localized": {
+    "message": "Failed to execute dotnet-trace command: {0}",
+    "comment": [
+      "dotnet-trace is a command name and should not be localized"
+    ]
+  },
+  "Language server process not found, ensure the server is running.": "Language server process not found, ensure the server is running.",
+  "Select Trace Folder": "Select Trace Folder",
+  "Select Folder to Save Trace File": "Select Folder to Save Trace File",
+  "Folder for trace file {0} does not exist": "Folder for trace file {0} does not exist",
+  "Enter dotnet-trace arguments/dotnet-trace is a command name and should not be localized": {
+    "message": "Enter dotnet-trace arguments",
+    "comment": [
+      "dotnet-trace is a command name and should not be localized"
+    ]
+  },
+  "You can modify the default arguments if needed": "You can modify the default arguments if needed",
+  "Recording trace...": "Recording trace...",
+  "Install": "Install",
+  "dotnet-trace not found, run \"{0}\" to install it?/dotnet-trace is a command name and should not be localized": {
+    "message": "dotnet-trace not found, run \"{0}\" to install it?",
+    "comment": [
+      "dotnet-trace is a command name and should not be localized"
+    ]
+  },
+  "Installing dotnet-trace.../dotnet-trace is a command name and should not be localized": {
+    "message": "Installing dotnet-trace...",
+    "comment": [
+      "dotnet-trace is a command name and should not be localized"
+    ]
+  },
+  "Failed to install dotnet-trace, it may need to be manually installed. See C# output for details./dotnet-trace is a command name and should not be localized": {
+    "message": "Failed to install dotnet-trace, it may need to be manually installed. See C# output for details.",
+    "comment": [
+      "dotnet-trace is a command name and should not be localized"
+    ]
+  },
   "C# configuration has changed. Would you like to reload the window to apply your changes?": "C# configuration has changed. Would you like to reload the window to apply your changes?",
   "Generated document not found": "Generated document not found",
   "Nested Code Action": "Nested Code Action",
--- a/package.json
+++ b/package.json
@@ -1899,6 +1899,12 @@
         "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
       {
+        "command": "csharp.recordTrace",
+        "title": "%command.csharp.recordTrace%",
+        "category": "CSharp",
+        "enablement": "dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'RoslynDevKit'"
+      },
+      {
         "command": "extension.showRazorCSharpWindow",
         "title": "%command.extension.showRazorCSharpWindow%",
         "category": "Razor",
--- a/src/lsptoolshost/profiling/profiling.ts
+++ b/src/lsptoolshost/profiling/profiling.ts
@@ -3,10 +3,16 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as fs from 'fs';
+import * as vscode from 'vscode';
 import { EOL } from 'os';
-import { LogOutputChannel } from 'vscode';
+import { RoslynLanguageServer } from '../server/roslynLanguageServer';
+import { showErrorMessageWithOptions } from '../../shared/observers/utils/showMessage';
+import { execChildProcess } from '../../common';
 
-export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.ProcessEnv {
+const TraceTerminalName = 'dotnet-trace';
+
+export function getProfilingEnvVars(outputChannel: vscode.LogOutputChannel): NodeJS.ProcessEnv {
     let profilingEnvVars = {};
     if (process.env.ROSLYN_DOTNET_EventPipeOutputPath) {
         profilingEnvVars = {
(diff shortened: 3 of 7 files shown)


src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,248 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+async function verifyOrAcquireDotnetTrace(
+    folder: string,
+    progress: vscode.Progress<{
+        message?: string;
+        increment?: number;
+    }>,
+    channel: vscode.LogOutputChannel
+): Promise<boolean> {
+    try {
+        await execChildProcess('dotnet-trace --version', folder, process.env);
```
dibarbet, line 155: Not all terminals will have shell execution, and it seems relatively common that we don't have it (which means we can't observe command output).  For example, in a windows sandbox this appears to be the case by default (also true for cmd).

So instead of fiddling with the terminal, for simple commands where we must have the output, we can just run them directly.

l10n/bundle.l10n.json
```diff
@@ -186,6 +186,20 @@
   "Sending request": "Sending request",
   "C# Project Context Status": "C# Project Context Status",
   "Active File Context": "Active File Context",
+  "dotnet-trace": "dotnet-trace",
```
JoeRobich: Should this be in our i10n bundle?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,248 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+    token: vscode.CancellationToken
+): Promise<void> {
+    // Use a terminal to execute the dotnet-trace.  This is much simpler and more reliable than executing dotnet-trace
+    // directly via the child_process module as dotnet-trace relies on shell input in order to stop the trace.
+    // Without using a psuedo-terminal, it is extremely difficult to send the correct signal to stop the trace.
+    //
+    // Luckily, VSCode allows us to use the built in terminal (a psuedo-terminal) to execute commands, which also provides a way to send input to it.
+
+    terminal.show();
+    const command = `dotnet-trace ${args.join(' ')}`;
```
JoeRobich, line 219: Do we need to take care to quote args?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,248 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+
+    const shellIntegration = terminal.shellIntegration;
+    if (shellIntegration) {
+        await new Promise<number>((resolve, _) => {
+            const execution = shellIntegration.executeCommand(command);
+
+            // If the progress is cancelled, we need to send a Ctrl+C to the terminal to stop the command.
+            const cancelDisposable = token.onCancellationRequested(() => {
+                terminal.sendText('^C');
```
JoeRobich, line 228: 😄

src/lsptoolshost/profiling/profiling.ts
dibarbet: Yeah, might be needed for paths with spaces, good point.

dibarbet: Actually I take it back.  The path is not part of the args.  We already quote the args properly when we generate them.  If they change the args, I think it is on them to quote them correctly.

package.json
```diff
@@ -1898,6 +1898,12 @@
         "category": "CSharp",
         "enablement": "dotnet.server.activationContext == 'OmniSharp'"
       },
+      {
+        "command": "csharp.recordTrace",
```
jasonmalinowski, line 1902: Should this be called csharp.recordLanguageServerTrace or something if we ever wanted to use this command name for an operation of recording traces of the user's app, for example?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,271 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+                    } catch (error) {
+                        const errorMessage = error instanceof Error ? error.message : String(error);
+                        showErrorMessageWithOptions(
+                            vscode,
+                            vscode.l10n.t({
+                                message: 'Failed to execute dotnet-trace command: {0}',
+                                args: [errorMessage],
+                                comment: 'dotnet-trace is a command name and should not be localized',
+                            }),
+                            { modal: true }
```
jasonmalinowski, line 65: Why modal?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,271 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+    progress: vscode.Progress<{
+        message?: string;
+        increment?: number;
+    }>,
+    outputChannel: vscode.LogOutputChannel,
+    cancellationToken: vscode.CancellationToken
+): Promise<void> {
+    const processId = languageServer.processId;
+
+    if (!processId) {
```
jasonmalinowski, line 85: Should this launch the process in this case? That way there's a good chance the trace might start earlier if the user needs to trace something right away?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,271 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+        throw new Error(vscode.l10n.t('Language server process not found, ensure the server is running.'));
+    }
+
+    let traceFolder: string | undefined = '';
+    if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders?.length >= 1) {
+        traceFolder = vscode.workspace.workspaceFolders[0].uri.fsPath;
```
jasonmalinowski, line 91: Is there a case where the workspaceFolder might not have a fsPath?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,271 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+    }
+
+    // Prompt the user for the folder to save the trace file
+    // By default, choose the first workspace folder if available.
+    const uris = await vscode.window.showOpenDialog({
+        canSelectFiles: false,
+        canSelectFolders: true,
+        canSelectMany: false,
+        defaultUri: traceFolder ? vscode.Uri.file(traceFolder) : undefined,
```
jasonmalinowski, line 100: Why reconstruct a file URI here if we already had a URI object we pulled the path out of?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -24,3 +30,271 @@ export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.Pro
...
+            comment: 'dotnet-trace is a command name and should not be localized',
+        }),
+        prompt: vscode.l10n.t('You can modify the default arguments if needed'),
+    });
+    if (userArgs === undefined) {
+        // User cancelled the input box
+        return;
+    }
+
+    const terminal = await getOrCreateTerminal(traceFolder, outputChannel);
```
jasonmalinowski, line 132: Should we do this after we've verified we have dotnet trace installed?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -3,10 +3,16 @@
...
+import * as fs from 'fs';
+import * as vscode from 'vscode';
 import { EOL } from 'os';
-import { LogOutputChannel } from 'vscode';
+import { RoslynLanguageServer } from '../server/roslynLanguageServer';
+import { showErrorMessageWithOptions } from '../../shared/observers/utils/showMessage';
+import { execChildProcess } from '../../common';
 
-export function getProfilingEnvVars(outputChannel: LogOutputChannel): NodeJS.ProcessEnv {
+const TraceTerminalName = 'dotnet-trace';
```
jasonmalinowski, line 13: Any chance this name could collide with something the user already has?

jasonmalinowski (approved):

src/lsptoolshost/profiling/profiling.ts
dibarbet: 1.  The command is only registered after the server starts (to get this processId).  So its not expected to be missing
2.  The most correct way to do a startup trace is to invoke the language server process via dotnet-trace.  But that requires a bit more refactoring that I didn't have time for my timebox.

src/lsptoolshost/profiling/profiling.ts
dibarbet: errors aren't expected - its only if running something failed, so we show a model dialog indicating the command failed.

This is an explicit user requested action, so a much more noticeable modal dialog seems better than an easily ignorable error toast.

package.json
dibarbet: Probably - will do in a followup

src/lsptoolshost/profiling/profiling.ts
dibarbet: Hmm... I don't know of one.  I don't think VSCode has virtual workspaces.

src/lsptoolshost/profiling/profiling.ts
dibarbet: Unlikely, typically you'll have terminals named like 'powershell', or 'git bash' for specific kinds of terminals.  We also already verify the terminal is in the correct working directory before we use it, so even if it came from somewhere else we could still likely use it.

PR #8401, opened 2025-07-07
Adding a build script

AbhitejJohn:


CONTRIBUTING.md
```diff
@@ -36,12 +36,15 @@ Follow these steps to build, run, and test the repository:
 
 #### Building
 
+If you have the ability to run powershell, you can invoke "build.ps1" from the root of the repo. If not, the following steps will get build going for you as well:
```
JoeRobich: I would prefer it named `init.ps1` as the script doesn't actually invoke a build. It would need to additionally run `npm run compileDev` or similar to build the extension.

AbhitejJohn: Done. 

As an aside, would it help having a build script?

JoeRobich: I don't think so. In web projects, `npm run` is a fairly universal way of running these build tasks and is only a single command. However, I am open to reviewing a PR if we thought it would improve onboarding.

JoeRobich (approved):

init.ps1
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env pwsh
+
+Write-Host "`nStarting project setup..." -ForegroundColor Cyan
+
+function Run-Command($cmd, $errorMsg) {
+    try {
+        Write-Host "Running: $cmd" -ForegroundColor Yellow
+        Invoke-Expression $cmd
```
jaredpar: `Invoke-Expression` will only throw an exception if the expression throws an exception. If the command fails this will just return a value that is ignored. If you want helpers that throw an exception when a command fails suggest looking at this

https://github.com/dotnet/roslyn/blob/main/eng/build-utils.ps1#L70-L153

init.ps1
```diff
@@ -0,0 +1,40 @@
...
+        Write-Host "$errorMsg" -ForegroundColor Red
+        exit 1
+    }
+}
+
+# 1. Install vsts-npm-auth globally
+Run-Command "npm install -g vsts-npm-auth" "Failed to install vsts-npm-auth."
+
+# 2. Authenticate with vsts-npm-auth
+if (Test-Path ".npmrc") {
```
jaredpar: Consider wrapping the body the following: 

```pwsh
Push-Location $PSScriptRoot
try {

}
finally {
  Pop-Location
}
```

Lacking that if `init.ps1` is run from a diff directory calls like `Test-Path` here won't do what you're expecting.

AbhitejJohn: Good catch, should have checked the copilot generated script a little better :)

AbhitejJohn:
@JoeRobich : Do you know if there's something up with the tests in CI ? I dont expect these changes to cause test failures as such.

dibarbet:
these are currently flaky.  its on my list to investigate

AbhitejJohn:
Looks like the auto-complete got this in based on previous approvals. I'm happy to change any of this up in a follow-up if anyone else has thoughts/feedback on the latest iteration.

PR #8408, opened 2025-07-09
Storage account for Copilot language server

AbhitejJohn:
This is primarily an implementation of David Barbet's suggestion that helps with our dev innerloop reducing the amount of dependencies across repos. With the roslyn copilot language server coming in from the storage account, changes are no longer required in C# Dev Kit to point to the right language server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -423,6 +423,24 @@
       "integrity": "9944EBD6EE06BD595BCADD3057CD9BEF4105C3A3952DAE03E54F3114E2E6661F"
     },
     {
+      "id": "RoslynCopilotLanguageServer",
+      "description": "Language server for Roslyn Copilot integration (Windows)",
+      "url": "https://roslynomnisharp.blob.core.windows.net/releases/1.39.12/omnisharp-linux-musl-arm64-net6.0-1.39.12.zip",
+      "installPath": ".roslyncopilotlanguageserver",
+      "platforms": [
+        "win32",
+        "linux",
+        "linux-musl",
+        "darwin"
+      ],
+      "architectures": [
+        "x86_64",
+        "arm64"
+      ],
+      "installTestPath": "./.roslyncopilotlanguageserver/Microsoft.VisualStudio.Copilot.Roslyn.LanguageServer.dll",
+      "integrity": "9944EBD6EE06BD595BCADD3057CD9BEF4105C3A3952DAE03E54F3114E2E6661F"
+    },
+    {
       "id": "Debugger",
       "description": ".NET Core Debugger (Windows / x64)",
       "url": "https://vsdebugger-cyg0dxb6czfafzaz.b01.azurefd.net/coreclr-debug-2-81-0/coreclr-debug-win7-x64.zip",
--- a/src/lsptoolshost/server/roslynLanguageServer.ts
+++ b/src/lsptoolshost/server/roslynLanguageServer.ts
@@ -8,6 +8,7 @@ import * as path from 'path';
 import * as cp from 'child_process';
 import * as uuid from 'uuid';
 import * as net from 'net';
+import * as util from '../../common';
 import {
     LanguageClientOptions,
     MessageTransports,
@@ -74,10 +75,6 @@ import { getProfilingEnvVars } from '../profiling/profiling';
 import { isString } from '../utils/isString';
 import { getServerPath } from '../activate';
 import { UriConverter } from '../utils/uriConverter';
-import {
-    copilotLanguageServerExtensionAssemblyName,
-    copilotLanguageServerExtensionComponentName,
-} from '../copilot/contextProviders';
 
 // Flag indicating if C# Devkit was installed the last time we activated.
 // Used to determine if we need to restart the server on extension changes.
@@ -107,6 +104,13 @@ export class RoslynLanguageServer {
      * The process Id of the currently running language server process.
      */
     private static _processId: number | undefined;
+
+    /**
+     * The folder name for the Roslyn Copilot language server extension.
+     */
+    private static readonly _copilotLanguageServerExtensionDirName = '.roslyncopilotlanguageserver';
+    private static readonly _copilotLanguageServerExtensionAssemblyName =
+        'Microsoft.VisualStudio.Copilot.Roslyn.LanguageServer.dll';
 
     /**
      * The solution file previously opened; we hold onto this so we can send this back over if the server were to be relaunched for any reason, like some other configuration
@@ -1079,11 +1083,15 @@ export class RoslynLanguageServer {
             await exports.setupTelemetryEnvironmentAsync(env);
         }
 
-        const copilotServerExtensionfolder = exports.components[copilotLanguageServerExtensionComponentName];
+        const copilotServerExtensionfolder = path.join(
+            util.getExtensionPath(),
+            RoslynLanguageServer._copilotLanguageServerExtensionDirName
+        );
+
         if (copilotServerExtensionfolder) {
             const copilotServerExtensionFullPath = path.join(
                 copilotServerExtensionfolder,
-                copilotLanguageServerExtensionAssemblyName
+                RoslynLanguageServer._copilotLanguageServerExtensionAssemblyName
             );
             additionalExtensionPaths.push(copilotServerExtensionFullPath);
             channel.trace(
--- a/src/lsptoolshost/copilot/contextProviders.ts
+++ b/src/lsptoolshost/copilot/contextProviders.ts
@@ -7,13 +7,7 @@ import * as vscode from 'vscode';
 import * as lsp from 'vscode-languageserver-protocol';
 import { RoslynLanguageServer } from '../server/roslynLanguageServer';
 import { CSharpExtensionId } from '../../constants/csharpExtensionId';
-import { csharpDevkitExtensionId, getCSharpDevKit } from '../../utils/getCSharpDevKit';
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -422,6 +422,24 @@
       "isFramework": false,
       "integrity": "9944EBD6EE06BD595BCADD3057CD9BEF4105C3A3952DAE03E54F3114E2E6661F"
     },
+    {
+      "id": "RoslynCopilotLanguageServer",
+      "description": "Language server for Roslyn Copilot integration (Windows)",
+      "url": "https://roslynomnisharp.blob.core.windows.net/releases/1.39.12/omnisharp-linux-musl-arm64-net6.0-1.39.12.zip",
```
AbhitejJohn: This url will change

package.json
```diff
@@ -422,6 +422,24 @@
...
+        "linux",
+        "linux-musl",
+        "darwin"
+      ],
+      "architectures": [
+        "x86_64",
+        "arm64"
+      ],
+      "installTestPath": "./.roslyncopilotlanguageserver/Microsoft.VisualStudio.Copilot.Roslyn.LanguageServer.dll",
+      "integrity": "9944EBD6EE06BD595BCADD3057CD9BEF4105C3A3952DAE03E54F3114E2E6661F"
```
AbhitejJohn: so will this.

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -1079,11 +1083,15 @@ export class RoslynLanguageServer {
             await exports.setupTelemetryEnvironmentAsync(env);
         }
 
-        const copilotServerExtensionfolder = exports.components[copilotLanguageServerExtensionComponentName];
+        const copilotServerExtensionfolder = path.join(
```
dibarbet: I do not remember if the download from the storage account promise is awaited before something calls into this - definitely verify that

package.json
```diff
@@ -422,6 +422,24 @@
       "isFramework": false,
       "integrity": "9944EBD6EE06BD595BCADD3057CD9BEF4105C3A3952DAE03E54F3114E2E6661F"
     },
+    {
+      "id": "RoslynCopilotLanguageServer",
```
dibarbet: nit - a better name is probably 'RoslynCopilot' - including language server makes me think its a separate language server , when really its just an extra dll loaded in the existing language server.

dibarbet (commented):
know this is a draft, just wanted to make a couple small notes

src/lsptoolshost/server/roslynLanguageServer.ts
JoeRobich: I can confirm that `installRuntimeDependencies` is awaited in main.ts before we start the language server.

AbhitejJohn: Yup, I think the debugger install logic is a little different in that a user could trigger an F5 and that should wait on install to happen. With the context provider logic though, we only register our handler with Copilot after the install is complete and there isn't any UI that invokes this logic outside of that.

AbhitejJohn:
> know this is a draft, just wanted to make a couple small notes

Thank you, feedback is welcome. This is change is mostly ready. I'm just trying to get the bits deployed, so I can update the url and the sha before its good to merge.

package.json
```diff
@@ -422,6 +422,24 @@
       "isFramework": false,
       "integrity": "9944EBD6EE06BD595BCADD3057CD9BEF4105C3A3952DAE03E54F3114E2E6661F"
     },
+    {
+      "id": "RoslynCopilot",
+      "description": "Language server for Roslyn Copilot integration (Windows)",
+      "url": "https://roslynomnisharp.blob.core.windows.net/releases/1.39.12/omnisharp-linux-musl-arm64-net6.0-1.39.12.zip",
+      "installPath": ".roslyncopilot",
+      "platforms": [
```
AbhitejJohn, line 430: if there is a better way of specifying support for all platforms and architectures the extension supports, I'd love to put that in so this list is a little more maintainable.

JoeRobich: We could introduce the concept of "neutral" which would be a catch-all. Are there no plans to generate R2R builds of these libraries?

AbhitejJohn: Perhaps, in the future? 
On the "neutral" piece, is that something we have the liberty to configure? I thought this was a fixed schema of some sort.

JoeRobich: It is used by the [PackageManager](https://github.com/dotnet/vscode-csharp/tree/main/src/packageManager) functionality which builds out of our repo. In particular we would update [our package filter](https://github.com/dotnet/vscode-csharp/blob/main/src/packageManager/packageFilterer.ts#L18) to also include these "neutral" packages

AbhitejJohn: Cool, gotcha. As long as that's the only place that needs a change, happy to make that update. Thought this was consumed by other tooling. Thanks for the confirmation 👍

azure-pipelines/PublishRoslynCopilot.yml
```diff
@@ -0,0 +1,54 @@
...
+  - pipeline: CI
+    project: DevDiv
+    source: VisualStudio.Conversations
+    branch: main
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    pool:
+      name: AzurePipelines-EO
+      image: AzurePipelinesWindows2022compliantGPT
```
dibarbet: What is this image?  Do we need it over the standard 1es images (IIRC like `1ESPT-Windows2022`)

azure-pipelines/PublishRoslynCopilot.yml
```diff
@@ -0,0 +1,54 @@
...
+      os: windows
+    stages:
+    - stage: Upload
+      jobs:
+      - job: UploadToServer
+        displayName: Upload to server
+        pool:
+          name: AzurePipelines-EO
+          image: AzurePipelinesWindows2022compliantGPT
+          os: windows
```
dibarbet: Not blocking, feel free to ignore if it causes too much churn: but consider using linux.  I've noticed that the windows pools often have more delay in acquiring machines

src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -1079,16 +1083,19 @@ export class RoslynLanguageServer {
             await exports.setupTelemetryEnvironmentAsync(env);
         }
 
-        const copilotServerExtensionfolder = exports.components[copilotLanguageServerExtensionComponentName];
-        if (copilotServerExtensionfolder) {
-            const copilotServerExtensionFullPath = path.join(
-                copilotServerExtensionfolder,
-                copilotLanguageServerExtensionAssemblyName
-            );
+        const copilotServerExtensionFullPath = path.join(
```
dibarbet: Instead of this and the hardcoded paths up top, consider adding this to the builtInComponents.  It will automatically handle having an option override for the path as well. https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/extensions/builtInComponents.ts#L16

you would then call it here like this https://github.com/dotnet/vscode-csharp/blob/6e2f0741a208a5684181d9636815647d1f162f91/src/lsptoolshost/server/roslynLanguageServer.ts#L1011

azure-pipelines/PublishRoslynCopilot.yml
AbhitejJohn: Pulled that from another pipeline I had. Could totally move to the standard we use.

azure-pipelines/PublishRoslynCopilot.yml
AbhitejJohn: gotcha, that's good to know. This one wouldn't be triggered as often, so I can flip to linux when that starts becoming a problem.

src/lsptoolshost/server/roslynLanguageServer.ts
dibarbet: For completion to work in the settings.json, also add the option name here - https://github.com/dotnet/vscode-csharp/blob/6e2f0741a208a5684181d9636815647d1f162f91/package.json#L1436 (we appear to be missing some already, missing ones just have missing completion, they still work if set in settings.json).

AbhitejJohn: Thanks for pointing that out, didn't know about the settings.json update. Fixed.

package.json
```diff
@@ -1459,6 +1459,10 @@
               "xamlTools": {
                 "description": "%configuration.dotnet.server.componentPaths.xamlTools%",
                 "type": "string"
+              },
+              "roslynCopilot": {
+                "description": "%configuration.dotnet.server.componentPaths.roslynCopilot%",
```
dibarbet, line 1460: this needs a corresponding string in package.nls.json
https://github.com/dotnet/vscode-csharp/blob/main/package.nls.json#L32

dibarbet (approved):
lgtm, pending the package.nls.json fix

package.json
AbhitejJohn: Added the concept of a neutral package @JoeRobich

dibarbet (approved):

package.json
AbhitejJohn: This has been updated.

package.json
AbhitejJohn: This has been updated.

JoeRobich (approved):

PR #8428, opened 2025-07-17
Register restore commands even if DevKit is being used

RikkiGibson:
The bug we observed with restore is: we have a new, specific restore behavior for file-based apps, which is implemented only in the base C# extension. When CDK is also loaded, the client simply never responds to `projectNeedsRestore` messages and similar from the server.

But CDK doesn't handle restore of file-based apps. So the effect we see is: when both C#+CDK are loaded together in VS Code (from a fresh "reload window" or fresh startup), restore will just not work. The client will silently not initiate the restore process.

I doubt that just always attaching the restore handlers anyways is the right solution. But, as a short term solution, we could simply attach the handlers and say: **if CDK is loaded, then only do the restore for file-based apps, while assuming CDK continues to handle projects and solutions.**

Wouldn't it be most reasonable to "forward" the message to CDK, and have some way of deciding if CDK handled it, and if they did not, then we handle it? Sorta like how UI event handlers bubble events up?

@jasonmalinowski @dibarbet

### DIFF
--- a/src/lsptoolshost/projectRestore/restore.ts
+++ b/src/lsptoolshost/projectRestore/restore.ts
@@ -21,7 +21,8 @@ let _restoreInProgress = false;
 export function registerRestoreCommands(context: vscode.ExtensionContext, languageServer: RoslynLanguageServer) {
     if (getCSharpDevKit()) {
         // We do not need to register restore commands if using C# devkit.
-        return;
+        // TODO: restore commands for FBPs still need to run, even if devkit is being used.
+        // return;
     }
     const restoreChannel = vscode.window.createOutputChannel(vscode.l10n.t('.NET NuGet Restore'));
     context.subscriptions.push(


src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -21,7 +21,8 @@ let _restoreInProgress = false;
 export function registerRestoreCommands(context: vscode.ExtensionContext, languageServer: RoslynLanguageServer) {
     if (getCSharpDevKit()) {
         // We do not need to register restore commands if using C# devkit.
-        return;
+        // TODO: restore commands for FBPs still need to run, even if devkit is being used.
+        // return;
     }
     const restoreChannel = vscode.window.createOutputChannel(vscode.l10n.t('.NET NuGet Restore'));
```
dibarbet: this is going to very unfortunately create a second nuget restore channel.  Other than having devkit do the restore though I don't have a great alternative.  But we should get an issue filed on this at the very least.

JoeRobich: Could we use the C# log for the time being? Had the O# extension not separated them, I am not sure I would have the distinction. OR, possibly I would have put all project system logging into the restore one.

RikkiGibson: My preference would actually be to put these in the C# log. When we split across output windows it just becomes harder to tell the order that things occur in. Even when there are timestamps (which I am adding in this PR.)

However to reduce scope I am going to refrain from renaming or moving which output channel we use here.

Also, what is `l10n`? Is this just a super weird abbreviation? An 'l' then 10 letters then an 'n' to make 'localization'?

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -19,27 +19,56 @@ import { getCSharpDevKit } from '../../utils/getCSharpDevKit';
...
 export function registerRestoreCommands(context: vscode.ExtensionContext, languageServer: RoslynLanguageServer) {
-    if (getCSharpDevKit()) {
-        // We do not need to register restore commands if using C# devkit.
-        return;
-    }
     const restoreChannel = vscode.window.createOutputChannel(vscode.l10n.t('.NET NuGet Restore'));
     context.subscriptions.push(
         vscode.commands.registerCommand('dotnet.restore.project', async (_request): Promise<void> => {
+            if (getCSharpDevKit()) {
+                appendLineWithTimestamp(restoreChannel, "Not handling command 'dotnet.restore.project' from C# extension, because C# Dev Kit is expected to handle it.");
```
RikkiGibson: Interestingly, when C#+CDK are loaded together to begin with, I cannot get these lines to run, but then if I unload+reload CDK, restart language server, (but do not reload window), I can get them to run.

I think this is related to some other mechanism unregistering the "dotnet restore" commands when CDK is loaded. The ".NET: Restore all projects" and ".NET: Restore projects" commands are just not there when CDK is loaded.

Thankfully since the last handler is for a language server message, not for a VS Code command, it will get hit regardless of whether CDK is loaded, when the project system says there are unresolved assets.

RikkiGibson: The general goal and philosophy of the change is: deviate behavior as late as possible, and as little as possible, and make it clear that the deviation is occurring, when the thing we are supposed to do, depends on Dev Kit being loaded or not. I think this is something we should look at doing with other calls to `getCSharpDevKit()` also.

RikkiGibson:
How do I get these [prettier errors](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1099449&view=logs&jobId=0380e083-8153-51ec-29f0-d7bc571e1653&j=0380e083-8153-51ec-29f0-d7bc571e1653&t=cd83eacf-8d3a-5145-dc28-c9b04dec57a2) to appear locally. I tried installing the prettier extension but they're not showing up.

```log
/Users/runner/work/1/s/src/lsptoolshost/projectRestore/restore.ts
  26:41  error  Replace `restoreChannel,·"Not·handling·command·'dotnet.restore.project'·from·C#·extension,·because·C#·Dev·Kit·is·expected·to·handle·it."` with `⏎····················restoreChannel,⏎····················"Not·handling·command·'dotnet.restore.project'·from·C#·extension,·because·C#·Dev·Kit·is·expected·to·handle·it."⏎················`  prettier/prettier
  36:41  error  Replace `restoreChannel,·"Not·handling·command·'dotnet.restore.all'·from·C#·extension,·because·C#·Dev·Kit·is·expected·to·handle·it."` with `⏎····················restoreChannel,⏎····················"Not·handling·command·'dotnet.restore.all'·from·C#·extension,·because·C#·Dev·Kit·is·expected·to·handle·it."⏎················`          prettier/prettier
  50:35  error  Replace `".cs"` with `'.cs'`                                                                                                                                                                                                                                                                                                                prettier/prettier
  53:45  error  Replace `restoreChannel,·`Not·restoring·'${path}'·from·C#·extension,·because·C#·Dev·Kit·is·expected·to·handle·restore·for·it.`` with `⏎························restoreChannel,⏎························`Not·restoring·'${path}'·from·C#·extension,·because·C#·Dev·Kit·is·expected·to·handle·restore·for·it.`⏎····················`          prettier/prettier
```

edit: looks like `npx prettier src/lsptoolshost/projectRestore/restore.ts --write` fixed things.

RikkiGibson:
Unsure what to do with the following failure: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1099482&view=ms.vss-test-web.build-test-results-tab&runId=30068902&resultId=100010&paneView=debug

> Expected value: not "dotnet.restore.project"
> Received array:     [...]

Basically what I am observing in the running product is that these commands aren't available to run when CDK is loaded. If you unload CDK they appear, reload and they disappear. So I am wondering if something about how we are detecting the commands in this context, is out of sync with the VS Code editor.

src/lsptoolshost/projectRestore/restore.ts
JoeRobich: > n 'l' then 10 letters then an 'n' to make 'localization'?

Yup!

dibarbet:
@RikkiGibson you need to remove the enablement in the package.json to ensure it is enabled in all contexts - https://github.com/dotnet/vscode-csharp/blob/341a80aa290112b5435f459ad4c05771e257eab3/package.json#L1852

src/lsptoolshost/projectRestore/restore.ts
dibarbet: Agreed, lets do the C# window then.  Better than a duplicate output window.

RikkiGibson:
@dibarbet @JoeRobich for review.

I'm wondering if it is simple to just use the C# channel instead for restore output. Would it work to just replace the line which is creating the NuGet channel with some other call that obtains a handle for an existing channel? (I'm sure I can look but those who know the APIs/factoring may know the best answer off the cuff.) Just want to avoid any significant amount of test churn etc.

dibarbet:
> Would it work to just replace the line which is creating the NuGet channel with some other call that obtains a handle for an existing channel?

Don't believe you can get an existing channel, but instead of creating one, when the command is registered you pass in the C# channel (_channel) here https://github.com/dotnet/vscode-csharp/blob/dc8c85b01650279390376b0b9934e9224d6ac495/src/lsptoolshost/activate.ts#L99

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -18,31 +18,72 @@ import { getCSharpDevKit } from '../../utils/getCSharpDevKit';
...
+            projectFilePaths = csharpFiles;
+        }
+
+        if (projectFilePaths.length > 0) {
+            await restore(languageServer, csharpOutputChannel, params.projectFilePaths, false);
+        }
     });
 }
 
+function appendLineWithTimestamp(outputChannel: vscode.OutputChannel, line: string) {
```
dibarbet: This shouldn't be necessary - the csharp channel is a log output channel, which should (IIRC) append the timestamp.  You can also set the severity (channel.info / channel.warn / etc)

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -18,32 +18,35 @@ import { getCSharpDevKit } from '../../utils/getCSharpDevKit';
...
+    languageServer: RoslynLanguageServer,
+    csharpOutputChannel: vscode.LogOutputChannel
+) {
     context.subscriptions.push(
         vscode.commands.registerCommand('dotnet.restore.project', async (_request): Promise<void> => {
             if (getCSharpDevKit()) {
                 appendLineWithTimestamp(
-                    restoreChannel,
+                    csharpOutputChannel,
                     "Not handling command 'dotnet.restore.project' from C# extension, because C# Dev Kit is expected to handle it."
```
JoeRobich: Should we preface these strings, `[.NET Restore]`, now that they will be mixed into the C# log?

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -18,31 +18,72 @@ import { getCSharpDevKit } from '../../utils/getCSharpDevKit';
...
+        if (getCSharpDevKit()) {
+            // Only restore '.cs' files (file-based apps) if CDK is loaded.
+            const csharpFiles = [];
+            for (const path of projectFilePaths) {
+                if (path.endsWith('.cs')) {
+                    csharpFiles.push(path);
+                } else {
+                    appendLineWithTimestamp(
+                        csharpOutputChannel,
+                        `Not restoring '${path}' from C# extension, because C# Dev Kit is expected to handle restore for it.`
```
AbhitejJohn: Would this show up under diagnostic logging? If I was just using C# ext intentionally, these messages wouldn't really contribute actionable data for me.

src/lsptoolshost/projectRestore/restore.ts
RikkiGibson: Oh, does that mean the "C#" channel is of a kind that already puts timestamps and such in the messages, while the ".NET Restore" channel was of some other kind that doesn't do that? I can rip this part out and change the existing usages to debug severity or something.

src/lsptoolshost/projectRestore/restore.ts
RikkiGibson: This message will only show when C#+Dev Kit is being used, when project system tells the client that restore is needed for an ordinary project. In my opinion, it is a useful message, because it would have made it much more clear to me that the component which receives the `_projectNeedsRestore` message, can actually choose to drop the message and not do a restore.

src/lsptoolshost/projectRestore/restore.ts
RikkiGibson: I verified this is the case, there are double timestamps for the restore messages with this function. Fixing.

src/lsptoolshost/projectRestore/restore.ts
RikkiGibson: That seems fine, though, I am wondering if there is a way to make a derived channel that bakes that in just like the timestamps seem to be getting baked in.

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -18,31 +18,63 @@ import { getCSharpDevKit } from '../../utils/getCSharpDevKit';
...
+    context: vscode.ExtensionContext,
+    languageServer: RoslynLanguageServer,
+    csharpOutputChannel: vscode.LogOutputChannel
+) {
     context.subscriptions.push(
         vscode.commands.registerCommand('dotnet.restore.project', async (_request): Promise<void> => {
-            return chooseProjectAndRestore(languageServer, restoreChannel);
+            if (getCSharpDevKit()) {
+                csharpOutputChannel.debug(
+                    "[.NET Restore] Not handling command 'dotnet.restore.project' from C# extension, because C# Dev Kit is expected to handle it."
```
JoeRobich: What if we just didn't register these commands if using CDK (also remove the enablement from the package.json) ? Seems likely to lead to confusion.

Could still hook the restore request.

RikkiGibson: I think I did this in a previous revision, but, an integration test failed IIRC due to registering the command, when the enablement expression in package.json was false. So it's necessary to have the enablement as well as the `registerCommand` in sync, I guess.

What I want is just to be certain that none of these commands will ever silently do nothing.

One thing which is confusing here is; does this function re-run when CDK is loaded/unloaded, then "Reload Extensions" is run? That does not seem to be the case, and neither does the "Restart Language Server" command I was prompted to run when loading/unloading CDK.

I'll have to jump in the debugger and check.

RikkiGibson: OK. `registerRestoreCommands` is called again when CDK changes from loaded->unloaded, but not when it changes from unloaded->loaded. So maybe enablements settings cause the commands to get filtered out for the unloaded->loaded case?

It seems like it is fine to adjust things to condition the `registerCommand` calls based on whether CDK is loaded. Just the `languageServer.registerOnRequest` call needs to happen regardless of CDK state. And in that case it seems good to delay the CDK check as the PR is currently doing.

RikkiGibson: > OK. `registerRestoreCommands` is called again when CDK changes from loaded->unloaded

The fact that I had to reload the window to repro the restore bug, after unloading CDK, makes it seem like the `languageServer.registerOnRequest` callback is not unregistered when CDK is unloaded. Seems funky. Does this mean that if we kept reloading->unloading CDK, we would register more and more callbacks?

RikkiGibson: Repeatedly calling `registerOnRequest` doesn't seem to cause problems. For example, if we send the message the handler is registered to handle, we don't get duplicate log messages indicating duplicate runs of the handler. So I don't think there is anything further that needs to be fixed.

test/lsptoolshost/integrationTests/expectedCommands.ts
```diff
@@ -11,16 +11,13 @@ const CommonCommands = [
     'csharp.listRemoteDockerProcess',
     'csharp.attachToProcess',
     'csharp.reportIssue',
-];
-
-// Commands used by both O# and Roslyn standalone activation contexts.
-const CommonStandaloneCommands = [
     'dotnet.restore.project',
     'dotnet.restore.all',
```
JoeRobich: I think this change can be reverted too

JoeRobich (approved):

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -18,31 +18,52 @@ import { getCSharpDevKit } from '../../utils/getCSharpDevKit';
...
-        vscode.commands.registerCommand('dotnet.restore.all', async (): Promise<void> => {
-            return restore(languageServer, restoreChannel, [], true);
-        })
-    );
 
     languageServer.registerOnRequest(ProjectNeedsRestoreRequest.type, async (params) => {
-        await restore(languageServer, restoreChannel, params.projectFilePaths, false);
+        let projectFilePaths = params.projectFilePaths;
+        if (getCSharpDevKit()) {
+            // Only restore '.cs' files (file-based apps) if CDK is loaded.
```
jasonmalinowski, line 43: Why do we need to have this check at all? Is there a case we'd be sending over a request for a restore if C# DevKit was the one loading projects?

RikkiGibson: If you load C#, then afterwards load CDK, without reloading the window, then this handler can receive requests for ordinary projects still. This may be a bug.

My goal here is to identify if we are receiving a request which doesn't seem to make sense under the current conditions, and, if so, log what is happening, rather than for example doing a redundant restore, or dropping the message silently. This involves making an assumption of what CDK "should be responsible for instead", when we receive this message.

It's possible we will want to make a change here in the future, either when we fix an associated bug or for some other reason. However, I feel current behavior is the right thing to do for now.

PR #8435, opened 2025-07-22
Support activating the extension with limited support in untrusted workspaces

JoeRobich:
Allows loading the C# extension in untrusted workspaces. Does not start any language server or provide any services. Hooks the trust granted event in order to prompt the user to reload the window and thus the C# extension.


jasonmalinowski:
@JoeRobich Is this fixing an issue where changes to trust normally wouldn't cause us to load, or is the goal having us in a partially loaded state will help some other scenario?

JoeRobich:
> @JoeRobich Is this fixing an issue where changes to trust normally wouldn't cause us to load, or is the goal having us in a partially loaded state will help some other scenario?

It allows the CDK to load in untrusted workspaces and provide limited functionality (e.g new project templating). Since they have a dependency on the C# extension, they cannot load if we do not load.

src/csharpExtensionExports.ts
```diff
@@ -11,7 +11,12 @@ import { GlobalBrokeredServiceContainer } from '@microsoft/servicehub-framework'
 import { PartialResultParams, ProtocolRequestType, RequestType } from 'vscode-languageclient/node';
 import { LanguageServerEvents } from './lsptoolshost/server/languageServerEvents';
 
+export interface LimitedExtensionExports {
+    isLimitedActivation: true;
+}
```
jasonmalinowski, line 16: Should there be a base type of this somehow? Otherwise I'm not sure how consumption of this might feel when you don't know what type you're getting back.

src/main.ts
```diff
@@ -110,8 +111,25 @@ export async function activate(
...
+        exports = { isLimitedActivation: true };
+        csharpChannel.trace('C# Extension activated in limited mode due to workspace trust not being granted.');
+        context.subscriptions.push(
+            vscode.workspace.onDidGrantWorkspaceTrust(() => {
+                const reloadTitle: CommandOption = {
+                    title: vscode.l10n.t('Reload Window'),
+                    command: 'workbench.action.reloadWindow',
+                };
+                const message = vscode.l10n.t(
+                    'Workspace trust has changed. Would you like to reload the window to activate the C# extension?'
```
jasonmalinowski: Would we be showing this even if DevKit is installed too?

src/csharpExtensionExports.ts
JoeRobich: Not necessarily. In TypeScript the value would be any of the three types before being narrowed. Assuming anyone who is using this is already using an in check or type guard function to distinguish between the CSharp and OmniSharp types. Those should still work to perform the narrowing.

package.json
```diff
@@ -33,7 +33,7 @@
   "capabilities": {
     "virtualWorkspaces": false,
     "untrustedWorkspaces": {
-      "supported": false
+      "supported": "limited"
```
jasonmalinowski, line 36: I'm looking at https://code.visualstudio.com/api/extension-guides/workspace-trust#contribution-points and have we gone through the various things listed to make sure we've got mitigation here?

jasonmalinowski:
> It allows the CDK to load in untrusted workspaces and provide limited functionality (e.g new project templating). Since they have a dependency on the C# extension, they cannot load if we do not load.

Makes sense. I admit it's tempting to also use this as an opportunity to expand our extension a bit to support limited mode in other ways. i.e. we can't load projects and can't do design time builds, but the base miscellaneous files experience could at least light up...

src/main.ts
```diff
@@ -110,8 +111,25 @@ export async function activate(
...
+    let exports: CSharpExtensionExports | OmnisharpExtensionExports | LimitedExtensionExports;
+    if (vscode.workspace.isTrusted !== true) {
+        activationEvent = TelemetryEventNames.CSharpLimitedActivation;
+        exports = { isLimitedActivation: true };
+        csharpChannel.trace('C# Extension activated in limited mode due to workspace trust not being granted.');
+        context.subscriptions.push(
+            vscode.workspace.onDidGrantWorkspaceTrust(() => {
+                const reloadTitle: CommandOption = {
+                    title: vscode.l10n.t('Reload Window'),
+                    command: 'workbench.action.reloadWindow',
```
dibarbet: `workbench.action.restartExtensionHost` might be the more modern way to do this, without reloading the entire window.

src/main.ts
```diff
@@ -110,8 +111,25 @@ export async function activate(
...
     };
 
-    let exports: CSharpExtensionExports | OmnisharpExtensionExports;
-    if (!useOmnisharpServer) {
+    let activationEvent = TelemetryEventNames.CSharpActivated;
+    let exports: CSharpExtensionExports | OmnisharpExtensionExports | LimitedExtensionExports;
+    if (vscode.workspace.isTrusted !== true) {
+        activationEvent = TelemetryEventNames.CSharpLimitedActivation;
+        exports = { isLimitedActivation: true };
+        csharpChannel.trace('C# Extension activated in limited mode due to workspace trust not being granted.');
```
dibarbet: This should be `info` or higher

dibarbet:
> Makes sense. I admit it's tempting to also use this as an opportunity to expand our extension a bit to support limited mode in other ways. i.e. we can't load projects and can't do design time builds, but the base miscellaneous files experience could at least light up...

only if you do the security review for that 😆

src/main.ts
jasonmalinowski: Is that what caused the issues we had while @RikkiGibson was debugging this? Because a window reload doesn't restart the extension host anymore?

jasonmalinowski:
> only if you do the security review for that 😆

It's "tempting" in the "forbidden apple" sense of tempting. :smile:

src/main.ts
JoeRobich: Yes, perhaps it could read differently when CDK is installed but the C# extension should own this dialog.

src/main.ts
dibarbet: I don't think so, reload window should still restart everything.  but restart extensionhost should be less intrusive as it doesn't reload the entire window.  when you install extensions, it used to tell you to reload window, now it tells you to restart the extension host.

JoeRobich: > workbench.action.restartExtensionHost might be the more modern way to do this

Maybe we want to update "reload on option change" too.

package.json
JoeRobich: Good call. Even though we are not registering any command handlers, it would be confusing to see them offered in the command prompt. I don't believe it is necessary to restrict our settings, since we are not using those settings for anything.

src/main.ts
dibarbet: Yes, likely we should

RikkiGibson: Reloading the window will definitely reload all the extensions, but, `restartExtensionHost` as best I can tell doesn't necessarily reload extensions which were already loaded. Or, at least, it doesn't nuke everything possible about those already-loaded extensions, which reloading the window will do.

That is why prior to dotnet/vscode-csharp#8428, if you loaded C# by itself and later on loaded CDK, the `projectNeedsRestore` handler for C# base would still be active. It seems like if that action was supposed to reload C# also, then, I would have expected that handler to not be attached. That was part of why the bug was a little harder to repro. You had to load C#+CDK and then reload the window, simply following the prompts to relaunch this and that was not enough.

dibarbet: > restartExtensionHost as best I can tell doesn't necessarily reload extensions which were already loaded. Or, at least, it doesn't nuke everything possible about those already-loaded extensions, which reloading the window will do.

that is surprising - because all extensions run in the extension host process, so they should get completely reloaded if the process is restarted.  And you can confirm this in the extension host logs, when you invoke the command you see something like
> 2025-07-23 11:05:49.090 [info] Extension host with pid 35084 exiting with code 0
2025-07-23 11:05:49.642 [info] Extension host with pid 54412 started

I wonder if something else was going on in your scenario, where on the restart extensions it still did not detect the CDK installation for some reason.  I don't see how an event subscription could stick around when the process is killed

RikkiGibson: I am suspecting that loading an additional extension doesn't trigger `restartExtensionHost`, but unloading an already loaded extension does. See also https://github.com/dotnet/vscode-csharp/pull/8428#discussion_r2220941396

RikkiGibson: Based on this, I am thinking that checking "is CDK loaded" at startup, is ok only for the special case of deciding to register a command or not. That's because the editor will continue to do the right thing if CDK is loaded later, by filtering out any unwanted commands based on enablements settings.

But it is not ok for pretty much anything else. You need to check if CDK is loaded at the time you are doing whatever action.

Sorry if this is getting somewhat irrelevant to the issue actually being addressed in this PR. But hopefully it sheds some light.

dibarbet (approved):

PR #8469, opened 2025-07-29
Update Razor cohosting to be 'on' by default.

phil-allen-msft:
Update Razor cohosting to be 'on' by default. 

Tested locally with a multi-project Blazor solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1560,7 +1560,7 @@
           },
           "razor.languageServer.cohostingEnabled": {
             "type": "boolean",
-            "default": false,
+            "default": true,
             "description": "%configuration.razor.languageServer.cohostingEnabled%",
             "tags": [
               "experimental"


davidwengier (approved):

dibarbet (changes requested):
Forgot one thing - can you also add a line in the changelog for this?  Likely want to mention how people can turn it off if they have issues etc...

I also see the normal (previously non-cohosting) Razor tests are all failing - possibly they need to explicitly turn cohosting off now in the test config - https://github.com/dotnet/vscode-csharp/blob/main/test/razor/razorIntegrationTests/testAssets/RazorApp/.vscode/devkit_RazorApp.code-workspace

davidwengier:
> possibly they need to explicitly turn cohosting off now in the test config

Thats a _very_ good point I should have thought of.

However, the fact that they're failing is surprising, as it implies that cohosting+devkit is having issues.

davidwengier:
Ran this branch up locally and manually verified, and everything works fine. Startup is pretty slow with devkit, so suspect that is what is going on. I think the failing tests are the only ones that don't retry. Also one of them is badly named :)

Have pushed up the settings change suggested, since it makes sense anyway.

JoeRobich:
Maybe we should be using https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options

phil-allen-msft:
Changelog updated.

PR #8476, opened 2025-07-31
Prompt to restart extensions when CDK is loaded

### LINKED ISSUE #8475: Change in the CDK loaded state should prompt to reload both C# + CDK extensions
It looks like currently we have a prompt to "Restart language server" when this happens. But I think it doesn't cause the C# extension itself to reload. Since the C# extension conditions some things on whether CDK is also loaded, it is best if we reload both. The "restart extension host" command would be fine here, or, if there is a more granular restart API we can use, we could use that instead.

This change is thought to also address issues around `isCdkLoaded()` checks in general. Since simply prompting to reload is going to push us toward patterns where CDK is either always loaded or never loaded, and not needing to compensate for the load state changing over time in a granular way.

RikkiGibson:
Closes #8475

Copied the pattern from #8435 to restart extensions

### DIFF
--- a/src/lsptoolshost/server/roslynLanguageServer.ts
+++ b/src/lsptoolshost/server/roslynLanguageServer.ts
@@ -962,7 +962,7 @@ export class RoslynLanguageServer {
                 };
                 if (csharpDevkitExtension && !_wasActivatedWithCSharpDevkit) {
                     // We previously started without C# Dev Kit and it's now installed.
-                    // Offer a prompt to restart the server to use C# Dev Kit.
+                    // Offer a prompt to restart extensions in order to use C# Dev Kit.
                     this._channel.info(`Detected new installation of ${csharpDevkitExtensionId}`);
                     const message = `Detected installation of C# Dev Kit. Please reload the C# extension to continue.`;
                     showInformationMessage(vscode, message, title);


src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -957,17 +957,17 @@ export class RoslynLanguageServer {
...
                 const title: CommandOption = {
-                    title: vscode.l10n.t('Restart Language Server'),
-                    command: 'dotnet.restartServer',
+                    title: vscode.l10n.t('Reload C# Extension'),
+                    command: 'workbench.action.restartExtensionHost',
                 };
                 if (csharpDevkitExtension && !_wasActivatedWithCSharpDevkit) {
-                    // We previously started without C# Dev Kit and its now installed.
+                    // We previously started without C# Dev Kit and it's now installed.
                     // Offer a prompt to restart the server to use C# Dev Kit.
```
dibarbet:
```suggestion
                    // Offer a prompt to restart extensions in order to use C# Dev Kit.
```

dibarbet (approved):

RikkiGibson:
I manually verified that hitting the button reloads the C# extension and causes all our activation code such as attaching restore handlers to run again from fresh.

jasonmalinowski (approved):

JoeRobich (approved):

PR #8489, opened 2025-08-04
Add configurations to enable file nesting in explorer

dibarbet:
Resolves https://github.com/microsoft/vscode-dotnettools/issues/1823

### DIFF
--- a/package.json
+++ b/package.json
@@ -5642,7 +5642,16 @@
       },
       "[csharp]": {
         "editor.formatOnType": true
-      }
+      },
+      "explorer.fileNesting.patterns": {
+        "*.cs": "${capture}.designer.cs,${capture}.g.cs,${capture}.generated.cs,${capture}.run.json,${capture}.settings.json,${capture}.settings.*.json",
+        "*.csproj": "${capture}.csproj.user",
+        "*.json": "${capture}.Development.json",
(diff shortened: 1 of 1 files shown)


JoeRobich (approved):

jasonmalinowski (approved):

package.json
```diff
@@ -5642,7 +5642,16 @@
...
+      },
+      "explorer.fileNesting.patterns": {
+        "*.cs": "${capture}.designer.cs,${capture}.g.cs,${capture}.generated.cs,${capture}.run.json,${capture}.settings.json,${capture}.settings.*.json",
+        "*.csproj": "${capture}.csproj.user",
+        "*.json": "${capture}.Development.json",
+        "*.cshtml": "${capture}.cshtml.cs,${capture}.cshtml.css",
+        "*.razor": "${capture}.razor.cs,${capture}.razor.css",
+        "*.xaml": "${capture}.xaml.cs"
+      },
+      "explorer.fileNesting.enabled": true
```
jasonmalinowski: Does this mean we're turning this behavior on generally for users? So other patterns from other extensions might also light up?

jasonmalinowski: I think I'm giving a similar answer to @timheuer at https://github.com/microsoft/vscode-dotnettools/issues/1823#issuecomment-3152749273. This is a user preference that might impact other extensions, let's not surprise them with that. The other part is fine.

PR #8503, opened 2025-08-06
Include logs when reporting an issue

dibarbet:
image

Test issue: https://github.com/dotnet/vscode-csharp/issues/8502

### DIFF
--- a/package.json
+++ b/package.json
@@ -5627,6 +5627,11 @@
           "when": "editorLangId == csharp && (dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp')",
           "group": "2_dotnet@2"
         }
+      ],
+      "issue/reporter": [
+        {
+          "command": "csharp.reportIssue"
+        }
       ]
     },
     "viewsWelcome": [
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -5627,6 +5627,11 @@
           "when": "editorLangId == csharp && (dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'OmniSharp')",
           "group": "2_dotnet@2"
         }
+      ],
+      "issue/reporter": [
```
dibarbet, line 5631: This allows the issue reporter dialog to call into us when we're not directly invoked via `csharp.reportIssue`

src/shared/reportIssue.ts
```diff
@@ -142,25 +146,47 @@ async function getMonoIfPlatformValid(monoResolver: IHostExecutableResolver): Pr
...
-        
-### C# log ###
-<details>Post the output from Output-->C# here</details>
+`;
 
-### C# LSP Trace Logs ###
-<details>Post the output from Output-->C# LSP Trace Logs here.  Requires the \`C#\` output window log level to be set to \`Trace\`</details>`;
+    for (const channel of logChannels) {
+        const contents = await getLogOutputChannelContents(context, channel);
+        logInfo += `### ${channel.name} log ###
```
JoeRobich, line 159: Maybe the Logs section should have a header, but the channel.name should go into the details `<summary>`.

JoeRobich (approved):

src/shared/reportIssue.ts
dibarbet: I do prefer this rendering as its easy to see the split between the sections, and you don't have to expand both if you dont need to
image

PR #8507, opened 2025-08-07
Add release comment handler workflow for issue labeling

claudiaregio:


JoeRobich:
Will there also be infra for automatically tagging issues when the publishing pipeline is run?

claudiaregio:
@JoeRobich Not at this time, I'm tagging them manually as the releases actually happen right now to keep track of where everything is. Was talking to David yesterday though and there may be a better way to handle these in batches in the future but for right now I'd like to manually keep the pulse.

JoeRobich (approved):

PR #8508, opened 2025-08-07
Pipeline to facilitate Roslyn Copilot publish

AbhitejJohn:
This pipeline ensures we get the latest roslyn copilot bits. IT needs to be approved explicitly from someone with a SAW, so I've left it as a manual trigger for now. I could imagine a release flow though that auto triggers some of these and that one person with a SAW makes all the approvals needed at once.


azure-pipelines/publish-roslyn-copilot.yml
```diff
@@ -0,0 +1,58 @@
+trigger: none
+pr: none
+
+resources:
```
dibarbet, line 4: Kind of a meta question here.  Should this pipeline even be in this repo?  Would conversations be a better place?

For example if you want to change the folder name where it finds the conversations build artifacts to upload.  Now you have to make the change across two repos.  If this was in conversations you just updated both at the same time.

AbhitejJohn: Good question, my thinking was having this closer to the component that actually needs it. The conversations repo has a slightly different set of processes tied to build and release pipelines which are tied to VS. Having it here gives us the freedom to iterate on this quicker. It also keeps this in our teams cross hairs a little more working with the processes we have - its been hard figuring out where things exist, so trying to keep some of the infra pieces atleast in fewer places.

AbhitejJohn:
@dibarbet : Is there an infra problem causing these test failures? I wouldn't expect these changes to fail the build. I could also use another review on this as well so I can close this out :)

dibarbet:
> Is there an infra problem causing these test failures? I wouldn't expect these changes to fail the build. I could also use another review on this as well so I can close this out :)

Yeah looks like there are some new flaky tests

dibarbet:
Have one question here - https://github.com/dotnet/vscode-csharp/pull/8508#discussion_r2261660884

AbhitejJohn:
> Have one question here - [#8508 (comment)](https://github.com/dotnet/vscode-csharp/pull/8508#discussion_r2261660884)

And responded : https://github.com/dotnet/vscode-csharp/pull/8508#discussion_r2262033634 :) I'm happy to chat if you feel strongly though.

azure-pipelines/publish-roslyn-copilot.yml
dibarbet: Understandable.  If we were totally disregarding process then I'd definitely prefer this lives in conversations (conversations produces the assets, so it should also own the code publishing them).  

But if its going to be a significant burden to put it there, I'm not that opposed to it being here.

dibarbet (approved):

dibarbet:
> /azp run dotnet.vscode-csharp

you should be able to just re-queue the individual job (either from the github checks page or AzDo page

PR #8510, opened 2025-08-07
Enhance dotnet.generateAssets command to accept skipPrompt parameter for asset generation

claudiaregio:
Bug: https://github.com/dotnet/vscode-csharp/issues/8414

### DIFF
--- a/src/lsptoolshost/debugger/debugger.ts
+++ b/src/lsptoolshost/debugger/debugger.ts
@@ -53,8 +53,8 @@ export function registerDebugger(
     );
 
     context.subscriptions.push(
-        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex) => {
-            if (!(await promptForDevKitDebugConfigurations())) {
+        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex, skipPrompt = false) => {
+            if (!(await promptForDevKitDebugConfigurations(skipPrompt))) {
                 return;
             }
 
@@ -63,8 +63,13 @@ export function registerDebugger(
     );
 }
 
-async function promptForDevKitDebugConfigurations(): Promise<boolean> {
+async function promptForDevKitDebugConfigurations(skipPrompt: boolean = false): Promise<boolean> {
     if (getCSharpDevKit()) {
+        // If skipPrompt is true, proceed with generating assets without showing the dialog
+        if (skipPrompt) {
+            return true;
+        }
+
         let result: boolean | undefined = undefined;
 
         while (result === undefined) {
--- a/src/lsptoolshost/debugger/debugger.ts
+++ b/src/lsptoolshost/debugger/debugger.ts
@@ -66,7 +66,7 @@ export function registerDebugger(
 async function promptForDevKitDebugConfigurations(skipPrompt: boolean = false): Promise<boolean> {
     if (getCSharpDevKit()) {
         // If skipPrompt is true, proceed with generating assets without showing the dialog
-        if (skipPrompt) {
+        if (options.skipPrompt) {
             return true;
         }
 
--- a/src/lsptoolshost/debugger/debugger.ts
+++ b/src/lsptoolshost/debugger/debugger.ts
@@ -63,7 +63,7 @@ export function registerDebugger(
     );
 }
 
-async function promptForDevKitDebugConfigurations(skipPrompt: boolean = false): Promise<boolean> {
+async function promptForDevKitDebugConfigurations(options: { skipPrompt?: boolean }): Promise<boolean> {
     if (getCSharpDevKit()) {
         // If skipPrompt is true, proceed with generating assets without showing the dialog
         if (options.skipPrompt) {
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/debugger/debugger.ts
```diff
@@ -53,8 +53,8 @@ export function registerDebugger(
     );
 
     context.subscriptions.push(
-        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex) => {
```
WardenGnaw, line 56: To future proof from additional options and just appending more and more parameters.

```suggestion
        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex: number, options: { skipPrompt?: boolean } = {}) => {
```

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -63,8 +63,13 @@ export function registerDebugger(
     );
 }
 
-async function promptForDevKitDebugConfigurations(): Promise<boolean> {
+async function promptForDevKitDebugConfigurations(skipPrompt: boolean = false): Promise<boolean> {
     if (getCSharpDevKit()) {
+        // If skipPrompt is true, proceed with generating assets without showing the dialog
+        if (skipPrompt) {
```
WardenGnaw:
```suggestion
        if (options.skipPrompt) {
```

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -63,8 +63,13 @@ export function registerDebugger(
     );
 }
 
-async function promptForDevKitDebugConfigurations(): Promise<boolean> {
+async function promptForDevKitDebugConfigurations(skipPrompt: boolean = false): Promise<boolean> {
```
WardenGnaw: Update to use new object, and no need to use a default here.

```suggestion
async function promptForDevKitDebugConfigurations(options: { skipPrompt?: boolean }): Promise<boolean> {
```

src/lsptoolshost/debugger/debugger.ts
claudiaregio: replaced!

claudiaregio:
@WardenGnaw put everything you suggested! Let me know if it's ok now

AbhitejJohn (approved):
LGTM provided @WardenGnaw is good with these updates. So please wait for his approval. 

Also tagging @dibarbet and @JoeRobich  to see if they have any concerns.

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -53,8 +53,8 @@ export function registerDebugger(
     );
 
     context.subscriptions.push(
-        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex) => {
-            if (!(await promptForDevKitDebugConfigurations())) {
+        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex: number, options: { skipPrompt?: boolean } = {}) => {
+            if (!(await promptForDevKitDebugConfigurations(skipPrompt))) {
```
WardenGnaw:
```suggestion
            if (!(await promptForDevKitDebugConfigurations(options))) {
```

src/lsptoolshost/debugger/debugger.ts
```diff
@@ -53,8 +53,8 @@ export function registerDebugger(
     );
 
     context.subscriptions.push(
-        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex) => {
-            if (!(await promptForDevKitDebugConfigurations())) {
+        vscode.commands.registerCommand('dotnet.generateAssets', async (selectedIndex: number, options: { skipPrompt?: boolean } = {}) => {
```
dibarbet: I assume the expectation is that the container tools extension is calling this programatically and can pass in the value they want here?

Additionally, generate assets can be called from the command palette directly, but generally command palette options don't pass in arguments.  Can you verify the generate assets command still works there?

dibarbet: image

WardenGnaw: Yep. The arguments shouldnt matter via command since its only used for calling it via the command API.

We already do this today with `selectedIndex`, now we are just creating another parameter for `options`

WardenGnaw:
@claudiaregio You will need to run the linter with `--fix`

src/lsptoolshost/debugger/debugger.ts
dibarbet: > Yep. The arguments shouldnt matter via command since its only used for calling it via the command API.

However the command is exposed today as a user-invokable command.  Should we remove that?

dibarbet (approved):

src/lsptoolshost/debugger/debugger.ts
WardenGnaw: Its intended to be invokable, and if users have DevKit installed, they should see the prompt.

PR #8511, opened 2025-08-07
Register Blazor WASM debugger in cohosting

davidwengier:
Fixes https://github.com/dotnet/vscode-csharp/issues/8494


davidwengier:
@dibarbet FYI CSharpIntegrationTests seems to have broken, and I'd be very surprised if it was because of this change :)

PR #8529, opened 2025-08-13
Capture separate logs for each lsptoolshost test suite

JoeRobich:


Directory.Build.props
```diff
@@ -5,6 +5,7 @@
             Defines the lowest supported target framework for the extension.
             Used by server download / integration tests to ensure they run when only this SDK is installed.
         -->
+        <CheckEolTargetFramework>false</CheckEolTargetFramework>
```
JoeRobich, line 8: Noticed project loading did not like running on a .NET 10 preview while building net6.0 projects.

test/jestHelpers.ts
```diff
@@ -0,0 +1,55 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { EOL, platform } from 'os';
+import { describe, expect, test } from '@jest/globals';
```
JoeRobich, line 8: Separated these helpers out because they use `@jest/globals` and importing it isn't allowed in the test runner (index.ts).

test/lsptoolshost/integrationTests/index.ts
```diff
@@ -3,11 +3,71 @@
...
+
+        const testFiles = (await fsExtra.readdir(__dirname))
+            .filter((file) => file.endsWith('.integration.test.js'))
+            .map((file) => path.join(__dirname, file));
+
+        for (let file of testFiles) {
+            // We have to fix up the file path because the test file was discovered in the /out/ directory.
+            file = file.substring(0, file.length - 2) + 'ts';
+            file = file.replace(/[\\/]out[\\/]/, path.sep);
+            file = workingDirectory[0] + file.substring(1);
```
JoeRobich, line 37: So locally `__dirname` is returning the drive letter lowercased and the filepaths also are lowercased. Jest fails to match the paths to a test suite if I don't bring in a correct drive letter. Thankfully the working directory has the correct casing. Figure on unix-like platforms it should basically be a no-op.

test/lsptoolshost/integrationTests/index.ts
```diff
@@ -3,11 +3,73 @@
...
+        anyFailures = anyFailures || !results.success;
+    } else {
+        const workingDirectory = process.cwd();
+
+        const testFiles = (await fsExtra.readdir(__dirname))
+            .filter((file) => file.endsWith('.integration.test.js'))
+            .map((file) => path.join(__dirname, file));
+
+        for (let file of testFiles) {
+            // We have to fix up the file path because the test file was discovered in the /out/ directory.
```
dibarbet, line 34: do sourcemaps still work if we do this?  e.g. f5 launching integration tests is fine?

test/lsptoolshost/integrationTests/index.ts
```diff
@@ -3,11 +3,73 @@
...
+    }
+
+    if (!exports.logDirectory) {
+        console.warn(`Failed to get log directory from C# extension exports`);
+        return;
+    }
+
+    const targetLogDir = path.join(
+        path.dirname(exports.logDirectory),
+        `${name ?? 'unknown'}_${path.basename(exports.logDirectory)}`
```
dibarbet, line 69: can remove unknown, name isn't nullable here

test/lsptoolshost/integrationTests/index.ts
```diff
@@ -3,11 +3,73 @@
...
+    let results: AggregatedResult | undefined;
+    if (process.env.TEST_FILE_FILTER) {
+        results = await runIntegrationTests(jestIntegrationTestProjectName);
+        await moveLogs(path.basename(process.env.TEST_FILE_FILTER, '.integration.test.ts'));
+        anyFailures = anyFailures || !results.success;
+    } else {
+        const workingDirectory = process.cwd();
+
+        const testFiles = (await fsExtra.readdir(__dirname))
+            .filter((file) => file.endsWith('.integration.test.js'))
```
dibarbet, line 30: instead of doing this, I think you could probably simplify all this by getting the [rootPath](https://github.com/dotnet/vscode-csharp/blob/main/tasks/projectPaths.ts#L10) here, and then just go directly to the typescript files?

test/lsptoolshost/integrationTests/index.ts
```diff
@@ -3,11 +3,73 @@
...
+            file = file.replace(/[\\/]out[\\/]/, path.sep);
+            file = workingDirectory[0] + file.substring(1);
+
+            console.log('');
+            console.log(`-- Running integration tests for ${path.basename(file)} --`);
+            console.log('');
+
+            process.env.TEST_FILE_FILTER = file;
+
+            results = await runIntegrationTests(jestIntegrationTestProjectName);
```
dibarbet, line 45: I'm wondering if this (or something above it) needs to catch errors and then move the logs - I noticed that in the 'Test Linux (.NET 9) CSharpIntegrationTests' leg, it looks like the `sourceGenerator.integration.test.ts` fail with an extension host crash, and there are no logs uploaded for it.

Image

JoeRobich: The current runs logs should still be in the logs folder even if they aren't moved.

test/lsptoolshost/integrationTests/index.ts
JoeRobich: our launch targets set the `TEST_FILTER_FILE` so it shouldn't run this bit of code.

PR #8539, opened 2025-08-14
Build all VSIX's on Linux and switch to pipeline artifacts

dibarbet:
Previously we built on all 3 platforms, and then uploaded to a unified vsix artifact (using azdo build artifacts).  But these are soon not being allowed, and we can only use immutable pipeline artifacts.

This changes us to build all vsix's on one OS, and upload all the vsixes at once to the same artifact.

We build on linux due to issues with bundling the linux executable in the vsix on windows.

Also resolves https://github.com/dotnet/vscode-csharp/issues/8497


azure-pipelines/build-vsix.yml
```diff
@@ -0,0 +1,193 @@
+parameters:
```
dibarbet, line 1: this is mostly the same as the build before, but without the platform parameter

azure-pipelines/validate-build.yml
```diff
@@ -0,0 +1,41 @@
+parameters:
```
dibarbet, line 1: new template for running builds on all platforms for validation in CI

azure-pipelines/release.yml
```diff
@@ -132,7 +133,7 @@ extends:
                     }
                     Else {
                       Write-Host "##[command]vsce $publishArgs"
-                      vsce @publishArgs
+                      #vsce @publishArgs
```
JoeRobich: Should this be commented out?

dibarbet: oops fixed

azure-pipelines/release.yml
```diff
@@ -4,7 +4,7 @@ pr: none
 parameters:
   - name: test
     type: boolean
-    default: true
+    default: false
```
dibarbet, line 7: was requested by Vikram to change

JoeRobich (approved):

PR #8556, opened 2025-08-15
Verify signatures of real signed vsixs

JoeRobich:


tasks/spawnNode.ts
```diff
@@ -27,5 +27,29 @@ export default async function spawnNode(args?: string[], options?: SpawnSyncOpti
 
     const buffer = spawnSync(nodePath, args, optionsWithFullEnvironment);
 
-    return { code: buffer.status, signal: buffer.signal };
+    return { code: buffer.status, signal: buffer.signal, stdout: buffer.stdout };
+}
+
+export async function spawnNodeWithOutput(args?: string[], options?: SpawnSyncOptions) {
```
dibarbet: can this be rolled into the prior function with an optional parameter?  Seems like the only difference is the `stdio` option?

Alternatively, allow the passed in options to set the `stdio` value?

azure-pipelines/build-vsix.yml
```diff
@@ -176,6 +176,12 @@ stages:
           SignType: $(signType)
           SYSTEM_ACCESSTOKEN: $(System.AccessToken)
 
+    - ${{ if eq(parameters.isOfficial, true) }}:
```
dibarbet: Potentially we should move this to after we upload, so someone can inspect the VSIX to debug why verification might have failed?

tasks/spawnNode.ts
JoeRobich: Yes!

azure-pipelines/build-vsix.yml
JoeRobich: Can't move to after upload because they are uploaded as outputs from the build. I did mark them as always publish and moved the verify step until after they are moved to Packages folder.

PR #8563, opened 2025-08-25
Dotnet vscode csharp insertion 

deepakrathore33:
* Using gulp tasks instead of bash script

### DIFF
--- a/tasks/insertionTasks.ts
+++ b/tasks/insertionTasks.ts
@@ -1,0 +1,308 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as gulp from 'gulp';
+import * as fs from 'fs';
+import * as path from 'path';
+import minimist from 'minimist';
+import { Octokit } from '@octokit/rest';
+import { exec } from 'child_process';
+import { promisify } from 'util';
+import fetch from 'node-fetch';
+import * as xml2js from 'xml2js';
+import * as AdmZip from 'adm-zip';
+
+const execAsync = promisify(exec);
+
+interface InsertionOptions {
+    roslynVersion?: string;
+    roslynStartSHA?: string;
+    roslynEndSHA?: string;
+    roslynBuildId?: string;
+    roslynBuildNumber?: string;
+    assetManifestPath?: string;
+    createPullRequest?: boolean;
+    targetBranch?: string;
+    githubPAT?: string;
+    dryRun?: boolean;
+}
+
+gulp.task('insertion:roslyn', async (): Promise<void> => {
+    const options = minimist<InsertionOptions>(process.argv.slice(2));
+    
+    console.log('Starting Roslyn insertion process...');
+    console.log(`Options: ${JSON.stringify(options, null, 2)}`);
+    
+    try {
+        // Step 1: Extract Roslyn version from AssetManifest
+        if (!options.assetManifestPath) {
+            throw new Error('assetManifestPath is required');
+        }
+        
+        const newVersion = await extractRoslynVersionFromManifest(options.assetManifestPath);
+        if (!newVersion) {
+            throw new Error('Failed to extract Roslyn version from asset manifest');
+        }
+        options.roslynVersion = newVersion;
+        console.log(`New Roslyn version: ${newVersion}`);
+        
+        // Step 2: Get current SHA from package
+        const currentSHA = await getCurrentRoslynSHA();
+        options.roslynStartSHA = currentSHA || '0000000000000000000000000000000000000000';
+        console.log(`Current Roslyn SHA: ${options.roslynStartSHA}`);
+        
+        // Step 3: Check if update needed
+        if (!options.roslynEndSHA) {
+            throw new Error('roslynEndSHA is required');
+        }
+        
+        if (options.roslynStartSHA === options.roslynEndSHA) {
+            console.log('No new commits to process - versions are identical');
+            return;
+        }
+        
+        console.log(`Update needed: ${options.roslynStartSHA}..${options.roslynEndSHA}`);
+        
+        // Step 4: Clone Roslyn repo and generate PR list
+        await cloneRoslynRepo();
+        const prList = await generatePRList(options.roslynStartSHA, options.roslynEndSHA);
+        console.log('PR List generated:', prList);
+        
+        // Step 5: Update files
+        await updatePackageJson(options.roslynVersion);
+        await updateChangelog(options.roslynVersion, prList, options.roslynBuildNumber, options.roslynBuildId);
+        
+        // Step 6: Create branch and PR if not dry run
+        if (!options.dryRun) {
+            const branchName = await createBranch(options.roslynEndSHA, options.roslynVersion);
+            console.log(`Branch created: ${branchName}`);
+            
+            if (options.createPullRequest) {
+                await createPullRequest(branchName, options);
+                console.log('Pull request created successfully');
+            }
+        } else {
+            console.log('Dry run mode - skipping branch and PR creation');
+        }
+        
+    } catch (error) {
+        console.error('Insertion failed:', error);
+        throw error;
+    } finally {
+        // Cleanup
+        if (fs.existsSync('roslyn')) {
+            await execAsync('rm -rf roslyn');
+        }
+    }
+});
+
+async function extractRoslynVersionFromManifest(manifestPath: string): Promise<string | null> {
+    const xmlFile = path.join(manifestPath, 'OfficialBuild.xml');
+    
+    if (!fs.existsSync(xmlFile)) {
+        console.error(`OfficialBuild.xml not found at ${xmlFile}`);
+        return null;
+    }
+    
+    const xmlContent = fs.readFileSync(xmlFile, 'utf8');
+    const parser = new xml2js.Parser();
+    const result = await parser.parseStringPromise(xmlContent);
+    
+    // Navigate the XML structure to find Microsoft.CodeAnalysis package
+    const packages = result?.Build?.Package || [];
+    for (const pkg of packages) {
+        const attrs = pkg.$;
+        if (attrs?.Id === 'Microsoft.CodeAnalysis' || attrs?.Id === 'Microsoft.CodeAnalysis.Common') {
+            return attrs.Version;
+        }
+    }
+    
+    return null;
+}
+
+async function getCurrentRoslynSHA(): Promise<string | null> {
+    const packageJsonContent = fs.readFileSync('package.json', 'utf8');
+    const packageJson = JSON.parse(packageJsonContent);
+    const currentVersion = packageJson.defaults?.roslyn;
+    
+    if (!currentVersion) {
+        console.log('No roslyn version in package.json, this is first run');
+        return null;
+    }
+    
+    console.log(`Current Roslyn version: ${currentVersion}`);
+    
+    // Download and extract commit SHA from NuGet package
+    const packageName = 'microsoft.codeanalysis.common';
+    const feed = 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/flat2';
+    const packageUrl = `${feed}/${packageName}/${currentVersion}/${packageName}.${currentVersion}.nupkg`;
+    
+    try {
+        const response = await fetch(packageUrl);
+        if (!response.ok) {
+            console.error(`Failed to download package: ${response.statusText}`);
+            return null;
+        }
+        
+        const buffer = await response.buffer();
+        const tempFile = path.join(process.cwd(), 'temp-package.nupkg');
+        fs.writeFileSync(tempFile, buffer);
+        
+        const zip = new AdmZip(tempFile);
+        const entries = zip.getEntries();
+        
+        for (const entry of entries) {
+            if (entry.entryName.endsWith('.nuspec')) {
+                const nuspecContent = zip.readAsText(entry);
+                const shaMatch = nuspecContent.match(/repository[^>]*commit="([a-f0-9]{40})"/);
+                fs.unlinkSync(tempFile);
+                return shaMatch ? shaMatch[1] : null;
+            }
+        }
+        
+        fs.unlinkSync(tempFile);
+    } catch (error) {
+        console.error('Error getting current SHA:', error);
+    }
+    
+    return null;
+}
+
+async function cloneRoslynRepo(): Promise<void> {
+    console.log('Cloning Roslyn repository...');
+    await execAsync('git clone --no-tags --filter=blob:none --depth=500 https://github.com/dotnet/roslyn.git roslyn');
+}
+
+async function generatePRList(startSHA: string, endSHA: string): Promise<string> {
+    console.log(`Generating PR list from ${startSHA} to ${endSHA}...`);
+    
+    // Setup auth for roslyn-tools
+    const homeDir = process.env.HOME || process.env.USERPROFILE;
+    const settingsDir = path.join(homeDir!, '.roslyn-tools');
+    if (!fs.existsSync(settingsDir)) {
+        fs.mkdirSync(settingsDir, { recursive: true });
+    }
+    
+    const authJson = {
+        GitHubToken: process.env.GITHUB_TOKEN || '',
+        DevDivAzureDevOpsToken: '',
+        DncEngAzureDevOpsToken: ''
+    };
+    const settingsFile = path.join(settingsDir, 'settings');
+    fs.writeFileSync(settingsFile, Buffer.from(JSON.stringify(authJson)).toString('base64'));
+    
+    try {
+        const { stdout } = await execAsync(
+            `cd roslyn && roslyn-tools pr-finder -s "${startSHA}" -e "${endSHA}" --format changelog --label VSCode`,
+            { maxBuffer: 10 * 1024 * 1024 } // 10MB buffer
+        );
+        return stdout || '(no PRs with required labels)';
+    } catch (error) {
+        console.warn('PR finder failed, using empty list:', error);
+        return '(no PRs with required labels)';
+    }
+}
+
+async function updatePackageJson(newVersion: string): Promise<void> {
+    console.log(`Updating package.json with Roslyn version ${newVersion}...`);
+    const packageJsonPath = 'package.json';
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+    
+    if (!packageJson.defaults) {
+        packageJson.defaults = {};
+    }
+    packageJson.defaults.roslyn = newVersion;
+    
+    fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2) + '\n');
+}
+
+async function updateChangelog(version: string, prList: string, buildNumber?: string, buildId?: string): Promise<void> {
+    console.log('Updating CHANGELOG.md...');
+    const changelogPath = 'CHANGELOG.md';
+    let changelogContent = fs.readFileSync(changelogPath, 'utf8');
+    
+    // Format PR list with proper indentation
+    const formattedPRList = prList.split('\n').map(line => `  ${line}`).join('\n');
+    
+    // Find and update the Roslyn bump line
+    const lines = changelogContent.split('\n');
+    const newLines: string[] = [];
+    let skipNext = false;
+    
+    for (let i = 0; i < lines.length; i++) {
+        if (skipNext && lines[i].startsWith('  *')) {
+            continue;
+        }
+        skipNext = false;
+        
+        if (lines[i].match(/^\* Bump Roslyn to/)) {
+            const prLink = buildNumber && buildId 
+                ? `[${buildNumber}](https://dev.azure.com/dnceng/internal/_build/results?buildId=${buildId})`
+                : '[#TBD](TBD)';
+            newLines.push(`* Bump Roslyn to ${version} (PR: ${prLink})`);
+            if (prList !== '(no PRs with required labels)') {
+                newLines.push(formattedPRList);
+            }
+            skipNext = true;
+        } else {
+            newLines.push(lines[i]);
+        }
+    }
+    
+    fs.writeFileSync(changelogPath, newLines.join('\n'));
+}
+
+async function createBranch(endSHA: string, version: string): Promise<string> {
+    console.log('Creating and pushing branch...');
+    
+    await execAsync('git config user.name "Azure Pipelines"');
+    await execAsync('git config user.email "azuredevops@microsoft.com"');
+    
+    const shortSHA = endSHA.substring(0, 8);
+    const branchName = `roslyn-bump/${shortSHA}`;
+    
+    await execAsync(`git checkout -b ${branchName}`);
+    await execAsync('git add package.json CHANGELOG.md');
+    await execAsync(`git commit -m "Bump Roslyn to ${version}"`);
+    await execAsync(`git push origin ${branchName}`);
+    
+    return branchName;
+}
+
+async function createPullRequest(branchName: string, options: InsertionOptions): Promise<void> {
+    console.log('Creating pull request...');
+    
+    if (!options.githubPAT) {
+        console.warn('No GitHub PAT provided, skipping PR creation');
+        return;
+    }
+    
+    const octokit = new Octokit({ auth: options.githubPAT });
+    
+    const prTitle = `Bump Roslyn to ${options.roslynVersion}`;
+    const prBody = `Automated Roslyn version bump.
+
+**Version:** \`${options.roslynVersion}\`
+**Commit Range:** \`${options.roslynStartSHA}...${options.roslynEndSHA}\`
+**Azure DevOps Build:** [${options.roslynBuildNumber}](https://dev.azure.com/dnceng/internal/_build/results?buildId=${options.roslynBuildId})
+
+See CHANGELOG.md for included PRs.`;
+    
+    try {
+        const { data } = await octokit.pulls.create({
+            owner: 'dotnet',
+            repo: 'vscode-csharp',
+            title: prTitle,
+            body: prBody,
+            head: branchName,
+            base: options.targetBranch || 'main'
+        });
+        
+        console.log(`Pull request created: ${data.html_url}`);
+    } catch (error) {
+        console.error('Failed to create PR:', error);
+        throw error;
+    }
+}
--- a/azure-pipelines/roslyn-version-bump.yml
+++ b/azure-pipelines/roslyn-version-bump.yml
@@ -19,29 +19,25 @@ resources:
     ref: refs/tags/release
   pipelines:
     - pipeline: officialBuildCI
-      source: 327
+      source: dotnet-roslyn-official
       project: internal
-      branch: refs/heads/main
+      branch: main
       trigger: none
 
 variables:
-  - name: RoslynStartSHA
-    value: ""
   - name: RoslynEndSHA
     value: $(resources.pipeline.officialBuildCI.sourceCommit)
   - name: RoslynBuildNumber
     value: $(resources.pipeline.officialBuildCI.runName)
   - name: RoslynBuildId
     value: $(resources.pipeline.officialBuildCI.runID)
-  - name: RoslynVersion
-    value: ""
 
 extends:
-  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  template: v1/1ES.Unofficial.PipelineTemplate.yml@1ESPipelineTemplates
   parameters:
     pool:
       name: AzurePipelines-EO
-      image: 1ESPT-Ubuntu22.04
+      image: AzurePipelinesUbuntu22.04compliant
       os: linux
     stages:
       - stage: BumpRoslyn
@@ -51,16 +47,8 @@ extends:
             displayName: Process Roslyn Bump
             pool:
               name: AzurePipelines-EO
-              image: 1ESPT-Ubuntu22.04
+              image: AzurePipelinesUbuntu22.04compliant
               os: linux
-            templateContext:
-              type: releaseJob
-              isProduction: false
-              inputs:
-              - input: pipelineArtifact
-                pipeline: officialBuildCI
-                artifactName: AssetManifests
-                destinationPath: $(Pipeline.Workspace)/officialBuildCI/AssetManifests
             steps:
               - checkout: self
                 persistCredentials: true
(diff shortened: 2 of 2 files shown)


tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+        // Step 5: Update files
+        await updatePackageJson(options.roslynVersion);
+        await updateChangelog(options.roslynVersion, prList, options.roslynBuildNumber, options.roslynBuildId);
+        
+        // Step 6: Create branch and PR if not dry run
+        if (!options.dryRun) {
+            const branchName = await createBranch(options.roslynEndSHA, options.roslynVersion);
+            console.log(`Branch created: ${branchName}`);
+            
+            if (options.createPullRequest) {
```
dibarbet: Is this option necessary?  When do we want to create the branch but not the pull request?

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+        const newVersion = await extractRoslynVersionFromManifest(options.assetManifestPath);
+        if (!newVersion) {
+            throw new Error('Failed to extract Roslyn version from asset manifest');
+        }
+        options.roslynVersion = newVersion;
+        console.log(`New Roslyn version: ${newVersion}`);
+        
+        // Step 2: Get current SHA from package
+        const currentSHA = await getCurrentRoslynSHA();
+        options.roslynStartSHA = currentSHA || '0000000000000000000000000000000000000000';
```
dibarbet: this should just throw if we can't find the current roslyn package sha

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+import { promisify } from 'util';
+import fetch from 'node-fetch';
+import * as xml2js from 'xml2js';
+import * as AdmZip from 'adm-zip';
+
+const execAsync = promisify(exec);
+
+interface InsertionOptions {
+    roslynVersion?: string;
+    roslynStartSHA?: string;
```
dibarbet: this doesn't seem to be passed in ever, probably can remove.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    fs.writeFileSync(settingsFile, Buffer.from(JSON.stringify(authJson)).toString('base64'));
+    
+    try {
+        const { stdout } = await execAsync(
+            `cd roslyn && roslyn-tools pr-finder -s "${startSHA}" -e "${endSHA}" --format changelog --label VSCode`,
+            { maxBuffer: 10 * 1024 * 1024 } // 10MB buffer
+        );
+        return stdout || '(no PRs with required labels)';
+    } catch (error) {
+        console.warn('PR finder failed, using empty list:', error);
```
dibarbet: use the azdo log warn syntax, e.g. https://github.com/dotnet/vscode-csharp/blob/4ca4cf4663480eb42a0e20e8e8a922711d6f8e95/tasks/createTagsTasks.ts#L190

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+        
+        fs.unlinkSync(tempFile);
+    } catch (error) {
+        console.error('Error getting current SHA:', error);
+    }
+    
+    return null;
+}
+
+async function cloneRoslynRepo(): Promise<void> {
```
dibarbet: I think it'd be better to not clone the repo as part of the task - instead pass in the path to the repo as a parameter.  And for the pipeline, clone the repo in the yaml.

That way
1.  You can use a an existing clone of roslyn when running locally.
2.  You don't have to worry about cleaning up the repo here (pipeline automatically cleans it up).

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+        } else {
+            console.log('Dry run mode - skipping branch and PR creation');
+        }
+        
+    } catch (error) {
+        console.error('Insertion failed:', error);
+        throw error;
+    } finally {
+        // Cleanup
+        if (fs.existsSync('roslyn')) {
```
dibarbet: as mentioned in another comment, repo cloning can be moved into the pipeline, and this can be deleted.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    
+    const xmlContent = fs.readFileSync(xmlFile, 'utf8');
+    const parser = new xml2js.Parser();
+    const result = await parser.parseStringPromise(xmlContent);
+    
+    // Navigate the XML structure to find Microsoft.CodeAnalysis package
+    const packages = result?.Build?.Package || [];
+    for (const pkg of packages) {
+        const attrs = pkg.$;
+        if (attrs?.Id === 'Microsoft.CodeAnalysis' || attrs?.Id === 'Microsoft.CodeAnalysis.Common') {
```
dibarbet: why do we need to check for both of these ids here?

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    if (!currentVersion) {
+        console.log('No roslyn version in package.json, this is first run');
+        return null;
+    }
+    
+    console.log(`Current Roslyn version: ${currentVersion}`);
+    
+    // Download and extract commit SHA from NuGet package
+    const packageName = 'microsoft.codeanalysis.common';
+    const feed = 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/flat2';
```
dibarbet: Instead of downloading the package yourself, the pipeline should run the 'gulp installDependencies' task.  Then the current roslyn package is downloaded by our existing infrastructure.

You can then read the nuspec file similarly to https://github.com/dotnet/vscode-csharp/blob/4ca4cf4663480eb42a0e20e8e8a922711d6f8e95/tasks/createTagsTasks.ts#L217

I'd extract out a helper you can re-use in both places to read the nuspec file.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    
+    // Setup auth for roslyn-tools
+    const homeDir = process.env.HOME || process.env.USERPROFILE;
+    const settingsDir = path.join(homeDir!, '.roslyn-tools');
+    if (!fs.existsSync(settingsDir)) {
+        fs.mkdirSync(settingsDir, { recursive: true });
+    }
+    
+    const authJson = {
+        GitHubToken: process.env.GITHUB_TOKEN || '',
```
dibarbet: should this be using `options.githubPAT` instead?

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+async function updatePackageJson(newVersion: string): Promise<void> {
+    console.log(`Updating package.json with Roslyn version ${newVersion}...`);
+    const packageJsonPath = 'package.json';
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+    
+    if (!packageJson.defaults) {
+        packageJson.defaults = {};
```
dibarbet: this should probably just throw if we couldn't get the defaults from the package.json

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    console.log('Cloning Roslyn repository...');
+    await execAsync('git clone --no-tags --filter=blob:none --depth=500 https://github.com/dotnet/roslyn.git roslyn');
+}
+
+async function generatePRList(startSHA: string, endSHA: string): Promise<string> {
+    console.log(`Generating PR list from ${startSHA} to ${endSHA}...`);
+    
+    // Setup auth for roslyn-tools
+    const homeDir = process.env.HOME || process.env.USERPROFILE;
+    const settingsDir = path.join(homeDir!, '.roslyn-tools');
```
dibarbet: @JoeRobich I didn't see anything for it obviously, but is it possible to pass the github token directly (either by an env var or parameter), instead of writing to  a settings file?

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+    fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2) + '\n');
+}
+
+async function updateChangelog(version: string, prList: string, buildNumber?: string, buildId?: string): Promise<void> {
+    console.log('Updating CHANGELOG.md...');
+    const changelogPath = 'CHANGELOG.md';
+    let changelogContent = fs.readFileSync(changelogPath, 'utf8');
+    
+    // Format PR list with proper indentation
+    const formattedPRList = prList.split('\n').map(line => `  ${line}`).join('\n');
```
dibarbet: do we always get these style new lines?  Potentially should use node's 'os' EOL here since this is from the console output

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+    
+    fs.writeFileSync(changelogPath, newLines.join('\n'));
+}
+
+async function createBranch(endSHA: string, version: string): Promise<string> {
+    console.log('Creating and pushing branch...');
+    
+    await execAsync('git config user.name "Azure Pipelines"');
```
dibarbet: Instead of doing all of this again, can we reuse the logic from https://github.com/dotnet/vscode-csharp/blob/4ca4cf4663480eb42a0e20e8e8a922711d6f8e95/tasks/localizationTasks.ts#L79 and extract out a helper that creates a branch and pull request from the local changes?

tasks/insertionTasks.ts
JoeRobich: It looks like I haven't updated pr-finder yet. Opened https://github.com/dotnet/roslyn-tools/pull/1500

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,102 @@
+trigger: none
+pr: none
+
+parameters:
+  - name: targetBranch
+    displayName: Target Branch for PR
```
AbhitejJohn, line 6: What is the difference between name and displayName here?

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,102 @@
...
+    ref: refs/tags/release
+  pipelines:
+    - pipeline: officialBuildCI
+      source: dotnet-roslyn-official
+      project: internal
+      branch: main
+      trigger: none
+
+variables:
+  - name: RoslynEndSHA
```
AbhitejJohn, line 27: How did we come up with the values for these variable? Can we confirm that these are the right choices?

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,102 @@
...
+    value: $(resources.pipeline.officialBuildCI.runName)
+  - name: RoslynBuildId
+    value: $(resources.pipeline.officialBuildCI.runID)
+
+extends:
+  template: v1/1ES.Unofficial.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    pool:
+      name: AzurePipelines-EO
+      image: AzurePipelinesUbuntu22.04compliant
```
AbhitejJohn: How did we get this image? Looks like this value is "1ESPT-Ubuntu22.04" from other places if you want an ubuntu image.

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,102 @@
...
+      os: linux
+    stages:
+      - stage: BumpRoslyn
+        displayName: Bump Roslyn Version
+        jobs:
+          - job: ProcessBump
+            displayName: Process Roslyn Bump
+            pool:
+              name: AzurePipelines-EO
+              image: AzurePipelinesUbuntu22.04compliant
```
AbhitejJohn: Similar image suggestion as above.

azure-pipelines/dotnet-vscode-csharp-insertion.yml
deepakrathore33: dispplayName is label shown in the Azure DevOps UI

azure-pipelines/dotnet-vscode-csharp-insertion.yml
deepakrathore33: RoslynEndSHA will be fetched from resources.pipeline.officialBuildCI.sourceCommit

azure-pipelines/dotnet-vscode-csharp-insertion.yml
deepakrathore33: Will update it

azure-pipelines/dotnet-vscode-csharp-insertion.yml
deepakrathore33: Will update it

deepakrathore33:
This will create pr like https://github.com/deepakrathore33/vscode-csharp/commit/3bbf97f79aac84a13ce7e72908e121aca0f60337

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,102 @@
...
+              os: linux
+            steps:
+              - checkout: self
+                persistCredentials: true
+
+              - task: UseDotNet@2
+                displayName: Install .NET SDK
+                inputs:
+                  version: 9.0.x
```
dibarbet: Use this variable template to retrieve the dotnet version - https://github.com/dotnet/vscode-csharp/blob/main/azure-pipelines/dotnet-variables.yml

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,102 @@
...
+
+              - task: NodeTool@0
```
dibarbet: replace this with a call to https://github.com/dotnet/vscode-csharp/blob/main/azure-pipelines/install-node.yml

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,196 @@
...
+    userName?: string;
+    email?: string;
+    commitSha: string;
+    targetRemoteRepo: string;
+    baseBranch: string;
+}
+
+export interface BranchAndPROptions extends GitOptions {
+    githubPAT: string;
+    dryRun?: boolean;
```
dibarbet: Can we make dryRun a required parameter?  Don't want callers to forget to specify a value for this

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,196 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import { Octokit } from '@octokit/rest';
+import { NugetPackageInfo } from './offlinePackagingTasks';
+import { PlatformInformation } from '../src/shared/platform';
+import { platformSpecificPackages } from './offlinePackagingTasks';
+
+export interface GitOptions {
+    userName?: string;
+    email?: string;
```
dibarbet: these two options appear to never be used.  I doubt they should even be options - should likely just hardcode to whatever the localization pipeline already uses

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    const stdout = git.stdout.toString();
+    if (printCommand) {
+        console.log(stdout);
+    }
+    return stdout;
+}
+
+export async function getCommitFromNugetAsync(packageInfo: NugetPackageInfo): Promise<string | null> {
```
dibarbet: I think we still need to update the localization tasks to use these helpers, not seeing any changes in the PR to the loc task

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,196 @@
...
+            body: body || title,
+            owner: 'dotnet',
+            repo: targetRemoteRepo,
+            title: title,
+            head: newBranchName,
+            base: baseBranch,
+        });
+        console.log(`Created pull request: ${pullRequest.data.html_url}.`);
+    } else {
+        console.log(`[DRY RUN] Would have created PR with title: ${title}`);
```
dibarbet: nit - consider adding the body here as well.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,358 @@
...
+            { maxBuffer: 10 * 1024 * 1024 } // 10MB buffer
+        );
+        return stdout || '(no PRs with required labels)';
+    } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        logWarning(`PR finder failed, using empty list: ${errorMessage}`);
+        if (error instanceof Error && error.stack) {
+            console.log(`##[debug]${error.stack}`);
+        }
+        return '(no PRs with required labels)';
```
dibarbet: in the error case, we should change this to something like '(failed to generate PR list, see pipeline for details)'

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,358 @@
...
+		: [];
+
+	const formattedPRList = prLines.length > 0 ? prLines.map((line) => `  ${line}`).join(NL) : '';
+
+	// Find the first top-level header "# ..."
+	const topHeaderRegex = /^# .*/m;
+	const headerMatch = topHeaderRegex.exec(text);
+	if (!headerMatch) {
+		// No top-level header; prepend a Roslyn bump
+		const prLink = buildNumber && buildId
```
dibarbet: this can all be replaced with just a throw - there should never be no header match.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,358 @@
...
+	} else {
+		leading = body.slice(0, starts[0]);
+		for (let i = 0; i < starts.length; i++) {
+			const s = starts[i];
+			const e = i + 1 < starts.length ? starts[i + 1] : body.length;
+			blocks.push(body.slice(s, e));
+		}
+	}
+
+	// Locate Roslyn and Razor blocks
```
dibarbet: I don't believe this is necessary - we can always just insert a new section at the top of the existing list.  It isn't really a problem if there are multiple updates to Roslyn/Razor listed in the changelog.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,358 @@
...
+		const firstLine = blocks[i].split(NL, 1)[0].trim();
+		if (/^\*\s*Bump\s+Roslyn\s+to/i.test(firstLine)) {
+			roslynIndex = i;
+		} else if (/^\*\s*Bump\s+Razor\s+to/i.test(firstLine)) {
+			razorIndex = i;
+		}
+	}
+
+	// Prepare new Roslyn block + optional PR list
+	const prLink = buildNumber && buildId
```
dibarbet: yeah this is a bit more complicated than I think its worth to support combining the roslyn changelog items together.  I think we can remove most of this and just insert a new item at the top of the current version changelog list.

If we want to aggregate all the roslyn insertions into one changelog item later on, I think we could change the format of the changelog to make it easier.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+
+	// Assemble new body and full text
+	const rebuiltBlocks = newBlocks.join(NL);
+	const newBody = leading + rebuiltBlocks;
+	const newTextNormalized = text.slice(0, headerLineEndIndex + 1) + newBody + text.slice(sectionEndIndex);
+
+	// Restore original newline style (normalize any newline to the original style)
+	const finalText = newTextNormalized.replace(/\r\n|\r|\n/g, originalHasCRLF ? '\r\n' : '\n');
```
dibarbet: I don't get the new line complications here.  Is it not enough to split the PR list output from roslyn-tools by os.EOL, and then insert into the changelog with os.EOL?  Why we do we have to do the normalization here?

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,100 @@
...
+              - pwsh: |
+                  npm ci
+                  npm install
+                  npm install -g gulp
+                  gulp installDependencies
+                displayName: 'Install npm dependencies and gulp'
+
+              - script: |
+                  echo "Cloning Roslyn repository..."
+                  git clone --no-tags --filter=blob:none --depth=500 https://github.com/dotnet/roslyn.git $(Pipeline.Workspace)/roslyn
```
JoeRobich: 500 *should* be enough, but I am not sure there is a reason to optimize cloning
```suggestion
                  git clone --no-tags --filter=blob:none https://github.com/dotnet/roslyn.git $(Pipeline.Workspace)/roslyn
```

JoeRobich: Thinking about this more, we should probably list the dotnet-roslyn internal mirror repo as a resource and simply checkout from there.

```
  - repository: RoslynMirror
    type: git
    name: internal/dotnet-roslyn
```

deepakrathore33: Good suggestion! It will be more efficient and reliable

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    const packages = result?.Build?.Package || [];
+    for (const pkg of packages) {
+        const attrs = pkg.$;
+        if (attrs?.Id === 'Microsoft.CodeAnalysis.Common') {
+            return attrs.Version;
+        }
+    }
+
+    return null;
```
JoeRobich, line 178: Log an Error that the Microsoft.CodeAnalysis.Common package could not be found. That would be quite a problem.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+            .filter((l) => l.trim() && !l.includes('View Complete Diff'))
+            .map((line) => line.trim())
+        : [];
+
+    // Format PR list as sub-items with proper indentation
+    const formattedPRList = prLines.length > 0
+        ? prLines.map((line) => {
+            // If the line already starts with *, keep it, otherwise add it
+            const cleanLine = line.startsWith('*') ? line.substring(1).trim() : line;
+            return `  * ${cleanLine}`;
```
dibarbet: this is weird - we're removing the `*` and adding it back?  I think it should be fine to just add the correct indentation as the tool should output the correct markdown list already

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    const headerEndLineIndex = headerMatch.index + headerMatch[0].length;
+
+    // Prepare new Roslyn block
+    const prLink = buildNumber && buildId
+        ? `[#${buildNumber}](https://dev.azure.com/dnceng/internal/_build/results?buildId=${buildId})`
+        : '[#TBD](TBD)';
+
+    let newRoslynBlock = `* Bump Roslyn to ${version} (PR: ${prLink})`;
```
dibarbet: This doesn't seem quite right.  I do like the link to the actual build, but that should probably be part of the PR description and doesn't need to be in the changelog itself.

The `(PR: <pr>)` part of the changelog typically includes a link to the PR in vscode-csharp that is responsible for the version bump - in this case that would be the PR we are creating.

In order to get that link however we have to create the PR first.  What we might have to do is create the PR in one step, then push to the branch again to update the PR # in the changelog (after we know what the PR number is).  We could potentially do the entire changelog update in a second commit after the PR is created.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+
+async function updateChangelog(version: string, prList: string, buildNumber?: string, buildId?: string): Promise<void> {
+    console.log('Updating CHANGELOG.md...');
+    const changelogPath = 'CHANGELOG.md';
+    const text = fs.readFileSync(changelogPath, 'utf8');
+    const NL = os.EOL;
+
+    // Prepare PR list (filter out 'View Complete Diff' lines)
+    const prLines = prList
+        ? prList
```
dibarbet: as far as I can tell `prList` is never empty - probably need to remove these conditionals

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    // Prepare new Roslyn block
+    const prLink = buildNumber && buildId
+        ? `[#${buildNumber}](https://dev.azure.com/dnceng/internal/_build/results?buildId=${buildId})`
+        : '[#TBD](TBD)';
+
+    let newRoslynBlock = `* Bump Roslyn to ${version} (PR: ${prLink})`;
+
+    // Add PR list as sub-items if available and valid
+    const shouldSkipPRList = !prList || prList === '(failed to generate PR list, see pipeline for details)';
```
dibarbet: I think we need to cleanup the handling of the `prList` a bit.  We're setting it to this complex failure string when it fails, but we don't even output that string anywhere. 

Couldn't we just make `prList` be nullable, and have the function that generates it return null when it fails?  And here we check if its null before adding the list to the changelog?

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    if (!shouldSkipPRList && formattedPRList) {
```
dibarbet: Not sure why we need both `shouldSkipPRList` and `formattedPRList`.  Related to the above suggestion, just make pr list nullable then put all the logic to convert the PR list into the changelog section into a condition on the list, e.g.

```
<create header block>
if (prList) {
    <convert to formatted pr list>
    <append formatted list to new changelog text
}
```

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    const stdout = git.stdout.toString();
+    if (printCommand) {
+        console.log(stdout);
+    }
+    return stdout;
+}
+
+export async function getCommitFromNugetAsync(packageInfo: NugetPackageInfo): Promise<string | null> {
```
dibarbet: can the createTags task also be updated to use this helper?
https://github.com/dotnet/vscode-csharp/blob/main/tasks/createTagsTasks.ts#L198

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,235 @@
...
+    packageJson.defaults.roslyn = newVersion;
+
+    fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2) + '\n');
+}
+
+async function updateChangelog(version: string, prList: string, buildNumber?: string, buildId?: string): Promise<void> {
+    console.log('Updating CHANGELOG.md...');
+    const changelogPath = 'CHANGELOG.md';
+    const text = fs.readFileSync(changelogPath, 'utf8');
+    const NL = os.EOL;
```
dibarbet, line 220: new line handling is much nicer, thanks for fixing that.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,260 @@
...
+            await git(['commit', '-m', `Update changelog with PR #${prNumber}`]);
+            await git(['push', 'target', `insertion/${options.roslynEndSHA}`]);
+
+            console.log(`Changelog updated with PR #${prNumber} link.`);
+        }
+
+    } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        logError(`Insertion failed: ${errorMessage}`);
+        if (error instanceof Error && error.stack) {
```
AbhitejJohn: Consider: Can printing the error stack trace be part of the logError function above?

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,260 @@
...
+            `cd "${roslynRepoPath}" && roslyn-tools pr-finder -s "${startSHA}" -e "${endSHA}" --format changelog --label VSCode`,
+            { maxBuffer: 10 * 1024 * 1024 } // 10MB buffer
+        );
+        return stdout && stdout.trim() ? stdout : null;
+    } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        logWarning(`PR finder failed: ${errorMessage}`);
+        if (error instanceof Error && error.stack) {
+            console.log(`##[debug]${error.stack}`);
+        }
```
AbhitejJohn: Similar comment to the above, consider placing some of these error message and error stack trace determinations in the `logError`/`logwarning` functions

AbhitejJohn (approved):
LGTM, heads-up: there's another PR [here](https://github.com/dotnet/vscode-csharp/pull/8584) that we'd want to de-dupe some of the git helpers with. Happy to chat offline.

tasks/insertionTasks.ts
```diff
@@ -0,0 +1,260 @@
...
+            newRoslynBlock += NL + formattedPRList;
+        }
+    }
+
+    // Insert the new block right after the header
+    const newText =
+        text.slice(0, headerEndLineIndex) +
+        NL + NL + // Add two newlines after the header
+        newRoslynBlock +
+        NL + // Add a newline after the block
```
JoeRobich: Please remove these newlines. We do not use emptylines.

https://github.com/dotnet/vscode-csharp/blob/99db45bc31e3f9e6897e5e1bb5587e62ba444c34/CHANGELOG.md?plain=1#L12-L20

deepakrathore33: Remvoed the new lines

tasks/insertionTasks.ts
deepakrathore33: Updated the PR

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,101 @@
...
+              - checkout: RoslynMirror
+                displayName: 'Checkout Roslyn repository'
+                path: roslyn
+
+              - script: |
+                  echo "Installing roslyn-tools..."
+                  dotnet tool install -g Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json
+                  echo "Verifying roslyn-tools installation..."
+                  dotnet tool list -g
+                  roslyn-tools --help
```
JoeRobich: If this is to test that the roslyn-tools were installed. Maybe switch to `--version` since that might be helpful during troubleshooting.

```suggestion
                  roslyn-tools --version
```

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,193 @@
...
+    newBranchName: string;
+    userName?: string;
+    email?: string;
+}
+
+export function logError(message: string): void {
+    console.log(`##vso[task.logissue type=error]${message}`);
+}
+
+export async function git(args: string[], printCommand = true): Promise<string> {
```
AbhitejJohn: Some of this is now similar to https://github.com/dotnet/vscode-csharp/blob/main/tasks/gitTasks.ts. Can we please move unique functions from here to there please? `getCommitFromNugetAsync` seems like the main thing that is missing. Although, I don't think that is a function of git in any way, so maybe that needs another home.

tasks/gitHelpers.ts
```diff
@@ -0,0 +1,193 @@
...
+    } catch (error) {
+        logError(`Error getting commit from NuGet package: ${error}`);
+        if (error instanceof Error && error.stack) {
+            console.log(`##[debug]${error.stack}`);
+        }
+        throw error;
+    }
+}
+
+export async function createBranchAndPR(
```
AbhitejJohn: The newer API here has different methods for pushing a branch, checking if it exists and creating PR, so when you adopt that you'd want to change the callers as well.

AbhitejJohn (approved):

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,101 @@
...
+                displayName: 'Install roslyn-tools CLI'
+
+              - task: Npm@1
+                displayName: Run Roslyn insertion
+                env:
+                  GITHUB_TOKEN: $(System.AccessToken)
+                  GitHubPAT: $(System.AccessToken)
+                inputs:
+                  command: custom
+                  customCommand: 'run gulp -- insertion:roslyn --assetManifestPath=$(Pipeline.Workspace)/AssetManifests --roslynRepoPath=$(Pipeline.Workspace)/roslyn --roslynEndSHA=$(RoslynEndSHA) --roslynBuildNumber=$(RoslynBuildNumber) --roslynBuildId=$(RoslynBuildId) --targetBranch=${{ parameters.targetBranch }} --githubPAT=$(System.AccessToken) --dryRun=false'
```
AbhitejJohn: How does this PAT flow in here? Do you set it on the pipeline or is this something that is coming in from a variable group?

`--githubPAT=$(System.AccessToken)`

JoeRobich: Since you are running in dnceng you should be able to use the same PAT as the localization pipeline - https://github.com/dotnet/vscode-csharp/blob/29cb53a66cbccfd79655e2825b4c2b82c668ef19/azure-pipelines/loc.yml#L95

azure-pipelines/dotnet-vscode-csharp-insertion.yml
```diff
@@ -0,0 +1,100 @@
...
+                  dotnet tool install -g Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json
+                  echo "Verifying roslyn-tools installation..."
+                  dotnet tool list -g
+                  roslyn-tools --version
+                displayName: 'Install roslyn-tools CLI'
+
+              - task: Npm@1
+                displayName: Run Roslyn insertion
+                env:
+                  GitHubPAT: $(BotAccount-dotnet-bot-repo-PAT)
```
JoeRobich: I see that we are passing this as a command line argument. Do we also need it as a variable?

deepakrathore33: I think we don't need it. Not sure if I have used env variable somewhere, let me check it and remove it

PR #8565, opened 2025-08-26
Register the CSharp context provider with Copilot Chat as well.

dbaeumer:
We are in the process of moving Inline completions from Copilot client to Copilot chat. For a certain time we therefore need to register the context provider with both Copilot client and Copilot chat. Which path is exercised will be determined by VS Code. So for the same [document,position] pair only one resolve request will occur.

Registering with Copilot Chat has also the advantage that context is resolved during NES. If you want to enable this you can use the setting "github.copilot.chat.advanced.inlineEdits.xtabProvider.languageContext.enabled"

I would also like to ask you to change your active experiment [C# completion context provider in VSCode with code snippets - Azure ExP Studio](https://exp.microsoft.com/feature/b06765d0-0ad4-4769-a86f-0d13f1fabee8?workspaceId=dbd12384-5001-4e90-a863-526321eaf233&group=/vscodeexpws/Copilot) to target `copilot-chat`

### DIFF
--- a/src/lsptoolshost/copilot/contextProviders.ts
+++ b/src/lsptoolshost/copilot/contextProviders.ts
@@ -2,7 +2,12 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


genlu:
FYI @akhera99 @AbhitejJohn

dbaeumer:
@dotnet-policy-service agree company="Microsoft"

AbhitejJohn (approved):

PR #8569, opened 2025-08-27
Move wiki documentation into the repo

JoeRobich:
Removed some docs for no longer being relevant. Updated links as appropriate. Hopefully we could remove the Wiki section following this change. 

CC: @gregg-miskelly for sign-off on debugger docs


dibarbet (approved):

gregg-miskelly:
Is the motivation for doing this that we want to make it easier for the communicate to contribute to documentation?

We can certainly do this, but we will need to work in stages as we have a bunch of aka.ms links that point at the Wiki. So, after this PR merges, we will need to edit the aka.ms links, and then you can delete the Wiki.

docs/C#-Extension-support-for-new-Linux-Distros-with-.NET-Core.md
```diff
@@ -0,0 +1,15 @@
+Linux distribution support is a little tricky for the C# extension's debugger as it needs a version of .NET Core to run on top of.
```
gregg-miskelly: I think we could probably get rid of this doc at this point -- we used to have lots of problems with people trying various distros, but ever since the .NET team moved two just two Linux builds (glibc and musl) we haven't had any problems.

gregg-miskelly:
Looks like there is one place in the code where we reference the wiki directly (not through an aka.ms link) -- https://github.com/search?q=repo%3Adotnet%2Fvscode-csharp+remoteDebugInfoUrl&type=code

I think that code path is from back when this extension didn't support ARM64. I can submit a PR to delete that code.

gregg-miskelly:
There are a few places in 'debugger.md' that have links to the wiki that need to be updated. Search for `https://github.com/dotnet/vscode-csharp/wiki`.

gregg-miskelly:
> Looks like there is one place in the code where we reference the wiki directly (not through an aka.ms link) -- https://github.com/search?q=repo%3Adotnet%2Fvscode-csharp+remoteDebugInfoUrl&type=code
>
> I think that code path is from back when this extension didn't support ARM64. I can submit a PR to delete that code.

Fixed with: https://github.com/dotnet/vscode-csharp/pull/8574

gregg-miskelly (approved):

JoeRobich:
> Is the motivation for doing this that we want to make it easier for the communicate to contribute to documentation?

Yes, also it has been my experience that it is generally easier for us to keep it up-to-date and accurate, since we can ask for updates in PRs that change documented behavior. I certainly don't plan to remove them until I get your OK to do so.

gregg-miskelly (approved):

PR #8584, opened 2025-09-04
Add automated PR creation for Roslyn Copilot updates using a gulp task

AbhitejJohn:
This should now automatically create a PR when we decide to upload a new version of the bits.

PR created: https://github.com/dotnet/vscode-csharp/pull/8583


package-lock.json
```diff
@@ -5146,15 +5146,6 @@
       "resolved": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/before-after-hook/-/before-after-hook-2.2.3.tgz",
       "integrity": "sha512-NzUnlZexiaH/46WDhANlyR2bXRopNg4F/zuSA3OpZnllCUgRaOF2znDioDWrmbNVsuZk6l9pMquQB38cfBZwkQ=="
     },
-    "node_modules/big.js": {
```
JoeRobich, line 5149: weird, was the package-lock.json out of sync? also weird that there is only deletions.

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+async function git(args: string[], printCommand = true): Promise<string> {
+    if (printCommand) {
+        console.log(`git ${args.join(' ')}`);
+    }
+
+    const git = spawnSync('git', args);
+    if (git.status != 0) {
+        const err = git.stderr ? git.stderr.toString() : '';
+        if (printCommand) {
+            console.error(`Failed to execute git ${args.join(' ')}.`);
```
JoeRobich: Instead of using `console.error`, grab [the logging functions](https://github.com/dotnet/vscode-csharp/pull/8563/files#diff-1552fb020f129f4eb76756179d467a47dc6d2e181d0ce33b02323bf4edae6bc0R28-R35) from the other auto-insert PR. Maybe create a PipelineLogger class we can all share.

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+            owner: 'dotnet',
+            repo: 'vscode-csharp',
+        });
+
+        if (listPullRequest.status != 200) {
+            throw `Failed get response from GitHub, http status code: ${listPullRequest.status}`;
+        }
+
+        const title = `Update RoslynCopilot url to ${version}`;
+        if (listPullRequest.data.some((pr) => pr.title === title)) {
```
JoeRobich: I think we should check this much earlier and avoid doing any work or pushing any branches.

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+
+    if (!updated) {
+        console.log('No changes required to package.json; aborting PR creation.');
+        return;
+    }
+
+    // Persist package.json
+    fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n', { encoding: 'utf8' });
```
JoeRobich: Does this preserve ordering of properties and objects?

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    const remoteRepoAlias = 'targetRepo';
+    await git(
+        ['remote', 'add', remoteRepoAlias, `https://${parsedArgs.userName}:${pat}@github.com/dotnet/vscode-csharp.git`],
+        // Note: don't print PAT to console
+        false
+    );
+    await git(['fetch', remoteRepoAlias]);
+
+    const lsRemote = await git(['ls-remote', remoteRepoAlias, 'refs/head/' + branch]);
```
JoeRobich: Extract to function or add comment that this is checking if a particular branch exists on the remote.

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+    const zipPath = path.join(stagingDir, zipFile);
+    console.log(`Using zip file: ${zipPath}`);
+    const zipName = zipFile;
+
+    // Extract version from file name
+    // Example: "Microsoft.VisualStudio.Copilot.Roslyn.LanguageServer-18.0.671-alpha.zip"
+    let version: string | null = null;
+    const m = zipName.match(/Microsoft\.VisualStudio\.Copilot\.Roslyn\.LanguageServer-(.+)\.zip$/i);
+    if (m && m[1]) {
+        version = m[1];
```
JoeRobich: I think it would help readability to extract this to a `getCopilotRoslynVersion` function.

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    if (!version) {
+        console.log(`Could not extract version from file name ${zipName}; skipping.`);
+        return;
+    }
+
+    console.log(`Extracted version: ${version}`);
```
JoeRobich, line 64: I would check here whether the PR exists and exit if so.

Then I would check if the branch exists on the remote and create it with the insertion if not.

Then finish by creating the PR for the existing or newly pushed branch.

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+        console.log(stdout);
+    }
+    return Promise.resolve(stdout);
+}
+
+gulp.task('publish roslyn copilot', async () => {
+    const parsedArgs = minimist<Options>(process.argv.slice(2));
+
+    // Get staging directory from environment variable passed from pipeline
+    const stagingDir = process.env['STAGING_DIRECTORY'];
```
dibarbet: any reason to use an env var here instead of an option passed in?

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+    if (!stagingDir) {
+        console.log('STAGING_DIRECTORY environment variable not set; skipping package.json update.');
```
dibarbet: is this ever expected to happen in normal use?  if not consider throwing instead so that we get actual pipeline failures

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+        return;
+    }
+
+    if (!fs.existsSync(stagingDir)) {
+        console.log(`Staging directory not found at ${stagingDir}; skipping package.json update.`);
```
dibarbet: same as above

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+        return;
+    }
+
+    // Find the Roslyn zip file in the staging directory (we know it was copied here)
+    const files = fs.readdirSync(stagingDir);
+    const zipFile = files.find((file) => /Roslyn\.LanguageServer.*\.zip$/i.test(file));
+
+    if (!zipFile) {
+        console.log(`No Roslyn LanguageServer zip file found in ${stagingDir}; skipping package.json update.`);
```
dibarbet: same as above

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+        console.log(`Could not extract version from file name ${zipName}; skipping.`);
+        return;
+    }
+
+    console.log(`Extracted version: ${version}`);
+
+    const pkg = getPackageJSON();
+    let updated = false;
+
+    if (pkg.runtimeDependencies && Array.isArray(pkg.runtimeDependencies)) {
```
dibarbet: not completely opposed to this since its relatively simple.  But IIRC we already had a task that could update package dependencies like O#/debugger.  Any chance we could just re-use that?

See https://github.com/dotnet/vscode-csharp/blob/99db45bc31e3f9e6897e5e1bb5587e62ba444c34/tasks/debuggerTasks.ts#L22

tasks/componentUpdateTasks.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    // Persist package.json
+    fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n', { encoding: 'utf8' });
+
+    // Prepare git
+    const safeVersion = version.replace(/[^A-Za-z0-9_.-]/g, '-');
+    const branch = `update/roslyn-copilot-${safeVersion}`;
+
+    // Make this optional so it can be tested locally by using dev's information. In real CI user name and email are always supplied.
+    if (parsedArgs.userName) {
```
dibarbet: definitely feels like we need to be sharing code with the other roslyn update.  Don't want to have multiple different implementations of commiting, pushing branch, creating pr, etc.

package-lock.json
AbhitejJohn: Possibly, this just showed up when I ran through the build steps.

tasks/componentUpdateTasks.ts
AbhitejJohn: Good thought, I think I'm going to save that as a follow-up. Did you have other places for these logs to go or would these primarily be output to console? If there were other places we'd want to see these logs, I find a little more promise to adding shared logging functions.

JoeRobich: It would just make using the AzDO logging commands easier. Since we would ideally get syntax coloring for warnings and errors to make logs easier to read.

tasks/componentUpdateTasks.ts
AbhitejJohn: Good callout, I missed updating the integrity check and the other API does do that, so I'll just reuse.

tasks/componentUpdateTasks.ts
AbhitejJohn: Switched to using an existing function for this

tasks/componentUpdateTasks.ts
AbhitejJohn: Agreed and changed, thank you.

.vscode/settings.json
```diff
@@ -12,7 +12,8 @@
         "vsix/": true
     },
     "[typescript]": {
-        "editor.defaultFormatter": "esbenp.prettier-vscode"
+        "editor.defaultFormatter": "esbenp.prettier-vscode",
+        "editor.formatOnSave": true
```
AbhitejJohn, line 16: I was just tired of fixing prettier errors, so I enabled `formatOnSave` just for typescript. Please scream if that isn't something we want to do.

tasks/componentUpdateTasks.ts
AbhitejJohn: Will work through this in a follow-up with Deepak.

Issue #8514, opened 2025-08-08
Default `enableFileBasedPrograms` to true in release channel


RikkiGibson:
Once we are satisfied with the level of stability of the file-based programs experience, we should default the setting to `true` in the release channel. We want to do this prior to .NET 10 RTM.

Currently it is `false` in release and `true` in prerelease.

@dibarbet for visibility

RikkiGibson:
~~We missed the august release train, let's get this in the next one.~~

It looks like this will get into the august release (coming this week.)

PR #8587, opened 2025-09-05
Default file-based programs support to true in release

RikkiGibson:
#8514

~~Want to get this in for September release cycle.~~

We are good to ship this in August release cycle, i.e. release it on Sept 10th.


jasonmalinowski (approved):

dibarbet (changes requested):
~~if you make this change now it will go out for the august release.  I assume we don't want that? ~~

sounds like we want this to go out

dibarbet (commented):
Can you update the changelog to mention this as well?

RikkiGibson:
@dibarbet for sign off

RikkiGibson:
From my last check, the failing tests are not caused by this PR. i.e. rolling builds and other PR builds had similar failed jobs.

PR #8588, opened 2025-09-06
Do not allow client to modify whitespace for auto insert edits

dibarbet:
Tell the client not to modify the whitespace we provided in auto insert snippets.
Requires https://github.com/dotnet/roslyn/pull/80075

### DIFF
--- a/src/lsptoolshost/autoInsert/onAutoInsert.ts
+++ b/src/lsptoolshost/autoInsert/onAutoInsert.ts
@@ -5,7 +5,12 @@
 
 import * as vscode from 'vscode';
 
-import { FormattingOptions, LanguageClient, TextDocumentIdentifier } from 'vscode-languageclient/node';
+import {
+    FormattingOptions,
+    InsertTextFormat,
+    LanguageClient,
+    TextDocumentIdentifier,
+} from 'vscode-languageclient/node';
 import * as RoslynProtocol from '../server/roslynProtocol';
 import { RoslynLanguageServer } from '../server/roslynLanguageServer';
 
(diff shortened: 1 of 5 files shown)


src/lsptoolshost/autoInsert/onAutoInsert.ts
```diff
@@ -50,7 +55,13 @@ export function registerOnAutoInsert(languageServer: RoslynLanguageServer, langu
         }
 
         // Regular expression to match all whitespace characters except the newline character
-        const changeTrimmed = change.text.replace(/[^\S\n]+/g, '');
+        let changeTrimmed = change.text.replace(/[^\S\n]+/g, '');
+        // When hitting enter between braces, we can end up with two new lines added (one to move the cursor down to an empty line,
+        // and another to move the close brace to a new line below that).  We still want to trigger on auto insert here, so remove
+        // duplicates before we check if the change matches a trigger character.
+        if (changeTrimmed.length > 1) {
+            changeTrimmed = [...new Set(changeTrimmed)].join('');
```
JoeRobich: `changeTrimmed` is a string. What is this intending to do?

src/lsptoolshost/autoInsert/onAutoInsert.ts
```diff
@@ -98,12 +109,22 @@ async function applyAutoInsertEdit(
         const textEdit = response._vs_textEdit;
         const startPosition = new vscode.Position(textEdit.range.start.line, textEdit.range.start.character);
         const endPosition = new vscode.Position(textEdit.range.end.line, textEdit.range.end.character);
-        const docComment = new vscode.SnippetString(textEdit.newText);
-        const code: any = vscode;
-        const textEdits = [new code.SnippetTextEdit(new vscode.Range(startPosition, endPosition), docComment)];
+
+        //const editor = vscode.window.activeTextEditor?.insertSnippet();
+
```
JoeRobich:
```suggestion
```

JoeRobich (commented):
Can we add a test to the onAutoInsert integration tests?

src/lsptoolshost/autoInsert/onAutoInsert.ts
dibarbet: The idea is to remove duplicate newline characters from the string.  Potentially this should instead just check if it the string is only new line characters, and remove duplicates then.

JoeRobich: We can set `changeTrimmed` directly to `'\n'` since we know it is all whitespace and has a length > 1.

JoeRobich (approved):
just be sure to remove the commented out code.

PR #8604, opened 2025-09-10
Collect full dumps on crash

dibarbet:
See https://learn.microsoft.com/en-us/dotnet/core/diagnostics/collect-dumps-crash#types-of-mini-dumps


src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionResolver.ts
```diff
@@ -99,7 +99,7 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
             // Enable dump collection
             env.DOTNET_DbgEnableMiniDump = '1';
             // Collect heap dump
```
jasonmalinowski, line 101: This comment might need updating. Should this also have a comment why we're including module images too?

PR #8624, opened 2025-09-12
Do not register the project context status item for Razor files

JoeRobich:
Reported in https://github.com/microsoft/vscode-dotnettools/issues/1900

The original fix went into the **prerelease** branch in https://github.com/dotnet/vscode-csharp/pull/7680. This must have been around the time we removed our automatic merge backs. The same fix was never made in main and it eventually merged forward again.


JoeRobich:
Opened https://github.com/dotnet/vscode-csharp/pull/8625 as an alternative that adds Razor support when cohosting.

JoeRobich:
Closing this out to take the alternative approach #8625

PR #8631, opened 2025-09-15
Update excludes

JoeRobich:


azure-pipelines.yml
```diff
@@ -16,12 +16,24 @@ pr:
...
-    - CHANGELOG.
+    - azure-pipelines/profiling.yml
+    - azure-pipelines/publish-roslyn-copilot.yml
+    - docs/*
+    - images/*
+    - RuntimeLicenses/*
     - 'l10n/bundle.l10n.*.json'
+    - themes/*
+    - CODEOWNERS
+    - 'version.json'
```
jasonmalinowski, line 44: Why not the version.json?

jasonmalinowski: (or generally: it might be good to have comments explaining some of these...)

dibarbet (approved):

PR #8655, opened 2025-09-25
Add copilot setup steps

dibarbet:


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,45 @@
...
+      # Install Node.js (required for TypeScript compilation and npm packages)
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      # Install .NET SDK (required for some build components and MSBuild tasks)
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8.0.x'
```
333fred, line 37: 8? Should we not do later?

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      # Install npm dependencies
+      - name: Install dependencies
+        run: npm ci
+
+      # Install gulp globally (required for build tasks)
+      - name: Install gulp
+        run: npm install -g gulp
```
333fred, line 45: What about vsce?

333fred: And `gulp installDependencies`?

.github/workflows/copilot-setup-steps.yml
333fred, line 1: Should we give it the ability to grab the latest roslyn and its dependencies as well? Would it need that to be able to integrate new roslyn APIs?

.github/workflows/copilot-setup-steps.yml
dibarbet: we're on 8 for general restore/signing/integration tests at the moment - https://github.com/dotnet/vscode-csharp/blob/main/azure-pipelines/dotnet-variables.yml

Probably works with 10, but would rather match rest of CI.

.github/workflows/copilot-setup-steps.yml
dibarbet: Good call on installDependencies, should be doing that if it wants to run integration tests (though I don't expect them to run well on these machines without additional changes).  vsce I suppose could potentially be useful if it wants to package something, but that is a pretty rare workflow.  Can add it anyway.

.github/workflows/copilot-setup-steps.yml
dibarbet: Grabbing prebuilt Roslyn bits is difficult, since we'd have to cross over to dnceng/internal to figure out which builds we should use, and would need permissions to pull them from the right upstream.

Potentially it could instead just checkout Roslyn and build it here, but going to punt on that for now - want to get the most basic scenario working first.

PR #8671, opened 2025-10-01
Fix update changelog script

JoeRobich:


tasks/snapTasks.ts
```diff
@@ -141,10 +141,10 @@ gulp.task('updateChangelog', async (): Promise<void> => {
     fs.writeFileSync(changelogPath, changelogLines.join(os.EOL));
 });
 
-const prRegex = /^\*.+\(PR: \[#(\d+)\]\(/g;
+const prRegex = /^\*.+\(PR: \[#(\d+)\]\(/;
```
JoeRobich, line 144: Using the `g` (global) flag was not only unnecessary since we are looking one line at a time but it also meant the regex was stateful and needed to be reset as it is intended to be rerun on the same string finding additional matches.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 # 2.94.x
+* Update RoslynCopilot url to 18.0.797-alpha (PR: [#8652](https://github.com/OmniSharp/omnisharp-vscode/pull/8652))
```
JoeRobich, line 7: The corrected regexes found these additional PRs.

PR #8720, opened 2025-10-24
Add group ids to configuration groups

dibarbet:
Related to https://github.com/microsoft/vscode/issues/273163#issuecomment-3444981423

### DIFF
--- a/package.json
+++ b/package.json
@@ -718,6 +718,7 @@
     "configuration": [
       {
         "title": "Project",
+        "id": "ms-dotnettools.csharp.project",
         "order": 0,
         "properties": {
           "dotnet.defaultSolution": {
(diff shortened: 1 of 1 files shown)


timheuer:
Sounds like we may also want to get ahead of this for csdevkit as well? I don't think we have grouping there (but I know there is a separate issue tracking putting grouping in to make more sense)

dibarbet:
> Sounds like we may also want to get ahead of this for csdevkit as well? I don't think we have grouping there (but I know there is a separate issue tracking putting grouping in to make more sense)

Could, but my read is it isn't necessary.  Dev Kit only has a single group, and so the one group getting the automatic `ms-dotnettools.csdevkit` id would not be problematic.  

Only would need it if they end up making the `id` a required field even if there is only a single group (but I suspect that would break a huge number of extensions out there and seems unlikely).

akhera99 (approved):

PR #8724, opened 2025-10-28
Add more completion tests for Razor

davidwengier:
Test for https://github.com/dotnet/razor/pull/12403 / https://github.com/dotnet/razor/issues/4351


davidwengier:
ugh, I can't get pass/fail to be reported locally, and the API for resolving completion items in VS Code is so bad that I don't think we can reasonably use this as a regression test for that issue anyway. We have unit tests in Razor that work, I'm giving up.

PR #8748, opened 2025-11-05
Wait for Html buffer updates before making requests

davidwengier:
Fixes https://github.com/dotnet/razor/issues/12409
Fixes https://github.com/dotnet/razor/issues/11786
Fixes https://github.com/dotnet/razor/issues/12410

We were doing all of the right things about updating Html, checking checksums, updating buffers before making requests etc., but we were still too quick and would sometimes make requests before the Html server had received the update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 - Debug from .csproj and .sln [#5876](https://github.com/dotnet/vscode-csharp/issues/5876)
 
 # 2.101.x
+* Update Roslyn to 5.3.0-2.25557.4 (PR: [#8759](https://github.com/dotnet/vscode-csharp/pull/8759))
+  * Implement canonical miscellaneous files project loader for non-file-based programs (PR: [#80748](https://github.com/dotnet/roslyn/pull/80748))
+  * Better handle if a BuildHost process crashes that prevents connection (PR: [#81041](https://github.com/dotnet/roslyn/pull/81041))
 * Wait for Html buffer updates before making requests (PR: [#8748](https://github.com/dotnet/vscode-csharp/pull/8748))
 
 # 2.97.x
--- a/docs/debugger/Troubleshoot-loading-the-.NET-Debug-Services.md
+++ b/docs/debugger/Troubleshoot-loading-the-.NET-Debug-Services.md
@@ -46,7 +46,7 @@ You can test for this condition by adding the following code to the start of you
             else if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX))
             {
                 nativeLibraryPrefix = "lib";
-                nativeLibraryExtension = ".so";
+                nativeLibraryExtension = ".dylib";
             }
             else
             {
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "workspace"
   ],
   "defaults": {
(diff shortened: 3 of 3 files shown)


src/lsptoolshost/razor/razorEndpoints.ts
```diff
@@ -252,6 +252,13 @@ export function registerRazorEndpoints(
                 return undefined;
             }
 
+            // We know that we've got the right document, and we've been told about the right content by the server,
+            // but all we can be sure of at this point is that we've fired the change event for it. The event firing
+            // is async, and the didChange notification that it would generate is a notification, so doesn't necessarily
+            // block. Before we actually make a call to the Html server, we should at least make sure that the document
+            // version is not still the same as before we updated the content.
+            await document.waitForBufferUpdate();
```
dibarbet: I don't know if this applies here.  But it is possible ensure something runs after the didChange (or other lsp notification) fires.  We do that for onAutoInsert because its important that the server receives the didChange before the onautoinsert request.

See https://github.com/dotnet/vscode-csharp/blob/main/src/lsptoolshost/autoInsert/onAutoInsert.ts#L23

Even if you don't block, if the html server is processing LSP requests correctly as long as it receives the didChange first it should process it before everything else in the queue.

davidwengier: Oohhh interesting! Will see if I can hook that up.

davidwengier: Oh wait, that would need us to get a reference to the Html language client. I think I've tried that before, and asked VS Code folks, and it's not possible. Will double check tomorrow though.

davidwengier: With some AI help, I got this working with the general workspace didChange event. Thanks for the idea!

src/lsptoolshost/razor/htmlDocumentManager.ts
```diff
@@ -111,10 +150,35 @@ export class HtmlDocumentManager {
...
+                try {
+                    // Wait for the update promise with a 5 second timeout
+                    await Promise.race([
+                        pendingUpdate.promise,
+                        new Promise((_, reject) =>
+                            setTimeout(() => reject(new Error('Document update timeout')), 5000)
+                        ),
+                    ]);
+                } catch (error) {
+                    this.logger.logWarning(`Failed to wait for document update: ${error}`);
```
dibarbet, line 174: what is the impact if this fails here?  request fails?  state isn't sync'd correctly?

dibarbet (approved):
seems mostly reasonable.  really wish we could just run it after the actual didChange from the html server, but I guess its just not possible

someone from razor should probably look too

src/lsptoolshost/razor/htmlDocumentManager.ts
davidwengier: It means that we potentially ask the Html server for completion (or anything else) on an old document, so results might be not what the user wants. eg, you type `<`, and if we don't wait for that to get to the Html document, you won't see Html elements in completion. If you wait a beat and hit Ctrl+Space then you do, because the system has caught up.

There should be no corruption of any state possible, because we're not actually trying to influence any state, just wait for it.

davidwengier:
I asked Razor folks for a review, but I guess the TypeScript scared them off. Merging this to avoid a conflict with the Razor insertion I'm about to open :P

PR #8758, opened 2025-11-07
Disable file based programs in release

dibarbet:
There are a few lingering issues with FBP causing problems with C#DK.  Has been requested that we turn this off in release until we can fix issues

1. Noisy restore notifications for files in an actual project
2. https://github.com/microsoft/vscode-dotnettools/issues/2443
(pending rest of the list)


JoeRobich (approved):

dibarbet:
(razor failures unrelated)

PR #8780, opened 2025-11-14
Handle restore on the server

dibarbet:


src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -32,33 +27,10 @@ export function registerRestoreCommands(
...
-                return restore(languageServer, csharpOutputChannel, [], true);
+                return restore(languageServer, csharpOutputChannel, []);
             })
         );
     }
-
-    languageServer.registerOnRequest(ProjectNeedsRestoreRequest.type, async (params) => {
-        let projectFilePaths = params.projectFilePaths;
-        if (getCSharpDevKit()) {
-            // Only restore '.cs' files (file-based apps) if CDK is loaded.
```
RikkiGibson, line 43: really nice to delete this also.

src/lsptoolshost/projectRestore/restore.ts
```diff
@@ -93,63 +65,34 @@ async function chooseProjectAndRestore(
...
 }
 
 export async function restore(
     languageServer: RoslynLanguageServer,
     outputChannel: vscode.LogOutputChannel,
-    projectFiles: string[],
-    showOutput: boolean
+    projectFiles: string[]
 ): Promise<void> {
     if (_restoreInProgress) {
```
RikkiGibson, line 76: Now this tracks only restores initiated by the client. Parallel restores on different things are still permitted on the server side. Perhaps that is what we want? At least there doesn't seem to be a need to make a change to that in this PR.

dibarbet: right.  don't want really multiple manually initiated restores running at once.  But server can run automatic restores for diff projects

dibarbet:
replaced by https://github.com/dotnet/vscode-csharp/pull/8799

PR #8789, opened 2025-11-15
Update from Ubuntu 20.04 to 22.04

Youssef1313:


JoeRobich:
@Youssef1313 Does this image exist yet? Also, thanks for being proactive in updating this!

Youssef1313:
@JoeRobich Apparently not. I thought it does when I looked at https://github.com/dotnet/templating/blob/8bd73d3f6659a2e28e79fadf83eecd2d3b066110/azure-pipelines.yml#L164 though.

JoeRobich:
> @JoeRobich Apparently not. I thought it does when I looked at https://github.com/dotnet/templating/blob/8bd73d3f6659a2e28e79fadf83eecd2d3b066110/azure-pipelines.yml#L164 though.

Wow that change is a year old. Something must be up. Are we using the same Pool I wonder.

JoeRobich:
@Youssef1313 Do you know the difference between `demands: ImageOverride -equals 1es-ubuntu-2004-open` and just `image: 1es-ubuntu-2204-open` ?

Youssef1313:
@JoeRobich Nope, unfortunately.

JoeRobich:
> @JoeRobich Nope, unfortunately.

**Key Differences:**
1. Syntax: `image` is cleaner and more direct
2. Purpose: `image` is specifically for VM images, while `demands` is a general capability matching system
3. Maintenance: `image` is easier to maintain and understand
4. Compatibility: `demands` with `ImageOverride` is mainly used for backwards compatibility or specific scenarios where you need the demands system

**Recommendation:**
Use `image` for new pipelines as it's the modern, cleaner approach. The `demands` approach is typically only needed for legacy compatibility or when you have specific agent capability requirements beyond just the image selection.

JoeRobich:
After digging I found issue https://github.com/dotnet/dnceng/issues/3587 which explains why the 1es- image is not present. Switching back to Build.Ubuntu.2204.AMD64.Open.

PR #8810, opened 2025-11-20
Remove cohosting setting and hardcode the value to true

davidwengier:
Part of https://github.com/dotnet/razor/issues/12491

Removes the bits of the C# extension that are for the non-cohosting editor. The OmniSharp setting still works as normal. I did it in bits, so commit-at-a-time might make sense, but most of it is just deleting whole files so GitHub hopefully will do a decent job.

There is a bit more cleanup that could happen, and could move some things around now that there is only one Razor system, could remove the word "cohost" from everything etc. but didn't want to make the PR too painful. Will need follow ups anyway, as bits form Roslyn and Razor are removed etc.

Reduces the VSIX size by about 40mb. I'll be honest, I was hoping for more :P

### DIFF
--- a/test/lsptoolshost/artifactTests/vsix.test.ts
+++ b/test/lsptoolshost/artifactTests/vsix.test.ts
@@ -18,7 +18,7 @@ describe('Vscode VSIX', () => {
     vsixFiles.forEach((element) => {
         // We're packaging the platform specific Roslyn server with ready to run in the vsix, so the size should be roughly ~50MB
         // We also publish the Razor server, which is roughly ~75MB
-        const sizeInMB = 240;
+        const sizeInMB = 200;
         const maximumVsixSizeInBytes = sizeInMB * 1024 * 1024;
 
         describe(`Given ${element}`, () => {


davidwengier:
Not as exciting as I would have hoped, but with both from local builds of the extension:
main:
| Extension                            | Eager | Load Code | Call Activate | Finish Activate | Event                                    | By                                 |
| ------------------------------------ | ----- | --------- | ------------- | --------------- | ---------------------------------------- | ---------------------------------- |
| ms-dotnettools.csharp                | true  | 516       | 2             | 131             | workspaceContains:**/*.{csproj,csx,cake} | ms-dotnettools.csharp              |


this branch:

| Extension                            | Eager | Load Code | Call Activate | Finish Activate | Event                                    | By                                 |
| ------------------------------------ | ----- | --------- | ------------- | --------------- | ---------------------------------------- | ---------------------------------- |
| ms-dotnettools.csharp                | true  | 456       | 2             | 125             | workspaceContains:**/*.{csproj,csx,cake} | ms-dotnettools.csharp              |

What's perhaps more interesting is that "Finish Activate" time was _dreadful_ until I manually deleted my local `.razor` and `.razorDevKit` folders. Which I can't really explain, as nothing should have been looking for files in them regardless. Maybe timing on local builds is just flaky.

dibarbet:
> What's perhaps more interesting is that "Finish Activate" time was dreadful until I manually deleted my local .razor and .razorDevKit folders. Which I can't really explain, as nothing should have been looking for files in them regardless. Maybe timing on local builds is just flaky.

interesting!  -50ms isn't too bad (still have a lot of O# code).  will try it out once we have an official build as well

PR #8819, opened 2025-11-28
Update TextMate grammars used in tests, and baselines

davidwengier:
In https://github.com/dotnet/razor/pull/12329 we updated the grammars we ship with VS. Forgot to update the grammars here, but we should probably have our tests match what we're shipping a little more closely.

This is a test-only change, and the grammars themselves came from the VS Code repo itself.


davidwengier:
@CyrusNajmabadi PTAL 😛

dibarbet (approved):

PR #8823, opened 2025-12-02
Add second opt-out flag for file-based programs

RikkiGibson:
dotnet/roslyn#81513

This also includes *Update Roslyn to 5.3.0-2.25603.1*.


package.json
```diff
@@ -1520,6 +1520,14 @@
               "preview"
             ]
           },
+          "dotnet.projects.enableFileBasedProgramsWithoutDirectives": {
+            "type": "boolean",
+            "default": true,
```
RikkiGibson, line 1525: In the release branch we will take another change on top to flip this default to false and the core enableFileBasedPrograms to true.

dibarbet: This will be important to remember since the next release will likely not be until January.

RikkiGibson:
Thanks @dibarbet!

PR #8839, opened 2025-12-05
Update extension to run on .NET 10

JoeRobich:
C# and C#DK are updating to run on .NET 10 runtime.


test/lsptoolshost/integrationTests/sourceGenerator.integration.test.ts
```diff
@@ -50,11 +50,11 @@ describe(`Source Generator Tests`, () => {
                 requestPosition
             )
         );
-        expect(referencesList.length).toEqual(8);
+        expect(referencesList.length).toEqual(5);
```
JoeRobich, line 53: Building against the net8.0 tfm causes the JsonSourceGenerator to generate fewer references.

PR #8885, opened 2026-01-14
update command prefix and create command that opens wiki page with explanation

claudiaregio:
This PR should update command prefixes from "CSharp" to "C#" and have one command named "CSharp: Commands now start with C#" and when clicked it should open a link in the browser (aka.ms/new-csharp-commands)

DRAFT as this should be merged into Feb Pre-release 1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1803,7 +1803,7 @@
     "commands": [
       {
 				"command": "csharp.openUrlInBrowser",
-				"title": "%commands.openUrlInBrowser%",
+				"title": "%command.openUrlInBrowser%",
 				"category": "CSharp"
 			},
       {


package.json
```diff
@@ -1801,6 +1801,11 @@
       }
     ],
     "commands": [
+      {
+				"command": "csharp.openUrlInBrowser",
+				"title": "%commands.openUrlInBrowser%",
+				"category": "CSharp"
+			},
```
dibarbet, line 1808: indentation seems wrong here (as compared to the other command definitions)

package.json
```diff
@@ -1801,6 +1801,11 @@
       }
     ],
     "commands": [
+      {
+				"command": "csharp.openUrlInBrowser",
```
dibarbet, line 1805: Possibly make the commandId here more specific?  e.g. `csharp.showNewCommandDocs` (or similar), since it is opening a specific page and not any URL.

package.json
```diff
@@ -1801,6 +1801,11 @@
...
+				"title": "%commands.openUrlInBrowser%",
```
JoeRobich:
```suggestion
				"title": "%command.openUrlInBrowser%",
```

PR #8886, opened 2026-01-14
Update required node version to v24 and required dotnet to 10.0

JoeRobich:
I saw warnings about using node v20 in CI logs. I see that there is about 3 months of support left (https://endoflife.date/nodejs) and wanted to get ahead of it.

### DIFF
--- a/test/vscodeLauncher.ts
+++ b/test/vscodeLauncher.ts
@@ -30,7 +30,7 @@ export async function prepareVSCodeAndExecuteTests(
     // Rather than having to uninstall Dev Kit between different test runs, we use workspace settings
     // to control which extensions are active - and we always install Dev Kit.
     const extensionsToInstall = [
-        'ms-dotnettools.vscode-dotnet-runtime@2.3.4',
+        'ms-dotnettools.vscode-dotnet-runtime@3.0.0',
         'ms-dotnettools.csharp',
         'ms-dotnettools.csdevkit@1.92.5',
     ];


CONTRIBUTING.md
```diff
@@ -23,10 +23,10 @@ Setting up your local development environment for the vscode-csharp repository i
 
 Before you start, make sure you have the following software installed on your machine:
 
-* Node.js v20 ([v20.17.0 LTS](https://nodejs.org/en/blog/release/v20.17.0)).
+* Node.js v22 ([22.17.0 LTS](https://nodejs.org/en/blog/release/v22.17.0)).
```
JoeRobich: Why 22.17? Wasn't sure whether to go with the absolute latest, so I chose a version a from a few months ago. Can update to v22.22 which is latest.

azure-pipelines/build-vsix.yml
```diff
@@ -133,11 +133,11 @@ stages:
       inputs:
         versionSpec: 3.11
 
-    # Non-windows signing requires .NET 8 installed on the machine.
+    # Non-windows signing requires .NET 10 installed on the machine.
```
dibarbet: I'm not sure this should change - specifically there was a .NET 8 requirement for non-windows signing on mac/linux.  If that is no longer the case then happy to move, otherwise this should stay .net8

azure-pipelines/loc.yml
```diff
@@ -59,16 +59,15 @@ extends:
             artifactName: l10n
         steps:
         - task: NodeTool@0
-          displayName: 'Install Node.js 18.x'
+          displayName: 'Install Node.js 22.x'
           inputs:
-            # Octokit client needs 18.x to have 'fetch' function.
-            versionSpec: '18.x'
+            versionSpec: '22.x'
```
dibarbet: could this be updated to use the install-node template

msbuild/global.json
```diff
@@ -1,9 +1,9 @@
 {
   "sdk": {
-    "version": "8.0.415",
+    "version": "10.0.101",
```
dibarbet, line 3: think this is fine, will just double check when CI is done that we ran the tests on the right .NET versions

CONTRIBUTING.md
dibarbet: We might want v22.22.0 at this point if we can

azure-pipelines/green-insertion.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     displayName: ⚙️ Install Node.js
     inputs:
       versionSource: spec
-      versionSpec: 16.x
+      versionSpec: 22.x
```
dibarbet: it feels like this should also be able to use the install-node template.  I don't think its necessary to use the vs-green registry here?

CONTRIBUTING.md
dibarbet: posted my comment before I saw this - yeah I think might as well go to v22.22 if we can

azure-pipelines/green-insertion.yml
JoeRobich: We aren't currently using this .yml. I will update it to use the install-node template but will leave the rest until we have time to revive it.

azure-pipelines/loc.yml
JoeRobich: Updated

azure-pipelines/build-vsix.yml
JoeRobich: Reverted!

.github/workflows/copilot-setup-steps.yml
```diff
@@ -28,13 +28,13 @@ jobs:
       # Install Node.js (required for TypeScript compilation and npm packages)
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
```
JoeRobich:
```suggestion
          node-version: '24'
```

PR #8907, opened 2026-01-22
Adjust Select Project Context editor menu item location

JoeRobich:
Give the Select Project Context its own item group in the editor context menu. This hopefully is enough to draw a distinction between `Tests in Context` a `Select Project Context`. We may want to additionally consider changing `Tests in Context` to `Tests at Cursor`.

Before:
image

After:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -5628,19 +5628,19 @@
       ],
       "editor/context": [
         {
+          "command": "csharp.changeProjectContextEditor",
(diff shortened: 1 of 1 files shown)


JoeRobich:
@claudiaregio Do you have any thoughts about whether we should change the labels for the `Run Tests in Context` / `Debug Tests in Context` to differentiate from `Select Project Context`. We could use `Run Tests at Cursor` / `Debug Tests at Cursor` instead.

dibarbet (approved):

PR #8933, opened 2026-01-30
Fix project context refresh handling on client

dibarbet:
Refreshes in the spec are requests, not notifications.  Because we're using the generic refresh infra in the server, project context refreshes are also requests, not notifications.

Without this, the server throws Remote method not found exceptions, and the refresh request never makes it to the client

should also fix https://github.com/dotnet/vscode-csharp/pull/8933


jasonmalinowski (approved):

JoeRobich (approved):

JoeRobich:
@dibarbet should we backport to prerelease so it can be in the next release?

RikkiGibson:
/backport to prerelease

Yes I think we should get this out sooner rather than later :)

PR #8935, opened 2026-01-30
Handle request error code `RoslynLspNonFatalRequestFailure`

RikkiGibson:
Related PR: dotnet/roslyn#81660
Fixes issue: dotnet/roslyn#81410
Fixes issue: https://github.com/microsoft/vscode-dotnettools/issues/2494

### DIFF
--- a/src/lsptoolshost/server/roslynProtocol.ts
+++ b/src/lsptoolshost/server/roslynProtocol.ts
@@ -365,5 +365,5 @@ export namespace RoslynLspErrorCodes {
      * is unlikely to be significant to the user (i.e. surface as an actual editor feature failing to function properly.)
      * For example, if pull diagnostics are requested for a virtual document that was already closed.
      */
-    export const nonFatalRequestFailure = -32799;
+    export const nonFatalRequestFailure = -30099;
 }


src/lsptoolshost/server/roslynLanguageClient.ts
```diff
@@ -34,13 +34,27 @@ export class RoslynLanguageClient extends LanguageClient {
...
 
+    /**
+     * Indicates that the server could not process the request, but that the failure shouldn't be surfaced to the user.
+     * (It's expected that the failure is still logged, however.)
+     *
+     * This is only meant to be used under conditions where we can't fulfill the request, but we think that the failure
+     * is unlikely to be significant to the user (i.e. surface as an actual editor feature failing to function properly.)
+     * For example, if pull diagnostics are requested for a virtual document that was already closed.
+     */
+    private static readonly RoslynLspNonFatalRequestFailure = -32799;
```
JoeRobich: Although this error code is just outside the reserved ranges for jsonrpc and lsp, I wonder if we should choose something a bit more distinct. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes

JoeRobich: Also, should we define this in RoslynProtocol and perhaps reserve our own range for errors?

RikkiGibson: > I wonder if we should choose something a bit more distinct

I'm open to more specific suggestions on the value.

> Also, should we define this in RoslynProtocol and perhaps reserve our own range for errors?

Were you referring to [roslynProtocol.ts](https://github.com/dotnet/vscode-csharp/blob/afc727e58c4e88febebd7b0c99663200c008c5af/src/lsptoolshost/server/roslynProtocol.ts)? Like a `RoslynLspErrorCodes` enumeration or something?

I could see us wanting to reserve... 0-1000 or something? It wasn't clear to me if the non-negative codes were meant to be used by specific language server implementation or what.

RikkiGibson: @dibarbet for any thoughts on this question

JoeRobich: > Were you referring to roslynProtocol.ts? Like a RoslynLspErrorCodes enumeration or something?

Doesn't have to be an enum, I know there are those that look down on typescripts implementation of that.

> I'm open to more specific suggestions on the value.

Maybe reserve ourselves -30099 to -30000.

dibarbet: > Were you referring to [roslynProtocol.ts](https://github.com/dotnet/vscode-csharp/blob/afc727e58c4e88febebd7b0c99663200c008c5af/src/lsptoolshost/server/roslynProtocol.ts)? Like a RoslynLspErrorCodes enumeration or something?

We should put it there, no need to reserve other ones I don't think.  It's not like a different server returning custom error codes could pollute ours (only reserved base LSP ones could).

RikkiGibson: I am open to any way of declaring the consts people may prefer.

Is the following appropriate:
```ts
export namespace RoslynLspErrorCodes {
    export const rangeStart = -30099;
    export const rangeEnd = -30000;

    export const nonFatalRequestFailure = -30099;
}
```

dibarbet: fine with that approach.  still not sure we need to necessarily reserve a range like this (as it only affects the communication between our client and server).  but not blocking

PR #8942, opened 2026-02-02
Add captureLogs command which will record C# and C# LSP log activity

JoeRobich:
Sets C# and LSP log level to Trace, Wrap LogOutputChannel to add a onMessageLog event, captures trace logging activity until the user cancels, generates a zip file with C# and C# LSP log files as well as captured trace logging. 

Demo:

https://github.com/user-attachments/assets/f3d640c2-386a-4a72-a370-507b0a65a1b1

Generated Zip: [csharp-logs-2026-02-02T19-50-03.zip](https://github.com/user-attachments/files/25024172/csharp-logs-2026-02-02T19-50-03.zip)

### DIFF
--- a/src/lsptoolshost/logging/captureLogs.ts
+++ b/src/lsptoolshost/logging/captureLogs.ts
@@ -61,58 +61,56 @@ async function captureLogsToZip(
                 // Wait for the user to cancel the progress
                 await waitForCancellation(token);
 
-                if (token.isCancellationRequested) {
-                    progress.report({
-                        message: vscode.l10n.t('Creating log archive...'),
-                    });
+                progress.report({
+                    message: vscode.l10n.t('Creating log archive...'),
+                });
 
-                    // Convert captured messages to log content
-                    const csharpLogContent = formatLogMessages(csharpLogMessages);
-                    const traceLogContent = formatLogMessages(traceLogMessages);
+                // Convert captured messages to log content
+                const csharpLogContent = formatLogMessages(csharpLogMessages);
+                const traceLogContent = formatLogMessages(traceLogMessages);
 
-                    // Prompt user for save location
-                    const saveUri = await vscode.window.showSaveDialog({
-                        defaultUri: getDefaultSaveUri(),
-                        filters: {
-                            // eslint-disable-next-line @typescript-eslint/naming-convention
-                            'Zip files': ['zip'],
-                        },
-                        saveLabel: vscode.l10n.t('Save Logs'),
-                        title: vscode.l10n.t('Save C# Logs'),
-                    });
+                // Prompt user for save location
+                const saveUri = await vscode.window.showSaveDialog({
+                    defaultUri: getDefaultSaveUri(),
+                    filters: {
+                        // eslint-disable-next-line @typescript-eslint/naming-convention
+                        'Zip files': ['zip'],
+                    },
+                    saveLabel: vscode.l10n.t('Save Logs'),
+                    title: vscode.l10n.t('Save C# Logs'),
+                });
 
-                    if (!saveUri) {
-                        // User cancelled the save dialog
-                        return;
+                if (!saveUri) {
+                    // User cancelled the save dialog
+                    return;
+                }
+
+                try {
+                    await createZipWithLogs(
+                        context,
+                        outputChannel,
+                        traceChannel,
+                        csharpLogContent,
+                        traceLogContent,
+                        saveUri.fsPath
+                    );
+                    const openFolder = vscode.l10n.t('Open Folder');
+                    const result = await vscode.window.showInformationMessage(
+                        vscode.l10n.t('C# logs saved successfully.'),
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/logging/captureLogs.ts
```diff
@@ -0,0 +1,237 @@
...
+            },
+            async (progress, token) => {
+                progress.report({
+                    message: vscode.l10n.t('Recording logs... Click Cancel to stop and save.'),
+                });
+
+                // Wait for the user to cancel the progress
+                await waitForCancellation(token);
+
+                if (token.isCancellationRequested) {
```
dibarbet: should waitForCancellation be inlined, and this bundled up in a single `token.isCancellationRequested` check

src/lsptoolshost/logging/captureLogs.ts
```diff
@@ -0,0 +1,237 @@
...
+        // Add existing log files to the archive
+        if (csharpLogContent) {
+            archive.append(csharpLogContent, { name: 'csharp.log' });
+        }
+        if (traceLogContent) {
+            archive.append(traceLogContent, { name: 'csharp-lsp-trace.log' });
+        }
+
+        // Add captured activity logs to the archive
+        archive.append(csharpActivityLogContent, { name: 'csharp.activity.log' });
```
dibarbet, line 199: Not quite getting the difference between these and the log content from the vscode log directory.  shouldn't they be the same, since we're setting the trace level?

If we need to know when recording started, potentially we could write something to the output channel indicating start/stop?

src/lsptoolshost/logging/observableLogOutputChannel.ts
```diff
@@ -0,0 +1,153 @@
...
+    public hide(): void {
+        this._channel.hide();
+    }
+
+    public dispose(): void {
+        this._onMessageLogged.dispose();
+        this._channel.dispose();
+    }
+
+    private formatMessage(message: string, args: any[]): string {
```
dibarbet, line 133: seems like we don't need two different format message functions (one here and one in the log collector)

src/lsptoolshost/logging/observableLogOutputChannel.ts
```diff
@@ -0,0 +1,153 @@
...
+    level: 'trace' | 'debug' | 'info' | 'warn' | 'error';
+    message: string;
+    timestamp: Date;
+}
+
+/**
+ * A wrapper around vscode.LogOutputChannel that emits events when messages are logged.
+ * This allows capturing log messages regardless of the UI log level filter.
+ */
+export class ObservableLogOutputChannel implements vscode.LogOutputChannel {
```
dibarbet, line 21: similar to before, want to understand why reading from the vscode output window log files isn't enough

src/lsptoolshost/logging/captureLogs.ts
JoeRobich: Although we are now telling the server to send us more logging, the Output Channels LogLevel is still in whatever state the user has choosen. This means more detailed logging will not be written to the files unless the user had already set their LogLevel manually to Trace.

src/lsptoolshost/logging/observableLogOutputChannel.ts
JoeRobich: The one turns Errors and Objects into strings that can be logged. The one in the log collector creates a `Timestamp | LogLevel | Message` string for writing to the activity log file.

JoeRobich: But I suppose since this is a single use event we can do all the formatting here assuming there is no need to use the other LogMessage fields for anything

src/lsptoolshost/logging/captureLogs.ts
dibarbet: Would it be reasonable to set the output channel log level directly perhaps?  Might simplify things here

JoeRobich: Unfortunately, LogOutputChannel.logLevel is a getter. This was my best compromise. Tell the server to report everything and then hook the channel to get all the logging.

src/lsptoolshost/logging/captureLogs.ts
JoeRobich: I am not sure why we are checking for cancellation again actually. waitForCancellation implies the token is cancelled

src/lsptoolshost/logging/observableLogOutputChannel.ts
JoeRobich: See earlier comment.

src/lsptoolshost/logging/observableLogOutputChannel.ts
JoeRobich: @dibarbet Was this a sticking point for you? I kinda think I would like to keep them separate.

PR #8946, opened 2026-02-03
Add the ability to open a solution file

JoeRobich:
An experiment to see what it would be like if opening a loose solution file in VSCode would then open the containing folder and launch roslyn lsp against the solution.

- If workspaceFolders are empty and the C# extension starts with the solution file as the active document, then we open the parent folder in the current window. 
- If workspacesFolders are not empty and the active document changes to a solution file outside the workspacesFolders, then we open the parent folder in a new window with the solution file open.
- When RoslynLanguageServer is determining the initial solution to open, we check whether the active file is a solution file and use that instead of any settings.

Downsides:
We don't want to activate C# outside of C# scenarios, since solution file are used for languages other than C# (for example C++). One thought is that a Solution Launcher extension could be made that always activate on startup and contained the code in solutionFileWorkspaceHandler.ts. Users could then opt-in to the solution loading behavior by installing this extension.

From within VS Code:

https://github.com/user-attachments/assets/b6a17396-52a4-4531-b0ce-6ea04db20893

From the command line:

https://github.com/user-attachments/assets/01020661-3cf8-43a7-94a8-0adcb39e6a4e

### DIFF
--- a/package.json
+++ b/package.json
@@ -699,7 +699,7 @@
     "onLanguage:csharp",
     "onCommand:o.showOutput",
     "onCommand:omnisharp.registerLanguageMiddleware",
-    "workspaceContains:**/*.{csproj,csx,cake}"
+    "workspaceContains:**/*.{csproj,csx,cake,sln,slnx,slnf}"
   ],
   "contributes": {
     "themes": [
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -699,7 +699,7 @@
     "onLanguage:csharp",
     "onCommand:o.showOutput",
     "onCommand:omnisharp.registerLanguageMiddleware",
-    "workspaceContains:**/*.{csproj,csx,cake}"
+    "workspaceContains:**/*.{csproj,csx,cake,sln,slnx,slnf}"
```
JoeRobich, line 702: This doesn't work as I expected and the C# extension doesn't load when only a loose solution file is opened in VS Code.

```suggestion
    "workspaceContains:**/*.{csproj,csx,cake}"
```

PR #8948, opened 2026-02-03
Log settings when capturing logs

JoeRobich:
Builds on https://github.com/dotnet/vscode-csharp/pull/8942

In addition to C# and C# LSP logs, capture the C# settings for the user.

### DIFF
--- a/src/lsptoolshost/logging/captureLogs.ts
+++ b/src/lsptoolshost/logging/captureLogs.ts
@@ -227,37 +227,35 @@ async function readLogFileContent(logFileUri: vscode.Uri): Promise<string | null
  */
 function gatherCurrentSettings(): string {
     const settings = {
-        commonOptions: {
-            waitForDebugger: commonOptions.waitForDebugger,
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/logging/captureLogs.ts
```diff
@@ -216,6 +221,45 @@ async function readLogFileContent(logFileUri: vscode.Uri): Promise<string | null
...
 }
 
+/**
+ * Gathers the current settings for CommonOptions, LanguageServerOptions, and RazorOptions.
+ * Returns a formatted JSON string.
+ */
+function gatherCurrentSettings(): string {
+    const settings = {
+        commonOptions: {
+            waitForDebugger: commonOptions.waitForDebugger,
```
dibarbet: is it possible to iterate the properties on the type, so it would automatically pick up new values?

JoeRobich: Good call, I think I could use the KeysOf helper

PR #8951, opened 2026-02-04
Capture logs when recording a server trace

JoeRobich:


src/lsptoolshost/profiling/profiling.ts
```diff
@@ -135,12 +144,62 @@ async function executeDotNetTraceCommand(
         return;
     }
 
-    const terminal = await getOrCreateTerminal(traceFolder, outputChannel);
+    // Create observers to collect log messages during capture
+    const csharpLogObserver = outputChannel.observe();
+    const traceLogObserver = traceChannel.observe();
+
+    // Set log levels to Trace for capture and get the restore function
+    const restoreLogLevels = await languageServer.setLogLevelsForCapture();
```
dibarbet, line 152: could we make observe set the log levels for capture, and then restore them when stop is called?

src/lsptoolshost/profiling/profiling.ts
```diff
@@ -135,12 +144,62 @@ async function executeDotNetTraceCommand(
...
+                vscode.l10n.t({
+                    message: 'Failed to save C# trace: {0}',
+                    args: [errorMessage],
+                    comment: ['{0} is the error message'],
+                })
+            );
+        }
+    } finally {
+        // Always restore log levels
+        await restoreLogLevels();
```
dibarbet, line 203: does this also need to call stop on the observers (can stop be called twice if it made it past the previous stop)?

src/lsptoolshost/logging/captureLogs.ts
```diff
@@ -116,30 +109,15 @@ async function captureLogsToZip(
             }
         );
     } finally {
-        // Always clean up subscriptions and restore log levels
-        csharpLogSubscription.dispose();
-        traceLogSubscription.dispose();
+        // Always restore log levels
```
dibarbet: same here, shouldn't this call stop (unsubscribe)?

src/lsptoolshost/profiling/profiling.ts
JoeRobich: I don't think it is a good fit for that. Observe is for a single channel and setLogLevels changes the levels for both. Would observer always need trace logging or is there a use where we just capture what the user has configured. What if multiple observers are created for the same channel, for whatever reason?

src/lsptoolshost/profiling/profiling.ts
JoeRobich: ah, good thinking. will separate into getLogs and dispose methods

dibarbet (approved):

PR #8970, opened 2026-02-07
Add support and change default to balanced sg execution

dibarbet:
Requires https://github.com/dotnet/roslyn/pull/82330


src/lsptoolshost/generators/sourceGeneratorsRefresh.ts
```diff
@@ -0,0 +1,47 @@
...
+                forceRegeneration: true,
+            });
+        })
+    );
+
+    // After a build task completes, refresh source generators without force regeneration.
+    context.subscriptions.push(
+        vscode.tasks.onDidEndTask(async (e) => {
+            if (e.execution.task.group === vscode.TaskGroup.Build) {
+                outputChannel.trace('Refreshing source generators on build start on build end');
```
JoeRobich:
```suggestion
                outputChannel.trace('Refreshing source generators on build end');
```

aradalvand:
The description of the Balanced mode says that it happens automatically on "save".
Is this really true? I only saw build-start and build-end event handlers being registered in the code.

dibarbet:
> The description of the Balanced mode says that it happens automatically on "save". Is this really true? I only saw build-start and build-end event handlers being registered in the code.

@aradalvand save is handled server side via LSP https://github.com/dotnet/roslyn/blob/main/src/LanguageServer/Protocol/Handler/DocumentChanges/DidSaveHandler.cs#L26

Build has no corresponding LSP hooks so must be done here via custom client code.

PR #8980, opened 2026-02-10
Set DOTNET_EnableDumpOnSigTerm when crash dumps are enabled

333fred:
This will ensure that when the OS sends the language server a sigterm, we also get a dump of that.

### DIFF
--- a/src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionResolver.ts
+++ b/src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionResolver.ts
@@ -102,6 +102,8 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
             env.DOTNET_DbgMiniDumpType = '4';
             // Collect crashreport.json with additional thread and stack frame information.
             env.DOTNET_EnableCrashReport = '1';
+            // Collect dumps when the process is sent a SigTerm as well
+            env.DOTNET_EnableDumpOnSigTerm = '1';
             // The dump file name format is <executable>.<pid>.dmp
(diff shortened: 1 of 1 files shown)


333fred:
The main thing I'm concerned about here is whether this will be "too" verbose, creating dumps for otherwise ok scenarios. @dibarbet @JoeRobich for thoughts.

src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionResolver.ts
```diff
@@ -102,6 +102,8 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
             env.DOTNET_DbgMiniDumpType = '4';
             // Collect crashreport.json with additional thread and stack frame information.
             env.DOTNET_EnableCrashReport = '1';
+            // Collect dumps when the process is sent a SigTerm as well
+            env.DOTNET_EnableDumpOnSigTerm = '1';
```
dibarbet, line 106: Slightly worried we might get a ton of useless dumps - like for example if vscode is shutting down and kills the process.  Can we check if that happens?

dibarbet: To some extent it might be reasonable in that the user has already set the crashdump path.  But it would be pretty bad if they leave it on and fill up their machine with useless dumps that aren't real crashes.

JoeRobich: Can we have the user set this on an as needed basis using the new env args setting?

PR #8982, opened 2026-02-11
Improve server initialization and error handling

dibarbet:
Paired with https://github.com/dotnet/roslyn/pull/82376

This came up as part of the overall C#DK reliability effort.  Currently the experience in the editor when the language server fails isn't great.  A couple issues:
1.  If the language server crashes, depending on the kind of error we do not always display a notification to the user, and its only visible in logs.
2.  If the language server crashes, the user gets spammed with meaningless notifications that have no actionable items

This PR does a few things
1.  Switch to using the built-in support for LSP processes using named pipes.  Previously we had our own support to handle a named pipe connection between the client and server.  This required us to manage the process ourselves.  However, VSCode LSP client has built in pipe support which we can easily utilize, and remove a bunch of our custom logic around pipe connections and process management.  We can delegate all of that to VSCode, which knows how to handle process crashes, killing the process, etc etc.
2.  De-duplicated toasts on crash.  The built in notifications created by the VSCode LSP client when things fail often overreport failures.  For example, on a crash it reports errors due to the process going away, errors due to the stream closing, and errors for indivdual requests failing.  The LSP client library however allows us to customize this so we can ensure we get a single notification instead of 3+.
3.  Added actions to the error toasts - they now have report issue buttons which open our issue reporter with logs pre-filled.  Server crashes can trigger a restart of extensions (temporary, to be deleted once we do rehydrate support).
4.  Deleted intellicode support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   * Fix `GetDeconstructionInfo` on converted deconstruction assignment (PR: [#82324](https://github.com/dotnet/roslyn/pull/82324))
   * Fix crash in simplify linq expression (PR: [#82392](https://github.com/dotnet/roslyn/pull/82392))
   * Do not suggest using with-element in pre-C# 15 (PR: [#82389](https://github.com/dotnet/roslyn/pull/82389))
+* Collect Razor logs (PR: [#8988](https://github.com/dotnet/vscode-csharp/pull/8988))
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/server/roslynLanguageServer.ts
```diff
@@ -1056,67 +935,16 @@ export class RoslynLanguageServer {
         return args;
     }
 
-    private static async getCSharpDevkitIntelliCodeExportArgs(
```
JoeRobich, line 1059: love it

PR #8988, opened 2026-02-11
Collect Razor logs

JoeRobich:
Adds Razor logs to our captureLogs, collectDump, and recordTrace commands.


JoeRobich:
> "Create Razor Issue" does this too, I wonder if we should remove that in favour of this, so its easier for users to understand what to do? Though that command does also capture a couple of extra Razor specific things too...

We are currently discussing how to unify log collection between C# and CDK. I can imagine this will be followed by working on a unified "Report Issue". I will be sure to loop you in when we get there.

PR #9006, opened 2026-02-14
Move from gulp to node scripts

JoeRobich:
This PR moves us from Gulp to running node scripts via `node run`. It exposes all our tasks as entries in the packages.json scripts section. This allows us to use `npm run {task}` in the same manner as we were using `npx gulp {task}`.

**_Problems with gulp_**
- Gulp didn't release a new version for multiple years until just recently
- Gulp or one of the gulp tasks we relied on limited our ability to take newer versions of some dependencies
- Also fewer dependencies is better assuming we can do it without reimplementing our own version of gulp

**_Aside_**
I investigated using node's built-in type stripping support to replace our use of ts-node, but it required changes inside the /src directory which I was trying to avoid. At some point I am sure we can move all our code to be compatible in order to avoid the compilation step. In particular the changes required meant `<type>` casts became `as type`, enum types became const lookups, and imports of Typescript `type` definitions were changed to `import type`.


src/omnisharp/omnisharpPackageCreator.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Package } from '../packageManager/package';
+import type { Package } from '../packageManager/package';
```
JoeRobich: An example of a pure typescript type being imported being made compatible with type stripping

src/packageManager/absolutePath.ts
```diff
@@ -6,7 +6,10 @@
 import * as path from 'path';
 
 export class AbsolutePath {
-    constructor(public value: string) {
+    public value: string;
+
+    constructor(value: string) {
+        this.value = value;
```
JoeRobich: Unfortunately some of the typescript sugar had to be transformed into a compatible form.

init.ps1
```diff
@@ -36,11 +36,8 @@ try {
     Write-Host "`n[3/5] Installing project dependencies..." -ForegroundColor Cyan
```
JoeRobich:
```suggestion
Write-Host "`n[1/4] Installing ado-npm-auth globally..." -ForegroundColor Cyan
Run-Command "npm" @("install", "-g", "ado-npm-auth") "Failed to install ado-npm-auth."

Write-Host "`n[2/4] Authenticating with Azure DevOps..." -ForegroundColor Cyan
if (Test-Path ".npmrc") {
Run-Command "ado-npm-auth" @("-c", ".npmrc") "Authentication failed."
} else {
Write-Host ".npmrc file not found in the current directory." -ForegroundColor Red
throw ".npmrc file not found in the current directory."
}

Write-Host "`n[3/4] Installing project dependencies..." -ForegroundColor Cyan
```

tasks/runTask.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export function runTask(task: () => Promise<void>): void {
```
JoeRobich: Since CJS doesn't support top-level await we need a little wrapper function so our async tasks can run.

tasks/profiling/profiling.ts
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { profilingTask } from './profilingTasks';
+import { runTask } from '../runTask';
+
+runTask(profilingTask);
```
dibarbet: nit - for ones that aren't sharing much with other tasks - should we just inline them in here too?

tasks/profiling/profilingTasks.ts
```diff
@@ -4,43 +4,41 @@
...
 
-createProfilingTasks();
+const profilingOutputFolder = path.join(outPath, 'profiling');
 
-function createProfilingTasks() {
-    const profilingOutputFolder = path.join(outPath, 'profiling');
-    gulp.task(`profiling:csharp:${basicSlnTestProject}`, async () => {
-        // Ensure the profiling output folder exists, otherwise the outputs will not get written.
-        fs.mkdirSync(path.join(outPath, 'profiling'), { recursive: true });
+export async function profilingCsharpTask(): Promise<void> {
```
dibarbet: does this need to be exported?

tasks/profiling/profilingTasks.ts
```diff
@@ -4,43 +4,41 @@
...
-        await mergeTraces(profilingOutputFolder);
-    });
+    const files = fs.readdirSync(profilingOutputFolder);
+    const nettraceFiles = files.filter((f) => f.endsWith('.nettrace'));
+    if (nettraceFiles.length === 0) {
+        throw new Error('No .nettrace files found in the profiling output folder.');
+    }
+}
+
+export async function mergeTracesTask(): Promise<void> {
```
dibarbet: does this need to be exported?

tasks/tags/createTagsTasks.ts
```diff
@@ -20,7 +19,7 @@ interface CreateTagsOptions {
     prerelease: string | null;
 }
 
-gulp.task('createTags:roslyn', async (): Promise<void> => {
+export async function createTagsRoslynTask(): Promise<void> {
```
dibarbet: Could remove the export, and same for the other two tags tasks below

CONTRIBUTING.md
```diff
@@ -38,9 +38,8 @@ Follow these steps to build, run, and test the repository:
 
 If you have the ability to run powershell, you can invoke "init.ps1" from the root of the repo. If not, the following steps will get build going for you as well:
 
-1. Run `npm i` - This command installs the project dependencies.
-2. Run `npm i -g gulp` - This command installs Gulp globally.
-3. Run `gulp installDependencies` - This command downloads the various dependencies as specified by the version in the [package.json](package.json) file.
+1. Run `npm ci` - This command installs the project dependencies.
+2. Run `npm run installDependencies` - This command downloads the various dependencies as specified by the version in the [package.json](package.json) file.
```
dibarbet: as a side note, I need to figure out how to run this as part of `npm ci` (can be done, but needs to be updated to no-op better if the version is already installed, and cleanup after  itself)

tasks/profiling/profilingTasks.ts
JoeRobich: I think they were individual gulp tasks previously, so was just natural to export them. But you are right, we don't have a task for them. Will inline.

tasks/profiling/profiling.ts
JoeRobich: Ok, went through and moved as much of the code into the individual tasks files as possible. Some code still needed to be shared so a few files of shared helpers is sticking around.

CONTRIBUTING.md
JoeRobich: I worked something up for this. Will post it as a follow up.

JoeRobich: Eh, I pushed it. Review last commit to see it in isolation

tasks/profiling/profiling.ts
JoeRobich: Import thing to remember is that if the task file is using `runTask` it shouldn't export anything because nothing should import it, since importing it is the same as executing it asynchronously.

tasks/packaging/offlinePackagingTasks.ts
```diff
@@ -204,10 +197,31 @@ async function acquireNugetPackage(
...
+    const outputPath = nugetPackageInfo.vsixOutputPath;
+    const versionFilePath = path.join(outputPath, versionFileName);
+    if (!fs.existsSync(versionFilePath)) {
+        return false;
+    }
+
+    const installedVersion = fs.readFileSync(versionFilePath, 'utf-8');
+    const versionFound = installedVersion === packageVersion;
+
+    if (!versionFound) {
```
dibarbet, line 210: nice!

package.json
```diff
@@ -63,47 +63,61 @@
...
     "l10nDevGenerateXlf": "npx @vscode/l10n-dev generate-xlf ./package.nls.json ./l10n/bundle.l10n.json --outFile ./loc/vscode-csharp.xlf",
-    "l10nDevImportXlf": "npx @vscode/l10n-dev import-xlf ./loc/vscode-csharp.*.xlf --outDir ./l10n && move l10n\\package.nls.*.json ."
+    "l10nDevImportXlf": "npx @vscode/l10n-dev import-xlf ./loc/vscode-csharp.*.xlf --outDir ./l10n && move l10n\\package.nls.*.json .",
+    "lint": "npx eslint ./",
+    "omnisharptest": "npm run packageDev && npx ts-node tasks/tests/omnisharptest.ts",
+    "omnisharptest:integration": "npm run packageDev && npx ts-node tasks/tests/omnisharptestIntegration.ts",
+    "omnisharptest:unit": "npm run compileDev && npx ts-node tasks/tests/omnisharptestUnit.ts",
+    "package": "npm run compile && npm run signJs && node esbuild.js --production",
+    "packageDev": "npm run compileDev && node esbuild.js",
+    "prepare": "npm run installDependencies",
```
dibarbet, line 85: 🥳

PR #9014, opened 2026-02-19
Don't show misc file toast for deleted file

dibarbet:
tentative, depending on approach
resolves https://github.com/microsoft/vscode-dotnettools/issues/2804

### DIFF
--- a/src/lsptoolshost/workspace/miscellaneousFileNotifier.ts
+++ b/src/lsptoolshost/workspace/miscellaneousFileNotifier.ts
@@ -67,15 +67,12 @@ export function registerMiscellaneousFileNotifier(
 }
 
 async function isDeletedFile(document: vscode.TextDocument): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


src/lsptoolshost/workspace/miscellaneousFileNotifier.ts
```diff
@@ -60,6 +66,18 @@ export function registerMiscellaneousFileNotifier(
     });
 }
 
+async function isDeletedFile(document: vscode.TextDocument): Promise<boolean> {
+    if (document.uri.scheme === 'file') {
```
JoeRobich: We have this same filter in the calling function.

JoeRobich (approved):

src/lsptoolshost/workspace/miscellaneousFileNotifier.ts
dibarbet: that we do!

PR #9019, opened 2026-02-21
Refactor build-vsix to set run variables in a templated step.

JoeRobich:
Noticed that we were paying a job queuing cost by having build-vsix be separate jobs. Extracted the Set Run Variables steps into a template so we only have to pay that cost once.

### DIFF
--- a/azure-pipelines/build-vsix.yml
+++ b/azure-pipelines/build-vsix.yml
@@ -22,75 +22,6 @@ stages:
   displayName: 'Build VSIXs'
   dependsOn: []
   jobs:
-  - job: SetRunVariables
-    displayName: 'Set Run Variables'
-    pool:
-      ${{ if eq(parameters.isOfficial, true) }}:
-        name: netcore1espool-internal
-        image: 1es-ubuntu-2204
-      ${{ else }}:
-        name: NetCore-Public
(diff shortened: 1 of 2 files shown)


azure-pipelines/build-vsix.yml
```diff
@@ -22,75 +22,6 @@ stages:
   displayName: 'Build VSIXs'
   dependsOn: []
   jobs:
-  - job: SetRunVariables
```
JoeRobich, line 25: Extracted to template so can run in the Build VSIXs job.

azure-pipelines/set-run-variables.yml
```diff
@@ -0,0 +1,74 @@
...
+        $isPrerelease = $false
+      } else {
+        Write-Host "Branch is not release, using prerelease channel."
+        $isPrerelease = $true
+      }
+    }
+
+    if ( $isPrerelease ) {
+      Write-Host "Setting pipeline channel variable to Prerelease."
+      Write-Host "##vso[task.setvariable variable=channel]Prerelease"
```
JoeRobich, line 40: Removed `;isoutput=true` from these because there is no dependent job that will use them. Jobs that use them will include the template.

PR #9026, opened 2026-02-23
Remove tooling runtime override validation

JoeRobich:


src/lsptoolshost/dotnetRuntime/dotnetRuntimeExtensionResolver.ts
```diff
@@ -39,7 +38,7 @@ export class DotnetRuntimeExtensionResolver implements IHostExecutableResolver {
         private getServerPath: (platform: PlatformInformation) => string,
         private channel: vscode.OutputChannel,
         private extensionPath: string
-    ) {}
+    ) { }
```
JoeRobich: Feel like prettier is going to complain abotu this
```suggestion
    ) {}
```

PR #9032, opened 2026-02-26
Unify logging commands under `csharp.collectLogs`

JoeRobich:
image

Combines the 3 logging commands into a single command which always captures the current log files. The user will be prompted to choose additional log sources like: Record Activity, Performance Trace, Memory Dump, and GC Dump. When dumps are combined with recording activity or a trace then they are taken before and after the recording.

### DIFF
--- a/src/lsptoolshost/logging/collectLogs.ts
+++ b/src/lsptoolshost/logging/collectLogs.ts
@@ -188,7 +188,7 @@ async function selectAdditionalLogs(): Promise<LogsToCollect | undefined> {
         canPickMany: true,
     });
 
-    if (!selected || selected.length === 0) {
+    if (!selected) {
         return undefined;
     }
 


package.json
```diff
@@ -1952,20 +1952,8 @@
...
-        "category": "CSharp",
-        "enablement": "isWorkspaceTrusted && (dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'RoslynDevKit')"
-      },
-      {
-        "command": "csharp.collectDump",
-        "title": "%command.csharp.collectDump%",
+        "command": "csharp.collectLogs",
+        "title": "%command.csharp.collectLogs%",
         "category": "CSharp",
         "enablement": "isWorkspaceTrusted && (dotnet.server.activationContext == 'Roslyn' || dotnet.server.activationContext == 'RoslynDevKit')"
```
JoeRobich, line 1958: When the C#DK provides dump collection with their logging, we can disable this command when running with C#DK.

```suggestion
        "enablement": "isWorkspaceTrusted && dotnet.server.activationContext == 'Roslyn'"
```

src/lsptoolshost/logging/collectLogs.ts
```diff
@@ -0,0 +1,505 @@
...
+            const args = await promptForToolArguments(config.toolName, defaultArgs);
+            if (args === undefined) {
+                toolsReady = false;
+                return;
+            }
+            dumpConfigs.push({ type: 'memory' as DumpType, args });
+        }
+
+        if (gcDump) {
+            progress.report({
```
dibarbet, line 269: A decent amount of mostly duplicative for these various collect options - could consider unifying into a a helper

src/lsptoolshost/logging/profiling.ts
```diff
@@ -44,319 +30,46 @@ export function getProfilingEnvVars(outputChannel: vscode.LogOutputChannel): Nod
...
+    outputChannel: ObservableLogOutputChannel,
     token: vscode.CancellationToken
 ): Promise<string | undefined> {
-    // Use a terminal to execute the dotnet-trace.  This is much simpler and more reliable than executing dotnet-trace
-    // directly via the child_process module as dotnet-trace relies on shell input in order to stop the trace.
-    // Without using a psuedo-terminal, it is extremely difficult to send the correct signal to stop the trace.
-    //
-    // Luckily, VSCode allows us to use the built in terminal (a psuedo-terminal) to execute commands, which also provides a way to send input to it.
-
+    const terminal = await getOrCreateTerminal(folder, outputChannel);
```
dibarbet, line 53: I think the comments here were useful, and should probably be kept (applies to pretty much all the comments below)

JoeRobich: Sorry, restored.

src/lsptoolshost/logging/collectLogs.ts
JoeRobich: The progress reporting and tool acquisition is going away in a follow up.

src/lsptoolshost/logging/collectLogs.ts
```diff
@@ -0,0 +1,505 @@
...
+        },
+    ];
+
+    const selected = await vscode.window.showQuickPick(items, {
+        title: vscode.l10n.t('Collect C# Logs'),
+        placeHolder: vscode.l10n.t('Select additional logging to collect'),
+        canPickMany: true,
+    });
+
+    if (!selected || selected.length === 0) {
```
JoeRobich: The logs are additional so we shouldn't cancel if none are selected.

```suggestion
    if (!selected) {
```

PR #9035, opened 2026-03-02
Use `dotnet dnx` to run dotnet tools used in logging

JoeRobich:
Builds on https://github.com/dotnet/vscode-csharp/pull/9032

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -78,6 +78,29 @@ ${stderr}`)
     });
 }
 
+export async function execFileChildProcess(
+    file: string,
+    args: string[],
+    workingDirectory: string | undefined,
+    env: NodeJS.ProcessEnv
+): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
+        cp.execFile(file, args, { cwd: workingDirectory, maxBuffer: 500 * 1024, env: env }, (error, stdout, stderr) => {
+            if (error) {
+                reject(
+                    new Error(`${error}
+${stdout}
+${stderr}`)
+                );
+            } else if (stderr && !stderr.includes('screen size is bogus')) {
+                reject(new Error(stderr));
+            } else {
+                resolve(stdout);
+            }
+        });
+    });
+}
+
 export async function getUnixChildProcessIds(pid: number): Promise<number[]> {
     return new Promise<number[]>((resolve, reject) => {
         cp.exec('ps -A -o ppid,pid', (error, stdout, stderr) => {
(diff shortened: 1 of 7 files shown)


src/common.ts
```diff
@@ -78,6 +78,29 @@ ${stderr}`)
     });
 }
 
+export async function execFileChildProcess(
```
JoeRobich, line 81: Added so I could directly run an executable without having to quote the path.

src/lsptoolshost/logging/collectLogs.ts
```diff
@@ -83,14 +86,18 @@ async function collectLogs(
         }
     }
 
-    // Step 2: Verify tools and let the user customize arguments for each selected tool
+    // Step 2: Resolve the dotnet path and let the user customize arguments for each selected tool
+
+    // We use the resolved dotnet path instead of relying on an install being available from
+    // the PATH because the resolved one is guarenteed to support `dotnet dnx` which will allow
```
dibarbet, line 92: ah interesting - is `dnx` included in the runtime install?

src/lsptoolshost/logging/profiling.ts
```diff
@@ -50,21 +51,36 @@ export async function runDotnetTraceInTerminal(
     //
     // Luckily, VSCode allows us to use the built in terminal (a psuedo-terminal) to execute commands, which also provides a way to send input to it.
 
-    const terminal = await getOrCreateTerminal(folder, outputChannel);
+    const dotnetFolder = path.dirname(dotnetPath);
+    const dotnetExecutableName = path.basename(dotnetPath);
+
+    // Generate an output file name in the form <appname>_<yyyyMMdd>_<HHmmss>.nettrace
+    const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
+    const outputFileName = `Microsoft.CodeAnalysis.LanguageServer_${timestamp}.nettrace`;
```
dibarbet, line 59: I think the default name includes the timestamp - do we need to recreate it?
>The output path for the collected trace data. If not specified it defaults to <appname>_<yyyyMMdd>_<HHmmss>.nettrace, for example, `myapp_20210315_111514.nettrace``.

https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-trace#dotnet-trace-collect

src/lsptoolshost/logging/collectLogs.ts
JoeRobich: No, will need to revisit this when we require a net10.0 SDK.

src/lsptoolshost/logging/profiling.ts
JoeRobich: Looks like we need to recalculate it ourselves.

PR #9051, opened 2026-03-06
Fancier CHANGELOG for version 2.130.x

JoeRobich:
This update includes significant improvements in reliability, diagnostics tooling, language server performance, and Razor editing. Key changes involve enhanced error reporting, balanced source generator execution, and new commands for capturing logs and dumps.


JoeRobich:
It is easier to read the file on its own https://github.com/dotnet/vscode-csharp/blob/4e5d74aacd87492f54608f14eff8bea996130b93/CHANGELOG.md

JoeRobich:
@dibarbet @davidwengier @claudiaregio Should we try a fancy changelog with this release?

davidwengier:
Looks great! Can update the skills so they produce this too? Or is it best to do individual insertions as normal, and then fancy up the changelog after the fact?

JoeRobich:
> Can update the skills so they produce this too? Or is it best to do individual insertions as normal, and then fancy up the changelog after the fact?

I think this format is better suited to describing the changes in our Release versions. The way we structure the prerelease changes makes generating this very easy.

JoeRobich:
> Can update the skills so they produce this too?

I added a skill for generating these in the future - https://github.com/dotnet/vscode-csharp/pull/9053

davidwengier:
> I added a skill for generating these in the future - #9053

Okay, so this is produced from the existing changlogs, so the skills we have to do the insertions still make sense as is. That's all I was wondering about. Thanks

PR #9096, opened 2026-03-24
Add setting `dotnet.fileBasedApps.enableAutomaticDiscovery`

RikkiGibson:
Also updates the Roslyn version to include the automatic discovery feature.

### DIFF
--- a/test/lsptoolshost/integrationTests/gotoImplementation.integration.test.ts
+++ b/test/lsptoolshost/integrationTests/gotoImplementation.integration.test.ts
@@ -38,10 +38,10 @@ describe(`Go To Implementation Tests`, () => {
         expect(implementationList).toHaveLength(2);
 
         expect(implementationList[0].uri.path).toContain('BaseClassImplementation.cs');
-        expect(implementationList[0].range).toStrictEqual(new vscode.Range(2, 17, 2, 40));
+        expect(implementationList[0].range).toStrictEqual(new vscode.Range(2, 17, 2, 17));
 
         expect(implementationList[1].uri.path).toContain('implementation.cs');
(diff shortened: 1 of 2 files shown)


RikkiGibson:
@akhera99 quick check--do changes in commit c26f35e match your expectations given dotnet/roslyn#82906?

package.json
```diff
@@ -1561,6 +1561,14 @@
             "tags": [
               "preview"
             ]
+          },
+          "dotnet.fileBasedApps.enableAutomaticDiscovery": {
```
jasonmalinowski, line 1565: Is the plan to migrate the other settings to this new prefix? It kinda makes sense, since it's odd the old switches were under a "projects" node which is odd given the entire point of them is they aren't projects. :smile:

test/lsptoolshost/integrationTests/gotoImplementation.integration.test.ts
```diff
@@ -38,10 +38,10 @@ describe(`Go To Implementation Tests`, () => {
         expect(implementationList).toHaveLength(2);
 
         expect(implementationList[0].uri.path).toContain('BaseClassImplementation.cs');
-        expect(implementationList[0].range).toStrictEqual(new vscode.Range(2, 17, 2, 40));
+        expect(implementationList[0].range).toStrictEqual(new vscode.Range(2, 17, 2, 17));
```
akhera99, line 41: test changes look correct!

package.json
RikkiGibson: yes #9136

RikkiGibson:
@dibarbet when this flows to release branch we just want to make sure to go into the setting definition and flip default to false.

PR #9155, opened 2026-04-08
Use new protocol method for sg content and normalize language client imports

dibarbet:
requires https://github.com/dotnet/roslyn/pull/83119


tsconfig.json
```diff
@@ -21,10 +21,10 @@
           "vscode-languageclient/selectionRange": ["./node_modules/vscode-languageclient/lib/common/selectionRange"],
           "vscode-languageclient/protocolConverter": ["./node_modules/vscode-languageclient/lib/common/protocolConverter"],
           "vscode-languageclient/node": ["./node_modules/vscode-languageclient/lib/node/main"],
-          "vscode-languageserver-protocol": ["./node_modules/vscode-languageserver-protocol/lib/common/api"],
-          "vscode-languageserver-protocol/node": ["./node_modules/vscode-languageserver-protocol/lib/node/main"],
+          "vscode-languageserver-protocol": ["./node_modules/vscode-languageclient/node_modules/vscode-languageserver-protocol/lib/common/api"],
+          "vscode-languageserver-protocol/node": ["./node_modules/vscode-languageclient/node_modules/vscode-languageserver-protocol/lib/node/main"],
           "vscode-jsonrpc": ["./node_modules/vscode-jsonrpc/lib/common/api"],
-          "vscode-jsonrpc/node": ["./node_modules/vscode-jsonrpc/lib/node/main"]
+          "vscode-jsonrpc/node": ["./node_modules/vscode-jsonrpc/lib/node/main"],
```
JoeRobich, line 27: Should these be removed or rewritten in terms of the vscode-languageclient package?

JoeRobich (approved):

tsconfig.json
dibarbet: We still use types from here - but via the transitive ref to vscode-jsonrpc from vscode-languageclient.  

We could possibly remove some of the usages from here (switch from RequestType to ProtocolRequestType), but its a little bit more involved, and I'm not sure we can get rid of them all (I think cancellation token comes from here for example).

JoeRobich: > We still use types from here - but via the transitive ref to vscode-jsonrpc from vscode-languageclient.

Oh, so this path is accurate and we don't need to reference from the node_modules under the vscode-languageclient package the way `vscode-languageserver-protocol` does?

dibarbet: Yeah - the protocol types seem to be shipped inside it.  But not vscode-jsonrpc

Image

PR #9219, opened 2026-04-21
Update .npmrc to disallow running install scripts

phil-allen-msft:
Update .npmrc to disallow running install scripts

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,5 @@
 registry=https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public-npm/npm/registry/
 always-auth=true
+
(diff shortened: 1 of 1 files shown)


WardenGnaw:
You are going to need to add a `--ignore-scripts=false` to the vsix publish command since they require scripts running.

E.g.
`npm i -g @vscode/vsce --userconfig $(Build.SourcesDirectory)/.npmrc --ignore-scripts=false`

I think its here: https://github.com/dotnet/vscode-csharp/blob/d2e0ee44613f3d6c79eb7292689f5aa3152d91a6/tasks/packaging/vsceTasks.ts#L13

PR #9277, opened 2026-05-06
Make Razor a built in feature of the Roslyn language server

davidwengier:
VS Code side of https://github.com/dotnet/roslyn/pull/83575, expect test failures without that.

We no longer need special infrastructure to bring Razor in as an extension, or provide it most of the configuration we used to do.

Part of https://github.com/dotnet/razor/issues/13140


davidwengier:
I didn't change the changelog, because the only commit this has that https://github.com/dotnet/vscode-csharp/pull/9293 doesn't is the packaging changes.

.github/skills/update-roslyn-version/SKILL.md
davidwengier, line 1: This is just reverted to whatever it was before the repo merge.

dibarbet (approved):

PR #9344, opened 2026-05-22
Commands to install/uninstall roslyn-language-server for the Copilot CLI

phil-allen-msft:
2 new commands (invokable from Command Palette/F1) to install and uninstall, which is really are edits on $HOME/.copilot/lsp-config.json

Reasonable error handling and notifications are present.  Initial integration with microsoft-tas-client as well.

To update the 'redist' copy of lsp-config.json which is merged in as part of install, run `npm run ingest`

Unittests added; manually tested with no file (file created, content added), existing file which is C# (no changes), and existing file which has C++-looking bindings (content added).


src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
     context.subscriptions.push(
         vscode.commands.registerCommand('csharp.showOutputWindow', async () => outputChannel.show())
     );
+    context.subscriptions.push(
+        vscode.commands.registerCommand(installCopilotLspCommand, async () => {
+            const lspConfigPath = path.join(os.homedir(), '.copilot', 'lsp-config.json');
```
dibarbet, line 105: Think we should probably also check if they have a plugin installed with an lsp.json configured for C#

test/lsptoolshost/unitTests/copilotLspConfig.test.ts
```diff
@@ -0,0 +1,144 @@
...
+        expect(firstRun.updatedContent).toBe(packagedContent);
+
+        const secondRun = getUpdatedCopilotLspConfigContent(firstRun.updatedContent, packagedContent);
+        expect(secondRun.shouldWrite).toBe(false);
+
+        const finalContent = secondRun.updatedContent ?? firstRun.updatedContent;
+        expect(finalContent).toBe(packagedContent);
+    });
+
+    test('does not modify config when any server maps .cs in fileExtensions', () => {
```
jasonmalinowski, line 28: Is this the behavior we want? If somebody is wanting to run us to upgrade from some other LSP to our own, don't we want to remove the other one?

tasks/components/ingestCopilotLspConfig.ts
```diff
@@ -0,0 +1,81 @@
...
+const outputPath = path.join(rootPath, 'redist', 'lsp-config.json');
+
+runTask(ingestCopilotLspConfig);
+
+async function ingestCopilotLspConfig() {
+    const content = await downloadText(sourceUrl);
+    validateLspConfig(content, sourceUrl);
+    await fs.mkdir(path.dirname(outputPath), { recursive: true });
+    await fs.writeFile(outputPath, content, 'utf8');
+    console.log(`Updated ${outputPath} from ${sourceUrl}`);
```
jasonmalinowski, line 22: Is there not some other npm package for downloading to a file easily?

package.json
```diff
@@ -68,6 +68,7 @@
     "createTags": "npx ts-node tasks/tags/createTags.ts",
     "fixLocUrls": "npx ts-node tasks/debugger/fixLocUrls.ts",
     "generateOptionsSchema": "npx ts-node tasks/debugger/generateOptionsSchema.ts",
+    "ingest": "npx ts-node tasks/components/ingestCopilotLspConfig.ts",
```
jasonmalinowski, line 71: Is this something we expect is happening automatically during a build or something we need to do on a regular basis?

src/lsptoolshost/commands.ts
jasonmalinowski, line 105: Rather than doing manual configuration this way, should we just be installing one of our existing plugins that contains the registration?

src/lsptoolshost/copilotLspConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+            csharpServerNames.push(serverName);
+        }
+    }
+
+    return csharpServerNames;
+}
+
+function containsCSharpServer(lspServers: { [key: string]: unknown }): boolean {
+    for (const [serverName, server] of Object.entries(lspServers)) {
+        if (serverName === 'csharp' || serverContainsCSharpFileExtension(server)) {
```
jasonmalinowski, line 95: Why is this checking the sever name but the other function above is just looking for the extension?

src/lsptoolshost/commands.ts
dibarbet, line 105: nit - could we extract this into a separate file & function (similar to other command registrations)

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+                return;
+            }
+            let currentContent: string | undefined;
+
+            try {
+                currentContent = await fs.readFile(lspConfigPath, 'utf8');
+            } catch (error) {
+                const nodeError = error as NodeJS.ErrnoException;
+                if (nodeError.code !== 'ENOENT') {
+                    void vscode.window.showErrorMessage(
```
dibarbet, line 124: Instead of manually showing an error message, consider throwing `throw Error(msg)` - the command infrastructure already seems to handle failures and throw up a modal dialog.

Image

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+    context.subscriptions.push(
+        vscode.commands.registerCommand(installCopilotLspCommand, async () => {
+            const lspConfigPath = path.join(os.homedir(), '.copilot', 'lsp-config.json');
+            const extensionLspConfigPath = path.join(context.extension.extensionPath, packagedCopilotLspConfigPath);
+            let packagedContent: string;
+            try {
+                packagedContent = await fs.readFile(extensionLspConfigPath, 'utf8');
+            } catch (error) {
+                const nodeError = error as NodeJS.ErrnoException;
+                void vscode.window.showErrorMessage(
```
dibarbet, line 112: nit - I don't necessarily think it's worth catching this error - we don't expect this to fail, and if it does the command execution should show a modal dialog if it does.  Doesn't feel necessary to wrap into a special localized exception

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+                    );
+                    return;
+                }
+            }
+
+            let updateResult: { shouldWrite: boolean; updatedContent?: string };
+            try {
+                updateResult = getUpdatedCopilotLspConfigContent(currentContent, packagedContent);
+            } catch (error) {
+                const message = error instanceof Error ? error.message : String(error);
```
dibarbet, line 135: Similar to before, I don't think we expect this to throw, so imho it is fine to let this bubble up without wrapping into a nice exception

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+                void vscode.window.showErrorMessage(vscode.l10n.t('Failed to update Copilot LSP config: {0}', message));
+                return;
+            }
+
+            if (!updateResult.shouldWrite || !updateResult.updatedContent) {
+                void vscode.window.showInformationMessage(
```
dibarbet, line 141: I feel like this should be a modal dialog - the user explicitly invoked a command, so it makes sense to be noisy

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+                    )
+                );
+                return;
+            }
+
+            try {
+                await fs.mkdir(path.dirname(lspConfigPath), { recursive: true });
+                await fs.writeFile(lspConfigPath, updateResult.updatedContent, 'utf8');
+                void vscode.window.showInformationMessage(
+                    vscode.l10n.t('Updated Copilot LSP config at {0}.', lspConfigPath)
```
dibarbet, line 153: would it be worth opening the config file after we write it, so the user can see the change?

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+                );
+            } catch (error) {
+                const nodeError = error as NodeJS.ErrnoException;
+                void vscode.window.showErrorMessage(
```
dibarbet, line 157: as above - could throw a nice exception and let it bubble up to the command handling

src/lsptoolshost/commands.ts
```diff
@@ -92,5 +100,118 @@ function registerExtensionCommands(
...
+        vscode.commands.registerCommand(uninstallCopilotLspCommand, async () => {
+            const lspConfigPath = path.join(os.homedir(), '.copilot', 'lsp-config.json');
+
+            let currentContent: string | undefined;
+            try {
+                currentContent = await fs.readFile(lspConfigPath, 'utf8');
+            } catch (error) {
+                const nodeError = error as NodeJS.ErrnoException;
+                if (nodeError.code !== 'ENOENT') {
+                    void vscode.window.showErrorMessage(
```
dibarbet, line 173: same nit on throwing (and below)

src/lsptoolshost/copilotLspConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+    const packagedCsharpConfig = packagedConfig.lspServers?.csharp;
+    if (!packagedCsharpConfig || typeof packagedCsharpConfig !== 'object') {
+        throw new Error('Packaged Copilot LSP config is missing lspServers.csharp.');
+    }
+
+    if (currentContent === undefined) {
+        return { shouldWrite: true, updatedContent: packagedContent };
+    }
+
+    const currentConfig = JSON.parse(currentContent) as CopilotLspConfig;
```
dibarbet, line 24: are these configs allowed to contain comments?  Elsewhere we use `jsonc` for that

src/lsptoolshost/copilotLspConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+    const updatedConfig: CopilotLspConfig = {
+        ...currentConfig,
+        lspServers:
+            currentConfig.lspServers && typeof currentConfig.lspServers === 'object'
+                ? { ...currentConfig.lspServers }
+                : {},
+    };
+
+    updatedConfig.lspServers!.csharp = packagedCsharpConfig;
+    return { shouldWrite: true, updatedContent: `${JSON.stringify(updatedConfig, null, 2)}\n` };
```
dibarbet, line 38: Could this change the formatting of their file?  Maybe not a huge issue, but we've done things like this - https://github.com/dotnet/vscode-csharp/blob/main/src/shared/assets.ts#L1017 before to ensure in place updates

src/lsptoolshost/copilotLspConfig.ts
dibarbet: Feels like they should both check for extension - I believe server name could be anything.  

Also curious what happens if there is already a c# server configured that doesn't use roslyn-language-server.

package.json
dibarbet: I don't think this should happen automatically on build - otherwise a build could randomly fail if the config changes.

src/lsptoolshost/copilotLspConfig.ts
dibarbet: I do also wonder if we could install the lsp config as a plugin (e.g. ship the plugin in the extension, then use CLI commands to install the plugin).  And rely on the cli for de-dupe?  Not sure if thats any better though

src/activateRoslyn.ts
```diff
@@ -78,6 +80,8 @@ export function activateRoslyn(
         logDirectory: context.logUri.fsPath,
         determineBrowserType: BlazorDebugConfigurationProvider.determineBrowserType,
         experimental: {
+            getTreatmentVariable: <T extends boolean | number | string>(name: string, configId = 'vscode') =>
```
dibarbet, line 83: What is actually using / plan to be using this export?  Doesn't appear to be for tests, so maybe we delete?

PR #9368, opened 2026-06-01
Remove razor logger

davidwengier:
I was going to do this in https://github.com/dotnet/vscode-csharp/pull/9367 but it was a bit bigger than I thought, so its separate and the first two commits of this can just be ignored.

This removes the razor logger, and just sends everything to the C# output window, since https://github.com/dotnet/vscode-csharp/pull/9367 removes the calls to the `razor/log` endpoint anyway. It also removes the "Report Razor Issue" command, because the collect logs command is the way to go now anyway. If we really need to see the C# and Html virtual documents, we can ask the user for them, and we can always update the command to include them. Can't remember the last time we used them.


JoeRobich (approved):

dibarbet (approved):
thanks!!!

PR #9426, opened 2026-06-12
Optional source based test discovery support in C# Dev Kit.

peterwald:


dibarbet (approved):

sebastianstudniczek:
@peterwald This looks interesting, out of curiosity how does it work? Is it already possible to test with devkit?

peterwald:
> @peterwald This looks interesting, out of curiosity how does it work? Is it already possible to test with devkit?

This is just laying the groundwork. The rest is on the way. This particular component is what enables  C# Dev Kit to get the access it needs to the Roslyn workspace for the feature to work.

PR #9447, opened 2026-06-18
Use local checkout for fetching package.json instead of accessing github directly

dibarbet:
Release pipeline is failing to access raw.githubusercontent.com due to network blocks when generating tags.  This modifies the pipeline to use the local checkout instead (note that the pipeline checks out vscode-csharp at the commit that is being released)

### DIFF
--- a/tasks/tags/createTags.ts
+++ b/tasks/tags/createTags.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as fs from 'fs';
+import { spawnSync } from 'child_process';
 import minimist from 'minimist';
 import { Octokit } from '@octokit/rest';
 import { allNugetPackages, NugetPackageInfo, platformSpecificPackages } from '../packaging/offlinePackagingTasks';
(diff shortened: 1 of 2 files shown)


tasks/tags/createTags.ts
```diff
@@ -258,3 +236,51 @@ async function getCommitFromNugetAsync(
...
+
+    try {
+        const headCommit = getGitOutput(['rev-parse', 'HEAD']);
+        if (headCommit?.toLowerCase() === normalizedReleaseCommit) {
+            console.log(`Reading package.json from checked-out commit ${releaseCommit}`);
+            return JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+        }
+
+        console.log(`Reading package.json from local git object ${releaseCommit}:${packageJsonPath}`);
+        const packageJsonString = getGitOutput(['show', `${releaseCommit}:${packageJsonPath}`]);
```
jasonmalinowski, line 250: It looks like you can do `git show HEAD:<filename>` and you don't need the intermediate rev-parse.

jasonmalinowski (approved):

JoeRobich (approved):

PR #9552, opened 2026-07-20
Add one-time feedback survey for Copilot Chat users (prerelease)

AbhitejJohn:
## What

Adds a one-time feedback survey prompt for C# extension users who are using
GitHub Copilot Chat.

## Behavior

- **Eligibility**: at least one Copilot-vendor chat model is available via
  `vscode.lm.selectChatModels({ vendor: ''copilot'' })`. Since Copilot Chat now
  ships with VS Code, an install check no longer distinguishes real users;
  requiring an available Copilot model means the user has Chat set up and is
  signed in. Probed once at registration, so we bail before wiring anything up
  when Chat isn''t usable.
- **Timing**: deferred until `ServerState.ProjectInitializationComplete` (via a
  self-disposing `onServerStateChange` listener) so the toast doesn''t contend
  with reload/restore/IntelliCode notifications at activation.
- **One-shot with snooze** (persisted in `globalState`):
  - **Take Survey** -> opens the survey URL; retires the survey *only after*
    `openExternal` succeeds (otherwise it snoozes so it can resurface).
  - **Not Now** -> retires the survey permanently.
  - **Ignored / closed** -> snoozed for 1 hour, then eligible again.
  - A passively-ignored toast never burns the one-shot.
- **Multi-window safe(r)**: the snooze path re-reads `globalState` and never
  downgrades a terminal `surveyShown: true`, so a choice made in one window
  isn''t clobbered by an ignored toast in another.

## Telemetry

- `roslyn/copilotChatSurvey` with `outcome`: `shown` | `accepted` | `dismissed`.
- `roslyn/copilotChatSurvey/error` when the prompt fails to present.

## Files

- `src/lsptoolshost/copilot/copilotChatSurvey.ts` (new)
- `src/lsptoolshost/activate.ts` (wire-up)
- `src/shared/telemetryEventNames.ts` (new event names)

### DIFF
--- a/src/lsptoolshost/activate.ts
+++ b/src/lsptoolshost/activate.ts
@@ -79,7 +79,7 @@ export async function activateRoslynLanguageServer(
     registerLanguageStatusItems(context, languageServer, languageServerEvents);
     registerMiscellaneousFileNotifier(context, languageServer);
     registerCopilotContextProviders(context, languageServer, _channel);
-    registerCopilotChatSurvey(context, languageServerEvents, reporter, _channel);
+    registerCopilotChatSurvey(context, languageServer, languageServerEvents, reporter, _channel);
 
     // Register any commands that need to be handled by the extension.
     registerCommands(context, languageServer, hostExecutableResolver, _channel, _traceChannel, reporter);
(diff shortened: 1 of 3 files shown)


AbhitejJohn:
Closing this prerelease port; it's not needed. The change is tracked in #9551 against main.

PR #9600, opened 2026-07-30
Migrate VSIX publisher to a C# file-based app

dibarbet:
## Summary
- move VSIX Marketplace publishing to a .NET 10 file-based app
- retain `Publish-Vsix.ps1` as a compatibility wrapper for releases of older build artifacts
- include the app in build artifacts and install the required SDK in the release job

## Validation
- exercised dry-run and pre-release argument forwarding
- verified stdout and stderr reach the pipeline console
- verified non-zero command exits fail the app

### DIFF
--- a/azure-pipelines/Publish-Vsix.cs
+++ b/azure-pipelines/Publish-Vsix.cs
@@ -1,10 +1,13 @@
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
 
 var path = GetSourceDirectory();
 var preRelease = false;
 var dryRun = false;
 var ci = false;
+const string AlreadyPublishedPattern =
+    @"^\s*ERROR\s+[A-Za-z0-9]+(?:[.-][A-Za-z0-9]+)*(?:\s+\([^)]+\))?\s+v[0-9.]+(?:-[0-9a-z.]+)*\s+already exists\.\s*$";
 
 for (var index = 0; index < args.Length; index++)
 {
@@ -111,7 +114,25 @@ foreach (var package in packages)
     }
     else
     {
-        RunVsce(arguments);
+        var publishArguments = GetVsceArguments(arguments);
+        var result = RunCommandWithOutput(npx, publishArguments);
+        if (result.ExitCode == 0)
+        {
+            continue;
+        }
+
+        if (Regex.IsMatch(
+            result.Output,
+            AlreadyPublishedPattern,
+            RegexOptions.Multiline))
+        {
+            Console.WriteLine(
+                $"{Path.GetFileName(package.PackagePath)} was already published.");
+            continue;
+        }
+
+        throw new InvalidOperationException(
+            $"'{npx}' exited with code {result.ExitCode}.");
     }
 }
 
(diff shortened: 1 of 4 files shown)


azure-pipelines/Publish-Vsix.cs
```diff
@@ -0,0 +1,165 @@
...
+            PackagePath: packagePath,
+            ManifestPath: $"{basePath}.manifest",
+            SignaturePath: $"{basePath}.signature.p7s");
+    })
+    .ToArray();
+
+if (packages.Length == 0)
+{
+    Console.WriteLine($"No .vsix files found in '{path}'. Nothing to publish.");
+    return 0;
```
JoeRobich: Should we make this return non-zero?

azure-pipelines/release.yml
```diff
@@ -73,6 +73,10 @@ extends:
                   Write-Host "##vso[build.updatebuildnumber]$buildNumberName"
                 displayName: Set BuildNumber
               - template: /azure-pipelines/install-node.yml@self
+              - task: UseDotNet@2
```
JoeRobich, line 76: I am surprised we have images without .NET 10 preinstalled

azure-pipelines/Publish-Vsix.cs
```diff
@@ -0,0 +1,165 @@
...
+
+    using var process = Process.Start(startInfo)
+        ?? throw new InvalidOperationException($"Failed to start '{fileName}'.");
+    process.WaitForExit();
+
+    if (process.ExitCode != 0)
+    {
+        throw new InvalidOperationException(
+            $"'{fileName}' exited with code {process.ExitCode}.");
+    }
```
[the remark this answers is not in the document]
JoeRobich: We def want to tolerate duplicate package.

azure-pipelines/release.yml
dibarbet: It very likely does, but imho I prefer to ensure its installed in case that ever changes.  can delete if you feel strongly

JoeRobich: nah, it isn't run often enough to be a bother.

azure-pipelines/Publish-Vsix.cs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+if (!Directory.Exists(path))
+{
+    throw new DirectoryNotFoundException($"Path '{path}' does not exist.");
+}
+
+if (!ci)
+{
+    RunCommand(GetCommandName("az"), ["account", "show"]);
```
JoeRobich, line 45: When not logged in the output looks like
```
> az account show
Please run 'az login' to setup account.
```
This is probably sufficient as long as it is echoed to the user.

azure-pipelines/Publish-Vsix.cs
```diff
@@ -0,0 +1,217 @@
...
+        throw new FileNotFoundException(
+            $"Signature file not found for '{package.PackagePath}'.",
+            package.SignaturePath);
+    }
+}
+
+var npx = GetCommandName("npx");
+if (dryRun)
+{
+    RunVsce(["verify-pat", "--azure-credential", "ms-dotnettools"]);
```
JoeRobich, line 88: When this error the output is similar to 
```
The Personal Access Token verification has failed.
```

PR #9721, opened 2026-09-02
Don't ask to report an issue if the server process was killed externally

dibarbet:
have seen a few issue reports where the server process was externally killed (e.g. sigkill).  There's nothing we can do with that information, so we don't need to ask for a issue report.

Note that this only works on linux/mac.  Windows doesn't have a way to differentiate.


test/lsptoolshost/unitTests/roslynLanguageClient.test.ts
```diff
@@ -18,26 +32,188 @@ jest.mock('vscode-languageclient', () => ({
...
 
-        client._hasShownConnectionClose = false;
-        client._telemetryReporter = { sendTelemetryEvent };
-        client.showCrashNotificationCore = showCrashNotificationCore;
+/** Lets the fire-and-forget notification promise chain settle. */
+async function flushPendingNotifications(): Promise<void> {
+    for (let i = 0; i < 5; i++) {
+        await Promise.resolve();
+    }
+}
```
JoeRobich, line 69: Why is 5 enough? Is there a better way?

JoeRobich (approved):

PR #9782, opened 2026-09-17
Install lsp plugin

dibarbet:


src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+    channel: Pick<vscode.LogOutputChannel, 'error' | 'info' | 'trace'>;
+};
+
+export function registerDotnetPlugin(
+    context: DotnetPluginHost['context'] & Pick<vscode.ExtensionContext, 'subscriptions' | 'extensionMode'>,
+    reporter: ITelemetryReporter,
+    channel: DotnetPluginHost['channel']
+): void {
+    const host: DotnetPluginHost = { context, reporter, channel };
+    const controller = new AbortController();
```
JoeRobich, line 61: Looks like AbortController is used instead of CancellationTokenSource so that we can return a cancellation reason.

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+function finishInstallation(host: DotnetPluginHost, result: InstallResult): void {
+    report(host, TelemetryEventNames.CopilotDotnetPlugin, result.outcome, result.source, false);
+    if (result.outcome === 'installed') {
+        void showInstalled();
+    }
+}
+
+function blockedReason(): Outcome | undefined {
```
JoeRobich: Maybe `determineIfBlocked`

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+        trace(host, `Caching ${result.cache.outcome} result for ${result.cache.source} source.`);
+        await host.context.globalState.update(dotnetPluginCacheKey, {
+            extensionVersion: host.context.extension.packageJSON.version,
+            ...result.cache,
+        } satisfies Cache);
+    }
+    finishInstallation(host, result);
+}
+
+function finishInstallation(host: DotnetPluginHost, result: InstallResult): void {
```
JoeRobich: inline?

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (commonOptions.disableAIFeatures) {
+        return 'aiDisabled';
+    }
+    if (!vscode.workspace.isTrusted) {
+        return 'untrustedWorkspace';
+    }
+    return undefined;
+}
+
+function readCache(context: DotnetPluginHost['context']): Cache | undefined {
```
JoeRobich: inline?

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+    const cache = context.globalState.get<Cache>(dotnetPluginCacheKey);
+    return cache?.extensionVersion === context.extension.packageJSON.version ? cache : undefined;
+}
+
+function deadline(signal: AbortSignal): { signal: AbortSignal; done: () => void } {
```
JoeRobich: Feel like this would be cleaner if it were done where we create the controller. We already add hook extension disabled from there.

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+        names.push(marketplace.name);
+    }
+
+    if (!names.includes(marketplaceName)) {
+        trace(host, `Registering ${marketplaceName} marketplace.`);
+        await runCopilotCli(cli, ['plugin', 'marketplace', 'add', marketplaceSource], signal);
+    }
+}
+
+function enabledOutcome(plugins: CopilotPlugin[]): CachedOutcome {
```
JoeRobich, line 228: inline?

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+function deadline(signal: AbortSignal): { signal: AbortSignal; done: () => void } {
+    const timer = new AbortController();
+    const handle = setTimeout(
+        () => timer.abort(named('TimeoutError', 'The Copilot CLI did not respond in time.')),
+        operationTimeoutMs
+    );
+    return { signal: AbortSignal.any([signal, timer.signal]), done: () => clearTimeout(handle) };
+}
+
+async function listPlugins(cli: CopilotCli, signal: AbortSignal): Promise<CopilotPlugin[]> {
```
JoeRobich, line 198: inline?

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,286 @@
...
+    context.subscriptions.push({
+        dispose: () => controller.abort(named('AbortError', 'The C# extension was deactivated.')),
+    });
+}
+
+/**
+ * Installs only when automatic installation and AI are enabled, the workspace is trusted, Copilot is available,
+ * and no existing or conflicting .NET plugin is found. Stable results are cached per extension version.
+ */
+export async function installDotnetPlugin(host: DotnetPluginHost, signal: AbortSignal): Promise<void> {
```
JoeRobich: Feel like this should be merged with registerDotnetPlugin. The separation seems to exist only for testing purposes.

package.nls.json
```diff
@@ -30,6 +30,7 @@
   "configuration.dotnet.autoInsert.enableAutoInsert": "Enable automatic adjustments of code constructs on typing, including documentation comment insertion, brace formatting adjustments, and raw string literal support.",
   "configuration.dotnet.formatting.organizeImportsOnFormat": "Specifies whether 'using' directives should be grouped and sorted during document formatting.",
   "configuration.dotnet.defaultSolution.description": "The path of the default solution to be opened in the workspace when multiple solutions are available.",
+  "configuration.dotnet.copilotDotnetPlugin.enableAutoInstall": "Automatically install the official .NET plugin for GitHub Copilot when a compatible Copilot installation is available. (Requires extension restart)",
   "configuration.dotnet.server.path": "Specifies the absolute path to the server (LSP or O#) executable. When left empty the version pinned to the C# Extension is used. (Requires extension restart)",
```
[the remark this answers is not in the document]
dibarbet: expected - will be updated by loc

l10n/bundle.l10n.json
```diff
@@ -69,6 +69,8 @@
   "Replace existing build and debug assets?": "Replace existing build and debug assets?",
   "Could not locate .NET Core project in '{0}'. Assets were not generated.": "Could not locate .NET Core project in '{0}'. Assets were not generated.",
   "Unable to generate assets to build and debug. {0}.": "Unable to generate assets to build and debug. {0}.",
+  "Learn More": "Learn More",
+  "Installed the C# LSP .NET plugin for GitHub Copilot": "Installed the C# LSP .NET plugin for GitHub Copilot",
```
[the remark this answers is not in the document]
dibarbet: expected - will be updated by loc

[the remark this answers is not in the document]
JoeRobich: I've gotten this feedback before as well. Perhaps we should update the repo instructions or code review skill.

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+    args: readonly string[],
+    token: CancellationToken
+): Promise<string> {
+    const controller = new AbortController();
+    const cancellation = token.onCancellationRequested(() => controller.abort());
+    try {
+        if (token.isCancellationRequested) {
+            controller.abort();
+        }
+        controller.signal.throwIfAborted();
```
dibarbet, line 143: kept this - this automatically handles aborting and killing the process if we pass it in to execFile

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+function platformPath(): typeof path.win32 {
+    return os.platform() === 'win32' ? path.win32 : path.posix;
+}
+
+function absolute(value: string | undefined): value is string {
+    if (!value || !platformPath().isAbsolute(value)) {
+        return false;
+    }
+    // A Windows rooted path without a drive still depends on the current drive.
+    return os.platform() !== 'win32' || /^[a-z]:[\\/]|^[\\/]{2}[^\\/]+[\\/][^\\/]+/i.test(value);
```
JoeRobich: How come we can't trust the win32 `isAbsolute` when it is `true`? If the path doesn't begin with a drive letter or network path, then it wouldn't make it past the earlier check. https://nodejs.org/api/path.html#pathisabsolutepath

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+function pathDirectories(): string[] {
+    return (process.env.PATH ?? '')
+        .split(platformPath().delimiter)
+        .map((directory) => directory.trim().replace(/^"(.*)"$/, '$1'))
+        .filter(absolute);
+}
+
+// The GitHub Copilot app does not expose the CLI directly; it downloads a pinned build into its cache.
+async function appCli(directories: readonly string[]): Promise<CopilotCli | undefined> {
+    const p = platformPath();
```
JoeRobich: Maybe name this `path` so that it shadows the import and is more apparent in this function what it is for.

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+        cache = environmentPath('XDG_CACHE_HOME') ?? p.join(home, '.cache');
+        // Tauri deb/rpm resources use productName, including its spaces. An extracted
+        // AppImage's usr/bin + usr/lib layout also works when usr/bin is on PATH.
+        for (const bin of new Set(['/usr/bin', '/usr/local/bin', ...directories])) {
+            apps.push({
+                executable: p.join(bin, 'github'),
+                resources: p.join(p.dirname(bin), 'lib', 'GitHub Copilot'),
+            });
+        }
+    }
```
JoeRobich: Would not be mad if this were pulled into a local function.

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+            'github-copilot-sdk',
+            'cli',
+            version.replace(/[^a-zA-Z0-9._-]/g, '_'),
+            platform === 'win32' ? 'copilot.exe' : 'copilot'
+        );
+        if (existsSync(command)) {
+            return { command, source: 'app' };
+        }
+    }
+    return undefined;
```
JoeRobich: This method needs some empty lines between following closing braces to let the code breathe

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+export async function findCopilotCli(): Promise<CopilotCli | undefined> {
+    const p = platformPath();
```
JoeRobich: same feedback as earlier the name

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+        if (kind && entry) {
+            plugins.push({ name: entry[1], enabled: !entry[2], kind });
+        }
+    }
+    if (!recognized) {
+        // Report a CLI output format change instead of silently reinstalling or skipping removal.
+        throw new Error('Unrecognized Copilot plugin inventory');
+    }
+    return plugins;
+}
```
JoeRobich: Same comment as earlier about letting the code breathe

src/shared/copilot/copilotCli.ts
```diff
@@ -0,0 +1,207 @@
...
+
+function pathDirectories(): string[] {
+    return (process.env.PATH ?? '')
+        .split(platformPath().delimiter)
+        .map((directory) => directory.trim().replace(/^"(.*)"$/, '$1'))
+        .filter(absolute);
+}
+
+// The GitHub Copilot app does not expose the CLI directly; it downloads a pinned build into its cache.
+async function appCli(directories: readonly string[]): Promise<CopilotCli | undefined> {
```
JoeRobich: For name maybe `discoverCli` or similar. Is it possible to have multiple installed? If so, do they share marketplaces and skills? Should we return multiple?

src/shared/copilot/dotnetPlugin.ts
```diff
@@ -0,0 +1,269 @@
...
+                host.channel.trace(`Copilot .NET plugin: Installing ${pluginSource} using ${source} source.`);
+                await runCopilotCli(cli, ['plugin', 'install', pluginSource], cancellation.token);
+                host.channel.trace(`Copilot .NET plugin: Installed ${pluginSource}.`);
+                outcome = 'installed';
+            }
+        }
+
+        // 6. Cache the stable result before returning it to the caller.
+        stage = 'cache';
+        await host.context.globalState.update(dotnetPluginCacheKey, {
```
JoeRobich: Should we cache the command to avoid future discovery, in the case it still exists?

JoeRobich (commented):
Should we include an unregister command?

src/shared/copilot/copilotCli.ts
JoeRobich: Typescript treats fixed length arrays as "tuples"

src/shared/copilot/dotnetPlugin.ts
dibarbet: Sorry - didn't quite understand this one - we do cache the result to avoid future discovery

dibarbet:
> Should we include an unregister command?

Tried, but we can't uninstall while vscode is open

src/shared/copilot/copilotCli.ts
dibarbet: As long as the copilot home directory for them is the same (which it generally is) they will share marketplaces and skills

src/shared/copilot/copilotCli.ts
dibarbet: technically `\tools` passes the isAbsolute check but is relative to the drive root (which could change).  But this actually doesn't matter we handle non-existent paths later anyway so will delete

src/shared/copilot/dotnetPlugin.ts
JoeRobich: Sorry, I guess I was thinking in terms of unregistering, which I now know we cannot do.

src/shared/copilot/copilotCli.ts
JoeRobich: Odd that a single slash is sufficient. Too bad that isn't one of their sample paths.

Issue #9791, opened 2026-09-18
Remove temporary .NET 11 CI container digest pin


JoeRobich:
The Linux .NET 11 PR CI container is temporarily pinned to digest `sha256:74206e5cb93b8a254539effe1b220c51ef6d97b7c488b9029335f1a62df53975`.

The mutable image moved from .NET SDK `11.0.100-preview.7.26381.103` to `11.0.100-rc.1.26425.128`. The RC1 SDK is in the file-based-app artifact-path compatibility window described by dotnet/roslyn#85204 and fixed by dotnet/sdk#56160. This causes the file-based program restore integration test to miss the post-restore `project.assets.json` change and fail.

Remove the digest pin after `mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-26.04-net11.0-vscode-csharp-amd64` is updated to an SDK containing dotnet/sdk#56160 and the Linux .NET 11 CSharp and DevKit integration test legs pass with the mutable tag.

JoeRobich:
The temporary digest pin is proposed in https://github.com/dotnet/vscode-csharp/pull/9792.

PR #9792, opened 2026-09-18
Pin .NET 11 CI container to last green digest

JoeRobich:
## Summary

- pin the Linux .NET 11 test container to the last green image digest
- avoid the .NET 11 RC1 file-based app artifact-path mismatch while the mutable image is being updated
- reference #9791 for removing the temporary pin

The mutable image changed from SDK `11.0.100-preview.7.26381.103` to `11.0.100-rc.1.26425.128`. The RC1 SDK falls in the compatibility window described by dotnet/roslyn#85204 and fixed by dotnet/sdk#56160, causing the file-based program restore integration test to miss the post-restore assets update.

## Validation

- `git diff --check`
- confirmed the pinned manifest resolves from MCR with HTTP 200


dibarbet (approved):
we're probably in a catch 22 here - devkit requires 11 rc, but this breaks fbp tests...

